PR #36, opened 2019-06-04
coming soon tweaks

rviscomi:
A bit of whitespace formatting and text tweaks. Also moved the GA JS into the document head per their best practices.

image


rviscomi:
Thanks @mikegeyser!

Issue #45, opened 2019-06-13
Create templates for content pages


rviscomi:
We don't have chapters written yet but now would be a good time to create the infrastructure for them so we can just drop in the content when they're ready.

I'm working with a designer to create the look and feel, which should be ready by July. Meanwhile we can get started on the backend routing and templating in preparation.

We should create the following templates:

- [x] rename the splash page at index.html to splash.html and update main.py
- [x] create a new index.html for the post-launch home page, stub things like navigation and introductory text
- [x] create /2019/outline.html where we show a table of contents for all chapters
- [x] create /2019/chapter.html as a template for each chapter
- [x] create /2019/methodology.html to provide technical info about our test methodology
- [x] create /2019/contributors.html to acknowledge everyone who contributed to the project

The Almanac is an annual report and I expect to do this again next year, so we should be proactive and separate the annual content from the static content. In this case, the home page is static to the project while things like table of contents, chapters, methodology, and contributors may change year to year. These should all be organized in a 2019 template directory.

The contributors page will be based on the data in [this sheet](https://docs.google.com/spreadsheets/d/1LR29JpI2XJSjan_z6uP1YpRy4EYYivYZPqCIG-9sfFY/view#gid=770177464). It might be a good idea to create a JSON metadata file of all contributors and generate this page dynamically. This way we can also dynamically add authors/reviewers on the bylines of their respective chapters.

Any volunteers from @HTTPArchive/developers  who can take part or all of this issue?

MSakamaki:
I part want to help.

Isn't it a problem to create a public web API use Google SpreadSheet script?

I made the next simple script. (I can not add this script to the original because I have no authority)
https://gist.github.com/MSakamaki/795b4c947fb0fdfc8e41d320e6e5ddc1

This points to my sheet where I copied the contributors page.
https://docs.google.com/spreadsheets/d/1DB0dhdgRiWfAeH5gTEd7nd3zsEuaNCMqp0gA1IGVzUA

When you hit the following API, you get json of name and hyperlink.
https://script.google.com/macros/s/AKfycbzuHDBRkaOlOi_iioNVKYQxVpQ4mZJy2V8xVS_l9FJM0eF3zhPq/exec

If there is no problem, we will start producing json API including the role.

anoblet:
I can take care of the scaffolding

rviscomi:
[...]

PR #47, opened 2019-06-14
Update scaffolding

anoblet:
Progress on #45

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -11,7 +11,7 @@ SUPPORTED_LANGS = ('en', 'ja')
 
 @app.route('/')
 def index():
-    return render_template('en/index.html')
+    return render_template('en/splash.html')
 
 @app.route('/<lang>/')
 def index_i18n(lang):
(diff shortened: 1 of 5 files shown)


src/main.py
```diff
@@ -18,7 +18,7 @@ def index_i18n(lang):
     if lang not in SUPPORTED_LANGS:
         abort(404)
 
-    return render_template('%s/index.html' % lang)
+    return render_template('%s/splash.html' % lang)
```
rviscomi: Need to also update the `index` method above.

src/templates/en/splash.html
```diff
@@ -0,0 +1,31 @@
+{% extends "en/base.html" %}
```
rviscomi: Also create a `ja/splash.html`

anoblet: Updated

src/main.py
anoblet: Updated

PR #49, opened 2019-06-17
add routing for scaffolding templates

rviscomi:
Progress on #45


rviscomi:
I'm torn about redirecting `/` to `/en` by default because it adds another layer of indirection to the initial page load. We could make it a permanent redirect so users only experience it once, but it still feels dirty.

> With the language-switcher it would just be a matter of taking the current path and swapping out the first part.

@mikegeyser's recent changes to unify the i18n-specific methods with their default counterparts would actually make this even easier. Flask enables us to [generate a URL](http://flask.pocoo.org/docs/1.0/api/#flask.url_for) for a given page and pass the desired lang attribute in.

anoblet:
Cool. Sounds good to me.

Issue #52, opened 2019-06-18
Consider users' Accept-Language preference when selecting the default language


MSakamaki:
When users first visit, I think that it will be a good experience for the user to be not conscious of language selection.

To do this, need to check the `Accept-Language` header sent by the client and route it to the appropriate language.
(Must be careful not to work with redirects after being selected by the switcher)

This needs to be addressed after #50.

rviscomi:
Yes! Looking at the [Accept-Language](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language) docs:

> The Accept-Language request HTTP header advertises which languages the client is able to understand, and which locale variant is preferred. (By languages, we mean natural languages, such as English, and not programming languages.) Using content negotiation, the server then selects one of the proposals, uses it and **informs the client of its choice with the Content-Language response header**. Browsers set adequate values for this header according to their user interface language and even if a user can change it, this happens rarely (and is frowned upon as it leads to fingerprinting).
>
> **This header is a hint to be used when the server has no way of determining the language via another way, like a specific URL**, that is controlled by an explicit user decision. It is recommended that the server never overrides an explicit decision. The content of the Accept-Language is often out of the control of the user (like when traveling and using an Internet Cafe in a different country); the user may also want to visit a page in another language than the locale of their user interface.

So we should only consider the `Accept-Language` request header for non `*_i18n` router methods, like the [`index`](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/main.py#L13) router method for the root page, `/`. If we recognize and support the language, we will return the contents in that language using the corresponding `*_i18n` method. If for example, someone has `Accept-Language: ja-JP`, we will recognize that the user prefers Japanese and route their `/` request to the `index_i18n` handler for `lang=ja`.

Alternatively, we might want to force a redirect to the `/ja/` URL. Otherwise I'm not sure what the caching behavior would be  with a single URL that can appear in multiple languages depending on the user's request headers. Open to other ideas and curious to hear if anyone has experienced this before.

Not quite sure what the docs are getting at regarding fingerprinting, but it sounds like we also want to return the [`Content-Language`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Language) response header for every `*_i18n` router method.

MSakamaki:
> So we should only consider the Accept-Language request header for non `*_i18n` router methods, like the `index` router method for the root page, `/`. 
[...]

PR #53, opened 2019-06-18
Combined routes for I18n

mikegeyser:
I've combined the routes for I18n using default parameters. I've also moved language to a hash map, and created a default for year. 

Finally, I created some of the templates for 'ja' so that I could test that the routing was working across languages as expected, I hope that's alright.

This is as a follow up to my suggestion in PR #49. If you disagree with the feedback, please feel free to disregard and close this. Also, submitting a PR for a PR seems a bit meta, but I wasn't sure how else to go about it. :)


src/main.py
```diff
@@ -1,92 +1,81 @@
 from flask import Flask, render_template, abort
 from flask_talisman import Talisman
 from csp import csp
+import logging
```
rviscomi: Unused import

src/main.py
```diff
@@ -1,92 +1,81 @@
...
-
-##
-## Internationalization (i18n)
-##
-
 @app.route('/<lang>/')
-def index_i18n(lang):
-    if lang not in SUPPORTED_LANGS:
-        abort(404)
+def index(lang=DEFAULT_LANG):
```
rviscomi: I _really_ like this approach to combine the implicit and explicit langs. One hesitation I have is that it obscures the use case @MSakamaki raised in #52. It'd be nice to overwrite the default implicit lang from English to the user's Accept-Language preference, if supported.

I think the solution would be to initialize `lang=None` and do the Accept-Languages check when that's the case. Otherwise, default to DEFAULT_LANG.

src/main.py
mikegeyser: Is it not used on line 81? Pycharm was complaining, which is why I assumed it was necessary, but it may just have been an overeager IDE warning. Happy to remove it though!

```
80:  def server_error(e):
81:      logging.exception('An error occurred during a request.')
```

src/main.py
mikegeyser: Got it! Should be easy enough to do, will give it a go in a few hours.

src/main.py
rviscomi: Oh you're right! Disregard.

src/main.py
rviscomi: Ok great. I'll go ahead and merge this as-is into the scaffolding branch. When you're ready you can make another PR for the Accept-Language change.

Issue #51, opened 2019-06-18
Generate the Contributors page


rviscomi:
The Contributors page will be generated based on a JSON file with all of the contributor metadata. Required info for all contributors:

- full name
- list of teams contributed

Not required but super nice to have info:

- avatar URL (Gravatar?)
- short personal tagline ("Engineer at Company", "Web dev thoughtleader", etc)
- GitHub profile
- Twitter profile

TODO: @HTTPArchive/developers 

- [x] contributors have verified their information is correct in the [Contributor sheet](https://docs.google.com/spreadsheets/d/1LR29JpI2XJSjan_z6uP1YpRy4EYYivYZPqCIG-9sfFY/edit#gid=770177464)
- [x] create a `src/config/contributors.json` file to organize all contributor metadata
- [ ] contributors have submitted a PR to update their metadata as needed
- [x] generate the `contributors.html` template based on the JSON metadata

rviscomi:
@HTTPArchive/almanac-contributors please take a look at the [Contributors sheet](https://docs.google.com/spreadsheets/d/1LR29JpI2XJSjan_z6uP1YpRy4EYYivYZPqCIG-9sfFY/edit#gid=770177464) and leave a comment in the sheet if any of your interests are inaccurate. If you're not listed at all, please let me know in this issue.

rviscomi:
@KJLarson this is a good first issue, do you want to give it a try? Happy to walk you through it.

KJLarson:
I would love to give this a try!

I'm going to take a look at stuff, and then I'm sure I will have a bunch of questions.

rviscomi:
Ok great!

KJLarson:
I think I just created a pull request with an empty contributors.json file. I want to make sure I'm doing the collaborative GitHub thing correctly (I haven't really used it for collaborative purposes yet, just personal projects). Did I do it right?

rviscomi:
Looks great! I left some general feedback in the PR.

Since it's just an empty file, let's start fleshing it out. The structure of the JSON object needs to contain basic info about each contributor and group contributors into teams. Here's one approach:

global object {}
- `teams` {}
  - `team_id` : team info {} (key : value)
    - name
    - members []
      - contributor_id
- `contributors` {}
  - `contributor_id` : contributor info {} (key : value)
    - name

[...]

PR #57, opened 2019-06-18
Create contributors.json

KJLarson:
Making sure I understand how to use Git/GitHub in a collaborative environment by submitting a pull request that adds an empty file (I've mostly just used GitHub for my own projects and practice repositories). That seems pretty safe. Am I doing this right?

Progress on #51


hyperpress:
Looks good, Kari.

KJLarson:
Thanks!

So, now do I keep doing it like this? Working on the add-contributors branch on my forked version of the repository and creating pull requests?

rviscomi:
Yep, your workflow is perfect: fork this repository, work on individual features in separate branches, file a PR to merge each branch with the upstream master. 🙌 

Only suggestion I can think of is to reference the issue you're addressing in the PR description. GitHub will add annotations linking them together to make it easier to find. For now just add `Progress on #57`. In the future if the issue should be closed after the PR is merged you can add `Fixes #57` and GitHub will do it automatically.

Let's leave this PR open and have you push a couple more commits to your branch to build out the feature. We'll coordinate in #57.

------

One general tip since you're not used to working collaboratively. Remember to pull from the upstream master before you file the PR, to merge in any changes others may have added. In this case everything looks great.

Check your remotes by entering `git remote -v` and you'll probably only see your fork labelled "origin". You can add the upstream repository as another remote by entering `git remote add upstream git@github.com:HTTPArchive/almanac.httparchive.org.git`. Now, whenever you want to merge changes from this repo with your local fork, you can just enter `git pull upstream master`. And you can keep your origin up to date with `git push`.

KJLarson:
I think I understand how to do the contributors.html file, but the python files are tripping me up! I'm probably totally off with that stuff, but I tried to work it out. I would love feedback and/or some hints (maybe something for me to research that would help me figure it out?). Thanks!

src/contributors.py
```diff
@@ -0,0 +1,13 @@
+import json
+
+contributors_json = {}
+
+
+def get_contributors():
```
rviscomi: We need two methods:
- one that updates the contributors_json variable based on the contents of the file
- one that returns the value of contributors_json

Rename this method to make it clearer that it's the first method. We also need to create the second method:

```suggestion
def get_contributors():
  return contributors_json

def update_contributors():
```

src/contributors.py
```diff
@@ -0,0 +1,13 @@
...
+  with open('config/contributors.json', 'r') as contributors_file:
+    contributors_json = json.load(contributors_file)
+# need to return data
+
+# Alphabatize data?
+
+get_contributors()
```
rviscomi:
```suggestion
update_contributors()
```

src/templates/en/2019/contributors.html
```diff
@@ -3,3 +3,16 @@
 {% block title %}
 Contributors | The Web Almanac by HTTP Archive
 {% endblock %}
+
+{% block main %}
+{% for contributor in contributors %}
+<section>
+  <h3>{{ contributor.name }}</h3>
+  <ul>
+    <li>{{ team.name }}</li>
```
rviscomi: We'll need another loop to render an `<li>` for each team that the contributor is a part of.
```suggestion
    {% for team in contributor.teams %}
    <li>{{ teams[team].name }}</li>
    {% endfor %}
```

rviscomi:
Thanks @KJLarson this is looking great! Very close to the finish line, just left a few comments and then it should be ready to go.

src/contributors.py
```diff
@@ -0,0 +1,13 @@
...
+
+contributors_json = {}
+
+
+def get_contributors():
+  with open('config/contributors.json', 'r') as contributors_file:
+    contributors_json = json.load(contributors_file)
+# need to return data
+
+# Alphabatize data?
```
rviscomi: Can remove these comments.

rviscomi:
@KJLarson let me know if you get stuck with any of this. I'm available to walk you through it over chat.

KJLarson:
Is there a way to view all the suggested changes for a file together rather than broken up into all these separate bits? I'm having a hard time making sense of all this and figuring out how it goes together. I'm sure several nights of poor quality sleep doesn't help (:-/)
Thanks!

rviscomi:
Try this view: https://github.com/HTTPArchive/almanac.httparchive.org/pull/57/files?utf8=%E2%9C%93&diff=split

KJLarson:
Another question...do you have any suggestions on how specific to be with commit messages? It seems it wouldn't be too helpful (in most cases) to just say "update filename." 

@rviscomi - I really appreciate your feedback, patience, and question answering!!! Thanks!

rviscomi:
> do you have any suggestions on how specific to be with commit messages? It seems it wouldn't be too helpful (in most cases) to just say "update filename."

Naming things is never easy 😁. For commits I like to have a short 1-liner describing what it is that I did. For more complex commits with many changes, I also include a list of each major change within the commit. For example:

```
add support for new widget

- optimizes widget performance on mobile
- removes redundant call to drawWidget in widget.js
- adds authentication to SuperWidget users
```

GitHub will name the commit based on the first line and add the rest of the commit message to the description. For small commits don't be afraid to be brief.

KJLarson:
I think I'm stuck on main.py and contributors.py. 

For contirbutors.py, I can't tell if it's done or if there are more changes needed.

For main.py, I can't tell if:
- the @app.route for contributors on line 28 is complete and I can delete the @app.route I added on line 53, OR
- if they need to be combined together, OR
- if the @app.route on line 31 isn't done, but the stuff I added isn't what it needs.

src/contributors.py
```diff
@@ -0,0 +1,16 @@
+import json
+
+contributors_json = {}
+
+
+def get_contributors():
+  return contributors_json
```
rviscomi: Oh, sorry we also need to declare the global property here too.

```suggestion
  global contributors_json
  return contributors_json
```

src/main.py
```diff
@@ -50,6 +51,17 @@ def chapter(year, chapter, lang):
...
 
+@app.route('/<year>/contributors')
+@app.route('/<lang>/<year>/contributors')
+reports = report_util.get_reports()
+
+  # TODO: Add any additional information about contributors to include
+  if get_format(request) == 'json':
+    return jsonify(name=contributors.name, teams=team.name)
+
+  return render_template('contributors.html', contributors=contributor_util.get_contributors())
```
rviscomi: Your changes on lines 54 through 64 can be deleted. What I was saying is that the router for the Contributors page is already defined earlier in this file on lines 29-34. The only change needed is to set the `contributors` parameter to the output from the util module on line 34:

```py
return render_template('contributors.html', contributors=contributor_util.get_contributors())
``` 

Which resolves this TODO on line 33, so it can be removed:

```py
# TODO: Get contributor data and pass into the template.
```

rviscomi:
Once these changes are finished, the only thing left to do is to test out the page by visiting http://127.0.0.1:8080/2019/contributors with the development server running. You should see the info for the 3 contributors in the config file.

KJLarson:
Well, shoot. I tested it and got this:
NameError: name 'contributor_util' is not defined

It looks like it's referring to line 33 in main.py.

src/main.py
```diff
@@ -29,8 +30,7 @@ def outline(year, lang):
 @app.route('/<lang>/<year>/contributors')
 @validate
 def contributors(year, lang):
-    # TODO: Get contributor data and pass into the template.
-    return render_template('%s/%s/contributors.html' % (lang, year), contributors={})
+    return render_template('%s/%s/contributors.html' % (lang, year), contributors=contributor_util.get_contributors())
```
rviscomi: Ah you have `contributor_util` but it's imported as `contributors_util`. Sorry if I suggested the wrong spelling!

KJLarson: No more error, but now it's just a blank page. I inspected it and the div with the id="main" is empty. Do we need to import the data? I don't see anything that refers to any of the other files.

KJLarson: In contributors.py, should "contributors_file" be "contributors_util"?

rviscomi: I just tested your PR locally and found the issue. I'll walk you through my debug process so you can see what I did and how I identified the issues.

First, I wanted to inspect what the value of the `contributors` object is as it's passed to the template. So I added this logging statement in `main.py`:

```py
def contributors(year, lang):
    contributors = contributors_util.get_contributors()
    logging.debug('Contributors: %r' % contributors)
    return render_template('%s/%s/contributors.html' % (lang, year), contributors=contributors)
```

Then after reloading the page I see this line in the terminal:

```
DEBUG:root:Contributors: {}
```

So the object is not populated with the contents of the JSON file. Looking in `contributors.py` I noticed that the invocation on line 17 is calling `get_contributors` instead of `update_contributors`.

After I fixed that I reloaded the page again and the page was still blank. The JSON was being logged as expected this time:

```
DEBUG:root:Contributors: {'teams': {'developers': {'name': 'Web Developers'}, 'brainstormers': {'name': 'Brainstormers'}, 'authors': {'name': 'Chapter Authors'}, 'reviewers': {'name': 'Peer Reviewers'}, 'analysts': {'name': 'Data Analysts'}, 'designers': {'name': 'Web Designers'}, 'translators': {'name': 'Content Translators'}, 'editors': {'name': 'Editors'}}, 'contributors': {'argyleink': {'name': 'Adam Argyle', 'github': 'https://github.com/argyleink', 'teams': ['brainstormers']}, 'kjlarson': {'name': 'Kari Larson', 'github': 'https://github.com/KJLarson', 'teams': ['developers']}, 'rviscomi': {'name': 'Rick Viscomi', 'github': 'https://github.com/rviscomi', 'teams': ['brainstormers', 'developers']}}}
```

But looking in the source HTML I see that two sections are being rendered with empty fields:

image

The name of the JSON object is `contributors` but it also has properties named `teams` and `contributors`. So it's rendering a section for each of these and not the individual contributors. I modified the logic in the template to this:

```html
{% for id, contributor in contributors.contributors.items() %}
<section>
  <h3>{{ contributor.name }}</h3>
  <ul>
    {% for team in contributor.teams %}
    <li>{{ contributors.teams[team].name }}</li>
    {% endfor %}
  </ul>
</section>
{% endfor %}
```

The major changes were:

- iterating over the `contributors` object within the variable named `contributors`
- calling `.items()` on the JSON object so we can access its key/value separately
- assigning the key/value pair `id, contributor` respectively
- referencing the `teams` object within the `contributors` variable

Now I'm seeing the page render as expected:

image

So go ahead and make the fixes to `contributors.py` and `contributors.html`. No need to commit the logging code, that's just for local debugging.

KJLarson: I'm trying to recreate your process, and when I try to call update_contributors on line 17, I get this error:
```
line 13, in update_contributors
    with open('config/contributors.json', 'r') as contributors_file:
FileNotFoundError: [Errno 2] No such file or directory: 'config/contributors.json'
```
I cannot figure out what is happening here. I double-checked the spelling and path a thousand times, and they seem right. Did you run into anything like this?

KJLarson: I added src/ in front of config/contributors.json and that seemed to work. That confuses me. I thought that when we use a relative path, we can just type a name that is on a sibling level in the directory tree. Doesn't this go up and then back down? Does it work differently here?

rviscomi: When we deploy the app to production, it takes everything under `src` and runs it on App Engine. So there will be no `src` directory.

When you're running it locally are you running `python src/main.py` or from the src directory are you running `python main.py`? The latter is the correct way. We should update our README.

src/contributors.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def get_contributors():
+  global contributors_json
+  return contributors_json
+
+def update_contributors():
+  global contributors_json
+  
+  with open('src/config/contributors.json', 'r') as contributors_file:
```
rviscomi:
```suggestion
  with open('config/contributors.json', 'r') as contributors_file:
```

src/main.py
KJLarson: Aha!

rviscomi:
image


Ok @KJLarson there comes a time in every open source contributor's life when they have to face the dreaded merge conflict. For you, today is that day! 😄

 #60 was just merged, which affects `main.py`. You've also made changes to that file, the same lines in fact, so now git doesn't know how to properly merge them after doing a `git pull git://github.com/HTTPArchive/almanac.httparchive.org.git master` (or similar, depending on how you set up your remote repos locally).

After syncing to master you'll see this error message in the terminal:

```
Auto-merging src/main.py
CONFLICT (content): Merge conflict in src/main.py
Automatic merge failed; fix conflicts and then commit the result.
```

If you open up `main.py` you'll see this:

```
def contributors(year, lang):
<<<<<<< HEAD
    # TODO: Get contributor data and pass into the template.
    return render_template('%s/%s/contributors.html' % (lang, year), contributors={}, supported_languages=supported_languages)
=======
    contributors=contributors_util.get_contributors()
    return render_template('%s/%s/contributors.html' % (lang, year), contributors=contributors)
>>>>>>> 4ec6c11f9ef25c73c71cff37f1d5e80a663cecd4
```

There are two sections to this, the part between HEAD and === (the latest contents of the file) and the part between === and >>> (your changes to the old file). To resolve the conflict, we need to merge the two sections while preserving old and new changes. Here's how I would merge them:

```py
def contributors(year, lang):
    contributors=contributors_util.get_contributors()
    return render_template('%s/%s/contributors.html' % (lang, year), contributors=contributors, supported_languages=supported_languages)
```

(Note that none of the <<< === >>> scaffolding should exist in the code after merging)

After saving that change, if you run `git status` you'll see this:

```
Unmerged paths:
  (use "git add <file>..." to mark resolution)

	both modified:   src/main.py
```

So run `git add src/main.py` and `git commit` which will generate a merge commit. You don't need to modify the commit message, it already contains info about what you're merging. After pushing, the conflict should be resolved and the PR will be ready to merge with the master branch.

src/main.py
```diff
@@ -30,8 +31,9 @@ def outline(year, lang):
 @app.route('/<lang>/<year>/contributors')
 @validate
 def contributors(year, lang):
-    # TODO: Get contributor data and pass into the template.
-    return render_template('%s/%s/contributors.html' % (lang, year), contributors={}, supported_languages=supported_languages)
+def contributors(year, lang):
```
rviscomi:
```suggestion
```

rviscomi: Accidentally copied this line. Should be removed.

KJLarson: Oops! Thanks!

src/main.py
```diff
@@ -31,9 +30,7 @@ def outline(year, lang):
 @app.route('/<lang>/<year>/contributors')
 @validate
 def contributors(year, lang):
-def contributors(year, lang):
-    contributors=contributors_util.get_contributors()
```
rviscomi: Whoops. This commit totally rolls back the progress on this issue. Can you add these changes back in and only delete the duplicate `contributors` function definition above?

KJLarson: 🤦‍♀️ Oh my, I think I'm getting confused. It's tricky looking at the changes and suggestions with the green and the red!
Is this what we want?:
```
def contributors(year, lang):
    contributors=contributors_util.get_contributors()
    return render_template('%s/%s/contributors.html' % (lang, year), contributors=contributors, supported_languages=supported_languages)
```

rviscomi: Yes, that's it! Also make sure to keep the "import contributors" bit above.

KJLarson: Where is the "import contributors" bit?

Thanks for your patience!

rviscomi: image

This part is required because the `contributors` method calls `contributors_util.get_contributors()`.

PR #58, opened 2019-06-19
Refactored the validating of year and language into a decorator

mikegeyser:
Refactored the i18n validation into a dedicated decorator that will respect the selection of language and year. It will also try to read the user language out of the 'Accept-Language' header as far
as possible. I've left a *lot* of debug logging in place, so that we can verify that it's doing the right thing. The only real compromise has been to accept the year as a parameter for the splash page, which I feel is reasonable.


src/validate_i18n.py
```diff
@@ -0,0 +1,77 @@
+import logging
+import re
+
+from flask import request
+from werkzeug.exceptions import abort
```
rviscomi: Was it a conscious decision not to use the abort method from the flask module?

src/validate_i18n.py
```diff
@@ -0,0 +1,77 @@
...
+from functools import wraps
+
+DEFAULT_YEAR = '2019'
+DEFAULT_LANG = 'en'
+SUPPORTED_YEARS = {
+    '2019': ('en', 'ja')
+}
+
+
+def validate_i18n(func):
```
rviscomi: Let's rename this module to a generic validator decoupled from i18n, since year is not necessarily tied to i18n. And soon we'll also have to validate chapters and maybe even contributors.

src/validate_i18n.py
```diff
@@ -0,0 +1,77 @@
...
+    logging.debug('Trying to extract Accept-Language header.')
+
+    header = request.headers.get('Accept-Language')
+    accepted_languages = re.findall('(?:^|\s|,)(\w+)', header)
+
+    logging.debug('Accepted languages: %s' % accepted_languages)
+
+    # The header could contain multiple languages, in order of precedence
+    for lang in accepted_languages:
+        if lang in supported_langs:
```
rviscomi: Does this take locales into consideration? For example, if the header value is `ja-JP` will that match the `ja` supported language?

src/validate_i18n.py
```diff
@@ -0,0 +1,77 @@
...
+
+    logging.debug('Using lang: "%s" and year: "%s" ' % (lang, year))
+
+    return {'lang': lang, 'year': year}
+
+
+def parse_accept_language(supported_langs):
+    # Try and extract the language out of the header. The regex below will pull the
+    # alpha characters out of the start of the string, after a comma, or after a space.
+    # It may not be exhaustive, and will require testing.
```
rviscomi: Would love to see tests for this :)

src/validate_i18n.py
```diff
@@ -0,0 +1,77 @@
...
+
+    if year not in SUPPORTED_YEARS:
+        logging.debug('Unsupported year requested: %s' % year)
+        abort(404)
+
+    supported_langs = SUPPORTED_YEARS[year]
+    logging.debug('Languages supported for %s: %s.' % (year, supported_langs))
+
+    # If an unsupported language code is passed in, abort.
+    # TODO: Should we still redirect back to default?
```
rviscomi: I don't think so. The request was bad so we should fail quickly.

src/validate_i18n.py
```diff
@@ -0,0 +1,77 @@
...
+    if lang is not None and lang not in supported_langs:
+        logging.debug('Unsupported language set: %s.' % lang)
+        abort(404)
```
rviscomi: The "unsupported language" error message would be helpful to display to the user.

src/validate_i18n.py
```diff
@@ -0,0 +1,77 @@
...
+        kwargs.update(updated_kwargs)
+
+        return func(*args, **kwargs)
+
+    return decorated_function
+
+
+def validate(lang, year):
+    if year is None:
+        logging.debug('Defaulting the year to: %s' % year)
```
rviscomi: When this is running on App Engine, do you know if these logs are available in Stackdriver or similar?

src/main.py
```diff
@@ -1,79 +1,53 @@
...
+logging.basicConfig(level=logging.DEBUG)
 
 
 @app.route('/')
 @app.route('/<lang>/')
-def index(lang=DEFAULT_LANG):
-    validate(lang)
-
+@validate_i18n
+def index(year, lang):
```
rviscomi: What happens if we remove the year from the parameters?

rviscomi:
Could you reply/resolve the comments from my last review? Easier to see what the outcomes were.

mikegeyser:
Of course, sorry!

src/validate_i18n.py
mikegeyser: No, it was just a mistake due to overeager IDE imports. Have changed it back.

src/validate_i18n.py
mikegeyser: It does. It ignores the locale as a part of the regex, so will match `ja-JP` as `ja`.

src/validate_i18n.py
mikegeyser: Done! (well, mostly) 

I've added tests for the `parse_accept_language` method only, because to test the decorator and other method I would have to mock out the http request. If we feel that it would be valuable, I can go ahead and figure out how best to accomplish that.

src/validate_i18n.py
mikegeyser: How would it be best to surface that error message? At the moment its a 404, should we have a custom 404 page that would show the user that they entered an unsupported language? This condition should only be hit if the user explicitly enters an incorrect language code into the URL. If it's a part of the `Accept-Language` header, it will default instead.

src/validate_i18n.py
mikegeyser: I am not 100% certain, but the docs lead me to believe so: https://cloud.google.com/logging/docs/setup/python

src/main.py
mikegeyser: This was the only compromise for the solution. If the year is not in the parameters, it will throw some kind of key exception. While this was a simplifying assumption (I thought all pages made sense to be versioned by year), if it doesn't hold then we can partially update `**kwargs` only with fields that were originally present.

src/validate_i18n.py
rviscomi: Yeah custom error pages would be great. For reference see https://github.com/HTTPArchive/httparchive.org/blob/86d844ebcaa11d85a9197b1d476b1cf178c513df/main.py#L235-L264

src/main.py
rviscomi: The home page is year-agnostic since it is the landing page for the current and all earlier editions, so it'd be nice if we can remove the param. I'm surprised that it's required because `args` should handle any input.

src/test_validate.py
```diff
@@ -0,0 +1,53 @@
+from validate import parse_accept_language
```
rviscomi: Nitpick: rename this to `validate_test.py` so it's always adjacent to the file it's testing when in sorted order.

src/validate_i18n.py
rviscomi: No need, this is great thanks!

src/test_validate.py
mikegeyser: Cool, done!

src/validate_i18n.py
mikegeyser: Ok cool, added TODOs in both places that we call abort(), and I'll address it in a follow-up PR.

src/requirements.txt
```diff
@@ -2,3 +2,5 @@ Flask==1.0.2
 Flask-Markdown==0.3
 flask-talisman
 gunicorn==19.9.0
+pytest
+pytest-watch
```
rviscomi: Could you add a section to src/README with instructions for running the tests?

src/main.py
rviscomi: This is my only lingering concern, otherwise this PR LGTM.

This PR is a big improvement to both code quality and functionality. Thanks so much for working on it!

mikegeyser: I used the `inspect` module to get the args that the route will accept, which will allow us to also optionally update chapter going forward.

```py
         accepted_args = inspect.getargspec(func).args

         lang, year = validate_lang_and_year(lang, year)

         if 'lang' in accepted_args:
            kwargs.update({'lang': lang})

         if 'year' in accepted_args:
            kwargs.update({'year': year})
```

It was the right change to make, thanks for insisting on it!

(Also: it's a pleasure. :) )

src/requirements.txt
mikegeyser: Done! I kept it really brief, I hope it's sufficient?

PR #60, opened 2019-06-23
Build a language switcher

### LINKED ISSUE #50: Build a language switcher
@HTTPArchive/developers 

Somewhere on the page we should have a dropdown field for users to change the language. The expected UX of the field is:

- the default/selected option is the current language
- clicking the field will open a switcher with the other supported languages displayed
- selecting any other language will reload the current page with the language specified in the URL

For example, if on `/` and the user selects Japanese, the new URL will be `/ja/`. Similarly, if on `/2019/outline` the new URL will be `/ja/2019/outline`. Switching to English will force the `/en` URL path prefix for simplicity.

In terms of implementation, it should be built into the `templates/base.html` template. We can look at the `{{ lang }}` property and set an option as `selected` if it matches. For accessibility, does this need to have anchor elements or are there equivalent ARIA attributes we can use?

Design questions:

- is there a common place for this UI to go? header/footer?
- should we show flags as the options as a way to convey the language in a way not dependent on translation? (need a text-based fallback anyway for accessibility)
- should the options be translated in the destination language?

tanhengyeow:
Fixes #50.

TODO/Discuss:
- [x] Update anchor tags' `href` attribute to handle prefix updates.
- [ ] Ensure options' fallback text are shown in the destination language.


tanhengyeow:
@rviscomi Ready for the initial review 😄 The flag icons look a little large but I'll be happy to update them after more design decisions have been set.

src/static/css/styles.css
```diff
@@ -39,17 +39,35 @@ a {
...
+    position: absolute;
+    cursor: pointer;
+}
 
+.language-switcher__dropdown {
+    display: none;
+    margin-top: 10px;
+}
+
+.language-switcher:hover .language-switcher__dropdown {
```
rviscomi: I have two concerns with displaying the dropdown on hover: there is no hover on mobile, and it's not discoverable to screen readers. I tried to switch languages using only my keyboard and it wasn't possible.

Could you try implementing this using native UI elements like `<select>` and `<option>`? This should address both of the usability concerns above.

src/templates/base.html
```diff
@@ -31,6 +31,12 @@
     <div id="navigation">
       {% block navigation %}{% endblock %}
     </div>
+    <div class="language-switcher">
+      {% block current_lang %}{% endblock %}
```
rviscomi: I'd love to see this generated dynamically if possible. We should know based on the current template what its current language is and we can pass in the `SUPPORTED_LANGUAGES` list to generate the alternate language options. This way as we add support for more languages, we don't need to update every page.

rviscomi:
@tanhengyeow let me know if you'd like to chat about my comments. Happy to explain in more detail.

src/static/css/styles.css
tanhengyeow: It would be difficult to include images inside native elements like `select` and `option` for the dropdown menu. Some references:
1. [putting-images-with-options-in-a-dropdown-list](https://stackoverflow.com/questions/4941004/putting-images-with-options-in-a-dropdown-list)
2. [how-to-add-small-thumbnails-to-option-elements-in-a-select-list/31697569](https://stackoverflow.com/questions/31697386/how-to-add-small-thumbnails-to-option-elements-in-a-select-list/31697569).

Emoji flags would be used in place of images.

src/static/js/language.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+const defaultOption = document.querySelector('.default-language');
+const alternateOptions = document.querySelectorAll('.alternate-language');
+
+// Generate default language option
+defaultOption.innerHTML = generateOptionText(defaultOption.value);
+
+// Generate alternate languages' options
+for (i = 0; i < alternateOptions.length; i++) {
```
rviscomi:
```suggestion
for (i = 0; i < languageOptions.length; i++) {
```

src/static/js/language.js
```diff
@@ -0,0 +1,42 @@
...
+  alternateOptions[i].innerHTML = generateOptionText(alternateOptions[i].value);
```
rviscomi:
```suggestion
  languageOptions[i].innerHTML = generateOptionText(languageOptions[i].value);
```

src/static/js/language.js
```diff
@@ -0,0 +1,42 @@
...
+  if (supportedLang.includes(langPrefix) && path[3] === '/') {
+    return path.replace(langPrefix, value);
+  }
+}
+
+const defaultOption = document.querySelector('.default-language');
+const alternateOptions = document.querySelectorAll('.alternate-language');
+
+// Generate default language option
+defaultOption.innerHTML = generateOptionText(defaultOption.value);
```
rviscomi: You can update all of the options at the same time, so this special case can be removed.

src/static/js/language.js
```diff
@@ -0,0 +1,42 @@
...
+
+// Generate alternate languages' options
+for (i = 0; i < alternateOptions.length; i++) {
+  alternateOptions[i].innerHTML = generateOptionText(alternateOptions[i].value);
+}
+
+function generateOptionText(language) {
+  switch(language) {
+    case 'en':
+      return '🇬🇧 English';
```
rviscomi: Small nit: our English language code/region in [en/base.html](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/templates/en/base.html) is explicitly set to `en-US` so could we show the US flag here?

I hate to make it so US-centric so I'm also open to an enhancement where we detect the user's locale and show the GB flag as needed. No need to do that here but something to keep in mind if we want.

src/static/js/language.js
```diff
@@ -0,0 +1,42 @@
...
+    case 'ja':
+      return '🇯🇵 Japanese';
```
rviscomi: We should localize for Japanese readers.

@MSakamaki can you provide the Japanese word for "Japanese" to show here?

src/templates/base.html
```diff
@@ -31,9 +31,26 @@
...
       {% block navigation %}{% endblock %}
     </div>
+    <select class="language-switcher">
+      <option
+        class="default-language"
+        selected="selected"
+        value={{ self.lang()[0:2] }}>
+      </option>
+      {% for language in supported_languages %}
+        {% if language != self.lang()[0:2] %}
```
rviscomi: I think we only support 2 character language codes, so the `[0:2]` slice shouldn't be necessary.
```suggestion
        {% if language != self.lang() %}
```

src/templates/base.html
```diff
@@ -31,9 +31,26 @@
     <div id="navigation">
       {% block navigation %}{% endblock %}
     </div>
+    <select class="language-switcher">
+      <option
+        class="default-language"
```
rviscomi: Can remove the classes on the options.

src/static/css/styles.css
```diff
@@ -27,7 +27,7 @@ h1 {
     line-height: 1.1;
 }
 
-a {
+section a {
```
rviscomi: Are these changes still needed? (here and on line 42)

src/static/css/styles.css
```diff
@@ -39,12 +39,24 @@ a {
...
 }
 
-a:hover {
+section a:hover {
     background-color: #fff;
     border-color: #667a7d;
     color: #667a7d;
 }
 
+select {
```
rviscomi: Is this style applicable to all `select` elements or only the language switcher? If the latter, I'd suggest merging with the style rule below.

src/static/js/language.js
```diff
@@ -0,0 +1,42 @@
...
+const langSwitcher = document.querySelector('.language-switcher');
+
+langSwitcher.addEventListener('change', (e) => {
+  if (e.target.value) {
+    window.location.href =
+      updatePath(window.location.pathname, e.target.value);
+  }
+});
+
+function updatePath(path, value) {
```
rviscomi: Does this only work for the home page, or does it also support subdirectories? For example, this URL: `/2019/outline`.

Ideally we would use the Jinja [url_for](http://flask.pocoo.org/docs/1.0/api/#flask.url_for) method to construct the new URLs dynamically in the base template.

src/templates/base.html
tanhengyeow: @rviscomi By using `self.lang()` we are reading the value that is passed into this [block](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/templates/base.html#L2). 

Both `base.html` files of [en](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/templates/en/base.html#L3) and [ja](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/templates/ja/base.html#L3) are not in 2 character language codes, thus I included the slice.

rviscomi: Oh I see, thanks for clarifying.

src/templates/base.html
```diff
@@ -31,9 +31,30 @@
     <div id="navigation">
       {% block navigation %}{% endblock %}
     </div>
+    <select class="language-switcher">
+      <option
+        selected="selected"
+        {% block default %}
+        value={{ url_for(request.endpoint, lang=self.lang()[0:2]) }}
```
tanhengyeow: I've looked into [url_for](http://flask.pocoo.org/docs/1.0/api/#flask.url_for) and this is one possible way to generate dynamic URLs.

However, pages below take in more variable arguments:
1. `/<lang>/<year>/outline`
2. `/<lang>/<year>/contributors`
3. `/<lang>/<year>/methodology` 
4. `/<lang>/<year>/<chapter>`

There might also be other new pages that take in more/different variable arguments, thus this block has to be repeatedly customized for the different types of pages. 

Not sure if this is the best approach, what do you think?

rviscomi: This looks super weird but I think we can build the arguments dynamically by using the `request.view_args` dict and updating it with the appropriate `lang` value:

```py
value={{ url_for(request.endpoint, **(request.view_args.update(lang=language) or request.view_args)) }}
```

Because the `update` method updates the dict in-place and returns `None` I added the `or request.view_args` to ensure the dict is passed into `url_for`.

Hope that makes sense.

src/main.py
```diff
@@ -10,12 +10,12 @@
...
 logging.basicConfig(level=logging.DEBUG)
 
-
 @app.route('/')
 @app.route('/<lang>/')
 @validate
 def index(lang):
-    return render_template('%s/splash.html' % lang)
+    supported_languages = SUPPORTED_YEARS.get('2019')
+    return render_template('%s/splash.html' % lang, supported_languages=supported_languages)
```
rviscomi: `supported_languages` must be passed into all templates that include the language switcher.

src/static/js/language.js
```diff
@@ -0,0 +1,20 @@
+const languageSelect = document.querySelector('.language-switcher');
+languageSelect.addEventListener('change', (e) => {
+  if (e.target.value) {
```
rviscomi: Does the change event ever fire when clicking on the default/selected option? If so we should add a check for that here because we don't want to reload the page and not change the current language.

src/static/js/language.js
rviscomi: Per my research for #66 let's use `日本語` for the text here.

src/static/js/language.js
tanhengyeow: I've checked and the change event only fires when a different option is selected, so we should be good here :)

src/templates/base.html
tanhengyeow: Yup that makes sense! Thanks for the explanation 😄

Issue #62, opened 2019-06-25
Write queries and add to the repo


rviscomi:
When the Analyst team generates queries for each metric, they should create a PR to merge it into the repo. This has two benefits: the PR process provides an opportunity for peer review, and it is a place to share and maintain the canonical queries. On the Almanac website we can link directly to the queries from each respective chapter/figure so readers can see exactly how it was calculated and fork it for their own analysis.

- [x] create a new directory system to organize queries (@KJLarson)
- [x] test queries (analysts)
- [x] file a PR to merge the queries into their respective directory (analysts)

For testing queries, you can query the new [almanac](https://console.cloud.google.com/bigquery?p=httparchive&d=almanac&page=dataset) dataset, which contains desktop/mobile sample tables for 1,000 websites. This smaller dataset should help you refine your queries without incurring the full cost for all ~5M websites.

Query guidelines:

- must specify `#standardSQL` on the first line and use [Standard SQL](https://cloud.google.com/bigquery/docs/reference/standard-sql/)
- must include a short description of the metric it's analyzing, eg:
```sql
# Percentage of requests that are third party requests
# broken down by third party category by resource type.
```
- must query the 2019_07_01 dataset (unless otherwise needed)
- must be reasonably optimized where possible
- file must be named according to its metric ID, eg `05_03.sql`
- file must be placed in the directory according to its chapter, eg `05_ThirdParties/05_03.sql`

rviscomi:
@HTTPArchive/developers any takers for the first task? We need to figure out a good home for the queries and create the directory structure. 

I think it should be named according to this pattern `/sql/2019/05/05.3.sql`, but I'm open to suggestions. Some open questions:

- should the chapter directories be named numerically or by their title, eg `05` vs `Third Parties`?
- should the `chapter.metric.sql` format use a `.` delimiter between chapter and metric, or does that interfere/confuse at all with the file extension?
- should the file names also include the chapter number, eg `05.3.sql`, or omit the (redundant) chapter, eg `3.sql`?

rviscomi:
@KJLarson FYI this is another good first issue (task 1 of 3, creating directory system)

KJLarson:
Oooh, oooh! I can do this! I've been doing records management for the last couple years and have a Masters of Library and Information Science...this is totally in my realm!

rviscomi:
Sold! Thanks @KJLarson!

[...]

PR #65, opened 2019-06-26
Create directory structure for metric queries

KJLarson:
Progress on #62


src/config/contributors.json
```diff
@@ -0,0 +1,52 @@
+{
```
rviscomi: This PR should be limited to the SQL directories only. Changes related to the contributors PR should be removed.

rviscomi:
It's also good practice to include a short note in the top comment about the issue this PR is addressing. For example, in this case `Progress on #62`. Or for example if you're resolving an issue you can write `Fixes #62`. Could you edit your first comment with a note like this?

src/config/contributors.json
KJLarson: Oh, yes. I was having some issues switching between branches, and I think I did some stuff in the wrong order. I think I figured out what I need to do for the future. As soon as I figure out how to remove the changes from the contributors PR, I will do so!

rviscomi: You should be able to straight up delete the files from within this branch and make a new commit that logs those deletions. They'll still exist in the other branch.

I think what might have happened was that you branched from that branch, which included all of its changes. Tip: I always sync/branch from master to ensure a clean branch.

rviscomi: `contributors.json` is still showing up in the PR. Can you delete it? Then this should be good to go.

PR #69, opened 2019-07-01
Language class

### LINKED ISSUE #66: Create a Language class
We have lots of language-related configs/logic all over the place and it'd be nice to define this all in a single place using a Language class. Some instance properties of the class should be:

- 2 char language code (eg `ja`)
- emoji flag (eg 🇯🇵)
- native name (eg `日本語` -- just guessing based on Google translate, we'll need the Translation team to help with these)
- lang/region code (eg `ja-JP`)

rviscomi:
Closes #66

### DIFF
--- a/src/language.py
+++ b/src/language.py
@@ -1,0 +1,39 @@
+# coding=utf-8
+class _Language(object):
+  def __init__(self, local_name, lang_code, region_code, emoji_flag):
+    self._local_name = local_name
+    self._lang_code = lang_code
+    self._region_code = region_code
+    self._emoji_flag = emoji_flag
+
+  def __eq__(self, other):
+    if isinstance(other, _Language):
+      return '%r' % self == '%r' % other
+    return False
+
+  def __str__(self):
+    return '%s %s' % (self._emoji_flag, self._local_name)
+
+  def __repr__(self):
+    return '%s, %s' % (self.__str__(), self.lang_attribute)
+
+  @property
+  def lang_attribute(self):
+    # Returns the language attribute (eg "en-US") to be used in HTML.
+    return '%s-%s' % (self._lang_code, self._region_code)
+
+  @property
+  def lang_code(self):
+    return self._lang_code
+
+class Language(object):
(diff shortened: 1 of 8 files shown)


src/language.py
```diff
@@ -0,0 +1,39 @@
+# coding=utf-8
+class _Language(object):
+  def __init__(self, local_name, lang_code, region_code, emoji_flag):
+    self._local_name = local_name
+    self._lang_code = lang_code
+    self._region_code = region_code
```
AymenLoukil: Why do we handle region code ? For countries flags ? 
If it is, language switcher best practices don't recommend displaying countries flags when targeting generic languages.
Example : English speaking people outside US will be confused to see the US flag in the language selector.

rviscomi: Because the content is written in American-style English. For example we would spell "color" rather than "colour".

rviscomi: In any case, I'm happy to follow up with UX improvements to the language switcher if we're not adhering to best practices. Feel free to file a new issue. cc @tanhengyeow

AymenLoukil: The lang in hreflang (or in html lang) is about to say which population am i targeting with this page : 
- Language speaking (en, fr, ja, es...etc) ? 
- Language speaking who are based somewhere (en-us, en-uk, fr-ca).
Search engines use these informations (Google uses hreflang, Bing uses html lang) to provide the page to users based on ths targeting. In our case i believe we should keep on generic language targeting.

rviscomi: Gotcha, thanks for explaining. Could you file a new issue so we can discuss those changes? For now, it shouldn't affect this PR, which is mostly just code organization.

src/templates/base.html
```diff
@@ -33,18 +37,12 @@
     </div>
     {% set view_args = request.view_args.copy() %}
```
AymenLoukil: the declaration in line 28 is sufficient ?

rviscomi: Resolved this by pulling out the view_args logic into a Python function that we can call from the templates.

AymenLoukil: Great !

PR #74, opened 2019-07-03
Added hreflang alternate tags

### LINKED ISSUE #61: Add alternate/hreflang link tags for translations
Per https://support.google.com/webmasters/answer/189077 we should make alternate translations discoverable with `<link rel="alternate" hreflang="..." href="...">` elements.

These should be available on every page for all supported translations.

This is basically the metadata/SEO equivalent of #50, so there may be some opportunities to reuse code. cc @tanhengyeow

AymenLoukil:
I checked out your branch "lang". Please tell me if i should wait your PR merge :) 

I loop over the available languages and add the link alternate tags to the head (on base.html). Sounds working 

Fixes #61

### DIFF
--- a/src/templates/base.html
+++ b/src/templates/base.html
@@ -25,6 +25,10 @@
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,600,700,800">
     <link rel="stylesheet" href="/static/css/styles.css">
     {% endblock %}
+		{% set view_args = request.view_args.copy() %}
+	{% for l in supported_languages %}
+	<link rel="alternate" href="{{ url_for(request.endpoint, **(view_args.update(lang=l.lang_code) or view_args)) }}" hreflang="{{ l.lang_code}}" />
+	      {% endfor %}
   </head>
 
   <body>
--- a/src/templates/base.html
+++ b/src/templates/base.html
@@ -25,10 +25,10 @@
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,600,700,800">
     <link rel="stylesheet" href="/static/css/styles.css">
     {% endblock %}
-		{% set view_args = request.view_args.copy() %}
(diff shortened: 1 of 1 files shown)


src/templates/base.html
```diff
@@ -25,6 +25,10 @@
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,600,700,800">
     <link rel="stylesheet" href="/static/css/styles.css">
     {% endblock %}
+		{% set view_args = request.view_args.copy() %}
+	{% for l in supported_languages %}
```
rviscomi: This includes the current language. Is that ok for alternate hreflangs?

src/templates/base.html
```diff
@@ -25,6 +25,10 @@
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,600,700,800">
     <link rel="stylesheet" href="/static/css/styles.css">
     {% endblock %}
+		{% set view_args = request.view_args.copy() %}
```
rviscomi: The indentation here is inconsistent. Could you match the rest of the file?

src/templates/base.html
AymenLoukil: Yes. The rules are : 
- You self reference (so if we are in /en/ we should point to /en/
- You reference tha alternate pages (so here we point to /ja/
- Bidirectional (A point to B, B should point to A)

src/templates/base.html
AymenLoukil: OK i'll correct.

rviscomi: Looks like they're one nested level too many. Everything should be aligned with the blocks/links before it.

AymenLoukil: Already corrected in the new commit no ?

rviscomi: See screenshot below. New changes are too far indented:

image

AymenLoukil: Thanks. I think it ok now :)

PR #77, opened 2019-07-06
Build custom error pages

### LINKED ISSUE #68: Build custom error pages
@HTTPArchive/developers 

If an unexpected error occurs on the server (5xx) or there's an issue with user input (4xx) we should show an error page with more information about what went wrong.

tanhengyeow:
Fixes #68.

The cases below are included:
1. Status code 400 - Bad Request
2. Status code 404 - Page Not Found
3. Status code 500 - Internal Server Error
4. Status code 502 - Bad Gateway

### DIFF
--- a/src/templates/base.html
+++ b/src/templates/base.html
@@ -25,9 +25,11 @@
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,600,700,800">
     <link rel="stylesheet" href="/static/css/styles.css">
     {% endblock %}
-    {% for l in supported_languages %}
-    <link rel="alternate" href="{{ url_for(request.endpoint, **get_view_args(lang=l.lang_code)) }}" hreflang="{{ l.lang_code}}" />
-    {% endfor %}
+    {% if supported_languages is iterable %}
+      {% for l in supported_languages %}
+      <link rel="alternate" href="{{ url_for(request.endpoint, **get_view_args(lang=l.lang_code)) }}" hreflang="{{ l.lang_code}}" />
+      {% endfor %}
+    {% endif %}
   </head>
 
   <body>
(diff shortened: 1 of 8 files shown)


tanhengyeow:
> Thanks for working on this @tanhengyeow!
> 
> How might we implement this TODO for which we would pass a custom error message to the error page:
> 
> https://github.com/HTTPArchive/almanac.httparchive.org/blob/49fdcf814f6f078c8cdab0e5797c3e6f923993a1/src/validate.py#L41-L45

@rviscomi Adding the error description in this [line](https://github.com/HTTPArchive/almanac.httparchive.org/pull/77/files#diff-fab5b7e3d7ff2dcbe82a5a0cf622fccbR43) does the trick 😄 

This is an example of the custom error page when the user accesses an unsupported year:
image

src/templates/base.html
```diff
@@ -28,28 +28,30 @@
...
-    </div>
+    {% block content %}
+      <div id="navigation">
+        {% block navigation %}{% endblock %}
+      </div>
+      {% set view_args = request.view_args.copy() %}
+      <select class="language-switcher">
+        <option
+          selected="selected"
+          value={{ url_for(request.endpoint, **(view_args.update(lang=self.lang()[0:2]) or view_args)) }}
```
rviscomi: FYI just merged #69 so this is outdated now.

rviscomi: Otherwise this looks good!

rviscomi:
Looks good. Curious why you made the `if supported_languages is iterable` change. Should always be a list or tuple, right?

tanhengyeow:
> Looks good. Curious why you made the `if supported_languages is iterable` change. Should always be a list or tuple, right?

So when an unsupported language is requested by the user, this [line](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/main.py#L17) returns a non-list/tuple, resulting in a `_Language object not iterable` error. That's the reason why I added that check 😄

tanhengyeow:
> > Looks good. Curious why you made the `if supported_languages is iterable` change. Should always be a list or tuple, right?
> 
> So when an unsupported language is requested by the user, this [line](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/main.py#L17) returns a non-list/tuple, resulting in a `_Language object not iterable` error. That's the reason why I added that check 😄

@rviscomi Come to think of it, should we do this in the backend instead? For e.g. return an empty list/tuple by default if we can't find the supported year. That way we can be sure that it is iterable on the frontend. But even so, we still have to check if `supported_languages` is empty because we do not want to trigger the for loop if `supported_languages` is an empty list/tuple.

rviscomi:
Yes I agree it should be addressed on the backend but I didn't want it to hold up this PR.

Issue #70, opened 2019-07-01
Create a chapter config


rviscomi:
When rendering a chapter page, we need to validate that the chapter exists (for the given year) and render metadata about that chapter including:

- part number (eg `IV`)
- chapter number (eg `20`)
- title (eg `HTTP/2`)
- author(s) (eg `[bazzadp]`)
- reviewer(s) (eg `[bagder, rmarx, dotjs]`)
- _description_
- _hero image_

A config object should be created that lists all possible chapters for a given year. The object should define all of the metadata listed above.

Authors and reviewers must be referenced by their contributor IDs, which are defined in [contributors.json](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/config/contributors.json).

- [ ] ~~create `src/config/2019/chapters.json`~~ create `src/config/2019.json` and combine chapter + contributor + team JSON into one file
- [ ] create a Python utility to load the config into memory and map contributor IDs to their metadata objects
- [ ] validate requests handled by the chapter router
- [ ] render metadata in the chapter template
- [ ] generate the table of contents in the outline template

rviscomi:
@HTTPArchive/translators What do you think about this approach to generate the chapter metadata and Outline page based on the JSON config? I'm not sure how easy it is to translate using that process. Alternatively, we could create HTML files for every chapter and hard-code the metadata. Translating the HTML file would be straightforward and that would also take care of validation. The advantage to generating based on metadata is that we'd have one template for all chapters.

c-torres:
I think the benefits of having _a single source of truth_ are better than focus on easing the translation process. In my personal opinion I like the proposed approach.

mikegeyser:
If we manage to land #59, we could put *some* of this metadata (such as author/reviewer) in the metadata section of the markdown file. Is that something that could be interesting?

rviscomi:
@KJLarson curious if you have any opinions about the organization of chapter URLs. A chapter page has the structure `/2019/<chapter_id>`. Should `chapter_id` be the numeric sort order, eg `/2019/1` for the first chapter? Or a short identifier, eg `/2019/js`? Or the full title of the chapter, eg `/2019/JavaScript`?

KJLarson:
[...]

Issue #67, opened 2019-06-30
Fill data viz placeholders


rviscomi:
For each metric, its results will be queried and saved as JSON. We need to build a tool that generates data visualizations from the JSON data into SVG. The type of visualization will depend on the metric and the context in which it's displayed in the chapter.

For now this is a placeholder metric to brainstorm ideas and discuss what the requirements of the tool may be. TODO: 

- types of visualizations needed
- mobile/desktop requirements
- interactivity requirements

mikegeyser:
Would it be possible to get a sample JSON file and visualisation, so that we can try and work on  plumbing it into the authoring?

rviscomi:
Sure, I've added the query, JSON, and a link to an example visualization for the 01_01 metric: https://gist.github.com/rviscomi/e50c893e3d045853215045e9dc62e47d

rviscomi:
We've got several open PRs to add queries for the metrics and one thing I wasn't expecting is that @HTTPArchive/data-analysts are returning multiple values in a single query. This makes sense because there are many ways to represent the results, for example a full histogram in addition to significant percentiles (25/50/75).

On the development side, I think we should plan for this when handling the JSON data. For example, in the markdown an author could annotate a placeholder for a histogram visualization:

```
getHistogram('04_18', 'distribution')
```

Where the metric ID is `04_08` (chapter 4, metric 18) and the field name in the JSON is `distribution` containing data formatted in a way that can be visualized as a histogram (array of x/y values).

If the metric JSON also has summary stats, those could also be extracted using similar syntax:

```
showBigNumber('04_18', 'p50')
```

This would find the field `p50` (the median) and render it in large text.

patrickhulce:
Oh yes big plus one there! There are also several circumstances in third-party where there's a multi-tiered breakdown and I would need the ability to pull out multiple percentiles from the same metric. I'll put up the PR later today so you can get a sense for it, but it might add another layer of complexity to the situations you've already mentioned.

rviscomi:
SG thanks for flagging @patrickhulce. However we implement the data rendering, it would need to be flexible for many data types and result schemas.

rviscomi:
Related part of the pipeline: #177

rviscomi:
Now that the analysis phase is just about done (:raised_hands:) , it's a good time to flesh out the data viz process a bit more. I have a few thoughts I'd just like to regurgitate for the record and gather feedback :)

### Consistency over simplicity
[...]

Issue #59, opened 2019-06-20
Authoring in markdown


mikegeyser:
There was some earlier discussion about allowing for authoring the chapter content using markdown. Would this still be something worth investigating as a spike task?

The benefits are that it seems easier to write in markdown than html, and that gitlocalize seems to have better support for markdown.

The challenges are that we will likely want some rich data visualisation in the chapters, which may not have be simple to achieve using markdown.

mikegeyser:
A suggestion that I have, that I'd like to table is using a potentially hybrid approach. Would it be feasible to us custom-elements inside the markdown, and then put the datavis content inside the custom element using slots? 

Example:

```
# Chapter Eleventeen

Some meaningful textual content.

<pie-chart data="data.json">
  <data value="25">Part of the pie chart</data>
 <data value="75">The rest of the pie chart</data>
</ pie-chart>

```

That way we have content that can be indexed and crawled, can fall back to a div if custom-elements aren't supported, but lets us write some more interactive visualisations using javascript as needed.

It's just an idea, not bound to it. :)

rviscomi:
Thanks for kicking this off @mikegeyser!

> lets us write some more interactive visualisations using javascript as needed

Does this mean that all visualizations are effectively client-side rendered? I think it'd be optimal to generate the SVG charts statically and embed that in the HTML. AFAICT embedding raw HTML in the markdown is what your proposal already is, so is that all we need to do?

mikegeyser:
> is that all we need to do?
Absolutely! That's what I'm suggesting. If something is fine as a static SVG element, then we should just put it in the markdown directly. 

I'm not suggesting that we client side render everything. The custom element would just be *if* (and only if) there was a compelling UX requirement to have the visualisation be interactive.

In that case, in my opinion, the ideal would be to have the slotted item contain the SVG and then progressively enhance it with javascript. If there's no requirement for interactivity, then the whole custom-element argument is irrelevant.

```
<drill-down-pie-chart data="data.json">
  <svg>
     <!-- default pie chart -->
  </svg>
</drill-down-pie-chart>
```

(Apologies for the contrived example.)

mikegeyser:
Also, if we're worried about embedding large SVGs in the markdown file, we could possibly encourage people to reference them using `<object data=""></object>`?

rviscomi:
`<drill-down-pie-chart data="data.json">`

[...]

PR #80, opened 2019-07-15
Authoring in Markdown

mikegeyser:
I've added a simplistic implementation of authoring in markdown to the almanac (#59), so that we can validate whether this is a direction we would like to continue down. I've included [mistune](https://github.com/lepture/mistune) rather than [flask-markdown](https://github.com/dcolish/flask-markdown) because it seemed more popular, more actively developed, and had a greater focus on performance. There are a couple of areas of concern that I have, which I'll highlight in line.

Depending on the feedback, the next steps are probably to:
- [ ] Move the markdown files out of the `templates` directory,
- [ ] Include chapter configuration (as highlighted in #70), 
- [ ] Validate the chapter against the configuration,
- [ ] Include the appropriate head meta tags,
- [ ] Generate the SVG images (#67, might be better as a separate PR).


src/chapters.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def validate_chapter(lang, year, chapter):
+    # TODO
+    return True
+
+
+# TODO: Should this be cached in memory to prevent having to repeatedly
+# open and read the file? It shouldn't lock, I'm just not sure how well
+# this will perform in python.
```
mikegeyser: As mentioned in the comment, I am nervous of the performance of this. It may be nothing, I'm just not very experienced with python in production. This shouldn't lock if concurrent threads are reading files, and I suppose it's a tradeoff between CPU (opening and reading) versus Memory (keeping the entire contents of the file in memory) when it's all running on app engine.

src/templates/ja/2019/chapters/1.md
```diff
@@ -0,0 +1,5 @@
+# これはテストです
```
mikegeyser: Please forgive the google-translate-generated text.

src/templates/en/2019/chapter.html
```diff
@@ -3,3 +3,7 @@
 {% block title %}
 Chapter {{ chapter }} | The Web Almanac by HTTP Archive
 {% endblock %}
+
+{% block main %}
```
mikegeyser: If we follow the current directory structure, this will need the templates to be duplicated across every language we support - which I don't think is necessarily a good idea. I think we should have templates that are language agnostic, and content that is translated and put in the template separately. I have a suspicion that's part of what #70 is addressing?

rviscomi:
Closing this per the new direction taken in https://github.com/HTTPArchive/almanac.httparchive.org/issues/59#issuecomment-511553852

Issue #87, opened 2019-07-23
Query metrics: Chapter 6. Fonts


rviscomi:
Part | Chapter | Authors | Reviewers | Tracking Issue
-- | -- | -- | -- | --
I. Page Content | 6. Fonts | @davelab6 @zachleat | @hyperpress @AymenLoukil | #7

# READ ME!
All of the metrics in the table below have been marked as `Able To Query` during the [metrics triage](https://github.com/HTTPArchive/almanac.httparchive.org/issues/33). The analyst assigned to each metric is expected to write the corresponding query and submit a PR to [have it reviewed and added to the repo](https://github.com/HTTPArchive/almanac.httparchive.org/issues/62).

In order to stay on schedule and have the data ready for authors, **please have all metrics reviewed and merged by August 5**.

## Assignments

ID | Metric description | Analyst | Notes
-- | -- | -- | --
06.01 | Local vs hosted | @tjmonsi |  
06.02 | Popular hosts | @tjmonsi |  
06.03 | Font formats | @tjmonsi |  
06.04 | Font-Display usage | @tjmonsi |  
06.09 | how many fonts are loaded but also how many type-faces (families) are used | @tjmonsi |  
06.10 | Related, group by weight/style: how many people use italics? those are often left off | @tjmonsi |  
06.11 | Font formats (how many people are still using Bulletproof font face syntax?; WOFF2 use specifically) | @tjmonsi | https://www.paulirish.com/2009/bulletproof-font-face-implementation-syntax/
06.13 | CSS Font Loading API use? | @tjmonsi |  
06.14 | unicode-range use (and range size, perhaps to glean some info on subsetting) | @tjmonsi |  
06.15 | uses preconnect for web font cdn? popular preconnect domains? | @tjmonsi | use of the custom metric written by ymschaap which extracts all `link` elements in the document head and enumerates all of their attributes. We can query for the preconnects and aggregate based on the href.
06.16 | Use of local() in src | @tjmonsi |  
06.18 | how many pages in the HTTPArchive link to a variable font via `@font-face`? | @tjmonsi |  
06.19 | what percent of total pages use VFs? | @tjmonsi |  
06.21 | of those pages linking to a VF, how many are using the 4 font selectors that select on a variable font family? | @tjmonsi |  
06.23 | how many pages link to a VF, but never use it beyond old CSS3 values? | @tjmonsi | Is a VF without VF values even a VF?
[...]

PR #105, opened 2019-07-29
Fonts queries

tjmonsi:
RE: #87 

What I have done for now is
06.01
06.02
06.03
06.04
06.09
06.13
06.16

All of the variable fonts were triaged by @rviscomi from last month so I hope I can get it so I can put in 06.18 - 06.30

I don't know yet where to find @ymschaap custom metric for 06.15 to get "preconnect for web font cdn" 

Here is the note
>> use of the custom metric written by @ymschaap which extracts all `link` elements in the document head and enumerates all of their attributes. We can query for the preconnects and aggregate based on the href.

I'll be adding unto this branch some edits and will remove the `[WIP]` sign at the header once approved for merging


sql/2019/06_Fonts/06_03.sql
```diff
@@ -0,0 +1,43 @@
...
+  summary_requests AS (
+    SELECT * FROM `httparchive.summary_requests.2019_07_01_desktop`
+    UNION ALL
+    SELECT * FROM `httparchive.summary_requests.2019_07_01_mobile`
+  ),
+  processedTable AS (
+    SELECT
+      (
+        IF(STRPOS(mimeType, "font") > 0,
+          -- get the type of font from the content-type
```
rviscomi: Ahh! We actually have an `ext` field for the file extension which is probably a good enough heuristic for the font format.

```sql
#standardSQL
SELECT
  DISTINCT LOWER(ext),
  COUNT(0) OVER (PARTITION BY LOWER(ext)) / COUNT(0) OVER () AS pct
FROM
  `httparchive.summary_requests.2019_07_01_*`
WHERE
  type = 'font'
ORDER BY
  pct DESC
```

Your logic would be good to include in the `format` field function: https://github.com/HTTPArchive/legacy.httparchive.org/blob/31a25b9064a365d746d4811a1d6dda516c0e4985/bulktest/batch_lib.inc#L994

sql/2019/06_Fonts/06_03.sql
```diff
@@ -0,0 +1,43 @@
...
+            )
+          )[OFFSET(0)]
+        )
+      ) AS font,
+      *
+    FROM summary_requests WHERE type = 'font'
+  )
+SELECT
+  DISTINCT(font),
+  COUNT(font) AS count
```
rviscomi: Again, prefer relative percents rather than absolute counts. This will help authors put the significance of the results in context.

sql/2019/06_Fonts/06_04.sql
```diff
@@ -0,0 +1,25 @@
+#standardSQL
+
+-- counts font-display value usage
+
+WITH
+  response_bodies AS (
```
rviscomi: Rather than inspecting the response bodies, let's rely on the Lighthouse "font-display" audit.

This is used on our website: https://httparchive.org/reports/state-of-the-web#fontDisplay

Here's the underlying timeseries query: https://github.com/HTTPArchive/bigquery/blob/master/sql/timeseries/fontDisplay.sql

So for 2019_07_01 it'd be:

```sql
#standardSQL
SELECT
  ROUND(SUM(IF(JSON_EXTRACT(report, '$.audits.font-display.score') IN ('true', '1'), 1, 0)) * 100 / COUNT(0), 2) AS percent
FROM
  `httparchive.lighthouse.2019_07_01_mobile`
WHERE
  report IS NOT NULL AND
  JSON_EXTRACT(report, '$.audits.font-display.score') IS NOT NULL
```

sql/2019/06_Fonts/06_09.sql
```diff
@@ -0,0 +1,32 @@
+#standardSQL
+
+-- counts the number of pages given number of font-face
+
+WITH
+  response_bodies AS (
```
rviscomi: For counting the number of font resources per page, let's use the `type=font` field of the `summary_requests` table:

```sql
#standardSQL
SELECT
  APPROX_QUANTILES(fonts_per_page, 1000)[OFFSET(250)] AS p25_fonts_per_page,
  APPROX_QUANTILES(fonts_per_page, 1000)[OFFSET(500)] AS median_fonts_per_page,
  APPROX_QUANTILES(fonts_per_page, 1000)[OFFSET(750)] AS p75_fonts_per_page
FROM (
  SELECT
    COUNT(0) AS fonts_per_page
  FROM
    `httparchive.summary_requests.2019_07_01_*`
  WHERE
    type = 'font'
  GROUP BY
    pageid)
```

I'm also aggregating the distribution into 25,50,75th percentiles (2/4/6 fonts per page accordingly)

In general, avoid the response_bodies table if you can help it.

sql/2019/06_Fonts/06_09.sql
```diff
@@ -0,0 +1,32 @@
...
+    SELECT * FROM `httparchive.response_bodies.2019_07_01_desktop`
+    UNION ALL
+    SELECT * FROM `httparchive.response_bodies.2019_07_01_mobile`
+  ),
+  temp AS (
+    SELECT ARRAY_LENGTH(REGEXP_EXTRACT_ALL(body, r"\@font-face\s*\{[^}]+\}")) AS font_display_arr, url, page
```
rviscomi: This metric description is asking about two different things: fonts resources and font faces. Let's put resources in 06_09a.sql and faces in 06_09b.sql so we can have them separated as needed. The font faces problem is probably best answered with a CSS-specific solution that we're still working out for the CSS chapter. Related: https://discuss.httparchive.org/t/analyzing-stylesheets-with-a-js-based-parser/1683

rviscomi: Chrome also has counters for every time it encounters a specific web API. We expose these in HTTP Archive as blinkFeatures.

There is a counter called [CSSAtRuleFontFace](https://www.chromestatus.com/metrics/feature/timeline/popularity/970) that could be helpful as an inexpensive way of detecting when those are used. Other relevant feature counters: [DocumentFonts](https://www.chromestatus.com/metrics/feature/timeline/popularity/440) (not sure what this one is detecting) and [FontFaceConstructor](https://www.chromestatus.com/metrics/feature/timeline/popularity/989) (JS API)

sql/2019/06_Fonts/06_13.sql
```diff
@@ -0,0 +1,35 @@
...
+
+WITH
+  response_bodies AS (
+    SELECT * FROM `httparchive.response_bodies.2019_07_01_desktop`
+    UNION ALL
+    SELECT * FROM `httparchive.response_bodies.2019_07_01_mobile`
+  ),
+  temp AS (
+    SELECT
+      ARRAY_LENGTH(REGEXP_EXTRACT_ALL(body, r"new FontFace\(")) AS fontface,
```
rviscomi: Let's use the blink feature counter for this metric.

```sql
SELECT
  client,
  pct_urls
FROM
  `httparchive.blink_features.usage`
WHERE
  feature = 'FontFaceConstructor' AND
  yyyymmdd = '20190701'
```

(20190701 data is unavailable until the first of next month, but you can try it with 20190601)

sql/2019/06_Fonts/06_16.sql
```diff
@@ -0,0 +1,22 @@
+#standardSQL
+
+-- counts the local() inside @font-face
+
+WITH
+  response_bodies AS (
```
rviscomi: I'd also wait to see the resolution of the CSS parser approach

sql/2019/06_Fonts/06_01.sql
```diff
@@ -0,0 +1,27 @@
+#standardSQL
+
+-- counts the local and hosted fonts
+
+WITH
```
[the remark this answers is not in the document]
tjmonsi: Oh wow, this is cleaner. I'll keep it in mind. But wait, why * at the end?

sql/2019/06_Fonts/06_03.sql
tjmonsi: But would it not get most of them? (given not all are set as type font?)

sql/2019/06_Fonts/06_03.sql
tjmonsi: copy will change

sql/2019/06_Fonts/06_04.sql
tjmonsi: ahhh lighthouse. I see. I will edit mine

sql/2019/06_Fonts/06_01.sql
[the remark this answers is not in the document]
tjmonsi: oh nevermind i get it now. I didn't see the summary_requests and summary_pages end

sql/2019/06_Fonts/06_03.sql
rviscomi: The results of your query vs mine are very similar so in that case I'd prefer the simpler solution.

sql/2019/06_Fonts/06_16.sql
tjmonsi: I made some changes on it using the parser approach

sql/2019/06_Fonts/06_13.sql
tjmonsi: changed to use blink_features

tjmonsi:
@rviscomi please see changes. And also I have some questions in tagged in the issue itself

ymschaap:
Re: 6.15 and the <link> please see my example code for identifying `<link rel="amphtml>`

See example 10.03:
https://github.com/HTTPArchive/almanac.httparchive.org/blob/2600eb43fdffad0df92f0a88fa0b821ddeb08c08/sql/2019/10_SEO/10_03.sql

In your case you would have to UNNEST based on the returned array `href` of `<link rel="preconnect">` elements.

10.01 does something similar:
> https://github.com/HTTPArchive/almanac.httparchive.org/blob/2600eb43fdffad0df92f0a88fa0b821ddeb08c08/sql/2019/10_SEO/10_01.sql

I am vacation, else I would've loved to help.

sql/2019/06_Fonts/06_09b.sql
```diff
@@ -0,0 +1,39 @@
...
+
+-- counts the number of pages given number of font-face
+
+CREATE TEMP FUNCTION parseCSS(stylesheet STRING)
+RETURNS INT64 LANGUAGE js AS '''
+   var css = parse(stylesheet, { silent: true })
+   var fontFace = {};
+
+   if (css.type === 'stylesheet' && css.stylesheet && css.stylesheet.rules) {
+     for (var rule of css.stylesheet.rules) {
```
rviscomi: Getting this error when I try to run the query:

> Error: UDF worker timed out during execution.; Unexpected abort triggered for worker worker-32607: request_timeout

Any ideas?

sql/2019/06_Fonts/06_09b.sql
```diff
@@ -0,0 +1,39 @@
...
+'''
+OPTIONS (library="gs://httparchive/lib/parse-css.js");
+
+SELECT
+  APPROX_QUANTILES(faces_per_page, 1000)[OFFSET(250)] AS p25_faces_per_page,
+  APPROX_QUANTILES(faces_per_page, 1000)[OFFSET(500)] AS median_faces_per_page,
+  APPROX_QUANTILES(faces_per_page, 1000)[OFFSET(750)] AS p75_faces_per_page
+FROM (
+  SELECT SUM(parseCSS(body)) as faces_per_page
+  FROM `httparchive.response_bodies.2017_01_01_*`
```
rviscomi: I think joining response bodies with requests could be causing the timeout. I'll look into ways to optimize this.

FWIW changing to use the 1k tables runs the query successfully:

```sql
CREATE TEMP FUNCTION parseCSS(stylesheet STRING)
RETURNS INT64 LANGUAGE js AS '''
   var css = parse(stylesheet, { silent: true })
   var fontFace = {};
    if (css.type === 'stylesheet' && css.stylesheet && css.stylesheet.rules) {
     for (var rule of css.stylesheet.rules) {
       if (rule.type === 'font-face' || rule.type === 'rule') {
         if (rule.declarations && rule.declarations.length) {
           for (var declaration of rule.declarations) {
             if (declaration.type === 'declaration' && declaration.property === 'font-family') {
               fontFace[declaration.value] = true
             }
           }
         }
       }
     }
   }
    return Object.keys(fontFace).length;
'''
OPTIONS (library="gs://httparchive/lib/parse-css.js");

 SELECT
  APPROX_QUANTILES(faces_per_page, 1000)[OFFSET(250)] AS p25_faces_per_page,
  APPROX_QUANTILES(faces_per_page, 1000)[OFFSET(500)] AS median_faces_per_page,
  APPROX_QUANTILES(faces_per_page, 1000)[OFFSET(750)] AS p75_faces_per_page
FROM (
  SELECT SUM(parseCSS(body)) AS faces_per_page
  FROM `httparchive.almanac.response_bodies_*`
  JOIN (SELECT _TABLE_SUFFIX, url, pageid FROM `httparchive.almanac.summary_requests_*` WHERE type = 'css')
  USING (_TABLE_SUFFIX, url)
  GROUP BY pageid
)
```

So it seems to be a memory issue with joining 370M^2 rows.

sql/2019/06_Fonts/06_01.sql
```diff
@@ -0,0 +1,38 @@
...
+SELECT
+  COUNTIF(NET.HOST(page) != NET.HOST(url)) / COUNT(0) AS hosted,
+  COUNTIF(NET.HOST(page) = NET.HOST(url)) / COUNT(0) AS local
+FROM
+  (SELECT _TABLE_SUFFIX, pageid, url FROM `httparchive.summary_requests.2019_07_01_*` WHERE type = 'font')
+JOIN
+  (SELECT _TABLE_SUFFIX, pageid, url AS page FROM `httparchive.summary_pages.2019_07_01_*`)
+USING (_TABLE_SUFFIX, pageid)
+
+-- old code but does the same thing as above
```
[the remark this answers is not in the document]
tjmonsi: Removing

sql/2019/06_Fonts/06_09b.sql
tjmonsi: I'll try joining them I guess after I parse the css?

rviscomi: Ok this has been haunting me all week but I think I've finally got a fix. See https://discuss.httparchive.org/t/analyzing-stylesheets-with-a-js-based-parser/1683/13?u=rviscomi

I've created a new table called `almanac.parsed_css` with the JSON output from the parser script for all ~80M stylesheets. It doesn't include inline styles, which was an oversight. I'll look into adding those to the table.

Here's an example of querying the parsed stylesheet JSON:

```sql
#standardSQL
CREATE TEMPORARY FUNCTION getAllValues(css STRING, property STRING)
RETURNS ARRAY<STRING> LANGUAGE js AS '''
try {
  var reduceValues = (values, rule) => {
    if (rule.type == 'media') {
      return reduceValues(values, rule.rules);
    }
    
    return values.concat(rule.declarations.filter(d => d.property == property).map(d => d.value));
  };
  var $ = JSON.parse(css);
  return $.stylesheet.rules.reduce(reduceValues, []);
} catch (e) {
  return [];
}
''';

SELECT
  APPROX_TOP_COUNT(color, 10)
FROM
  UNNEST(getAllValues(css, 'color')) AS color
```

tjmonsi:
@rviscomi apologies for putting this behind. I just came back to the Philippines from Korea and was just getting myself back to my regular rhythm. I will put what I have by end of week and would like to ask for some things to be offloaded because my PhD will require me to focus on some things for this semester.

I will update though what I can do after my meetings today.

I am sorry if I am lagging behind on this one and was not properly communicating :(

sql/2019/06_Fonts/06_01.sql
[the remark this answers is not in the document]
rviscomi: @tjmonsi I'm still seeing the previous query. Could you update this PR and ping me when you're ready for another review? Hope to finish these queries this week.

tjmonsi:
@rviscomi Hi, I just pushed my changes. I may not be able to finish the others because of my load. :( As for 06_15.sql. I am quite near as this is my current code...

```
#standardSQL

# sample: `httparchive.almanac.pages_desktop_1k`
# dataset: `httparchive.pages.2019_07_01_desktop`

CREATE TEMP FUNCTION analyse(payload STRING)
RETURNS STRING
LANGUAGE js AS """
try {
    var $ = JSON.parse(payload);
    var almanac = JSON.parse($._almanac);
    if(almanac && almanac['link-nodes']) {
        var found = almanac['link-nodes'].findIndex(node => {
            if(node.rel && node.rel.toLowerCase() == 'preconnect') {
                return true;
            }
        });
        return found >= 0 ? almanac['link-nodes'][found].href : 0;
    }
    return 0;
} catch (e) {
  return 0;
}
""";

SELECT
  DISTINCT(analyse(payload)) as href,
  COUNT(DISTINCT(analyse(payload))) as count
--     SUM(analyse(payload)) AS `scoreSum`,
--     AVG(analyse(payload)) AS `scoreAverage`,
--     (SUM(analyse(payload)) / COUNT(url)) as `scorePercentage`
FROM
    `httparchive.almanac.pages_desktop_1k`
    GROUP BY href
```

But my head can't think anything to check if the url is a host for fonts.

Anyway, I apologize for not being able to finish all of the tasks

rviscomi:
Thanks @tjmonsi. I'll take over your branch from here and make changes directly on it to resolve any outstanding feedback.

Issue #86, opened 2019-07-23
Query metrics: Chapter 5. Third Parties


rviscomi:
Part | Chapter | Authors | Reviewers | Tracking Issue
-- | -- | -- | -- | --
I. Page Content | 5. Third Parties | @patrickhulce | @simonhearne @flowlabs @jasti @zeman | #8

# READ ME!
All of the metrics in the table below have been marked as `Able To Query` during the [metrics triage](https://github.com/HTTPArchive/almanac.httparchive.org/issues/33). The analyst assigned to each metric is expected to write the corresponding query and submit a PR to [have it reviewed and added to the repo](https://github.com/HTTPArchive/almanac.httparchive.org/issues/62).

In order to stay on schedule and have the data ready for authors, **please have all metrics reviewed and merged by August 5**.

## Assignments

ID | Metric description | Analyst | Notes
-- | -- | -- | --
05.01 | Percentage of pages that include at least one third-party resource. | @patrickhulce |  
05.02 | Percentage of pages that include at least one ad resource. | @patrickhulce |  
05.03 | Percentage of requests that are third party requests broken down by third party category by resource type. | @patrickhulce |  
05.04 | Percentage of total bytes that are from third party requests broken down by third party category by resource type. | @patrickhulce |  
05.05 | Percentage of total script execution time that is from third party scripts broken down by third party category. | @patrickhulce |  
05.06 | Top 100 third party domains by request volume | @patrickhulce |  
05.07 | Top 100 third party domains by total byte weight | @patrickhulce |  
05.08 | Top 100 third party domains by total script execution time | @patrickhulce |  
05.09 | Top 100 third party requests by request volume | @patrickhulce |  
05.10 | Top 100 third party requests by total script execution time | @patrickhulce |  
05.11 | Percentile breakdown page-relative percentage of requests that are third party requests broken down by third party category. | @patrickhulce |  
05.12 | Percentile breakdown page-relative percentage of total bytes that are from third party requests  broken down by third party category. | @patrickhulce |  
05.13 | Percentile breakdown page-relative percentage of total script execution time that is from third party scripts. | @patrickhulce |  

## Checklist of metrics to be merged

- [x] 	05.01	Percentage of pages that include at least one third-party resource.
- [x] 	05.02	Percentage of pages that include at least one ad resource.
- [x] 	05.03	Percentage of requests that are third party requests broken down by third party category by resource type.
[...]

PR #107, opened 2019-07-29
Chapter 5: Add All Byte and Request Count Queries

patrickhulce:
I hope this is the expected format. From https://github.com/HTTPArchive/almanac.httparchive.org/issues/62#issuecomment-506403182 it sounds like we prefer duplication for easy of reading (many queries differ only in their `ORDER BY` clause or could easily be combined when run). Should this be called out in a comment perhaps?

ref https://github.com/HTTPArchive/almanac.httparchive.org/issues/86

### DIFF
--- a/sql/2019/05_Third_Parties/05_01.sql
+++ b/sql/2019/05_Third_Parties/05_01.sql
@@ -1,9 +1,9 @@
 #standardSQL
 # Percentage of pages that include at least one third party resource.
 SELECT
-  COUNT(*) AS numberOfPages,
+  COUNT(0) AS numberOfPages,
   COUNTIF(numberOfThirdPartyRequests > 0) AS numberOfPagesWithThirdParty,
-  COUNTIF(numberOfThirdPartyRequests > 0) / COUNT(*) AS percentOfPagesWithThirdParty
+  COUNTIF(numberOfThirdPartyRequests > 0) / COUNT(0) AS percentOfPagesWithThirdParty
 FROM (
   SELECT
     pageUrl,
--- a/sql/2019/05_Third_Parties/05_01.sql
+++ b/sql/2019/05_Third_Parties/05_01.sql
@@ -3,7 +3,7 @@
 SELECT
   COUNT(*) as numberOfPages,
   COUNTIF(numberOfThirdPartyRequests > 0) AS numberOfPagesWithThirdParty,
-  COUNT(*) / COUNTIF(numberOfThirdPartyRequests > 0) AS percentOfPagesWithThirdParty
+  COUNTIF(numberOfThirdPartyRequests > 0) / COUNT(*) AS percentOfPagesWithThirdParty
 FROM (
   SELECT
     pageUrl,
--- a/sql/2019/05_Third_Parties/05_01.sql
+++ b/sql/2019/05_Third_Parties/05_01.sql
@@ -1,15 +1,18 @@
 #standardSQL
 # Percentage of pages that include at least one third party resource.
 SELECT
+  client,
   COUNT(0) AS numberOfPages,
   COUNTIF(numberOfThirdPartyRequests > 0) AS numberOfPagesWithThirdParty,
-  COUNTIF(numberOfThirdPartyRequests > 0) / COUNT(0) AS percentOfPagesWithThirdParty
+  ROUND(COUNTIF(numberOfThirdPartyRequests > 0) * 100 / COUNT(0), 2) AS percentOfPagesWithThirdParty
 FROM (
   SELECT
+    client,
     pageUrl,
     COUNTIF(thirdPartyDomain IS NOT NULL) AS numberOfThirdPartyRequests
   FROM (
     SELECT
+      client,
       page AS pageUrl,
       DomainsOver50Table.requestDomain AS thirdPartyDomain
     FROM
@@ -19,5 +22,8 @@ FROM (
     ON NET.HOST(url) = DomainsOver50Table.requestDomain
   )
   GROUP BY
+    client,
     pageUrl
 )
+GROUP BY
+  client
--- a/sql/2019/05_Third_Parties/05_01.sql
+++ b/sql/2019/05_Third_Parties/05_01.sql
@@ -1,7 +1,7 @@
 #standardSQL
 # Percentage of pages that include at least one third party resource.
 SELECT
-  COUNT(*) as numberOfPages,
+  COUNT(*) AS numberOfPages,
   COUNTIF(numberOfThirdPartyRequests > 0) AS numberOfPagesWithThirdParty,
   COUNTIF(numberOfThirdPartyRequests > 0) / COUNT(*) AS percentOfPagesWithThirdParty
 FROM (
@@ -11,7 +11,7 @@ FROM (
   FROM (
     SELECT
       page AS pageUrl,
-      DomainsOver50Table.requestDomain as thirdPartyDomain
+      DomainsOver50Table.requestDomain AS thirdPartyDomain
     FROM
       `httparchive.almanac.summary_requests`
     LEFT JOIN
--- a/sql/2019/05_Third_Parties/05_01.sql
+++ b/sql/2019/05_Third_Parties/05_01.sql
@@ -1,0 +1,23 @@
+#standardSQL
+# Percentage of pages that include at least one third party resource.
+SELECT
+  COUNT(*) as numberOfPages,
+  COUNTIF(numberOfThirdPartyRequests > 0) AS numberOfPagesWithThirdParty,
+  COUNT(*) / COUNTIF(numberOfThirdPartyRequests > 0) AS percentOfPagesWithThirdParty
+FROM (
+  SELECT
+    pageUrl,
+    COUNTIF(thirdPartyDomain IS NOT NULL) AS numberOfThirdPartyRequests
+  FROM (
+    SELECT
+      page AS pageUrl,
+      DomainsOver50Table.requestDomain as thirdPartyDomain
+    FROM
+      `httparchive.almanac.summary_requests`
+    LEFT JOIN
+      `lighthouse-infrastructure.third_party_web.2019_07_01_all_observed_domains` AS DomainsOver50Table
+    ON NET.HOST(url) = DomainsOver50Table.requestDomain
+  )
+  GROUP BY
+    pageUrl
+)
--- a/sql/2019/05_Third_Parties/05_02.sql
+++ b/sql/2019/05_Third_Parties/05_02.sql
@@ -1,15 +1,18 @@
 #standardSQL
 # Percentage of pages that include at least one ad resource.
 SELECT
+  client,
   COUNT(0) AS numberOfPages,
   COUNTIF(numberOfAdRequests > 0) AS numberOfPagesWithAd,
-  COUNTIF(numberOfAdRequests > 0) / COUNT(0) AS percentOfPagesWithAd
+  ROUND(COUNTIF(numberOfAdRequests > 0) * 100 / COUNT(0), 2) AS percentOfPagesWithAd
 FROM (
   SELECT
+    client,
     pageUrl,
     COUNTIF(thirdPartyCategory = 'ad') AS numberOfAdRequests
   FROM (
     SELECT
+      client,
       page AS pageUrl,
       ThirdPartyTable.category AS thirdPartyCategory
     FROM
@@ -19,5 +22,8 @@ FROM (
     ON NET.HOST(url) = ThirdPartyTable.domain
   )
   GROUP BY
+    client,
     pageUrl
 )
+GROUP BY
+  client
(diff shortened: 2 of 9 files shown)


patrickhulce:
@simonhearne @flowlabs @jasti @zeman any desire to review these query PRs, or shall I only ping y'all for the article content phase?

sql/2019/05_Third_Parties/05_06.sql
```diff
@@ -0,0 +1,23 @@
...
+  SUM(requestBytes) AS totalBytes
+FROM (
+  SELECT
+      SAFE_CAST(REGEXP_EXTRACT(payload, r'_bytesIn":(\d+)') AS INT64) AS requestBytes,
+      NET.HOST(url) AS requestDomain,
+      DomainsOver50Table.requestDomain as thirdPartyDomain
+    FROM
+      `httparchive.requests.2019_07_01_mobile`
+    LEFT JOIN
+      `lighthouse-infrastructure.third_party_web.2019_07_01_all_observed_domains` AS DomainsOver50Table
```
rviscomi: Any chance you can make this table publicly queryable? Otherwise we should add it to our scratchspace dataset.

sql/2019/05_Third_Parties/05_06.sql
```diff
@@ -0,0 +1,23 @@
...
+    ON NET.HOST(url) = DomainsOver50Table.requestDomain
+)
+ORDER BY
```
rviscomi: GROUP BY must come before ORDER BY (validation error)

sql/2019/05_Third_Parties/05_06.sql
```diff
@@ -0,0 +1,23 @@
+#standardSQL
+# Top 100 third party domains by request volume
+SELECT
+  thirdPartyDomain,
+  COUNT(*) AS totalRequests,
+  SUM(requestBytes) AS totalBytes
+FROM (
+  SELECT
+      SAFE_CAST(REGEXP_EXTRACT(payload, r'_bytesIn":(\d+)') AS INT64) AS requestBytes,
```
rviscomi: If all you need is the response size, let's use `respSize` from the much smaller `summary_requests` dataset.

sql/2019/05_Third_Parties/05_06.sql
```diff
@@ -0,0 +1,23 @@
...
+      NET.HOST(url) AS requestDomain,
+      DomainsOver50Table.requestDomain as thirdPartyDomain
+    FROM
+      `httparchive.requests.2019_07_01_mobile`
```
rviscomi: Do you also want to include desktop in your analysis?

sql/2019/05_Third_Parties/05_06.sql
patrickhulce: Oh sure I didn't realize it wasn't!

sql/2019/05_Third_Parties/05_06.sql
patrickhulce: I was anticipating doing everything on mobile for consistency because some metrics are only available on lighthouse dataset which is mobile.

sql/2019/05_Third_Parties/05_06.sql
patrickhulce: Doh! I didn't copy back the validation fixes thanks!

sql/2019/05_Third_Parties/05_06.sql
patrickhulce: Ah yeah everything got removed here I guess this is just response size at this point thanks! :)

sql/2019/05_Third_Parties/05_03.sql
```diff
@@ -0,0 +1,19 @@
+#standardSQL
+# Percentage of requests that are third party requests broken down by third party category by resource type.
```
patrickhulce: @rviscomi for something like this, I was assuming the percentage is trivially computed by hand but based on https://github.com/HTTPArchive/almanac.httparchive.org/issues/67#issuecomment-517350413 it seems as though this sort of `GROUP BY` result won't be very well handled. Do I need to manually explode this set of options to arrive at one static query?

rviscomi: Could you calculate the percent in the query? Something like this:

```sql
#standardSQL
# Percentage of requests that are third party requests broken down by third party category by resource type.
SELECT
  thirdPartyCategory,
  contentType,
  COUNT(*) AS totalRequests,
  ROUND(COUNT(0) * 100 / SUM(COUNT(0)) OVER (), 2) AS pctRequests
FROM (
  SELECT
      resp_content_type AS contentType,
      ThirdPartyTable.category as thirdPartyCategory
    FROM
      `httparchive.almanac.summary_requests`
    LEFT JOIN
      `lighthouse-infrastructure.third_party_web.2019_07_01` AS ThirdPartyTable
    ON NET.HOST(url) = ThirdPartyTable.domain
)
GROUP BY
  thirdPartyCategory,
  contentType
ORDER BY
  totalRequests DESC
```

sql/2019/05_Third_Parties/05_01.sql
```diff
@@ -0,0 +1,23 @@
+#standardSQL
+# Percentage of pages that include at least one third party resource.
+SELECT
+  COUNT(*) as numberOfPages,
+  COUNTIF(numberOfThirdPartyRequests > 0) AS numberOfPagesWithThirdParty,
+  COUNT(*) / COUNTIF(numberOfThirdPartyRequests > 0) AS percentOfPagesWithThirdParty
```
rviscomi: This needs to be flipped so the total is the denominator:

```suggestion
  COUNTIF(numberOfThirdPartyRequests > 0) / COUNT(*) AS percentOfPagesWithThirdParty
```

sql/2019/05_Third_Parties/05_02.sql
```diff
@@ -0,0 +1,23 @@
+#standardSQL
+# Percentage of pages that include at least one ad resource.
+SELECT
+  COUNT(*) as numberOfPages,
+  COUNTIF(numberOfAdRequests > 0) AS numberOfPagesWithAd,
+  COUNT(*) / COUNTIF(numberOfAdRequests > 0) AS percentOfPagesWithAd
```
rviscomi:
```suggestion
  COUNTIF(numberOfAdRequests > 0) / COUNT(*) AS percentOfPagesWithAd
```

sql/2019/05_Third_Parties/05_04.sql
```diff
@@ -0,0 +1,20 @@
+#standardSQL
+# Percentage of total bytes that are from third party requests broken down by third party category by resource type.
```
rviscomi: A little help for calculating the %:

```sql
#standardSQL
# Percentage of total bytes that are from third party requests broken down by third party category by resource type.
SELECT
  thirdPartyCategory,
  contentType,
  SUM(requestBytes) AS totalBytes,
  ROUND(SUM(requestBytes) * 100 / SUM(SUM(requestBytes)) OVER (), 2) AS pctBytes
FROM (
  SELECT
      resp_content_type AS contentType,
      respBodySize AS requestBytes,
      ThirdPartyTable.category as thirdPartyCategory
    FROM
      `httparchive.almanac.summary_requests`
    LEFT JOIN
      `lighthouse-infrastructure.third_party_web.2019_07_01` AS ThirdPartyTable
    ON NET.HOST(url) = ThirdPartyTable.domain
)
GROUP BY
  thirdPartyCategory,
  contentType
ORDER BY
  totalBytes DESC
```

I'm also seeing many similar content types, for example `text/javascript` vs `application/javascript`. The `type` field is more coarse and folds these together into a "script" type, if that's more helpful.

sql/2019/05_Third_Parties/05_07.sql
```diff
@@ -0,0 +1,23 @@
+#standardSQL
+# Top 100 third party domains by total byte weight
+SELECT
+  thirdPartyDomain,
```
rviscomi: The top thirdPartyDomain is null, do you want to filter that out?

sql/2019/05_Third_Parties/05_09.sql
```diff
@@ -0,0 +1,19 @@
...
+      url AS requestUrl,
+      respBodySize AS requestBytes
+    FROM
+      `httparchive.almanac.summary_requests`
+)
+GROUP BY
+  requestUrl
+ORDER BY
+  totalRequests DESC
+LIMIT 1000
```
rviscomi:
```suggestion
LIMIT 100
```

sql/2019/05_Third_Parties/05_09.sql
```diff
@@ -0,0 +1,19 @@
+#standardSQL
+# Top 100 third party requests by request volume
+SELECT
+  requestUrl,
+  COUNT(*) AS totalRequests,
```
rviscomi: You may also want to add a pctRequests field. Out of context `5662539` could either be a very large or very small number, relatively speaking.

sql/2019/05_Third_Parties/05_04.sql
patrickhulce: thanks very much!!

re: text/javascript

I thought it cuts both ways and it's always easier to manually interpret groups if we have multiple, but given the leaning toward everything being done in the query I think that's a good change 👍

sql/2019/05_Third_Parties/05_07.sql
patrickhulce: I guess we already have the total byte weight 1p vs. 3p comparison, so yeah SG 👍

sql/2019/05_Third_Parties/05_09.sql
patrickhulce: BigQuery wasn't able to execute the over query and failed with a memory limit exceeded on this one, just make a separate SELECT in the from for the overall count then?

rviscomi:
Is this ready for another review?

patrickhulce:
Oh sorry I missed your thumbs up! Let me apply the last bit then.

patrickhulce:
> Oh sorry I missed your thumbs up! 

Aside: is there any GH setting I don't know about that lets you get notifications about that? I imagine it could get annoying for certain types, but code review comments would be nice.

patrickhulce:
Yep I think it's ready @rviscomi!

sql/2019/05_Third_Parties/05_01.sql
```diff
@@ -0,0 +1,23 @@
+#standardSQL
+# Percentage of pages that include at least one third party resource.
+SELECT
+  COUNT(*) as numberOfPages,
```
rviscomi: There are some conventions in the [Analyst wiki](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide) that can be applied here. (eg, uppercase `AS`)

I'd also like to see `COUNT(0)` used rather than `COUNT(*)` as a micro optimization since we're counting the rows and the fields don't matter.

rviscomi: For each of these queries, I think we should also split out by desktop/mobile rather than grouping together. See this conversation on the Slack channel for context: https://httparchive.slack.com/archives/CGFLV2S06/p1565208006011800

sql/2019/05_Third_Parties/05_01.sql
```diff
@@ -0,0 +1,23 @@
...
+  COUNTIF(numberOfThirdPartyRequests > 0) AS numberOfPagesWithThirdParty,
+  COUNTIF(numberOfThirdPartyRequests > 0) / COUNT(*) AS percentOfPagesWithThirdParty
```
rviscomi: Format the percents as `ROUND(numerator * 100 / denominator, 2)`

sql/2019/05_Third_Parties/05_06.sql
```diff
@@ -0,0 +1,23 @@
...
+  thirdPartyDomain,
+  COUNT(*) AS totalRequests,
+  SUM(requestBytes) AS totalBytes
+FROM (
+  SELECT
+      respSize AS requestBytes,
+      NET.HOST(url) AS requestDomain,
+      DomainsOver50Table.requestDomain as thirdPartyDomain
+    FROM
+      `httparchive.summary_requests.2019_07_01_mobile`
```
rviscomi: Should this be both desktop and mobile? You can change `mobile` to `*`. This will create a special `_TABLE_SUFFIX` field that you can alias as `client` and group by.

sql/2019/05_Third_Parties/05_01.sql
patrickhulce: great points, done!

patrickhulce: SG, done!

sql/2019/05_Third_Parties/05_06.sql
patrickhulce: I think the signficance of desktop vs. mobile here is much less important than the other metrics and complicates the `LIMIT` of this query as well, so I'm inclined to just combine both without distinguishing, there was also no reason not to use `alamanac.summary_requests` here so switched to that 👍

PR #108, opened 2019-07-30
Add ability to filter contributors by team

### LINKED ISSUE #73: Add ability to filter contributors by team

tanhengyeow:
Fixes #73


src/templates/en/2019/contributors.html
```diff
@@ -5,14 +5,26 @@
...
-  <ul>
-    {% for team in contributor.teams %}
-    <li>{{ contributors.teams[team].name }}</li>
+  <div class="filter-container">
+  {% for id, team in contributors.teams.items()  %}
+    <button class="{{ id }}">{{ team.name }}</button>
+  {% endfor %}
+  </div>
+
+  <div class="contributors">
```
rviscomi:
```suggestion
  <div class="contributors {{ ' '.join(contributors.teams.keys()) }}">
```

src/static/js/filter-teams.js
```diff
@@ -0,0 +1,61 @@
+const buttons = document.querySelectorAll('button');
```
rviscomi: I think this logic could be simplified. What do you think of an approach like this:

- `div.contributors` is initialized with a class for each team name
```html
<div class="contributors analysts authors brainstormers designers developers editors reviewers translators">
```
- in the CSS we have rules for each team so that if `.contributors.<team-id>` is set, then `.contributor.<team-id>` is set to `display: block` (`.contributor` could be initialized to `display: none`)
```css
.contributor {
  display: none;
}
/* For each team... */
.contributors.developers .contributor.developers {
  display: block;
}
```
- the toggle state of each filter button is linked to the classes on `.contributors`, so every button click results in a `toggleClass(<team-id>)` call
```js
const buttons = document.querySelector('.filter-container');
const contributors = document.querySelector('.contributors')
buttons.addEventListener('click', e => {
  const teamId = e.target.className;
  contributors.classList.toggle(teamId);
});
```
- `div.filter-container` and the toggle buttons could be moved inside of `div.contributors` so they can inherit their styles depending on the selected teams
```css
.contributors.developers button.developers {
  background: #0053ba;
  color: white;
}

Comment (tanhengyeow) [reply src/static/js/filter-teams.js]: > /* For each team... */
.contributors.developers .contributor.developers {
  display: block;
}

This might mean having to manually add//edit/delete lines of different teams should there be any changes in the future. 

However, I don't think there will be lots of changes to the current teams' (perhaps none of at all) from time to time. So I agree this approach is much better to read and it also simplifies the logic which is a good thing 😄

Comment (rviscomi) [reply src/static/js/filter-teams.js]: We could also generate the CSS in an inline style block in the template, which should be more resilient.

File: src/static/js/contributors.js [hunk 3]
```diff
@@ -0,0 +1,24 @@
+const style = document.querySelector('style');
+const buttons = document.querySelectorAll('.filter-container button');
+const contributors = document.querySelector('.contributors');
+
+for (let button of buttons) {
+  const buttonClassName = button.className;
+
+  // Dynamically generate CSS that reacts to contributors' view state and buttons' toggle state
```
rviscomi: Rather than generating the styles in JS, you can do the iteration in the `styles` block of the contributors template, so it's pure CSS.

src/templates/base.html
```diff
@@ -54,6 +55,7 @@
     {% endblock %}
     {% block scripts %}
       <script src="/static/js/language.js"></script>
+      <script src="/static/js/contributors.js"></script>
```
rviscomi: Move this out of the base template (we don't want to load contributor-specific JS on all pages).

src/static/js/contributors.js
```diff
@@ -0,0 +1,10 @@
+const style = document.querySelector('style');
```
rviscomi: This variable is unused

Issue #82, opened 2019-07-23
Query metrics: Chapter 1. JavaScript


rviscomi:
Part | Chapter | Authors | Reviewers | Tracking Issue
-- | -- | -- | -- | --
I. Page Content | 1. JavaScript | @addyosmani @housseindjirdeh @mathiasbynens | @rwaldron @RReverser | #3

# READ ME!
All of the metrics in the table below have been marked as `Able To Query` during the [metrics triage](https://github.com/HTTPArchive/almanac.httparchive.org/issues/33). The analyst assigned to each metric is expected to write the corresponding query and submit a PR to [have it reviewed and added to the repo](https://github.com/HTTPArchive/almanac.httparchive.org/issues/62).

In order to stay on schedule and have the data ready for authors, **please have all metrics reviewed and merged by August 5**.

## Assignments

ID | Metric description | Analyst | Notes
-- | -- | -- | --
01.01 | Distribution of JS bytes | @rviscomi |  
01.02 | Distribution of first party JS bytes vs. third party | @rviscomi |  
01.03 | Number of JS requests | @rviscomi |  
01.04 | Number of first-party JS requests vs. third-party | @rviscomi |  
01.05 | % of gzip-compressed scripts | @rviscomi |  
01.06 | % of brotli-compressed scripts | @rviscomi |  
01.07 | Breakdown of V8 CPU times (if feasible) | @rviscomi | Confirm with pmeenan that the CPU timing is enabled
01.08 | Top N JS libraries | @rviscomi |  
01.09 | Notable changes in popularity since last year | @rviscomi |  
01.10 | Top N JS client-side frameworks (React, Vue, etc…) | @rviscomi |  
01.11 | Distribution of JS bytes on site per JavaScript framework | @rviscomi |  
01.12 | % of pages that use `<script type=module>` | @rviscomi | Added custom metric
01.13 | % of pages that use `<script nomodule>` | @rviscomi | Added custom metric
01.14 | % of pages that use `<link rel=preload>` for JS resources | @rviscomi | Resolved by link-nodes custom metric
01.15 | % of pages that use `<link rel=modulepreload>` | @rviscomi | Easiest as a DOM query
01.16 | % of pages that use `<link rel=prefetch>` for JS resources | @rviscomi | Easiest as a DOM query
01.18 | Atomics | @rviscomi |  
01.19 | Intl | @rviscomi |  
01.20 | Proxy | @rviscomi |  
[...]

PR #110, opened 2019-08-01
JS chapter queries

rviscomi:
#82


rviscomi:
@addyosmani FYI for our 1:1, I want to make sure the format of the results matches your expectations

cc FYI @housseindjirdeh @mathiasbynens

sql/2019/01_JavaScript/01_02a.sql
```diff
@@ -0,0 +1,21 @@
+#standardSQL
+# 01_02a: Distribution of 1P/3P JS bytes
+SELECT
+  percentile,
+  ROUND(APPROX_QUANTILES(first_party, 1000)[OFFSET(percentile * 10)], 2) AS first_party_js_kbytes,
+  ROUND(APPROX_QUANTILES(third_party, 1000)[OFFSET(percentile * 10)], 2) AS third_party_js_kbytes
+FROM (
+  SELECT
+    SUM(IF(NET.HOST(page) = NET.HOST(url), respSize, 0) / 1024) AS first_party,
+    SUM(IF(NET.HOST(page) != NET.HOST(url), respSize, 0) / 1024) AS third_party
```
patrickhulce: would you be interested in using `lighthouse-infrastructure.third_party_web.2019_07_01_all_observed_domains` at all?

sql/2019/01_JavaScript/01_09.sql
```diff
@@ -0,0 +1,48 @@
...
+    client,
+    total)
+JOIN (
+  SELECT
+    app,
+    _TABLE_SUFFIX AS client,
+    COUNT(DISTINCT url) AS freq_2019,
+    COUNT(DISTINCT url) / total AS pct_2019
+  FROM
+    (SELECT _TABLE_SUFFIX, COUNT(DISTINCT url) AS total FROM `httparchive.summary_pages.2019_07_01_*` GROUP BY _TABLE_SUFFIX)
```
patrickhulce: maybe I misunderstood that convo about clients, but shouldn't the `DISTINCT url` be unnecessary here once we're grouping by client and it's a query on a pages table not requests?

sql/2019/01_JavaScript/01_09.sql
```diff
@@ -0,0 +1,48 @@
...
+  USING (_TABLE_SUFFIX)
+  WHERE
+    category = 'JavaScript Libraries'
+  GROUP BY
+    app,
+    client,
+    total)
+USING (app, client)
+WHERE
+  freq_2019 > 10
```
patrickhulce: would we want to see js libs that fell off the map too?

sql/2019/01_JavaScript/01_18.sql
```diff
@@ -0,0 +1,37 @@
+#standardSQL
+# 01_18-23: % of pages using JS APIs
+SELECT
+  client,
+  COUNT(0) AS total,
+  COUNTIF(atomics > 0) AS atomics,
+  ROUND(COUNTIF(atomics > 0) * 100 / COUNT(0), 2) AS pct_atomics,
+  COUNTIF(intl > 0) AS intl,
+  ROUND(COUNTIF(intl > 0) * 100 / COUNT(0), 2) AS pct_intl,
```
patrickhulce: should I be moving all of my query names to `pct_*`?

sql/2019/01_JavaScript/01_09.sql
rviscomi: No you're right. These tables should only have one url per page. `DISTINCT` is redundant.

sql/2019/01_JavaScript/01_09.sql
rviscomi: Ones that went to < 10 are probably not very notable, although I can remove this limitation if the authors feel like there's anything remarkable in the tail.

sql/2019/01_JavaScript/01_18.sql
rviscomi: Not necessary. As long as queries for a given chapter are internally consistent, that's mostly what matters. Especially since you're analyzing _and_ authoring, the naming shouldn't matter. It's just to give authors a unique ID to include/visualize the results.

In this case I suffix with the metric because there are multiple percent values in this query.

sql/2019/01_JavaScript/01_02a.sql
rviscomi: Hmm that might not be a bad idea. I'll play with that dataset and see how much the results differ.

rviscomi: Updated

Issue #95, opened 2019-07-23
Query metrics: Chapter 14. CMS


rviscomi:
Part | Chapter | Authors | Reviewers | Tracking Issue
-- | -- | -- | -- | --
III. Content Publishing | 14. CMS | @amedina | @westonruter @mor10 @sirjonathan | #16

# READ ME!
All of the metrics in the table below have been marked as `Able To Query` during the [metrics triage](https://github.com/HTTPArchive/almanac.httparchive.org/issues/33). The analyst assigned to each metric is expected to write the corresponding query and submit a PR to [have it reviewed and added to the repo](https://github.com/HTTPArchive/almanac.httparchive.org/issues/62).

In order to stay on schedule and have the data ready for authors, **please have all metrics reviewed and merged by August 5**.

## Assignments

ID | Metric description | Analyst | Notes
-- | -- | -- | --
14.01 | Top CMSs by # websites | @rviscomi |  
14.02 | Version of the plugin | @rviscomi | amp-plugin custom metric
14.03 | Template mode used (paired vs. native) | @rviscomi | amp-plugin custom metric
14.04 | Suggestion: WordPress.com enables AMP by default so it would be interesting to see how many sites have disabled it in addition to how many self-hosted sites have enabled it. Not sure if this is possible or not but would also help | @rviscomi | amp-plugin custom metric
14.07 | Comparison against non-CMS cases would also shed light on demographics, geography (together with device usage per region) | @rviscomi | CrUX form factor
14.08 | Connection types | @rviscomi | CrUX ECT
14.09 | Is it happening? | @rviscomi | Need TTFB metric
14.10 | Breakdown of CMS FCP from CrUX | @rviscomi |  
14.11 | Breakdown of CMS FID from CrUX | @rviscomi |  
14.12 | Delightfulness | @rviscomi | CLS

## Checklist of metrics to be merged

- [x] 	14.01	Top CMSs by # websites
- [x] 	14.02	Version of the plugin
- [x] 	14.03	Template mode used (paired vs. native)
- [ ] 	14.04	Suggestion: WordPress.com enables AMP by default so it would be interesting to see how many sites have disabled it in addition to how many self-hosted sites have enabled it. Not sure if this is possible or not but would also help
- [ ] 	14.07	Comparison against non-CMS cases would also shed light on demographics, geography (together with device usage per region)
- [ ] 	14.08	Connection types
- [ ] 	14.09	Is it happening?
[...]

PR #123, opened 2019-08-19
CMS queries

rviscomi:
Progress on #95

### DIFF
--- a/sql/2019/14_CMS/14_01.sql
+++ b/sql/2019/14_CMS/14_01.sql
@@ -1,0 +1,22 @@
+#standardSQL
+# 14_01: Top CMSs
+SELECT
+  _TABLE_SUFFIX AS client,
+  app AS cms,
+  COUNT(0) AS freq,
+  total,
+  ROUND(COUNT(0) * 100 / total, 2) AS pct
+FROM
+  `httparchive.technologies.2019_07_01_*`
+JOIN
+  (SELECT _TABLE_SUFFIX, COUNT(0) AS total FROM `httparchive.summary_pages.2019_07_01_*` GROUP BY _TABLE_SUFFIX)
+USING
+  (_TABLE_SUFFIX)
+WHERE
+  category = 'CMS'
+GROUP BY
+  client,
+  total,
+  cms
+ORDER BY
+  freq DESC
--- a/sql/2019/14_CMS/14_02.sql
+++ b/sql/2019/14_CMS/14_02.sql
@@ -1,0 +1,24 @@
+#standardSQL
+# 14_02: AMP plugin version
+SELECT
+  client,
+  amp_plugin_version,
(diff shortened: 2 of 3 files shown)


sql/2019/14_CMS/14_02.sql
```diff
@@ -0,0 +1,24 @@
...
+    `httparchive.almanac.summary_response_bodies`
+  WHERE
+    firstHtml)
+JOIN
+  (SELECT _TABLE_SUFFIX AS client, url FROM `httparchive.technologies.2019_07_01_*` WHERE app = 'WordPress')
+USING
+  (client, url)
+GROUP BY
+  client,
+  amp_plugin_version
```
obto: 1. Results are showing there are only 386 mobile websites out of 3 million wordpress sites which use this plugin... when the plugin has 400k installs. Looking closer it appears the meta generator tag only appears when the plugin is set to AMP-first mode.

    Is this enough data to draw any conclusions?

2. Will we be treating AMP versions like the following as being the same version, or different?

    `1.2.0; mode=standard; experiences=website`
    `1.2.0; mode=standard; experiences=website,stories`

sql/2019/14_CMS/14_03.sql
```diff
@@ -0,0 +1,26 @@
...
+    firstHtml)
+INNER JOIN
+  (SELECT _TABLE_SUFFIX AS client, url FROM `httparchive.technologies.2019_07_01_*` WHERE app = 'WordPress')
+USING
+  (client, url)
+GROUP BY
+  client,
+  amp_plugin_mode
+ORDER BY
+  freq / total DESC
```
obto: Same question as above regarding the few amount of sites we can detect.

sql/2019/14_CMS/14_01.sql
```diff
@@ -0,0 +1,22 @@
...
+  app AS cms,
+  COUNT(0) AS freq,
+  total,
+  ROUND(COUNT(0) * 100 / total, 2) AS pct
+FROM
+  `httparchive.technologies.2019_07_01_*`
+JOIN
+  (SELECT _TABLE_SUFFIX, COUNT(0) AS total FROM `httparchive.summary_pages.2019_07_01_*` GROUP BY _TABLE_SUFFIX)
+USING
+  (_TABLE_SUFFIX)
```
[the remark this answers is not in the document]
rviscomi: The sum of `pct`s is around 79% because not every page uses a CMS. The ones detected as having multiple CMSs is definitely a bug and would be worth investigating on the side.

sql/2019/14_CMS/14_02.sql
rviscomi: Definitely true. I chatted with @amedina about this and he agreed that there's not enough representation of the plugin in the dataset to draw meaningful conclusions.

PR #142, opened 2019-09-07
Fonts queries

rviscomi:
Progress on #87 

[Results sheet](https://docs.google.com/spreadsheets/d/108g6LXdC3YVsxmX1CCwrmpZ3-DmbB8G_wwgQHX5pn6Q/edit?usp=sharing)

### DIFF
--- a/sql/2019/06_Fonts/06_31.sql
+++ b/sql/2019/06_Fonts/06_31.sql
@@ -1,0 +1,29 @@
+#standardSQL
+# 06_31: Popular fonts (see also 02_23)
+CREATE TEMPORARY FUNCTION getFontFamilies(css STRING)
+RETURNS ARRAY<STRING> LANGUAGE js AS '''
+try {
+  var $ = JSON.parse(css);
+  return $.stylesheet.rules.filter(rule => rule.type == 'font-face').map(rule => {
+    var family = rule.declarations && rule.declarations.find(d => d.property == 'font-family');
+    return family && family.value.replace(/['"]/g, '');
+  }).filter(family => family);
+} catch (e) {
+  return [];
+}
+''';
+
+SELECT
+  client,
+  font_family,
+  COUNT(0) AS freq,
+  SUM(COUNT(0)) OVER (PARTITION BY client) AS total,
+  ROUND(COUNT(0) * 100 / SUM(COUNT(0)) OVER (PARTITION BY client), 2) AS pct
+FROM
+  `httparchive.almanac.parsed_css`,
+  UNNEST(getFontFamilies(css)) AS font_family
+GROUP BY
+  client,
+  font_family
+ORDER BY
+  freq / total DESC
--- a/sql/2019/06_Fonts/06_39.sql
+++ b/sql/2019/06_Fonts/06_39.sql
@@ -1,0 +1,45 @@
+#standardSQL
+# 06_39-41: Font formats declared together
+CREATE TEMPORARY FUNCTION getFontFormats(css STRING)
+RETURNS ARRAY<STRING> LANGUAGE js AS '''
+try {
+  var reduceValues = (values, rule) => {
+    if ('rules' in rule) {
+      return rule.rules.reduce(reduceValues, values);
+    }
+    if (!('declarations' in rule)) {
+      return values;
+    }
+    if (rule.type != 'font-face') {
+      return values;
+    }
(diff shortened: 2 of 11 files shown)


sql/2019/06_Fonts/06_42.sql
```diff
@@ -0,0 +1,23 @@
...
+  (SELECT _TABLE_SUFFIX AS client, COUNT(0) AS total FROM `httparchive.summary_pages.2019_07_01_*` GROUP BY _TABLE_SUFFIX)
+USING
+  (client)
+WHERE
+  type = 'font' AND
+  # Color fonts have any of sbix, cbdt, svg or colr tables.
+  JSON_EXTRACT_SCALAR(payload, '$._font_details.table_sizes.sbix') IS NOT NULL OR
+  JSON_EXTRACT_SCALAR(payload, '$._font_details.table_sizes.cbdt') IS NOT NULL OR
+  JSON_EXTRACT_SCALAR(payload, '$._font_details.table_sizes.svg') IS NOT NULL OR
+  JSON_EXTRACT_SCALAR(payload, '$._font_details.table_sizes.colr') IS NOT NULL
```
obto: Was getting only a single occurrence here. Turns out we need to use case insensitivity here.

```suggestion
  ARRAY_LENGTH(REGEXP_EXTRACT_ALL(JSON_EXTRACT(payload, '$._font_details.table_sizes'), '(?i)(sbix|cbdt|svg|colr)')) > 0
```

sql/2019/06_Fonts/06_46.sql
```diff
@@ -0,0 +1,23 @@
+#standardSQL
+# 06_46: % of pages linking to a Google Fonts stylesheet
+CREATE TEMP FUNCTION preloadsGoogleFont(payload STRING)
+RETURNS BOOLEAN LANGUAGE js AS '''
+  try {
+    var $ = JSON.parse(payload);
+    var almanac = JSON.parse($._almanac);
+    return !!almanac['link-nodes'].find(e => e.rel.toLowerCase() == 'stylesheet' && e.href.startsWith('https://fonts.googleapis.com'));
```
obto: Don't forget about http://
```suggestion
    return !!almanac['link-nodes'].find(e => e.rel.toLowerCase() == 'stylesheet' && (e.href.startsWith('https://fonts.googleapis.com') || e.href.startsWith('http://fonts.googleapis.com')));
```

sql/2019/06_Fonts/06_31.sql
```diff
@@ -0,0 +1,29 @@
+#standardSQL
+# 06_31: Popular fonts (see also 02_23)
+CREATE TEMPORARY FUNCTION getFontFamilies(css STRING)
+RETURNS ARRAY<STRING> LANGUAGE js AS '''
+try {
+  var $ = JSON.parse(css);
+  return $.stylesheet.rules.filter(rule => rule.type == 'font-face').map(rule => {
+    var family = rule.declarations && rule.declarations.find(d => d.property == 'font-family');
```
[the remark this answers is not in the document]
rviscomi: > This is a bit pedantic, but in the case where there are multiple font-families supplied in a block, this returns the first occurrence instead of the last (which is what would be applied).

That's true. I'll make a note for the authors to be aware of this, but it's probably so subtle as to not make a difference in the aggregate.

> Will we be combining different versions of fonts like FontAwesome and Font Awesome 5 Free?

I'll leave that for the authors to interpret manually :innocent:

sql/2019/06_Fonts/06_39.sql
```diff
@@ -0,0 +1,45 @@
...
+    if ('rules' in rule) {
+      return rule.rules.reduce(reduceValues, values);
+    }
+    if (!('declarations' in rule)) {
+      return values;
+    }
+    if (rule.type != 'font-face') {
+      return values;
+    }
+    // From `url(basic-sans-serif.ttf) format("truetype")` get `truetype`.
```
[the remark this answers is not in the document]
rviscomi: Doh. Thanks!

sql/2019/06_Fonts/06_42.sql
rviscomi: Good idea/catch

sql/2019/06_Fonts/06_46.sql
rviscomi: Good catch. This actually increased the results from ~25 to ~30%.

PR #150, opened 2019-09-09
Accessibility queries [1 of 2]

dougsillars:


sql/2019/04_Media/04_01.sql
```diff
@@ -0,0 +1,36 @@
+#standardSQL
```
rviscomi: Can you revert all 04_Media changes in this PR?

PR #153, opened 2019-09-11
A11Y queries

obto:


obto:
Querying for 09.30 in regards to the % of input elements that meet this criteria is impossible without another custom metric. Instead I opted to split the metric into 2 parts:

1. % of sites which have properly accessible inputs via Axe in lighthouse
2. Usage rates of aria-label and aria-labeledby. Both by % of inputs, and % of sites

Issue #106, opened 2019-07-29
Implement home page design


rviscomi:
@HTTPArchive/developers 

Using the design in #46, implement the HTML/CSS/JS. For any assets that we don't have (specific fonts or images) use placeholders for now.

tanhengyeow:
@rviscomi I'm interested in this 😄 Do we want to go with mobile-first or desktop-first implementation?

rviscomi:
Desktop first, just because we're still waiting on the mobile designs 😁

tanhengyeow:
@rviscomi Do we want to consider using [normalize.css](https://necolas.github.io/normalize.css/) or is custom reset preferred?

rviscomi:
Yeah normalize SGTM 👍

tanhengyeow:
@rviscomi Are the mobile designs/assets needed for the home page finalized? If so I can factor that in the current PR as well. 

Another approach would be that I finish up the current PR taking care of the desktop design first and follow up the mobile design in a separate PR :)

rviscomi:
Yes, the mobile home page design is final and ready to be implemented. I think it'd make sense to do both desktop/mobile responsively in one PR since they will share the same code. Does that sound good?

tanhengyeow:
Yep sounds great since they are all finalized 😄

rviscomi:
Here's the link to the interactive designs with colors/fonts/etc annotated: https://projects.invisionapp.com/d/main?origin=v7#/console/18174420/379965024/inspect

rviscomi:
@HTTPArchive/developers: is anyone available to help finish the implementation of the home page design? @tanhengyeow and I have made some progress on this in the [design branch](https://github.com/HTTPArchive/almanac.httparchive.org/tree/design). It's not quite fully responsive and there are some graphical elements still to be implemented, like the header background (see the [original designs](https://projects.invisionapp.com/d/main?origin=v7#/console/18174420/379965024/inspect)). If you're able to contribute, feel free submit a PR into the branch.

The goal is to have two stylesheets: one specifically for the home page and another more general stylesheet that is for the site-wide design.

MSakamaki:
I can will help.
python is a beginner, javascript/css/html is good.

I will try to find implementation I can do it.

MSakamaki:
@rviscomi 
sorry.
The link of the above link (original design) does not seem to be seen due to insufficient authority.
How can I view it?

rviscomi:
[...]

PR #156, opened 2019-09-13
Implement home page design (mobile)

tanhengyeow:
Part of #106

### DIFF
--- a/src/static/css/index.css
+++ b/src/static/css/index.css
@@ -34,7 +34,12 @@ button:hover {
 /* Shared */
 #home {
 	display: grid;
-	grid-template-rows: 2fr 1fr 1fr 2fr 1fr;
+	grid-template-areas:
+		'header'
+		'chapter'
+		'contributors'
+		'methodology'
+		'footer';
 	height: 100vh;
 }
 
@@ -53,7 +58,6 @@ footer,
 }
 
 .nav-items {
-	flex-shrink: 0;
 	display: flex;
 	align-items: center;
 }
@@ -75,7 +79,6 @@ footer,
 
 .featured-chapter,
 .methodology-container {
-	width: 100%;
 	display: flex;
 	justify-content: center;
 }
@@ -119,6 +122,8 @@ footer,
 /* Header */
 header {
 	background-color: #5C687D;
+	grid-area: header;
+	width: 100vw;
 }
 
 .select-year {
@@ -140,16 +145,14 @@ header {
 	margin: 0;
 }
 
-.logo,
-.explore {
-	flex-shrink: 0;
-}
-
+.menu-icon,
+.menu-close-icon,
 .menu {
 	display: none;
 }
 
 .intro-container {
+	position: relative;
 	display: grid;
 	grid-template-areas:
 		'intro year'
@@ -177,25 +180,31 @@ header {
 
 .intro-image {
 	grid-area: image;
-	position: relative;
+	display: grid;
+	grid-template-columns: repeat(3, 1fr);
+	grid-template-rows: repeat(3, 1fr);
+	width: 600px;
+	height: 500px;
+}
+
+.home-browser {
+	grid-row: 1 / 3;
+	grid-column: 1 / 3;
 }
 
 .camada-751 {
-	position: absolute;
-	bottom: -50px;
-	left: 0px;
+	grid-area: 3 / 1;
 }
 
 .camada-759 {
-	position: absolute;
-	bottom: 0px;
-	left: 200px;
+	grid-area: 3 / 2;
+	align-self: flex-end;
 }
 
 .camada-752 {
-	position: absolute;
-	bottom: 0px;
-	right: -115px;
+	grid-area: 3 / 3;
+	align-self: flex-end;
+	margin-left: 120px;
 }
 
 .explore {
@@ -230,6 +239,11 @@ header {
 }
 
 /* Featured Chapter */
+.featured-chapter {
+	grid-area: chapter;
+	width: 100vw;
+}
+
 .featured-chapter h1 {
 	margin-bottom: 0;
 	font-size: 3rem;
@@ -280,9 +294,11 @@ header {
 
 /* Contributors */
 .contributors-container {
+	grid-area: contributors;
 	display: flex;
 	align-items: center;
 	justify-content: space-between;
+	width: 100vw;
 }
 
 .rectangles {
@@ -302,7 +318,7 @@ header {
 
 .camada-755 {
 	position: absolute;
-	top: -200px;
+	top: -210px;
 	right: 100px;
 }
 
@@ -338,6 +354,11 @@ header {
 }
 
 /* Methodology */
+.methodology-container {
+	grid-area: methodology;
+	width: 100vw;
+}
+
 .methodology {
 	display: flex;
 	flex-direction: column;
@@ -371,7 +392,7 @@ header {
 .data-processed div {
 	font-size: 14px;
 	font-weight: bold;
-	margin: 0 10px;
+	margin: 10px 10px;
 }
 
 .websites-tested div:last-of-type,
@@ -392,8 +413,17 @@ header {
 }
 
 /* Footer */
-.footer-container {
-	margin: 0 200px;
+.footer {
+	grid-area: footer;
+	width: 100vw;
+}
+
+.footer hr:last-of-type {
+	display: none;
+}
+
+.footer-mobile-nav {
+	display: none;
 }
 
 .misc {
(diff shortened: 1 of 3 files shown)


tanhengyeow:
@rviscomi With this PR, the home page design is near completion. There is definitely room for improvement that we can iterate on but the UI/UX is closely matched according to the mockups 😄

tanhengyeow:
@rviscomi With this PR, the home page design is near completion. There is definitely room for improvement that we can iterate on but the UI/UX now is closely matched according to the mockups 😄

rviscomi:
Testing this out locally. Even at desktop widths, I'm getting the mobile layout. Is that intentional?
image

tanhengyeow:
@rviscomi Yes, right now the media query is:
```
/* Mobile View */
@media (max-width: 1440px) {
```
So the desktop view would be shown at 1440px. I was thinking we can tweak accordingly based on iterations 😄

src/static/js/home.js
```diff
@@ -0,0 +1,15 @@
+const menu = document.querySelector(".menu");
```
rviscomi: Let's use single quotes for consistency throughout the JS
```suggestion
const menu = document.querySelector('.menu');
```

src/static/js/home.js
```diff
@@ -0,0 +1,15 @@
...
+const menuIcon = document.querySelector(".menu-icon");
+const menuCloseIcon = document.querySelector(".menu-close-icon");
+
+menuIcon.addEventListener('click', () => {
+  menu.style.display = "flex";
```
rviscomi: Rather than changing the specific styles in JS, let's toggle a "hidden" class on the necessary elements, which could be a generic CSS rule to `display: none`. Removing the class should revert it back to its original display value, so the JS doesn't need to care whether it was flex or block or inline etc.

src/templates/en/index.html
```diff
@@ -69,11 +70,50 @@ <h2>
...
-        <img class="camada-752" src="/static/images/camada-752.png" />
+      <div class="intro-image-container">
+        <div class="intro-image">
+          <img class="home-browser" src="/static/images/home-browser.png" />
+          <img class="camada-751" src="/static/images/camada-751.png" />
+          <img class="camada-759" src="/static/images/camada-759.png" />
+          <img class="camada-752" src="/static/images/camada-752.png" />
+        </div>
+      </div>
+      <div class="menu">
```
rviscomi: Any reason why the menu isn't part of the nav section, adjacent to the menu icons?

src/templates/en/index.html
```diff
@@ -69,11 +70,50 @@ <h2>
...
-        <img class="home-browser" src="/static/images/home-browser.png" />
-        <img class="camada-751" src="/static/images/camada-751.png" />
-        <img class="camada-759" src="/static/images/camada-759.png" />
-        <img class="camada-752" src="/static/images/camada-752.png" />
+      <div class="intro-image-container">
+        <div class="intro-image">
+          <img class="home-browser" src="/static/images/home-browser.png" />
+          <img class="camada-751" src="/static/images/camada-751.png" />
+          <img class="camada-759" src="/static/images/camada-759.png" />
+          <img class="camada-752" src="/static/images/camada-752.png" />
```
rviscomi: The red character appears stretched out and misaligned with the background graphic

image


Would it be easier if we flatten the browser graphic with the three characters into a single image? Or is there an advantage to having them separate for desktop/mobile layout?

src/static/css/index.css
```diff
@@ -422,3 +452,257 @@ header {
 	width: 25px;
 	opacity: 0.3;
 }
+
+/* Mobile View */
+@media (max-width: 1440px) {
```
rviscomi: Let's use a 600px breakpoint for mobile. I'm also noticing that smaller desktop viewport sizes appear to have overflowing images:

image

src/templates/en/index.html
```diff
@@ -52,7 +52,8 @@
         </select>
         <img src="/static/images/dropdown-arrow.png" />
       </div>
-      <img class="menu" src="/static/images/menu.png" />
+      <img class="menu-icon" src="/static/images/menu.png" />
+      <img class="menu-close-icon" src="/static/images/cross.png" />
     </section>
     <section class="intro-container">
       <div class="intro">
```
rviscomi: Paragraphs of text should use `<p>` and have a font family of Lato.

src/templates/en/index.html
```diff
@@ -69,11 +70,50 @@ <h2>
...
+        <li>
+          <a href="#">Contributors</a>
+        </li>
+        <li>
+          <a href="#">Methodology</a>
+        </li>
+        <div>
+          <img src="/static/images/ha-home.png" />
+          <div class="social-media">
+            <img src="/static/images/twitter.png" />
```
rviscomi: Link to https://twitter.com/HTTPArchive

src/templates/en/index.html
```diff
@@ -69,11 +70,50 @@ <h2>
...
+            <img src="/static/images/github.png" />
```
rviscomi: Link to https://github.com/HTTPArchive/almanac.httparchive.org

src/templates/en/index.html
rviscomi: For each of these social links, they should be gray/transparent by default and yellow on hover. We should be able to use CSS filters to adjust the color overlay if we're not already using SVGs.

image

src/templates/en/index.html
```diff
@@ -69,11 +70,50 @@ <h2>
...
+          <a href="#">Table of Contents</a>
+        </li>
+        <li>
+          <a href="#">Contributors</a>
+        </li>
+        <li>
+          <a href="#">Methodology</a>
+        </li>
+        <div>
+          <img src="/static/images/ha-home.png" />
```
rviscomi: Link to https://httparchive.org/

Also add `alt` attributes for all `img` elements, even if empty for now.
```suggestion
          <a href="https://httparchive.org/">
            <img src="/static/images/ha-home.png" alt="HTTP Archive home" />
          </a>
```

rviscomi: Ditto for each of these images in the footer

src/templates/en/index.html
```diff
@@ -191,6 +231,25 @@ <h1>Methodology</h1>
         </div>
       </section>
       <hr>
```
rviscomi: Two changes to the `hr`s in the footer:

- lower the opacity to 20%
- should have the same horizontal padding as other footer sections, not extending to the edges of the viewport

src/static/css/index.css
```diff
@@ -422,3 +452,257 @@ header {
...
+	.intro-container .menu li,
+	footer .footer-mobile-nav li {
+		list-style: none;
+		margin: 30px;
+	}
+
+	.intro-container .menu a,
+	footer .footer-mobile-nav a {
+		text-decoration: none;
+		color: white;
```
rviscomi: This and other footer text should be #f2f2f2

src/templates/en/index.html
```diff
@@ -191,6 +231,25 @@ <h1>Methodology</h1>
...
+          <div class="social-media">
+            <img src="/static/images/twitter.png" />
+            <img src="/static/images/github.png" />
+          </div>
+        </div>
+      </div>
+      <hr>
       <div class="misc">
         <div>
           Copyright © 2019 Web Almanac. All rights reserved.
```
rviscomi: Lower the font size to 12px

src/templates/en/index.html
```diff
@@ -69,11 +70,50 @@ <h2>
...
+                {% for l in supported_languages %}
+                  <option
+                    {% if l == language %}selected="selected"{% endif %}
+                    value="{{ url_for(request.endpoint, **get_view_args(lang=l.lang_code)) }}">
+                    {{ l }}
+                  </option>
+                {% endfor %}
+              {% endif %}
+            </select>
+            <img src="/static/images/dropdown-arrow.png" />
```
rviscomi: Use a filter to make this appear white

src/templates/en/index.html
tanhengyeow: Yup, that would be easier 😄 The only difference between the desktop and mobile layout is that their sizes differ.

So either we have 1 image that can be scaled nicely for both desktop and mobile layouts or 2 images (one optimized for desktop and one optimized for mobile). WDYT?

src/templates/en/index.html
tanhengyeow: image

The menu now appears below the nav bar and placed on top of the `intro-container` section.

src/static/css/index.css
tanhengyeow: > I'm also noticing that smaller desktop viewport sizes appear to have overflowing images

Yup, that's because I used a fixed height and width for these images currently. 

I think the suggestion to flatten the browser graphic with the three characters into a single image would be great so that responsive images can be accounted more easily here.

rviscomi:
Merging as is and we can iterate further

Issue #89, opened 2019-07-23
Query metrics: Chapter 8. Security


rviscomi:
Part | Chapter | Authors | Reviewers | Tracking Issue
-- | -- | -- | -- | --
II. User Experience | 8. Security | @arturjanc @ScottHelme | @paulcalvano @bazzadp @ghedo @ndrnmnn | #10

# READ ME!
All of the metrics in the table below have been marked as `Able To Query` during the [metrics triage](https://github.com/HTTPArchive/almanac.httparchive.org/issues/33). The analyst assigned to each metric is expected to write the corresponding query and submit a PR to [have it reviewed and added to the repo](https://github.com/HTTPArchive/almanac.httparchive.org/issues/62).

In order to stay on schedule and have the data ready for authors, **please have all metrics reviewed and merged by August 5**.

## Assignments

ID | Metric description | Analyst | Notes
-- | -- | -- | --
08.01 | SSLv2 / SSLv3 / TLSv1.0 / TLSv1.1 / TLSv1.2 / TLSv1.3 | @dotjs | request -> _securityDetails
08.02 | Unique CA issuers | @dotjs | request -> _securityDetails
08.03 | RSA certificates | @dotjs | request -> _securityDetails
08.04 | ECDSA certificates | @dotjs | request -> _securityDetails
08.06 | Suites supporting Forward Secrecy (ECDHE / DHE) | @dotjs |  
08.07 | Authenticated suites (GCM / CCM) | @dotjs |  
08.08 | Modern suites (AES GCM, ChaCha20-Polyc1305) | @dotjs |  
08.09 | Legacy suites (AES CBC, 3DES, RC4) | @dotjs |  
08.13 | Sites with degraded HTTPS UI (mixed-content) | @dotjs | possibly find http resources on a https origin request
08.14 | Policies with frame-ancestors | @dotjs | request -> headers regex
08.15 | Policies with 'nonce-*' | @dotjs | request -> headers regex
08.16 | Policies with 'hash-*' | @dotjs | request -> headers regex
08.17 | Policies with 'unsafe-inline' | @dotjs | request -> headers regex
08.18 | Policies with 'unsafe-eval' | @dotjs | request -> headers regex
08.19 | Policies with 'strict-dynamic' | @dotjs | request -> headers regex
08.20 | Policies with 'trusted-types' | @dotjs | request -> headers regex
08.21 | Policies with 'upgrade-insecure-requests' | @dotjs | request -> headers regex
[...]

PR #157, opened 2019-09-14
Security queries revisited

rviscomi:
Revised security queries. Progress on #89.

All results saved to [this sheet](https://docs.google.com/spreadsheets/d/1Zq2tQhPE06YZUcbzryRrBE6rdZgHHlqEp2XcgS37cm8/edit?usp=sharing).


sql/2019/08_Security/08_23-24.sql
```diff
@@ -0,0 +1,22 @@
+#standardSQL
```
obto: What about a metric testing for how many sites meet the requirements of being in chrome's preload list? Could be interesting to see how many are ready to be added vs how many are in the preload list.

https://hstspreload.org/

sql/2019/08_Security/08_29.sql
```diff
@@ -1,22 +1,31 @@
...
+} catch (e) {
+  return null;
+}
+''';
 
 SELECT
-  _TABLE_SUFFIX AS client,
-  policy,
+  client,
+  NORMALIZE_AND_CASEFOLD(extractHeader(payload, 'X-Content-Type-Options')) AS value,
```
obto: Lot of interesting unicode characters get used here, like , huh... How'd you think to look for this?

sql/2019/08_Security/08_35.sql
```diff
@@ -1,23 +1,35 @@
 #standardSQL
```
obto: Looks good, but when I run this in BQ I'm getting different results than what you have in the google doc

| client | http_only | total	 | pct   |
|--------|-----------|---------|-------|
| desktop| 467463    | 4371973 | 10.69 |
| mobile | 520748    | 5297442 | 9.83  |

sql/2019/08_Security/08_39.sql
```diff
@@ -1,20 +1,13 @@
...
+  COUNTIF(REGEXP_CONTAINS(body, '(?i)<(?:link|script)[^>]*integrity=')) AS freq,
+  COUNT(0) AS total,
+  ROUND(COUNTIF(REGEXP_CONTAINS(body, '(?i)<(?:link|script)[^>]*integrity=')) / COUNT(0), 2) AS pct
 FROM
   `httparchive.almanac.summary_response_bodies`
 WHERE
   firstHtml
 GROUP BY
-  client  
+  clien
```
obto: Should be `client`

sql/2019/08_Security/08_07.sql
```diff
@@ -1,7 +1,18 @@
...
+  ROUND(COUNTIF(REGEXP_CONTAINS(cipher, r'GCM|CCM|POLY1305')) * 100 / COUNT(0), 2) AS pct
+FROM (
+  SELECT
+    _TABLE_SUFFIX,
+    JSON_EXTRACT(payload, '$._securityDetails.cipher') AS cipher,
+    JSON_EXTRACT_SCALAR(payload, '$._securityDetails.protocol') AS protocol
+  FROM
+   `httparchive.requests.2019_07_01_*`)
+WHERE
+  protocol IS NOT NULL
```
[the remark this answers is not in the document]
rviscomi: Good call.

sql/2019/08_Security/08_23-24.sql
rviscomi: Good idea!

sql/2019/08_Security/08_29.sql
rviscomi: Could have used `LOWER`. Normalization may be overkill.

sql/2019/08_Security/08_35.sql
rviscomi: Really good catch.

After some debugging, this has to do with `return header.split(':')[1].trim();`. Header values may have a `:` in their value (like an expires directive) and this would be cutting off the return value too short.

Many other metrics use this splitting approach, so I'll have to go through them all with a fix.

The reason for the difference in results is because earlier I had used `return header.split(': ')[1]` which included a space after the colon, and was less likely to split up things like expires directives, which have numbers after it. I updated all queries to use the trim approach, assuming it wouldn't affect the results.

A proper fix is to take a substring after the first colon rather than splitting on any colons.


Thanks for catching this!!

sql/2019/08_Security/08_39.sql
rviscomi: 🤦‍♂ bad copypasta

PR #160, opened 2019-09-15
Use 9_26 for 12_04 instead

obto:
12_04 as is, uses the lighthouse viewport test to determine if the viewport is scalable or not. However, this LH test can fail on more than just the viewport being unscalable.

Let's use the query for 9.26 here instead.


rviscomi:
I've updated 12_04 in the [results sheet](https://docs.google.com/spreadsheets/d/1dPBDeHigqx9FVaqzfq7CYTz4KjllkMTkfq4DG4utE_g/edit#gid=586907091) with the results from 09_26.

Issue #134, opened 2019-09-03
Write content: Chapter 5. Third Parties


rviscomi:
Part | Chapter | Authors | Reviewers | Analysts | Metrics | Analysis
-- | -- | -- | -- | -- | -- | --
I. Page Content | 5. Third Parties | @patrickhulce | @simonhearne @flowlabs @jasti @zeman | @patrickhulce | #8 | #86

### Due by: October 7

- [x] Write the contents of your chapter and get it peer reviewed (@simonhearne @flowlabs @jasti @zeman).
- [ ] File a PR to merge your written contents in markdown format to the [src/content/en/2019](https://github.com/HTTPArchive/almanac.httparchive.org/tree/master/src/content/en/2019) directory.

Refer to the [Authors' Guide](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide) for writing advice.

rviscomi:
👋 Greetings! Just checking in to make sure this is progressing and remind you of the October 7 deadline timed so we can launch at Chrome Dev Summit.

The checklist above indicates that you still need to write content and get it reviewed. Feel free to reach out if you have any questions about the process.

patrickhulce:
Alright @simonhearne @flowlabs @jasti @zeman, it's your time to shine! :)

Comments/feedback/reviews/edits welcome: https://docs.google.com/document/d/1q52f7pRSO0g6J1zbiVq-iVCYdGs6UdYQTCLOAFglz94/edit?usp=sharing

patrickhulce:
Hey reviewers! Friendly reminder to try to get your thoughts in on the chapter **by EOD tomorrow** if you'd like your feedback to be included, so I can turn it around over the weekend.

rviscomi:
🛎 @simonhearne @zeman @flowlabs @jasti

rviscomi:
@patrickhulce could you start a PR with the markdown version of your doc and perhaps we can get it peer reviewed on the PR?

patrickhulce:
Sure @rviscomi  👍

patrickhulce:
Reviewers please see https://github.com/HTTPArchive/almanac.httparchive.org/pull/189 instead of the Google doc now :)

PR #175, opened 2019-09-26
Chapter 5: Put unidentified third parties into "Other"

patrickhulce:
While writing up https://github.com/HTTPArchive/almanac.httparchive.org/issues/134 I bumped into this issue with my queries where things weren't adding up, realized I forgot to also put the unidentified domains into the "Other" bucket :)

### DIFF
--- a/sql/2019/05_Third_Parties/05_03.sql
+++ b/sql/2019/05_Third_Parties/05_03.sql
@@ -10,12 +10,17 @@ FROM (
   SELECT
       client,
       type AS contentType,
-      ThirdPartyTable.category AS thirdPartyCategory
+      IFNULL(ThirdPartyTable.category,
+        IF(DomainsOver50Table.requestDomain IS NULL, 'first-party', 'other')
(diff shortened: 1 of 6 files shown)


sql/2019/05_Third_Parties/05_03.sql
```diff
@@ -10,12 +10,17 @@ FROM (
   SELECT
       client,
       type AS contentType,
-      ThirdPartyTable.category AS thirdPartyCategory
+      IFNULL(ThirdPartyTable.category,
+        IF(DomainsOver50Table.requestDomain IS NULL, 'first-party', 'other')
```
obto: Why are requests in which the domain is unknown (null) treated as first-party? Wouldn't this be the type of request you'd put in the other category?

patrickhulce: `DomainsOver50Table` contains a list of the all the third-party domains. If the request domain is not a third-party domain, it's assumed to be first-party.

Issue #84, opened 2019-07-23
Query metrics: Chapter 3. Markup


rviscomi:
Part | Chapter | Authors | Reviewers | Tracking Issue
-- | -- | -- | -- | --
I. Page Content | 3. Markup | @bkardell | @zcorpan | #5

# READ ME!
All of the metrics in the table below have been marked as `Able To Query` during the [metrics triage](https://github.com/HTTPArchive/almanac.httparchive.org/issues/33). The analyst assigned to each metric is expected to write the corresponding query and submit a PR to [have it reviewed and added to the repo](https://github.com/HTTPArchive/almanac.httparchive.org/issues/62).

In order to stay on schedule and have the data ready for authors, **please have all metrics reviewed and merged by August 5**.

## Assignments

ID | Metric description | Analyst | Notes
-- | -- | -- | --
03.01 | Deprecated elements | @rviscomi |  
03.02 | Popular elements | @rviscomi |  
03.03 | Custom elements (“slang”) | @rviscomi |  
03.05 | count of shadowRoots | @rviscomi |  

## Checklist of metrics to be merged

- [x] 	03.01	Deprecated elements
- [x] 	03.02	Popular elements
- [x] 	03.03	Custom elements (“slang”)
- [x] 	03.05	count of shadowRoots

bkardell:
Sorry, there's a lot of tasks and issues and things here, I keep popping over once in a while to see the status of this issue, but I'd like to make sure I'm not confused.. we don't have data for this yet, right?

rviscomi:
> we don't have data for this yet, right?

That's right, we're a bit behind on the analysis. This chapter only has 4 metrics so I'll see if I can unblock you.

rviscomi:
@bkardell I'm writing queries and saving the results to [this sheet](https://docs.google.com/spreadsheets/d/1WnDKLar_0Btlt9UgT53Giy2229bpV4IM2D_v6OM_WzA/edit?usp=sharing). I've also included a comment for each metric explaining how to read the results.

PR #178, opened 2019-09-26
More markup queries

rviscomi:
Addition to #84 

image

image


bkardell:
Tagging in @zcorpan who is better with the big table sql than I to review the actual queries, and also so he is aware I am adding these to the doc which he did his initial reviews on just today

rviscomi:
As long as the spreadsheet output is what you're looking for, no need for a detailed review of the underlying SQL.

PR #182, opened 2019-10-01
A11Y query updates + bug fixes

obto:
This PR contains two things:

1. A couple fixes to the A11Y chapter queries - **Still in progress**
2. Various bugs I've found in other queries

I have also noticed there are a lot of queries not using case-insensitive regex when querying the page body. Would you like me to update them in this PR or make another?


obto:
Alright sounds good. I'll start another PR for those.

And yes, I've updated the affected metrics in the datasheet. Still have some A11Y queries to look at, before I merge this PR however

PR #183, opened 2019-10-02
fix responsive of home page

### LINKED ISSUE #106: Implement home page design

MSakamaki:
fix #106

Remaining challenges:
There is a problem with the number of people in the contributor section between resolution 901 ~ 1100.

This problem is difficult to solve with the current layout.


src/templates/en/2019/index.html
```diff
@@ -25,8 +25,15 @@ <h2>
...
-    <img class="intro-image" src="/static/images/home-hero.png" alt="">
+
+    <svg class="intro-image" width="100%" height="100%" viewBox="0 0 715 812"
+    xmlns="http://www.w3.org/2000/svg" 
+    xmlns:xlink="http://www.w3.org/1999/xlink">
+      <text x="20" y="250" class="intro-year">2019</text>
+      <image xlink:href="/static/images/home-hero.png"
+      x="0" y="230"
+      height="562" width="715" />    
+    </svg>
```
rviscomi: I haven't seen this technique before. What's the advantage of using an SVG wrapper for the image?

MSakamaki: I did this because I wanted to make the image and text responsive in the same width units.
Certainly, this is a bit tricky and has side effects.

Stop this hack, I think it is best to include letters in the illustrations.
Prepare new images with illustrations and 2019 text.
I fix it right away because.

MSakamaki:
Additional actions:
Fixed text center when mobile  (I checked from the design shared on slack).

src/templates/en/2019/index.html
rviscomi: > I think it is best to include text in the illustrations

It might be easier to implement but I'm not sure about things like a11y and i18n. @obto WDYT?

obto: Just to clarify, we're talking about this "2019" text correct?
screenshot-projects invisionapp com-2019 10 03-12_20_56

Assuming we are:
1. Text inside SVGs is accessible so there's no problem there, but I think it'd be easiest to put that inside the image itself, and just make sure it's readable when displayed at smaller sizes.
2. Is this graphic informative to readers? Or is it merely presentational? If it's the latter (which I feel it is) we can just give it `alt=""`

rviscomi: I think @MSakamaki's most recent suggestion is to [put the text in the PNG itself](https://github.com/HTTPArchive/almanac.httparchive.org/blob/51650c655ed4ebdda90a8e26d3ed3462bd6f59b1/src/static/images/home-hero-2019.png). It's true that we can use alt to make it accessible, but I do feel like the "2019" text itself is important information. In the future when we have multiple editions of the Almanac, this will be the most prominent (only?) indicator of which version the user is currently viewing.

Another consideration is that the responsive mobile design separates the text from the graphic, so we need to have the ability to position them independently.

obto: Both versions can be accessible. For the images an alt of "The Web Almanac 2019 edition", or similarly a <title> inside the svg with the same text would work well.

So I think either option is good so long as its marked up correctly.

MSakamaki: Thanks @obto @rviscomi  for the review.

The reason I create the image is that prevent the 2019 word from becoming instability while loading the web fonts.

I agree "2019" text of important information.
Update alt attribute of image.

rviscomi: For the mobile view we want to have the "2019" prominently at the top of the page, above the H1. Here's how it looks in the [design](https://projects.invisionapp.com/d/?origin=v7?origin=v7#/console/18174420/379965024/preview):

image

Here's how it appears in this branch currently: 

image

MSakamaki: @rviscomi 
Sorry, I was looking at the old mobile layout.
I fixed the intro part to match the shared URL page design.

PR #188, opened 2019-10-04
Both '0' and 'no' will prevent zooming

obto:
Changes have been applied the A11Y and Mobile web datasheets


rviscomi:
See https://twitter.com/rick_viscomi/status/1172012824819916801. Browsers fall back to "no" very leniently.

obto:
Wow... That's crazy. I'll update the query and see how that affects things

obto:
I investigated the usage of all the viewport directives, and found =0 and =no cover nearly all of the cases where scaling would get disabled.

https://docs.google.com/spreadsheets/d/1dPBDeHigqx9FVaqzfq7CYTz4KjllkMTkfq4DG4utE_g/edit#gid=2087670861 

What ended up having a significant impact,however, was including maximum-scale=1. Numbers jumped to 33% of sites disable zooming via the viewport

Issue #130, opened 2019-08-30
Implement Chapter page design


rviscomi:
See below for the Chapter mockup. This will be the design template for non-home pages.

@HTTPArchive/developers 

![chapter-001f](https://user-images.githubusercontent.com/1120896/63997472-9b59bf00-cacc-11e9-9c34-b9f722e05d1d.jpg)

tanhengyeow:
Some questions I have in mind:
1. Is the side navigation menu's position intended (some margin to the right)? Or should we flush it to the left?
2. I've not looked at `generate_chapters.py` in detail yet but from the looks of it it
   - HTML rendering is covered for this part:
![image](https://user-images.githubusercontent.com/25320782/64004342-a6e6cd80-cb40-11e9-8035-479c1634b6a3.png)
   - Graph visualization is covered for this part:
![image](https://user-images.githubusercontent.com/25320782/64004361-b49c5300-cb40-11e9-9ffa-d85a358c3d50.png)

Are we missing anything needed to implement this page? For e.g.
1. Mapping of authors to chapters
2. Source query link for the visualization

rviscomi:
> Is the side navigation menu's position intended (some margin to the left)? Or should we flush it to the left?

There's a grid layout being used where the main body of the page is fixed width and centered. Also see how it aligns with the footer content. I think we might also want to change the header to fit within this same grid, so the background is 100% but the header contents are fixed width and centered like the footer.

> HTML rendering is covered for this part:

This is just an example of embedding a code sample, you can disregard for now.

> Chart visualization is covered for this part:

Data visualization will be covered by #114 so it can be left out of this issue's scope.

> Mapping of authors to chapters

Handled by #70 

> Source query link for the visualization

The actual contents of the chapter are still being written, so for now you can use the placeholder content written by @mikegeyser: https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/content/en/2019/javascript.md

rviscomi:
Mobile designs are also available here: https://drive.google.com/open?id=1ud4iZWxUB7vMZlcKUrPUkLuT1haBV_B6

rviscomi:
@HTTPArchive/developers is anyone available to work on this?

tyohan:
We need this before able to work on #199. So I'll try to do this as well. @tanhengyeow did you work on this already? do you want to push on a working branch so I can continue it from there?

tanhengyeow:
Hi @tyohan 

[...]

PR #213, opened 2019-10-18
Generate chapter contents

rviscomi:
- Edited chapter metadata.
- Ran `(env) almanac.httparchive.org/src$ python generate_chapters.py` to generate chapters from markdown.
- Basic layout styles for chapter page
- First pass at embedding data viz

Progress on #130 cc @tyohan @bkardell 

Still needs table of contents sidebar, but here's how it looks now: (Markup chapter for example)

image

image


tyohan:
I made some changes on templating. Just pull this one and will see how it works with my current code.

Issue #210, opened 2019-10-18
Add hover/focus states for interactive elements


rviscomi:
It should be easy to navigate the website using only a keyboard. Some UI like the header links don't have focus states, so it's not clear when they receive focus. This is for desktop only.

Links could have an underline on hover and/or an outline on focus. Open to ideas.

cc @HTTPArchive/developers

MSakamaki:
I am interested task.

If the UI that is the target of this hover focus is already on the home page, it seems that it can be unified with the following concept.

1. White background: black (#1A2B49) to white or white to black
2. Gray background: yellow (#F7F779) transparent or transparent yellow

-The design of ghost buttons should be unified on home page.
-Just add text color and underline according to the above rules when text is on focus.

And, Should this issues discuss the following?

- index of methodology page.
- Toggle button and social icon on contributors page.
- I not find the user-action that returns to the home page now from the methodology page or the contributors page.
Is return home page with a click on the logo?
And in that case, does the logo need hover/focus?

PR #217, opened 2019-10-22
added hover and focus for interactive element (header/footer)

MSakamaki:
#210 
A unified hover/focus has been applied to the interactive elements of the header and footer.

### DIFF
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -62,6 +62,9 @@ header.alt-bg a:hover, footer.alt-bg a:hover,
 header.alt-bg a:focus, footer.alt-bg a:focus {
   color: #F7F779;
   outline: none;
+}
+header.alt-bg a.interactive:hover, footer.alt-bg a.interactive:hover,
+header.alt-bg a.interactive:focus, footer.alt-bg a.interactive:focus {
   text-decoration: underline;
 }
 
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -62,14 +62,7 @@ header.alt-bg a:hover, footer.alt-bg a:hover,
 header.alt-bg a:focus, footer.alt-bg a:focus {
   color: #F7F779;
   outline: none;
-}
-header.alt-bg a.interactive:hover::after, footer.alt-bg a.interactive:hover::after,
-header.alt-bg a.interactive:focus::after, footer.alt-bg a.interactive:focus::after {
-  content: '';
-  border-bottom: solid 1px #F7F779;
-  display: block;
-  position: relative;
-  bottom: -4px;
+  text-decoration: underline;
 }
 
 .decorative-line-center {
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -53,15 +53,13 @@ a.navigation-logo {
   outline: none;
   color: white;
 }
-a.navigation-logo img:hover {
-  filter: brightness(0) saturate(100%) invert(86%) sepia(4%) saturate(4357%) hue-rotate(16deg) brightness(136%) contrast(94%);
-}
+a.navigation-logo img:hover,
 a.navigation-logo:focus img {
   filter: brightness(0) saturate(100%) invert(86%) sepia(4%) saturate(4357%) hue-rotate(16deg) brightness(136%) contrast(94%);
 }
 
-header.alt-bg a.interactive:hover, footer.alt-bg a.interactive:hover,
-header.alt-bg a.interactive:focus, footer.alt-bg a.interactive:focus {
+header.alt-bg a:hover, footer.alt-bg a:hover,
+header.alt-bg a:focus, footer.alt-bg a:focus {
   color: #F7F779;
   outline: none;
 }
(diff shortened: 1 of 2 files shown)


src/static/css/2019.css
```diff
@@ -49,6 +49,31 @@ select {
...
+  color: white;
+}
+a.navigation-logo img:hover {
+  filter: brightness(0) saturate(100%) invert(86%) sepia(4%) saturate(4357%) hue-rotate(16deg) brightness(136%) contrast(94%);
+}
+a.navigation-logo:focus img {
+  filter: brightness(0) saturate(100%) invert(86%) sepia(4%) saturate(4357%) hue-rotate(16deg) brightness(136%) contrast(94%);
+}
+
+header.alt-bg a.interactive:hover, footer.alt-bg a.interactive:hover,
```
rviscomi: Why is `a.interactive:hover` needed and not simply `a:hover`? Are there some anchors that shouldn't inherit these interactive styles?

src/static/css/2019.css
```diff
@@ -49,6 +49,31 @@ select {
...
+a.navigation-logo {
+  outline: none;
+  color: white;
+}
+a.navigation-logo img:hover {
+  filter: brightness(0) saturate(100%) invert(86%) sepia(4%) saturate(4357%) hue-rotate(16deg) brightness(136%) contrast(94%);
+}
+a.navigation-logo:focus img {
+  filter: brightness(0) saturate(100%) invert(86%) sepia(4%) saturate(4357%) hue-rotate(16deg) brightness(136%) contrast(94%);
+}
```
rviscomi: Combine
```suggestion
a.navigation-logo img:hover,
a.navigation-logo:focus img {
  filter: brightness(0) saturate(100%) invert(86%) sepia(4%) saturate(4357%) hue-rotate(16deg) brightness(136%) contrast(94%);
}
```

src/static/css/2019.css
```diff
@@ -235,6 +260,12 @@ footer .nav-items *:last-child {
 .social-media img:hover {
```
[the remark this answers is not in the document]
MSakamaki: ok, fix it.

src/static/css/2019.css
MSakamaki: ok, fix it.

src/static/css/2019.css
MSakamaki: Certainly, this is simple text a hover/focus, so this useless selector (`interactive`) can be removed.
I agree and fix it.

src/static/css/2019.css
```diff
@@ -49,6 +49,29 @@ select {
...
+
+header.alt-bg a:hover, footer.alt-bg a:hover,
+header.alt-bg a:focus, footer.alt-bg a:focus {
+  color: #F7F779;
+  outline: none;
+}
+header.alt-bg a.interactive:hover::after, footer.alt-bg a.interactive:hover::after,
+header.alt-bg a.interactive:focus::after, footer.alt-bg a.interactive:focus::after {
+  content: '';
+  border-bottom: solid 1px #F7F779;
```
rviscomi: What's the advantage of a border on the `::after` pseudo-element versus `text-
decoration: underline` on the anchor itself?

Underline:
image

Border:
image

I also noticed with the border, the link shifts vertically by a few px on hover. Is that intentional?

rviscomi:
One question about border vs underline, otherwise this looks great. Thanks for working on it!

src/static/css/2019.css
MSakamaki: @rviscomi 
I'm sorry, there was no deep thought in this, it was my mistake.
(This is a remnant of trying out focus decorations)
I agree this should fix be text-underlined.

rviscomi: No worries! This is looking great! 🙌

src/static/css/2019.css
```diff
@@ -49,6 +49,22 @@ select {
...
+a.navigation-logo {
+  outline: none;
+  color: white;
+}
+a.navigation-logo img:hover,
+a.navigation-logo:focus img {
+  filter: brightness(0) saturate(100%) invert(86%) sepia(4%) saturate(4357%) hue-rotate(16deg) brightness(136%) contrast(94%);
+}
+
+header.alt-bg a:hover, footer.alt-bg a:hover,
```
rviscomi: The "Start Exploring" button in the header appears to have an underline when hovered/focused, which I like:

image

But the same button in the intro section does not:

image

Could you make sure this behavior is consistent for all buttons and an underline is shown?

rviscomi: To clarify, _I do like the underline_ for buttons, I just want to make sure they're shown consistently 😄

MSakamaki: @rviscomi 
I will start this task!
Try to underline the button explicitly.

MSakamaki: @rviscomi 
During the focus test, I found problem in my environment (chrome and firefox for mac os).

At a certain resolution(width: 1480px), when there is a overflow content of position: relative/absolute, can not touch it, so the button to disabled focus.

Setting  `z-index: 1` to button avoids the problem, but do you know a smarter way?

スクリーンショット 2019-10-24 2 12 51

MSakamaki: notebook:
Even if width 599px, can not press some of Contributors buttons.

rviscomi: Yeah I noticed that too. I have a fix in #212: https://github.com/HTTPArchive/almanac.httparchive.org/pull/212/files#diff-dcfbc15530630ce0591f21b62fbb1237R217 (disabling pointer events for the "91" text)

src/static/css/2019.css
```diff
@@ -49,6 +49,25 @@ select {
...
+header.alt-bg a:focus, footer.alt-bg a:focus {
+  color: #F7F779;
+  outline: none;
+}
+header.alt-bg a.interactive:hover, footer.alt-bg a.interactive:hover,
```
rviscomi: I think this can be simplified to `a:hover, a:focus { text-decoration: underline; }`

Are there any cases where we don't want a link to have an underline?

src/static/css/2019.css
rviscomi: ^ That PR was just merged so this should be fixed for you now.

src/static/css/2019.css
MSakamaki: I agreed :+1: 
Yes, I working on it now.

src/static/css/2019.css
MSakamaki: @rviscomi 
I was saved!
this I good experience and I will complete the rest of the tasks.

rviscomi:
Looks great! Thanks for working on this @MSakamaki!

MSakamaki:
Thank @rviscomi  for checking at the code details!

PR #215, opened 2019-10-21
Performance chapter

rviscomi:
Closes #171

Converted from https://docs.google.com/document/d/1S0De_lBFCuqLWa2ZlB31SKQZE_zkboweeAUWLrcDuLw/edit?usp=sharing

Looks like the md tables aren't rendering to html correctly, I'll look into that separately (cc @mikegeyser). Ignore that for now.

Looking to get a final LGTM from the reviewers.

PR #222, opened 2019-10-23
Converted the markdown generation to showdown.

### LINKED ISSUE #219: MD tables not rendering correctly
The markdown tables don't appear to be rendering correctly, as @rviscomi mentions in the PR #215. This is just a bug to track the investigation / resolution.

mikegeyser:
This closes #219 

This includes markdown rendering via [showdown](https://github.com/showdownjs/showdown) instead of mistune (because of the reasons mentions [in this comment](https://github.com/HTTPArchive/almanac.httparchive.org/issues/219#issuecomment-545281951). Importantly, the `chapter.html` template had to change to accommodate ejs, so the escaped curly braces has to be removed (which looks like a weird change). I haven't deleted the 'old' generate_chapters script, but think that it should probably also be removed.

I haven't included regenerated jinja templates, so that the PR was leaner and easier to review, but I have tested it against the contents of the #215 PR. If it would help, I am happy to run it against these templates.

Please take a look, and let me know if there are any problems or suggestions?

### DIFF
--- a/src/templates/en/2019/chapter.html
+++ b/src/templates/en/2019/chapter.html
@@ -1,6 +1,6 @@
 {# IMPORTANT!
 
-- `chapter.html` is a "template for templates" used by the `generate_chapters.js` script, hence the strange template syntax (eg, mixing ejs and jinja syntax `<%- ... %>`)
+- `chapter.html` is a "template for templates" used by the `generate_chapters.js` script, hence the strange template syntax (eg, mixing ejs and jinja syntax)
 - if you want to modify `chapter.html`, you must also:
   - translate the corresponding language-specific templates (eg `src/templates/<lang>/<year>/chapter.html`)
   - run the generation script to update each chapter template
--- a/src/templates/en/2019/chapter.html
+++ b/src/templates/en/2019/chapter.html
@@ -1,26 +1,26 @@
-{{# IMPORTANT!
+{# IMPORTANT!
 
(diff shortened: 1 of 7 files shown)


src/templates/en/2019/chapter.html
```diff
@@ -1,26 +1,26 @@
-{{# IMPORTANT!
+{# IMPORTANT!
 
-- `chapter.html` is a "template for templates" used by the `generate_chapters.py` script, hence the strange template syntax (eg, double braces `{{% ... %}}`)
+- `chapter.html` is a "template for templates" used by the `generate_chapters.py` script, hence the strange template syntax (eg, double braces `{% ... %}`)
```
rviscomi:
```suggestion
- `chapter.html` is a "template for templates" used by the `generate_chapters.py` script
```

src/tools/generate/index.js
```diff
@@ -0,0 +1,5 @@
+let { generate_chapters } = require('./generate_chapters');
+
+// TODO: Generate visualisations 
+generate_chapters();
```
rviscomi: Could you also add docs to the README with steps to install and use the generation script? You can add it under a new section "Generating chapters" since it's distinct from running the website locally.

src/templates/en/2019/chapter.html
mikegeyser: I've updated the warning to include a bit about ejs, instead of removing it.

Issue #199, opened 2019-10-14
Add the ability to permalink all chapter headings and figures


rviscomi:
@HTTPArchive/developers 

All headings in the chapter should have a unique ID to which we can deep link. It would also be nice if a clickable link icon appeared on hover so users can copy the URL.

In addition, figures like charts and data tables should be able to be deep linked. Since the Almanac is a reference document, we should expect and support third party websites to link citations to specific figures. Ideally, deep linking figures should be built into the process for #67.

If not statically baked into the markup, we could consider implementing this as a JS function at runtime.

tyohan:
I'll work on this.

rviscomi:
Thanks @tyohan!

One thing I forgot to mention is that when a user clicks on a deep link into the chapter, the page must jump down to that section. So if the IDs are not baked into the markup the browser might not make the jump immediately on page load and our JS solution would have to parse the URL hash and scroll manually.

Which approach do you prefer: static markup or dynamic JS?

tyohan:
I prefer to use static markup, so I'll make sure we can make this built in the process for #67 and other sections too. We can add JS function later for smooth scrolling if needed.

rviscomi:
@tyohan as we wrap up the chapter page design, I had a thought about how we could implement the static markup approach for this issue. [generate_chapters.py](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/generate_chapters.py) is a script that converts markdown into markup, based on [mistune](https://mistune.readthedocs.io/en/latest/).

We could create a custom renderer similar to their [Table of Contents](https://github.com/lepture/mistune-contrib/blob/master/mistune_contrib/toc.py) example to take markdown headings and inject ID attributes into their markup.

For example, given something like `## Intro` I'd expect it to generate `<h2 id="intro">Intro</h2>` and a corresponding table of contents.

WDYT?

obto:
Awesome. In case you run into some edge cases and want to see how others have done it, here's the function WP uses to "slugify" titles as reference

https://core.trac.wordpress.org/browser/tags/5.2.3/src/wp-includes/formatting.php#L2206

rviscomi:
[...]

PR #232, opened 2019-10-24
Generate Table of Contents for each chapter

mikegeyser:
This generates a table of contents from the output of the html generation, with links to individual sections. It is currently completely unstyled, so that we can use the work already done as a part of #199.

The code below is based on #222, so at present it looks a lot larger than it is, and that PR needs to be merged first. (I've kept them separate, so that they're easier to review - I hope that's ok.)

Please let me know what you think, and I'll make whatever changes are necessary.


src/tools/generate/generate_table_of_contents.js
```diff
@@ -0,0 +1,81 @@
...
+    dom.window.document.querySelectorAll('h1, h2, h3, h4, h5, h6')
+  );
+  const starting_level = get_level(all_headings[0]);
+  const nested_headings = nest_headings(all_headings, starting_level);
+  const toc = generate_html(nested_headings.children);
+
+  return toc;
+};
+
+const generate_html = (headings) => {
```
mikegeyser: Because this recursively generates the table of contents, I've kept it in javascript. If we want to move it into the ejs template, we would need to create it as an partial inside the templates directory.

src/tools/generate/generate_table_of_contents.js
```diff
@@ -0,0 +1,81 @@
...
+    const children = heading.children ? generate_html(heading.children) : '';
+    const li = `<li>${a}${children}</li>`;
+    list.push(li);
+  }
+
+  return `<ul>${list.join('')}</ul>`;
+};
+
+// This is a recursive function to nest the headings.
+const nest_headings = (source, current_level = 1) => {
```
mikegeyser: I know this is a little complicated, and I'm open to suggestions as to how to simplify it.

rviscomi:
Getting this error when running locally:

```
 Generating chapter: en, 2019, performance
(node:9069) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'children' of undefined
    at nest_headings (/Users/rviscomi/git/google/almanac.httparchive.org/src/tools/generate/generate_table_of_contents.js:61:37)
    at nest_headings (/Users/rviscomi/git/google/almanac.httparchive.org/src/tools/generate/generate_table_of_contents.js:60:39)
    at generate_table_of_contents (/Users/rviscomi/git/google/almanac.httparchive.org/src/tools/generate/generate_table_of_contents.js:9:27)
    at parse_file (/Users/rviscomi/git/google/almanac.httparchive.org/src/tools/generate/generate_chapters.js:28:15)
    at generate_chapters (/Users/rviscomi/git/google/almanac.httparchive.org/src/tools/generate/generate_chapters.js:18:45)
    at <anonymous>
(node:9069) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 3)
(node:9069) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

src/tools/generate/generate_table_of_contents.js
rviscomi: Hmmm yeah this is tricky. I think my preference is to have all HTML in the templates. What do you think of generating a JSON object of the TOC and injecting that into the chapter template, similar to the metadata? Then we can run that through a Jinja macro to generate the TOC markup.

src/tools/generate/generate_table_of_contents.js
rviscomi: Whatever works 😄

mikegeyser:
I've spotted the error, going to need to rework something.

tyohan:
I got some error on generating chapters locally

```Generating chapter: pwa, 2019, en
Traceback (most recent call last):
  File "generate_chapters.py", line 70, in <module>
    generate_chapters()
  File "generate_chapters.py", line 24, in generate_chapters
    write_template(language, year, chapter, metadata, body)
  File "generate_chapters.py", line 58, in write_template
    file_to_write.write(template.format(body=body, metadata=metadata))
ValueError: expected ':' after conversion specifier```

tyohan:
@mikegeyser i saw you're generated the toc on the client side with JS. I've done this with mistune so the chapter templates will have toc in generated templates. Also, trying to make sure the CSS styles will shared with methodology pages.

This is how Javascript chapter looks like in [chapter-page branch](https://github.com/HTTPArchive/almanac.httparchive.org/tree/chapter-page)
image
 And in mobile
image

Sorry i reused your avatar @rviscomi  for Addy :D, need to see how the page looks like.

I can finish this chapter page by today so we can see how the chapter pages can be generated config file and markdown contents. So no need for extra json or other files. WDYT @rviscomi

mikegeyser:
@tyohan The table of contents is being statically generated, not on the client side. 

I agree regarding the CSS, I deliberately haven't added styling to this because of the other work that you were doing. :)

rviscomi:
Sorry we have two very similar PRs doing similar things so there's some confusion. Let's try to divide the work up:

@tyohan focus on styling the Chapter page. For now you can create a static "lorem ipsum" table of contents for styling.

@mikegeyser continue working on the showdown migration and TOC generation.


Does that work for everyone?

tyohan:
@rviscomi @mikegeyser looks good for me. Will continue to work on on styling

rviscomi:
@mikegeyser are these changes working and ready for review?

mikegeyser:
@rviscomi I believe it's working, will rebase off of master and test again. The only outstanding change is to move the html generation into a template, which I will also do now.

src/tools/generate/generate_table_of_contents.js
mikegeyser: I've moved this into a separate ejs template, instead of doing it in template literal.

mikegeyser:
Sorry for the churn, I believe that this is ready to go now. I have some more that I want to add (formatting and linting, to pick up errors more easily) but I'll do it in a separate PR.

rviscomi:
@mikegeyser I've merged #251 which created a merge conflict in `chapter.html`. I was going to try to resolve it but I'll leave that to you since it depends on how you want to implement the TOC markup.

mikegeyser:
@rviscomi @tyohan I've merged master into this PR and resolved the conflict. It looks right to me, but could you take a look and see if it's right? 

The Author information and the Previous/Next isn't implemented yet, should that happen in this PR or can it be in a follow up? (It's not strictly speaking related to this.)

PR #236, opened 2019-10-25
Chapter page

tyohan:
Notes on this PR:
- chapter template design implementation ( still need some touch on details )
- table of content automatically generated
- not include permalink to headings( #199 ) but with the current implementation it will be easy to generate a permalink to each heading
- authors box, the authors will automatically generated from config data. Will not shows up if the authors metadata can't found a match from contributors data in config
- previous and next chapter button will automatically generated from config, the script will iterate through all chapters in config and will set previous and next button once it found the same chapter number from chapter metadata.
- change the methodology page CSS file to page.css and reuse the CSS file with this chapter page


tyohan:
Got some error after merge with master

```Traceback (most recent call last):
  File "generate_chapters.py", line 160, in <module>
    generate_chapters()
  File "generate_chapters.py", line 76, in generate_chapters
    write_template(language, year, chapter, metadata, body, tochtml)
  File "generate_chapters.py", line 125, in write_template
    file_to_write.write(template.format(body=body, metadata=metadata, toc=tochtml, authors=authors, prevnext=prevnext))
ValueError: expected ':' after conversion specifier```

**Comment by tyohan** [comment]
Ok, all good now. It should be safe.

File: src/templates/en/2019/chapter.html [hunk 1]
```diff
@@ -1,26 +1,36 @@
-{# IMPORTANT!
+{{# IMPORTANT!
```
mikegeyser: Hi @tyohan, the changes that were made to the `chapter.html` in #222 were so that we can use Node + Showdown + EJS to statically generate the templates - because [mistune wasn't generating tables (amongst other potential problems)](https://github.com/HTTPArchive/almanac.httparchive.org/issues/219#issuecomment-545281951).

If we add the extra curly braces back, it will break the ejs rendering of the markdown. The work done in #232 is supposed to try and help with this, but I'm not sure how best to proceed?

tyohan: @mikegeyser sorry accidentally click re-request review button. 
 
Noted on showdown. Let me try to work around this thing and see if we can combine our works.

tyohan:
@rviscomi there are 2 ways to merge my code and @mikegeyser :
1. Let showdown generate the body through Node but others generate with Python and mistune. We can use [Naked](https://pypi.org/project/Naked/) to run Node from Python, so we just need to call it once from Python script.
2. Move some of my code to Flash app route, so contributors and previous & next navigation will render there.

WDYT?

src/requirements.txt
```diff
@@ -1,9 +1,44 @@
+argh==0.26.2
```
rviscomi: Wow this is a lot of new requirements! Where did they all come from?

src/templates/en/2019/chapter_author.html
```diff
@@ -0,0 +1,24 @@
+{% for author in authors %}
```
rviscomi: Rather than individual template files for each of these components, we can generate them in the existing chapter template file.

Similar to the render_byline macro in base_chapter.html.

src/templates/en/2019/chapter_prevnext.html
```diff
@@ -0,0 +1,25 @@
+<nav id="chapter-navigation">
```
rviscomi: I don't think we need to generate these as part of the markdown rendering. We have the chapter info in `2019.json` and we can render them server-side.

Also we might not launch with all 20 chapters, so some are marked `todo` in the config. These links should go to the next/previous sequential chapter that isn't still marked "todo". That should be done dynamically on the server-side because it could change after the markdown is initially rendered.

src/templates/ja/2019/chapter_author.html
```diff
@@ -0,0 +1,24 @@
+{% for author in authors %}
```
rviscomi: No need to update Japanese templates, we can leave that for the final translation step.

src/templates/en/2019/chapter.html
rviscomi: We should only maintain one markdown parsing script and given that mistune doesn't support markdown tables within HTML, we should deprecate this mistune version in favor of the showdown script @mikegeyser is building.

Sorry that you put so much work into this script and for any confusion as we migrated tools. Having only one script will simplify things a lot.

src/requirements.txt
tyohan: I was installing the mistune contrib module and trying to update the the requirements but didn't realized it somehow generated all of these :|

tyohan:
Noted, i'll push in a few hours.

tyohan:
I have a new PR #251. Feel free to delete this branch.

PR #235, opened 2019-10-25
add toc with desktop resolution footer

MSakamaki:
Fixes #216 

- Added table of contents
- Moved HTTP Archive home logo image to second line

スクリーンショット 2019-10-25 15 07 20

PR #248, opened 2019-10-29
simplify footer markup

### LINKED ISSUE #238: Simplify footer markup
As of now we have two sets of footer markup for desktop and mobile. This should be rewritten to have one set of markup that is styled responsively to accommodate either.

https://github.com/HTTPArchive/almanac.httparchive.org/blob/5c0e002eacb0dace823300b5f6a5f12f0f13f596/src/templates/en/2019/base.html#L55-L101

Anyone on @HTTPArchive/developers available to take this?

MSakamaki:
fix #238 .

360px
スクリーンショット 2019-10-29 14 14 26

760px
スクリーンショット 2019-10-29 14 14 38

1200px
スクリーンショット 2019-10-29 14 14 48

### DIFF
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -199,47 +199,64 @@ blockquote::before {
 }
 
 /* Footer */
-footer hr:last-of-type {
-  display: none;
+footer {
+  display: grid;
+  grid-template-areas:
+    'home-logo ha-logo nav-items language-switcher'
+    'hr hr hr hr'
+    'copyright copyright copyright social-media';
+  grid-template-rows: 60px 1fr 50px;
 }
 
+footer .home-logo {
+  grid-area: home-logo;
+  display: flex;
+  justify-content: start;
+  align-items: center;
+}
 footer .nav-items {
+  grid-area: nav-items;
   display: flex;
-  justify-content: space-between;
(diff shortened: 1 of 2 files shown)


src/static/css/2019.css
```diff
@@ -199,47 +199,64 @@ blockquote::before {
 }
 
 /* Footer */
-footer hr:last-of-type {
-  display: none;
+footer {
+  display: grid;
```
rviscomi: Is grid the best fit for this layout? I would have thought flexbox given the fluidity. Does it work better in your opinion?

MSakamaki: Thank @rviscomi  for the review!

When I implemented it responsively,  I decided that the next elements would be complicated if flexbox, and that the grid layout would may be simple.

1. copyright and social element order
2. Language switch and navigation element order
3. http archive home logo position

The layout is confirmed with firefox/chrome/safari on mac (and safari on iPhone).

This judgment may be my lack of knowledge.
I am interested in the possibility that flexbox is more suitable.

MSakamaki: @rviscomi 
I'm sorry, I forgot to issues #235.
I will fix it soon.

PR #250, opened 2019-10-29
Lint templates

mikegeyser:
This PR includes prettier to do html reformatting and linting, so that we can catch errors in the markup at generation-time (both visually and programmatically). I've had to change some of the content files, so that the errors in them didn't break parsing.

I'm not sure if we want this at such a late stage, and I understand if this gets rejected, but I think that it will help catch and action errors sooner. (This change is built off of #232).


src/templates/en/2019/chapter.html
```diff
@@ -1,12 +1,12 @@
-{# IMPORTANT!
+<!--  IMPORTANT!
```
rviscomi: Not sure we want this. As a Jinja comment, it is excluded from rendering. As an HTML comment, won't it be included in the page source and visible to users?

rviscomi:
This also includes TOC changes. Should the destination branch be TOC as opposed to master?

src/templates/en/2019/chapter.html
mikegeyser: You're right, but it results in a parsing error. I'll try and resolve it another way.

mikegeyser:
@rviscomi I can change the target, I did mention it in the PR message. My rationale was that I  didn't want to delay #232 on the chance that we didn't want this change.

src/templates/en/2019/chapter.html
mikegeyser: I've embedded the jinja comment inside the html comment, so that jinja will strip it out, but we will still send `<!---->` down as a part of the document.

The `<!-- prettier-ignore -->` functionality is also a code comment, so would send more characters down with the document, for the same net effect.

Do you think this is an acceptable compromise, or shall I keep looking for a different solution?

rviscomi: I think this is a reasonable tradeoff. My only concern is that the comment appears as the first line of the document:

```html
<!---->

<!doctype html>
<html lang="en-US">
  <head>
    <meta charset="UTF-8">
```

I think that would screw up the `doctype` declaration.

rviscomi:
I do think it's a good change! Let's make it work.

And sorry I accidentally pushed a commit to this branch, but it's actually semi-relevant so let's include it. I'm just rerunning the generation script and saving the markup changes.

src/templates/en/2019/chapter.html
mikegeyser: I moved the `{% extends %}` to the top of the file (above the comment), and now jinja is not generating the empty html comment at all. It's a bit bizarre, but I think that solves the issue?

mikegeyser:
I've merged #232 into this again, and regenerated the templates to check the output. I **think** it's working alright now?

rviscomi:
More merge conflicts 😭

mikegeyser:
Fixed conflicts. Really sorry for the churn around these two PRs. :(

Issue #239, opened 2019-10-25
Gather bios for each author


rviscomi:
At the top of each chapter we have a short byline to say who wrote it. For example, "By Person One and Person Two" where each name is a link to the author on the Contributors page.

At the bottom we have space for a few sentences per author for a bio. This is a good place to list your qualifications for the chapter like what your day job is or other related work (projects, books, etc).

![image](https://user-images.githubusercontent.com/1120896/67606239-a9376380-f74e-11e9-879d-fedab15f9f97.png)

@HTTPArchive/authors can you each comment on this issue with your preferred bio? If you're writing multiple chapters it's also ok to have a custom version for each, if you'd like.

- [x] 	1. JavaScript	@housseindjirdeh
- [x] 	2. CSS	@una @argyleink
- [x] 	3. Markup	@bkardell
- [x] 	4. Media	@dougsillars @colinbendell
- [x] 	5. Third Parties	@patrickhulce
- [x] 	6. Fonts @zachleat
- [x] 	7. Performance	@rviscomi
- [x] 	8. Security	@arturjanc @ScottHelme
- [x] 	9. Accessibility	Nektarios Paisios @obto @kleinab
- [x] 	10. SEO	@rachellcostello @ymschaap @AVGP
- [x] 	11. PWA	@tomayac @jeffposnick
- [x] 	12. Mobile web	@obto
- [x] 	13. Ecommerce	@samdutton @alankent
- [x] 	14. CMS	@amedina @ernee
- [x] 	15. Compression	@paulcalvano
- [x] 	16. Caching	@paulcalvano
- [x] 	17. CDN	@andydavies @colinbendell
- [x] 	18. Page Weight	@khempenius @tammyeverts
- [x] 	19. Resource Hints	@khempenius
- [x] 	20. HTTP/2	@bazzadp

ymschaap:
SEO, @ymschaap, bio: "Founder at technical SEO consultancy [build.amsterdam](https://build.amsterdam/)"

dougsillars:
Doug Sillars is a freelance digital nomad working on the intersection of performance and media.  He tweets @dougsillars, and blogs regularly at dougsillars.com.

tammyeverts:
Short bio:

[...]

Issue #226, opened 2019-10-23
Add all contributors


rviscomi:
Over a hundred people have expressed interest in contributing to the project. About 70 people have already contributed in some way so far, by my count. The [Contributors](#214) page is the place for us to highlight each person's contributions. Chapter pages will also link each  author/reviewer/translator to their respective card on the Contributors page.

![image](https://user-images.githubusercontent.com/1120896/67437705-7ca11180-f5bf-11e9-8164-b704c0bbddd9.png)

We need to make sure all contributors are reflected on this page by adding them to the `contributors` collection in [config/2019.json](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/config/2019.json).

Also the authors/reviewers listed in the issues may not reflect people's actual contributions. Some people went AWOL and others have stepped up last-minute, so we should update the JSON config accordingly. (_Hey, it's a volunteer project and I'm not passing judgment, I'm just happy to have the help!_ 😁)

[This spreadsheet](https://docs.google.com/spreadsheets/d/1LR29JpI2XJSjan_z6uP1YpRy4EYYivYZPqCIG-9sfFY/edit?ts=5d13c7d0#gid=770177464) is the most up-to-date version of everyone who has actually contributed and for which team. I'm least sure about the author and reviewer teams, which have been the hardest to keep track of in real time. We should find a point of contact from each chapter to verify who contributed to writing and reviewing it.

rviscomi:
Also, if you're reading this, feel free to submit a PR to add yourself to the config file with whatever teams you contributed to!

Issue #214, opened 2019-10-21
Implement the Contributors page design


rviscomi:
![image](https://user-images.githubusercontent.com/1120896/67223708-c3470e00-f3fd-11e9-84c1-992d340b6454.png)

Desktop: https://projects.invisionapp.com/d/main#/console/18009643/386025636/inspect
Mobile: https://projects.invisionapp.com/d/main#/console/18174420/387303024/inspect

cc @HTTPArchive/developers

rviscomi:
The avatars in https://github.com/HTTPArchive/almanac.httparchive.org/tree/master/src/static/images/avatars have a white bg. We'll also need to make them transparent. I have the source PSD file if needed.

ymschaap:
Question, I saw the contributor lists supports [github and twitter](https://github.com/HTTPArchive/almanac.httparchive.org/blob/bf2f9f68a2d294007913d942ddfb076f4e224b1f/src/templates/en/2019/contributors.html#L115), would it make sense to add any website/blog as an option too for contributors?

rviscomi:
@ymschaap possibly if people feel strongly, otherwise I think we should limit it to just these two.

rviscomi:
Any @HTTPArchive/developers available to take this on?

ymschaap:
@rviscomi looking at the bio's at #239 I see more external websites then github or twitter mentions :) this is the *web* almanac right ;) But I understand this will be extra work/designs/time.

catalinred:
@rviscomi Is there anything I can do to help at this point?

rviscomi:
Ok you've convinced me. But you have to promise to build the functionality! We'll need an icon (generic globe icon used everywhere) and the plumbing to render it from the config file into the Contributors page.

ymschaap:
Haha, it was a trap.  Working on it.

PR #252, opened 2019-10-29
Contributors config and design

ymschaap:
Re: #226
Re: #214

To speed things up, I've taken the spreadsheet here:
> https://docs.google.com/spreadsheets/d/1LR29JpI2XJSjan_z6uP1YpRy4EYYivYZPqCIG-9sfFY/edit?ts=5d13c7d0#gid=770177464

And transformed it to json using a script.

If we add the bio's from #239 as a column I could quickly update this file to also include those. Else we have to manually copy+pasta.

I did lose any existing edits, and hence the fields: tagline, gravatar, twitter, etc. We could manually add that (I think there were ~6 contributors).

Given we have github accounts for 66 of the users, I can run it through the Github User API and extract that image & bio. Ok?

**Update:** I did that just now, let me know if that works (e.g. maybe contributors don't expect the Github bio to appear here).


rviscomi:
Cool! Thanks for working on this @ymschaap!

A few suggestions:

- use the contributor's GitHub ID as the default ID (fall back to first/last name if no GitHub ID is given)
- remove `bio` field since that only applies to authors
- for any existing contributors in the config, try to merge their previous info like twitter fields and taglines

bazzadp:
> use the contributor's GitHub ID as the default ID (fall back to first/last name if no GitHub ID is given)

Any reason why? I think it's actually nicer being firstname-lastname like @ymschaap has done. Better for searchability and SEO too. But then I'm probably weird cause I don't really use my GitHub handle (@bazzadp) in my other online profiles (I really should change it but don't want to lose my history). On plus die it would allow consistent naming convention even for those not on GitHub. On the downside it would mean editing all the chapters to alter the first section but don't think that's too difficult.

rviscomi:
No particular reason. I can be convinced. But yes it would require syncing with the chapter metadata.

ymschaap:
@rviscomi 
* implemented changes (by github Id as key, if available, bio gone, merge previous added missing contributors (1))
* I don't have access to the invisionapp mockup document, so I just winged it.
* I tried to sort all contributors by name (instead of key), maybe anybody could help me sort by name key here ?
* I did some UX work on the button functionality
* I dropped tagline and twitter (we don't have twitter handles, and the tagline wasn't in the contributor.html yet)
*

rviscomi:
Thanks for working on this.

Shared the Invision designs with you.

I'd like to keep the `twitter` field. The old config had it in a few places and could be merged with the new GitHub-based data.

src/templates/en/2019/contributors.html
```diff
@@ -107,20 +171,21 @@ <h1>
     <div class="contributor-grid">
     {% for id, contributor in config.contributors.items() %}
       <div class="contributor {{ ' '.join(contributor.teams) }}">
-        <img class="contributor-avatar" src="{{ contributor.avatar_url }}" height="100" width="100" alt="{{ contributor.name }} avatar">
```
rviscomi: `avatar_url` is a pre-generated URL that falls back to a random character illustration from the Almanac design. We should try to keep that functionality.

src/templates/en/2019/contributors.html
```diff
@@ -26,31 +26,63 @@
...
-  padding: 8px;
+  padding: 8px 16px;
+  border: none;
+  -webkit-font-smoothing: antialiased;
 }
 
 .contributor-grid {
   display: flex;
   align-items: stretch;
+  flex-direction: row;
```
rviscomi: On mobile, contributors should appear one per row.

image

src/templates/en/2019/contributors.html
```diff
@@ -107,20 +171,21 @@ <h1>
     <div class="contributor-grid">
     {% for id, contributor in config.contributors.items() %}
       <div class="contributor {{ ' '.join(contributor.teams) }}">
-        <img class="contributor-avatar" src="{{ contributor.avatar_url }}" height="100" width="100" alt="{{ contributor.name }} avatar">
+        <img class="contributor-avatar" src="{% if contributor.avatar %}{{contributor.avatar}}&s=200{% else %}data:image/gif;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNkYAAAAAYAAjCB0C8AAAAASUVORK5CYII=
+{% endif %}" height="100" width="100" alt="{{ contributor.name }} avatar" loading="lazy">
         <div class="contributor-name">{{ contributor.name }}</div>
         {% if contributor.twitter or contributor.github %}
         <div class="contributor-social">
```
rviscomi: Some contributors have no social links. If possible this space would be empty so the team names are aligned across the row of contributors.

image

ymschaap:
Implemented your feedback.

PR #255, opened 2019-10-30
Chapter style tweaks

rviscomi:
image


src/templates/en/2019/chapters/markup.html
```diff
@@ -647,6 +647,10 @@ <h3 id="products-and-libraries-and-their-custom-markup">
         the 5% bar, for perspective.
       </p>
       <figure id="fig9">
+<<<<<<< HEAD
```
tanhengyeow: Some traces of merge conflicts to be removed :)

rviscomi: Fixed, thanks!

PR #256, opened 2019-10-30
Chapter author prevnext

tyohan:
The author is using macro but the previous and next navigation too messy to put in Jinja template so decided to move it in router level.


src/templates/en/2019/table_of_contents.html
```diff
@@ -56,15 +56,23 @@
 }
 
 .chapter .todo {
+<<<<<<< HEAD
```
mikegeyser: There seem to be merge conflicts committed here?

rviscomi: Fixed.

PR #275, opened 2019-10-31
Chapter consistency edits

bazzadp:
Minor edits for consistency.

@patrickhulce I've also thrown together a Conclusion for your Third Party chapter as part of this. Feel free to edit that some more if you disagree with it.


src/content/en/2019/third-parties.md
```diff
@@ -163,9 +168,12 @@ The interest of site owners in understanding and analyzing user behavior is not
...
-- Roughly 84% of all third-party domains by request volume have been identified and categorized. The remaining 16% fall into the “Other” category.
+This chapter has shown that use of third-party content is rampant. This is entirely unsurprising - the entire basis of the web is to allow interconnectiveness and linking.
+
+[Reuse in computer technologies](https://en.wikipedia.org/wiki/Code_reuse) is generally a good thing, though it does introduce depencies which can impact performance (in positive and negative ways!) and has privacy implications, both of which we have touched on here. We have also not discussed the security implications of adding such dependencies to your site, though the security chapter has a [section on vulnerable JS libraries](./security#vulnerable-js-libraries) which is relevant and may be of interest to readers. Even if code has no security concerns it can introduce a single point of failure [which has been recognised as a potenital problem for some time now](https://www.stevesouders.com/blog/2010/06/01/frontend-spof/).
```
patrickhulce:
```suggestion
[Reuse in computer technologies](https://en.wikipedia.org/wiki/Code_reuse) is generally a good thing, though it does introduce dependencies which can impact performance (in positive and negative ways!) and has privacy implications, both of which we have touched on here. We have also not discussed the security implications of adding such dependencies to your site, though the security chapter has a [section on vulnerable JS libraries](./security#vulnerable-js-libraries) which is relevant and may be of interest to readers. Even if code has no security concerns it can introduce a single point of failure [which has been recognized as a potential problem for some time now](https://www.stevesouders.com/blog/2010/06/01/frontend-spof/).
```

patrickhulce: two typos, but otherwise looks great thank you for doing this! I may make a few tweaks when I go back for the domain/CDN explanation

rviscomi:
LGTM. @bazzadp can you also run `npm run generate` to generate the corresponding markup?

bazzadp:
> LGTM. @bazzadp can you also run `npm run generate` to generate the corresponding markup?

Done. Merge away if you're happy!

BTW any way to ignore .DS_Store files when running `npm run generate`? Just deleted them to avoid errors but would be nice to not have to do that each time for us Mac Users! Can look at making some edits to the scripts in tools directory if you want me to?

rviscomi:
I'm also on Mac but haven't had an issue with `.DS_Store`.

It should be ignored by .gitignore:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/.gitignore#L4

cc @mikegeyser

bazzadp:
Yeah it is ignored by git but not the generate scripts. Probably should just avoid going into those directories in finder and that'll solve that! 😀

rviscomi:
Ah ok. Yeah in that case we could add some logic to the script to only handle md files and ignore the rest.

Issue #274, opened 2019-10-31
Add ability to permalink contributor team filter


rviscomi:
When using the team filter on the Contributors page, update the URL so that it reflects the current state of the filter. For example, `?teams=authors,reviewers`. Visiting a contributor page with this parameter should pre-filter to match.

cc @ymschaap

PR #277, opened 2019-10-31
Created a filter based on the hash #274

ymschaap:
See #274


rviscomi:
On mobile so hard for me to verify, but it looks like this uses hashes (`#authors`) which might interfere with the contributor permalink. Is that right? If so could we switch to query string params (`?teams=authors`)?

ymschaap:
Now this all works:
* contributors#rviscomi
* contributors#ymschaap
* contributors#designers
* contributors#designers,authors

Do you have an case where you would want to link to a filter view AND an author?
Else it would not interfere (it supports both).

With your suggestion, if you toggle filters, it would not reflect the url anymore, unless I do stuff with the history.updateState().

rviscomi:
> Do you have an case where you would want to link to a filter view AND an author?

FWIW I was thinking that when linking to an author from the chapter page, we show the contributor in the context of the other authors, eg `?teams=authors#ymschaap`.

I think there's also a case to be made for using querystring params for configuring the UI state and hashes for jumping to specific elements.

PR #289, opened 2019-10-31
Revert "Add Calibre App image actions to automatically optimize images"

JMPerez:
Reverts HTTPArchive/almanac.httparchive.org#287

Assuming we can't enable [GitHub actions](https://github.com/features/actions) on this repo.

@rviscomi since GitHub Actions are not enabled on this repo, this approach is not going to work, and [ImgBot](https://imgbot.net/) is only for free on public repos and needs the organization owner to install it.


rviscomi:
I think I just enabled it. How can we check?
image

JMPerez:
@rviscomi It triggers on PR so I'm going to make a minor change in the existing Action yaml so it triggers it.

rviscomi:
Seems to be working. If there are any config changes feel free to open a new PR.

Issue #291, opened 2019-11-01
Broken links [Contributors page]


AymenLoukil:
In contributors page, there are these broken links : 

- [x] https://almanac.httparchive.org/en/2019/www.bendell.ca 
- [x] https://almanac.httparchive.org/en/2019/dougsillars.com

rviscomi:
Removed @colinbendell's website from his config. Even if we add https, I still get an ERR_TOO_MANY_REDIRECTS error. Colin, if there's a better URL or if you're able to fix the redirect problem we can add it back in.

PR #295, opened 2019-11-01
Fix reference to Doug Sillars' website

JMPerez:
This fixes partly #291


rviscomi:
@JMPerez also feel free to add `developers` to your teams :)

PR #298, opened 2019-11-01
Ignorelist on the generator

mikegeyser:
Added an ignorelist on the generator to cater for `.DS_Store` files (and anything else that might come up).

([In reference to this conversation.](https://github.com/HTTPArchive/almanac.httparchive.org/pull/275#issuecomment-548432158) )


bazzadp:
Not near my computer right now to review but would you not be better with a whitelist than an ignorelist for directories? And then only looking for .md files in each directory?

mikegeyser:
That's a fair point, but the directory structure itself is being used to generate the templates (languages and years). The enumerating of directories within directories is as important as finding the chapter files. (Not sure if that makes sense?)

src/tools/generate/generate_chapters.js
```diff
@@ -11,8 +11,14 @@ converter.setFlavor('github');
 
 const generate_chapters = async () => {
   for (let language of await fs.readdir('content')) {
+    if (ignore(language)) continue;
+
     for (let year of await fs.readdir(`content/${language}`)) {
+      if (ignore(year)) continue;
```
rviscomi: Since you're using `readdir` would it ever try to read ds store files as a directory, or is it ignored by default anyway? If anything would we only need to handle these files at leaf?

Even still, I think I agree with @bazzadp that all we need is an `isMarkdown` check on the leaf files since that's the only acceptable input anyway. WDYT?

mikegeyser: When I was testing using finder to access the content directory, it created `.DS_Store` files in the parent directories too, which was also causing an error.  It seems as if  `readdir` still tries to enumerate the `.DS_Store` as a directory - causing it to break messily.

There does need to be a check at each level, but I am totally open to alternatives.

bazzadp:
`readdir` reads the directories for all files and directories. We could filter for directories or .md files? Something like https://stackoverflow.com/questions/18112204/get-all-directories-within-directory-nodejs

mikegeyser:
We definitely can. We'll just have to use `(await fs.stat()).isDirectory` to check whether or not it's a directory and then at the leaf node ensure that it has an `md` suffix or `continue`.

Happy to do that if we feel it's better?

bazzadp:
Personally I think that would be better. You never know what sort of random files will end up in the directories - especially given the dispersed nature of this project.

On the other hand this is only a small issue, so if others feel this is over engineering the solution then happy to go with a simpler solution - if we even need one at all?

rviscomi:
This LGTM. It works and we can always iterate on improvements as needed.

PR #299, opened 2019-11-01
Line breaks

### LINKED ISSUE #282: Markdown compiler adding line breaks
The showdown compiler should be more tolerant of line breaks in md and not convert them to `<br>` tags in the markup.

For example:

```md
this long line would be 
split up in the HTML
```

```html
<p>
this long line would be <br />
split up in the HTML
</p>
```

(contrived example, not actual compiler output)

For a real example see https://github.com/HTTPArchive/almanac.httparchive.org/pull/273/files#r341347538

mikegeyser:
This PR closes #282.

Fixed the weird linebreak thing, which is supposed to be off by default. Apparently changing the dialect to 'github' flavoured markdown enables it by default, and it[ must just be switched off again](https://github.com/showdownjs/showdown/issues/602#issuecomment-429547738).


mikegeyser:
> Images automagically compressed by [Calibre](https://calibreapp.com)'s [image-actions](https://github.com/marketplace/actions/image-actions) ✨
> 

What wizardry is this?! 🧙🏻‍♂️

PR #306, opened 2019-11-03
Tidied up generator code

mikegeyser:
This PR changes the recursive directory traversal to be a little tidier, inline with the feedback on #298.


bazzadp:
Wow thanks @mikegeyser. Won’t be online till this evening to test it but from a quick scan of the code it looks good to me!

PR #302, opened 2019-11-02
SEO fixes for canonical and meta info (inc Meta Descriptions, OG, Twitter and Article Structured Data)

bazzadp:
Fixes some of https://github.com/HTTPArchive/almanac.httparchive.org/issues/286, sets up basis for https://github.com/HTTPArchive/almanac.httparchive.org/issues/292 and fixes https://github.com/HTTPArchive/almanac.httparchive.org/issues/294.

Still to do:

- Add descriptions to each Chapter (only HTTP/2 done here as an example).
- Get descriptions reviewed by SEO experts
- Add Published Date and Last Updated timestamps to each Chapter meta data
- Do similar for non-chapter pages (Methodology, ToC...etc.)

Issue #292, opened 2019-11-01
Create Meta descriptions [All pages]


AymenLoukil:
We should create meta descriptions for each page.

(edit - @bazzadp added table below)

- [x]	1. JavaScript
- [x]	2. CSS
- [x]	3. Markup
- [x]	4. Media
- [x]	5. Third Parties
- [x]	6. Fonts
- [x]	7. Performance
- [x]	8. Security
- [x]	9. Accessibility
- [x]	10. SEO
- [x]	11. PWA
- [x]	12. Mobile web
- [x]	13. Ecommerce
- [x]	14. CMS
- [x]	15. Compression
- [x]	16. Caching 
- [ ]	17. CDN
- [x]	18. Page Weight
- [x]	19. Resource Hints
- [x]	20. HTTP/2
- [x] Methodology
- [x] Table of Contents
- [x] Contributors
- [x] Home

rviscomi:
Hmm this is a good reminder that we should have a short description for each chapter.

bazzadp:
> Hmm this is a good reminder that we should have a short description for each chapter.

I think these should go in each Chapter Markdown to allow different languages. Have done that for HTTP/2 chapter as part of #302. Let me know your thoughts on that.

rviscomi:
Yes, good idea!

rviscomi:
@bazzadp we're still missing CDN (landing any moment) but how are the JS and CSS chapters looking? Can we check them off?

bazzadp:
> @bazzadp we're still missing CDN (landing any moment) but how are the JS and CSS chapters looking? Can we check them off?

Will file a PR to add them. Was waiting for the CDN chapter to do them all at once. After that I will list them all here in a table so SEO experts can view and suggest any improvements, but I think they are pretty good:
- < 160 chars except one IIRC
- descriptive

bazzadp:
OK here are the English Meta Descriptions:

**Chapters**

|Page|Length|Description|
|---|---|---|
|accessibility|145|Accessibility chapter of the 2019 Web Almanac covering ease of reading, media, ease of navigation, and compatibility with assistive technologies.|
|caching|154|Caching chapter of the 2019 Web Almanac covering cache-control, expires, TTLs, validitaty, vary, set-cookies, AppCache, Service Workers and opportunities.|
|cdn|*TBC*|*CDN chapter of the 2019 Web Almanac covering ... (TBC).*|
[...]

Issue #286, opened 2019-10-31
Audit the website for SEO issues


rviscomi:
Since we have SEO experts on-hand it'd be great if we can make sure that we're following our own best practices on the website. 

The version pushed to https://almanac.httparchive.org has most of the site structure except for the chapter content (the most important part) so if needed we can revisit the website when that's pushed in ~a week or we can audit the local website. The website is also missing translations, so we won't be able to check for that class of SEO issues yet.

CC @ymschaap @rachellcostello @AVGP @clarkeclark @andylimn @voltek62 @AymenLoukil @catalinred

AymenLoukil:
Already started ;)

AymenLoukil:
I made a first check pass and wrote down my findings in draft mode. I'll create an issue for each one + details on how to correct / enhance :) 
Here is the draft document : feel free for every one who want to add things

bazzadp:
> http => https 307 temporary redirect should be 301 one

This one is not quite correct. 307 is because of HSTS which the browser has cached. It’s an internal browser redirect added by your browser. If you use an online tool like https://wheregoes.com/ on http://almanac.httparchive.org then you’ll see it does a 302 to https://almanac.httparchive.org/ (this should probably be a 301) and then a further 302 https://almanac.httparchive.org/en/2019/ (which probably needs to stay as a 302 because next year it will redirect to the 2020 directory so it is a temporary redirect -albeit a year long temporary redirect).

AymenLoukil:
Yeah, the point is http=>https should be 301 and not 302, 307 or any other redirection type.

AymenLoukil:
The second 302 is not because 1 year redirect. It is mainly for language detection. When the translations will be added, if your accept-language is for example es, it should redirect to /es/2019 with 302.

rachellcostello:
I ran a crawl of the site in its current state, and everyone can access the findings [here](https://app.deepcrawl.com/accounts/22614/projects/299664/crawls/2425042?token=qPu5rVvQqK3Xlkfc9r1LboJOH3TLpTcc3D94W-tkX_MVFosTyrtNRhpP0LwkgkHQNRBBnLBKTdDKc_tBkAu5lQ).

In addition to the points in @AymenLoukil's Google doc:

* Multiple H1 tags on the home page and methodology page (not a huge issue for SEO but would recommend maintaining one H1 per page and moving the rest to H2s / H3s etc for more clarity for screen readers)
[...]

PR #308, opened 2019-11-03
SEO Fixes part 2

bazzadp:
Continuing on the work from https://github.com/HTTPArchive/almanac.httparchive.org/pull/302 I've now added:

- Breadcrumb Structured Data
- Meta Descriptions and Timestamps to all existing chapters in master
- Similar changes for non-chapter pages
- Corrected Heading issues in Methodology page
- Corrected Heading issues in Home page

Fixes more of https://github.com/HTTPArchive/almanac.httparchive.org/issues/286 and does some of https://github.com/HTTPArchive/almanac.httparchive.org/issues/292


src/content/en/2019/third-parties.md
```diff
@@ -120,9 +123,7 @@ The category breakdowns among script execution largely follow that of resource c
 
 While much could be said about every individual provider’s popularity and performance impact, this more opinionated analysis is left as an exercise for the reader and other purpose-built tools such as the previously mentioned [third-party-web](https://thirdpartyweb.today).
 
-## Analysis
-
-### Usage Patterns
+## Usage Patterns
```
rviscomi: The heading structure looked ok to me. Each of these headings fit under "Analysis":

Usage Patterns
Generate and Consume Data
Monetize Web Traffic
Simplify Development

src/content/en/2019/performance.md
```diff
@@ -34,7 +37,7 @@ Experiences are measured monthly including paint, load, and interactivity metric
 
 So let's dive in and see what insights we can find.
 
-### FCP
+### First Contentful Paint (FCP)
```
rviscomi: Not sure I agree with expanding these abbreviations. They're all defined earlier in the document. So now they're defined in both "The state of performance" and each heading. Is this a best practice that I wasn't aware of?

src/content/en/2019/third-parties.md
bazzadp: I'm reading it as those three types (Generate..., Monetize and Simplify Development) are the Usage patterns. Hence they were sub headings for that heading. At which point it didn't seem to make sense to have Analysis as a heading, Usage as a sub-heading and the three under that as sub-sub-headings. Re-reading I still think this is the case so do think we should make this change.

rviscomi: @patrickhulce can you clarify your intent?

src/content/en/2019/performance.md
bazzadp: Well as someone less familiar with at least some of these acronyms I struggled to understand them when trying to pull out the key details for the Meta Descriptions. The Table of Contents for this chapter isn't the friendliest with so many abbreviations and I think first use of them in ToC should be expanded - in the same way that first use of them in the text is.

src/content/en/2019/third-parties.md
patrickhulce: The `Usage Patterns` section was intended to be like an overview/intro for the `Analysis` section.

It'd make sense to me for any of the following to happen...

- Rename "Usage Patterns" heading to "Overview".
- Eliminate "Usage Patterns" heading.
- Rename "Analysis" to "Usage Patterns" and eliminate existing "Usage Patterns" heading (this change)

I'd have a slight preference for 2 to keep the top level headings fairly parallel but am fine with whatever y'all decide.

bazzadp: I have a slight preference for 3 as think "Analysis" is a bit generic and "Usage Preferences" is more meaningful section name. Isn't the whole chapter (in fact the whole Almanac) "Analysis" after all? 

Will leave as is and resolve this, but feel free to re-open and add further suggestions. Or I could be persuaded to do 2 if you really feel strongly about it.

src/content/en/2019/performance.md
bazzadp: Going to have a second go at persuading you on this one for the following reasons:

- [Our style guide](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Wiki) emphasis accessibility, writing for global audience and writing inclusive documentation. I think a ToC full of acronyms goes against that spirit.
- The Chapter ToC is the first text you see - before the explanations embedded in the text.
- There is precedence is [avoid abbreviations in headings in other style guides](https://blog.apastyle.org/apastyle/abbreviations/#Q8), though I think avoiding them completely in headings and writing them all out would be a step too far given how many you've used here.
- Some of the terms (e.g. TTFB) are familiar in web performance circles but others are newish (e.g. FCP) and some (e.g. ECT) would be less well known (I didn't know this last one and had to hunt for it!)

Persuaded you yet? 😀

src/content/en/2019/third-parties.md
patrickhulce: the analysis generalism is a fair point. 3 is fine with me! :)

src/content/en/2019/performance.md
rviscomi: Ok ok let's do it your way 😁 

But for conciseness could we remove the acronym from the parenthesis? So just `First Contentful Paint` as the heading rather than `First Contentful Paint (FCP)`.

bazzadp: Done!

PR #311, opened 2019-11-04
Prevent generating ids for Table Headers to fix #310

### LINKED ISSUE #310: HTML validation - duplicate ids when tables with same column names are used
Looking at the [HTTP/2 chapter](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/templates/en/2019/chapters/http2.html) for example, it has generated multiple instances of the following (as I used pretty much the same headings in all my tables):

```
      <table>
        <thead>
          <tr>
            <th id="protocol">Protocol</th>
            <th id="desktop">Desktop</th>
            <th id="mobile">Mobile</th>
            <th id="both">Both</th>
          </tr>
        </thead>
```

HTML states that there should only be one `id` of each name. Additionally using the column name like this could break the ToC chapter links if a h1-h6 heading has the same name as a table column heading.

Can we prefix these ids with the table name? Or just not use ids for column headings at all? Not familiar enough with the library we are using to convert the markdown to know how to fix this myself hence opening this issue.

bazzadp:
Fixes https://github.com/HTTPArchive/almanac.httparchive.org/issues/310

This turned out to be easier to fix than I thought as there's an [option to turn this off](https://github.com/showdownjs/showdown/wiki/Showdown-Options#tablesheaderid). Weirdly it says it's off by default and can't see where we turned this on. Unless documentation is misleading and once tables is on, this is automatically enabled. Any ideas @mikegeyser?

Either way this fixes our issue by explicitly turning it off.

### DIFF
--- a/src/templates/en/2019/chapters/http2.html
+++ b/src/templates/en/2019/chapters/http2.html
@@ -149,10 +149,10 @@
       <table>
         <thead>
           <tr>
-            <th id="protocol">Protocol</th>
-            <th id="desktop">Desktop</th>
-            <th id="mobile">Mobile</th>
-            <th id="both">Both</th>
+            <th>Protocol</th>
+            <th>Desktop</th>
+            <th>Mobile</th>
+            <th>Both</th>
           </tr>
         </thead>
         <tbody>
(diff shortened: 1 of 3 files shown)


bazzadp:
Ah I see it's turned on for [github flavour that we are using](https://github.com/showdownjs/showdown/blob/master/src/showdown.js). That explains that mystery at least 😀

PR #314, opened 2019-11-04
Author bios

### LINKED ISSUE #284: Add author bios to chapter pages
The bios in #239 should be made visible on authors' respective chapter pages similar to how they appear in this mockup:

image

- [x] gather bios (#239)
- [ ] add bios to 2019.json
- [ ] ensure bios can support markdown like links
- [ ] add bios to the chapter template

cc @HTTPArchive/developers

mikegeyser:
This closes #284.

I've added a section for author bios in the `base_chapter.html`, following the pattern set by everything else. I've updated the contributor file with the from #239. The bios had a lot of hyperlinks in them, so I've allowed the template to render them - I hope that's right.

I've left the default styling, because I think it looks ok. The screen mockup looked like the text needed to be a bit smaller than the body text, so I can add that if required.

### DIFF
--- a/src/config/2019.json
+++ b/src/config/2019.json
@@ -171,6 +171,7 @@
     },
     "argyleink": {
       "name": "Adam Argyle",
+      "bio":"Adam Argyle is a Google Chrome developer relations member focused on CSS; He's a web addict with an insatiable lust for great UX &amp; UI; Find him on the web @argyleink> or <a href=\"https://nerdy.dev\">https://nerdy.dev</a>;",
       "teams": [
         "brainstormers",
         "authors"
@@ -204,6 +205,7 @@
     },
     "alankent": {
       "name": "Alan Kent",
+      "bio": "Alan Kent is a Developer Advocate at Google focusing on e-commerce and content ecosystems. He blogs at alankent.me and tweets as @akent99.",
       "teams": [
         "brainstormers",
         "authors"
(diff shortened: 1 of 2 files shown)


bazzadp:
Actually I’ve had a second thought on this. I think the bios should be in the Chapter Markdown files as that would allow them to be translated along with chapter contents. It would also have the secondary minor benefit of avoiding non-authors wasting time writing bios to match other examples they see. And finally it would allow different bios per chapter if an author is involved in multiple chapters and wants to emphasis different things in each bio.

This does raise other questions then. Should it be in the Chapter meta data? Can that handle more complicated content? Should it be in Markdown or HTML?

WDYT?

mikegeyser:
The suggestion definitely has merit, but I'm not sure anyone is going to have different bios for different chapters (or at least they don't yet). Putting it into the chapters could also make the contributor pages harder to generate, depending on how we do it.

I was thinking that it would be cool to have all contributor details in individual md files, and then generate the 2019.json from that before generating the chapters. Although, while I think that is a nice idea, there may be better places to put the dev effort in at this stage.

bazzadp:
I think the translation issue would be the main reason to do this. The others are more nice to haves. Maybe we go with what you have for now and iterate when we come to our first translation?

rviscomi:
> Maybe we go with what you have for now and iterate when we come to our first translation?

+1. Let's ship it!

Issue #319, opened 2019-11-04
Link author names in bio to Contributors


rviscomi:
Add links on each author's photo and name in the bio section of the chapter page to go to their respective part of the contributor page, similar to the byline at the top of the chapter page.

![image](https://user-images.githubusercontent.com/1120896/68153526-7d5c7080-fefa-11e9-992a-ada1507b6f80.png)

@HTTPArchive/developers anyone available to work on this small feature?

PR #320, opened 2019-11-04
Seo fixes3

bazzadp:
- Some more fixes for https://github.com/HTTPArchive/almanac.httparchive.org/issues/286 (should be last of these - except for chapter specific ones but will try to add those manually).
- Fixes https://github.com/HTTPArchive/almanac.httparchive.org/issues/319
- Also changed page to say "Author" instead of "Authors" for chapters with a single Author (OCD strikes again!).
- Fixed the Alt text for Author images in Chapters.
- Added a couple of missing Bios from https://github.com/HTTPArchive/almanac.httparchive.org/issues/239


src/templates/en/2019/base_chapter.html
```diff
@@ -24,14 +24,16 @@
...
   {% for author in metadata.get('authors') %}
-    {% set authordata = config.contributors[author] if author in config.contributors else None %}
+  {% if loop.length == 1 %}<h4>Author</h4>{% endif -%}
+  {% if loop.length > 1 and loop.index == 1 %}<h4>Authors</h4>{% endif %}
+  {% if loop.index == 1 %}<ul>{% endif %}
+      {% set authordata = config.contributors[author] if author in config.contributors else None %}
     {% if authordata %}
       <li>
-          <img class="avatar" alt="Author name" src="{{ authordata.avatar_url if authordata.avatar_url else '/static/images/avatars/'+str(randrange(1, 12))+'.jpg'  }}">
+          <a href="https://almanac.httparchive.org{{ url_for('contributors', year=year, lang=lang, _anchor=author) }}"><img class="avatar" alt="{{ authordata.name }}" src="{{ authordata.avatar_url if authordata.avatar_url else '/static/images/avatars/'+str(randrange(1, 12))+'.jpg'  }}" /></a>
```
rviscomi:
```suggestion
          <a href="https://almanac.httparchive.org{{ url_for('contributors', year=year, lang=lang, _anchor=author) }}"><img class="avatar" alt="{{ authordata.name }} avatar" src="{{ authordata.avatar_url if authordata.avatar_url else '/static/images/avatars/'+str(randrange(1, 12))+'.jpg'  }}" /></a>
```

rviscomi: I think we should do the avatar fallback URL somewhere in the py

bazzadp: Not sure what you mean.

The Avatar URL? Isn't that already there with this?
`src="{{ authordata.avatar_url if authordata.avatar_url else '/static/images/avatars/'+str(randrange(1, 12))+'.jpg'  }}`
Just looked at the Accessibility Chapter in my branch (the only one with an Author without an Avatar!) and I see a fallback Avatar for Nektarios.

Or do you mean the fallback link URL? Well if the author isn't found then it will just link to the Contributors page with an unknown hash ref, and that's the page we'd probably want to fallback to isn't it? So currently that would take you to https://almanac.httparchive.org/en/2019/contributors#authorwhodoesnotexist - which is fine as will still go to contributors page and just can't scroll to the `authorwhodoesnotexist` id that doesn't exist. Or do you think we should fallback to another page? Or remove the link altogether in this situation?

rviscomi: Oh sorry, I meant that the templates shouldn't have to decide what to do if there isn't an avatar_url. I see that it wasn't actually changed in this PR but I just noticed it.

For example when we assign the avatar_url a value in https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/config.py#L23 we should make sure that in all cases it will have either a gravatar URL, GitHub URL, or randomly-generated character URL. AFAICT in that code it should cover all of our bases so maybe we can just remove the fallback logic from the template?

```suggestion
          <a href="https://almanac.httparchive.org{{ url_for('contributors', year=year, lang=lang, _anchor=author) }}"><img class="avatar" alt="{{ authordata.name }}" src="{{ authordata.avatar_url }}" /></a>
```

bazzadp: OK tested that and also worked. Pushed now!

PR #333, opened 2019-11-05
Design tweaks

ymschaap:
I'm not sure who is now working on fine tuning the styling. But e.g. for mobile it needed a few fixes to not overflow the viewport.

I propose these changes:
* smoother font
* less large h1 + return of the stripe
* mobile h1 within viewport (smaller)
* don't overflow viewport (hack for `<table>`

I could pick up some more mobile changes if interested.


src/static/css/page.css
```diff
@@ -268,6 +267,8 @@ table {
   margin: 0 auto;
   max-width: 100%;
   border-collapse: collapse;
+  display: block;
```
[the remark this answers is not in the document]
ymschaap: Yes, I didn't test properly. To make the table scrollable on mobile, we need to wrap it in a div that has overflow: auto; Let me look for another solution.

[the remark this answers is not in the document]
rviscomi: Open to other ideas. Ping me when this is ready for another review.

ymschaap:
@rviscomi I wrap each <table> node with a <div> using `npm run generate`.

I did my best of also cleaning up much of the styling that is in different files, and made it more consistent.

Also - for wide screens - I added the content with a max-width. Else it's unreadable :)

I could not center the <tables>, I would need another nested div for that.

catalinred:
@rviscomi @ymschaap 
On the responsive `<table>`  matter, what do you guys think about this solution?
https://codepen.io/lukepeters/pen/bfFur

It's based on using custom HTML attributes `data-*` and CSS pseudo-elements `::before` or `::after` in order to stack tables when on a mobile view. It's a classic approach for tables with fixed number of columns and I did use this solution with success in some of my projects.

src/templates/en/2019/chapters/accessibility.html
```diff
@@ -249,204 +249,209 @@ <h1 class="title">
...
-      <p>The next frontier, in our opinion, is making widgets that are available through UI frameworks more accessible. Since many complex widgets used in the wild (e.g., calendar pickers) are sourced from a UI library, it would be great for these widgets to be accessible out of the box. We hope that when we collect our results next time, the usage of more properly implemented complex ARIA roles is on the rise—signifying more complex widgets have also been made accessible. In addition, we hope to see more accessible media, like images and video, so all users can enjoy the richness of the web.</p>
+      <html>
```
rviscomi: This doesn't look right. Shouldn't be generating `<html>` inside of the body :)

src/tools/generate/wrap_tables.js
```diff
@@ -0,0 +1,21 @@
...
+  const tables = dom.window.document.querySelectorAll('table');
+  [...tables].forEach(table => {
+      const wrapper = dom.window.document.createElement('div');
+      wrapper.className = 'table-wrap';
+
+      table.parentNode.insertBefore(wrapper, table);
+      wrapper.appendChild(table);
+  });
+
+  return dom.window.document.documentElement.outerHTML;;
```
rviscomi: I think this is causing html/head/body to be written into the chapter templates.

src/templates/en/2019/chapters/accessibility.html
ymschaap: Yes, didn't catch that. Fixed.

rviscomi:
On chapter pages `<aside>` and `<figcaption>` text appears as 18px while paragraphs are 17px. Is this intentional?

rviscomi:
Resolved the merge conflict. Let's get this in and we can iterate on improvements.

PR #335, opened 2019-11-05
Copyediting for Performance Chapter

rachellcostello:
Some very minor edits of the performance copy.

### DIFF
--- a/src/content/en/2019/performance.md
+++ b/src/content/en/2019/performance.md
@@ -80,7 +80,7 @@ In Figures 3 and 4 above, the FCP distributions are broken down by desktop and p
 
 According to PSI's classification, 17% of websites have fast FCP experiences overall for desktop users, compared to 11% for mobile users. The entire distribution is skewed to being slightly faster for desktop experiences, with fewer slow websites and more in the fast and moderate category.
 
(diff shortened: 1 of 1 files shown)


src/content/en/2019/performance.md
```diff
@@ -71,16 +71,16 @@ To help us understand how users experience FCP across different devices, let's s
...
 
-Why might desktop users experience fast FCP on a higher proportion of websites than phone users? We can only speculate, after all this dataset is meant to answer how is the web performing and not necessarily _why_ it's performing that way. But one guess could be that desktop users are connected to the internet on faster, more reliable networks like WiFi rather than cell towers. To help answer this question we can also explore how user experiences vary by ECT.
+Why might desktop users experience fast FCP on a higher proportion of websites than phone users? We can only speculate, after all, this dataset is meant to answer how the web is performing and not necessarily _why_ it's performing that way. But one guess could be that desktop users are connected to the internet on faster, more reliable networks like WiFi rather than cell towers. To help answer this question, we can also explore how user experiences vary by effective connection type.
```
rviscomi: We will have already defined ECT by this point, so I think it's ok to use the abbreviation. WDYT?
```suggestion
Why might desktop users experience fast FCP on a higher proportion of websites than phone users? We can only speculate, after all, this dataset is meant to answer how the web is performing and not necessarily _why_ it's performing that way. But one guess could be that desktop users are connected to the internet on faster, more reliable networks like WiFi rather than cell towers. To help answer this question, we can also explore how user experiences vary by ECT.
```

rviscomi:
Committed my suggestion to move this along. Merging. Thanks @rachellcostello!

Issue #318, opened 2019-11-04
Create a sitemap.xml file


bazzadp:
We should create a sitemap.xml file per supported language (e.g. https://almanac.httparchive.org/sitemap-en.xml for English) as part of the SEO recommendations discussed in https://github.com/HTTPArchive/almanac.httparchive.org/issues/286.

This should be added to `robots.txt` being generated as part of https://github.com/HTTPArchive/almanac.httparchive.org/issues/293 like so ([multiple sitemaps are allowed](https://developers.google.com/search/reference/robots_txt#google-supported-non-group-member-lines)):

```
User-agent: *
Allow: /
Sitemap: https://almanac.httparchive.org/sitemap-en.xml
Sitemap: https://almanac.httparchive.org/sitemap-jp.xml
```

This should be of the following format with one section per page for that language:

```
<urlset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9 http://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd">
<url>
  <loc>https://almanac.httparchive.org/en/2019/</loc>
  <lastmod>2019-11-04</lastmod>
</url>
<url>
  <loc>https://almanac.httparchive.org/en/2019/table-of-contents</loc>
  <lastmod>2019-11-04</lastmod>
</url>
...etc. for rest of non-chapter pages
<url>
  <loc>https://almanac.httparchive.org/en/2019/markup</loc>
  <lastmod>2019-11-04</lastmod>
</url>
...etc. for rest of chapter pages
</urlset>
```

The `lastmod` date can be derived from the date discussed in https://github.com/HTTPArchive/almanac.httparchive.org/issues/317 so this could be dynamically generated on the fly, or as part of `npm run generate`.

AymenLoukil:
@bazzadp i would prefer to have [a sitemap index](https://support.google.com/webmasters/answer/75712?hl=en) that lists all the sitemaps.
So in the robots.txt we put just the sitemap index URL.

I wouldn't add the  <changefreq></changefreq> attribute (not used by Google and not useful n our case for the other engines.)

So in the sitemap-en.xml we will loop through all the years pages ?

bazzadp:
[...]

PR #344, opened 2019-11-06
Serve a robots.txt file

### LINKED ISSUE #293: Create a robots.txt file
We should create a [robots.txt file ](https://support.google.com/webmasters/answer/6062596?hl=en)

rviscomi:
Fixes #293 
Progress on #318 

I created an empty sitemap.xml file to show how it can be served, similar to robots.txt. We still need to populate the file. If it's dynamically generated we'll need a different approach.

### DIFF
--- a/src/static/robots.txt
+++ b/src/static/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Allow: /
--- a/src/main.py
+++ b/src/main.py
@@ -111,7 +111,6 @@ def get_chapter_nextprev(config, chapter_slug):
 
 
 @app.route('/robots.txt')
-@app.route('/sitemap.xml')
 def static_from_root():
     return send_from_directory(app.static_folder, request.path[1:])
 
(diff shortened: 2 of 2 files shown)


src/static/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
```
AymenLoukil: Ensure its UTF-8 encoded

rviscomi: Is the action item to explicitly serve a content encoding header with this file? Or is there something else we need to do?

AymenLoukil: Just to be sure when creating the txt file, should be UTF-8 encoded. In the answer header we should have : content-type: text/plain; charset=utf-8

rviscomi: Ok yeah Flask returns a header with this charset.

> `Content-Type: text/plain; charset=utf-8`

rviscomi:
I've removed the sitemap parts of this PR since @bazzadp is working on that in #345

Issue #317, opened 2019-11-04
Autogenerate timestamps


bazzadp:
As part of https://github.com/HTTPArchive/almanac.httparchive.org/pull/302 and https://github.com/HTTPArchive/almanac.httparchive.org/pull/308 we added Article Structured Data which depends on having a Published Date (required or it fails the [Google Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool)) and a Last Updated Date (recommended by the [Google Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool)). These have been implemented in the chapter markdown files as meta data. For example:

```
---
part_number: IV
chapter_number: 20
title: HTTP/2
description: HTTP/2 chapter of the 2019 Web Almanac covering adoption and impact of HTTP/2, HTTP/2 Push, HTTP/2 Issues, and HTTP/3
authors: [bazzadp]
reviewers: [bagder, rmarx, dotjs]
published: 2019-11-04T12:00:00+00:00:00
last_updated: 2019-11-04T12:00:00+00:00:00 
---
```

This allows for different timestamps for different publication dates and also for different timestamps for the translations which might be published at different dates - hence why I put it in the markdown.

At present these dates must be manually maintained which @rviscomi and I have been discussing.

I think that makes sense for the Published Date as this should be set once and forgotten about, but for the last_updated this is more error prone and people may forget to update them when changing the contents.

It should **not** be generated as part `npm run generate` as the contents my not have changed for every chapter (or even any chapter!) when it is run, and also generic formatting changes, which are not specific to this chapter (e.g. when base_chapter.html or chapter.html is updated) probably don't count as a "real edit".

Requirements:
- Auto update `last_updated` meta data in chapter md files when a merge is accepted into that chapter’s markdown. Is it possible to do this as part of a GitHub action?

The format currently has to be as per above example, however this will be converted as part of `npm run generate` as per https://github.com/HTTPArchive/almanac.httparchive.org/pull/345 so less of a concern (and we should use an easier format for authors!).
- ~~Bonus points 1 - have a global "earliest publish date" per year and auto flag any date (`published` or `last_updated`) before then (or when a date is not present?) to prevent errors creeping in and prevent the merge. Set this global constant to 2019-11-04T12:00:00+00:00:00 for now and then we can update it later once we have an actual publish date.~~ This has been implemented as part of https://github.com/HTTPArchive/almanac.httparchive.org/pull/345
[...]

PR #345, opened 2019-11-06
Auto generate sitemap

bazzadp:
Fixes https://github.com/HTTPArchive/almanac.httparchive.org/issues/318

Also does some of https://github.com/HTTPArchive/almanac.httparchive.org/issues/317 to set min dates (would still be good to set the actual date on a git merge though!).

It generates the sitemap as part of `npm run generate`.

@rviscomi unlike https://github.com/HTTPArchive/almanac.httparchive.org/pull/344 I've put it in the templates folder because it's not completely static and don't want people to think they should edit that file manually. WDYT?

Would appreciate a thorough code review here as my JavaScript ain't the best!

Thanks,
Barry


AymenLoukil:
Hello Barry ! 
I will check that

rviscomi:
This looks great to me. I'll defer to @AymenLoukil for SEO correctness and @mikegeyser for the script integration.

mikegeyser:
Would it not make sense to move the sitemap generate code into a distinct file, much like `generate_table_of_contents` and `generate_figure_ids`? I feel like it makes the solution a lot tidier and more manageable.

src/tools/generate/generate_chapters.js
```diff
@@ -12,7 +12,21 @@ converter.setFlavor('github');
...
+const sitemap_path = `templates/sitemap.xml`;
+
+var config = {};
+var chapters_config = [];
+var todo_chapters = [];
+
 const generate_chapters = async () => {
+
+  //Read in the Chapters to see which ones are still to do (needed for sitemap)
+  await read_config('2019');
```
mikegeyser: I know we only have one year's config at the moment, but should this be hard coded?

src/tools/generate/generate_chapters.js
```diff
@@ -12,7 +12,21 @@ converter.setFlavor('github');
...
+  chapters_config = config.outline.map(function(item) { return item["chapters"]; }).flat();
```
mikegeyser: Instead of mutating the module global `var`s, I find it's more sustainable to initialise them as `const` and then pass the reference into subsequent function calls? It makes the interdependencies between sections of functionality far more explicit.

bazzadp:
> Would it not make sense to move the sitemap generate code into a distinct file, much like `generate_table_of_contents` and `generate_figure_ids`? I feel like it makes the solution a lot tidier and more manageable.

Potentially, though currently they are done once per chapter, whereas this needs to do some work before (to start the file and add the non-chapters), some as per chapter (which could be pulled out in a similar way as you suggest) and some after (to add the end tag) so not sure you'd actually pull much of the code out and make it that much clearer? Especially as some of the code is shared with the chapter and non-chapters.

Open to suggestions on how we do this though? Or could accept and iterate once you're happy with the code, even if not exactly happy with the approach?

src/tools/generate/generate_chapters.js
```diff
@@ -22,14 +36,53 @@ const generate_chapters = async () => {
...
       const { metadata, body, toc } = await parse_file(markdown);
+      check_meta_dates(metadata);
       await write_template(language, year, chapter, metadata, body, toc);
+      await write_sitemap(language, year, chapter, metadata);
     } catch (error) {
       console.error(error);
       console.error('  Failed to generate chapter, moving onto the next one. ');
     }
   }
+  await write_sitemap_end();
```
mikegeyser: I know it's an awkward challenge, but instead of `write_sitemap_start`, a number of  `write_sitemap` and then `write_sitemap_end`, would it make sense to have the `write_sitemap` return the data (which we can keep in an array) and then only open and write the file once? That way we don't have to keep opening and appending to the same file.

src/tools/generate/generate_chapters.js
```diff
@@ -72,10 +125,67 @@ const write_template = async (language, year, chapter, metadata, body, toc) => {
   });
 
   await fs.outputFile(path, fomatted_html, 'utf8');
-
   await size_of(path);
 };
 
+const write_sitemap_start = async (language, year) => {
+  
+  await fs.outputFile(sitemap_path, '<urlset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9 http://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd">\n', 'utf8');
```
mikegeyser: @rviscomi It's xml and not html, but would it make sense to use the an ejs template for this - like we do for the chapter/ toc itself?

src/tools/generate/generate_chapters.js
```diff
@@ -22,14 +36,53 @@ const generate_chapters = async () => {
...
       console.error(error);
       console.error('  Failed to generate chapter, moving onto the next one. ');
     }
   }
+  await write_sitemap_end();
 };
 
+const read_config = async (year) => {
+  try {
+  config_file = fs.readFileSync('config/'+year+'.json');
```
mikegeyser: I know it doesn't *really* matter in this context, but perhaps consider `await fs.readFile` instead of `fs.readFileSync`?

mikegeyser:
I can help with some of the feedback items, if you'd like? The PR is from your fork of the repo, though. I'm actually not sure how to go about it, though, would I PR into your fork?

src/tools/generate/generate_chapters.js
bazzadp: Yes - should have done this like this originally.

I also think it would be a much better process for the future anyway as alt languages should be handled differently than I do currently (should be `<xhtml:link ...>` links rather than separate entries as I have currently (see: https://support.google.com/webmasters/answer/189077?hl=en). This change would make doing that much easier!

src/tools/generate/generate_chapters.js
bazzadp: I actually implemented that! But that means it would be dynamically generated? So I ripped it out as we would need to iterate through each chapter to get the last_updated timestamp, so that would be slow. Unless we moved that date into config? But as per our other converstion I quite like it where it is as can be different per chapter and language...

To me it is simple enough and static so not sure there is need for ejs template.

src/tools/generate/generate_chapters.js
bazzadp: And there's where my JS knowledge stumbles.

bazzadp:
> I can help with some of the feedback items, if you'd like? The PR is from your fork of the repo, though. I'm actually not sure how to go about it, though, would I PR into your fork?

That would be awesome if you could! I think I bit off more than I could chew with this one 😢 Though happy that I eventually got it working!

We can either PR into my branch, or I can give you direct commit access to that repo, or I can change this merge request into a new branch in the main repo?

Or might be easier to just merge this as is and then you open your own PR to improve it? It's only changing the generate script so not like that will be called by the website itself. So approve and iterate?

src/tools/generate/generate_chapters.js
bazzadp: I know it's not an excuse but there is precence here 😊
https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/config.py

mikegeyser:
Give me an hour or so, and let's see if we can get it closer? :) Then we can decide to take it in and iterate.

src/tools/generate/generate_chapters.js
mikegeyser: It would be static. We're just using ejs to help with the code generation.

src/tools/generate/generate_chapters.js
```diff
@@ -72,10 +125,67 @@ const write_template = async (language, year, chapter, metadata, body, toc) => {
...
+  await write_sitemap_for_file('templates/' + language + '/' + year + '/contributors.html', 'https://almanac.httparchive.org/en/2019/contributors.html');
+
+};
+
+const write_sitemap_for_file = async (file, url) => {
+  
+  var lastmod;
+
+  let static_file = fs.readFileSync(file).toString();
+  var rx = /"datePublished": "([0-9\-\+\:T]*)"/;
```
mikegeyser: For this, `datePublished` is being generated by the `published` frontmatter attribute on the md file. If we just use the data from the md file, then we can forego the regex. Would that be ok?

mikegeyser: Oh wait, nevermind. This is only for the static files.

bazzadp: Yeah there must be a better way to do this too, but that's the best way I could come up with for now.

bazzadp: Alternative would be to do all the .md files like this so only one way of doing it?

mikegeyser:
Ok, I've made the changes. @bazzadp, would you please take a look and see if you're happy? The biggest change is using ejs to generate the static `sitemap.xml` file and moving it into a dedicated module. I believe that this simplifies things greatly.

The changes will also use the content (source md) files to determine which static files to include, so that we can accommodate different years and languages. 

I've left the utility functions as is.

bazzadp:
This looks great thanks so much @mikegeyser !

The only comment I have is instead of storing these as and array of `urls`, should we store them as an array of `pages` (based off the English ones as the master ones) and then each page can have an array of other languages per page (with an object giving the url and  where they exist for other languages? This would allow us to implement a proper localised sitemap as per [Google's suggested way of doing this](https://support.google.com/webmasters/answer/189077?hl=en):

```
  <url>
    <loc>https://almanac.httparchive.org/en/2019/caching</loc>
    <lastmod>2019-11-05</lastmod>
    <xhtml:link 
               rel="alternate"
               hreflang="de"
               href="https://almanac.httparchive.org/de/2019/caching"/>
    <xhtml:link 
               rel="alternate"
               hreflang="en"
               href="https://almanac.httparchive.org/en/2019/caching"/>
    <xhtml:link 
               rel="alternate"
               hreflang="jp"
               href="https://almanac.httparchive.org/jp/2019/caching"/>
  </url>
```

Which with your new ejs templat would be quite simple if we had the `pages` array.

I can't see a way to add lastmod ates to alternative languages (maybe there is an assumption that they are all modified when the main page is).

But then I didn't do this either so maybe one for future work?

src/tools/generate/generate_chapters.js
rviscomi: Yeah it's not perfect but we can iterate on it after launch for new editions.

mikegeyser:
That's interesting (I had no idea), and *definitely* possible. I assume that we would always use `en` as the base language, and then put all of the pages for a chapter in the `<link>`. The `<lastmod>` only being on the parent is bizarre, though.

Would you like me to do it as a part of this PR, or should we iterate once we have another translation. Is there an SEO impact one way or another?

bazzadp:
Depends how much effort :-)

There would be a better SEO of doing it with alternatives - it basically links the pages so the search engines know they are the same page, whereas listing them separately would not.

On that note we should also add a `<link rel=“alternative”... >` in the HEAD of each page when we have it for similar reasons which should be easy enough but will hold off on that until we have some examples.

rviscomi:
> On that note we should also add a <link rel=“alternative”... > in the HEAD of each page when we have it for similar reasons which should be easy enough but will hold off on that until we have some examples.

Already done!

https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/templates/base.html#L27

bazzadp:
@mikegeyser in case you are still working on this, I've pulled in latest master so I could add the sitemap link to Robots.txt file as part of this. You might wanna do a `git pull` to get the latest.

AymenLoukil:
> Depends how much effort :-)
> 
> It would be a better SEO by doing it with `alternatives` - it basically links the pages so the search engines know they are the same page, whereas listing them separately would not.
> 
> On that note we should also add a `<link rel=“alternative”... >` in the HEAD of each page when we have it for similar reasons which should be easy enough but will hold off on that until we have some examples.

:D I saw this in mail email and i said : No we already implemented alternate hreflang stuff in <head>

AymenLoukil:
When i run npm run generate i get an error   

UnhandledPromiseRejectionWarning: TypeError: file.match is not a function or its return value is not iterable
    at generate_chapters (C:\Users\Aymen\almanac.httparchive.org\src\tools\generate\generate_chapters.js:19:50)
    at <anonymous>
(node:6360) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:6360) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code. @bazzadp

bazzadp:
What version of node are you running? I'm running v12.13.0 and it works for me.

Can you generate in the master branch or is it all branches?

AymenLoukil:
> What version of node are you running? I'm running v12.13.0 and it works for me.
> 
> Can you generate in the master branch or is it all branches?

Yeah same V12.13.0 and it's on master.

This line seems to be the origin of the error 
const [path, language, year, chapter] = file.match(re);

bazzadp:
> Yeah same V12.13.0 and it's on master.

So nothing to do with this PR which is on bazzadp:generate_sitemap?

mikegeyser:
> Yeah same V12.13.0 and it's on master.
> 
> This line seems to be the origin of the error
> const [path, language, year, chapter] = file.match(re);

Hm, weird. I'm not sure wha's going on, but it should work. It likely means that either the regex isn't matching properly (perhaps picking up a rogue file that's in the wrong place?), but I can't reproduce it on master.

bazzadp:
Did you get a chance to look https://github.com/HTTPArchive/almanac.httparchive.org/pull/345#issuecomment-550441488 @mikegeyser ? Or should we just merge this as is and iterate when we get our first translation?

mikegeyser:
I haven't had a chance yet (today hasn't gone according to plan). I'll let you know by this evening.

src/templates/sitemap.xml
```diff
@@ -0,0 +1,318 @@
...
+            href="https://almanac.httparchive.org/en/2019/http2"/>
+           
+    </url>
+
+    <url>
+        <loc>
+            https://almanac.httparchive.org/en/2019/javascript
+        </loc>
+        <lastmod>
+
```
mikegeyser: `en/2019/javascript.md` doesn't have a `last_modified`. I'll try and fix this up in the script.

mikegeyser:
@bazzadp The implementation is a lot more complicated, unfortunately, but I *think* it produces the right output. Please check it and see if you're happy that we go ahead with this?

src/templates/sitemap.xml
mikegeyser: Actually, can we do this as a part of #317, please ? This PR is becoming heavy. :)

bazzadp: Fine with me. This will be fixed for this chapter before go live anyway and a blank lastmod tag is not actually a bad way to handle it anyway.

And as you say #317 will possibly avoid this happening anyway.

bazzadp:
LGTM! Let's merge I say if you're happy @rviscomi !

src/templates/sitemap.ejs.xml
```diff
@@ -0,0 +1,18 @@
...
+    <url>
+        <loc>
+            https://almanac.httparchive.org/<%= url.loc %>
+        </loc>
+        <lastmod>
+            <%= url.lastmod %>
+        </lastmod>
+        <% for (let alt of url.alternatives ) { %>
+        <xhtml:link 
+            rel="alternate"
```
AymenLoukil: no need for this because we already handly hreflang tags in the HTML which is recommended.

src/templates/sitemap.xml
```diff
@@ -0,0 +1,318 @@
...
+    <url>
+        <loc>
+            https://almanac.httparchive.org/en/2019/index.html
+        </loc>
+        <lastmod>
+            2019-11-05
+        </lastmod>
+        
+        <xhtml:link 
+            rel="alternate"
```
AymenLoukil: Same. No alternate tags in sitemap.xml

src/templates/sitemap.ejs.xml
mikegeyser: A question: is it unnecessary, or harmful? Do we have to remove it, or can we leave it in both places?

bazzadp: Why not both? Can argue don’t need a site map at all as have links on the page but it’s best practice. As is this.

AymenLoukil: It is unnecessary to have both of the methods for hreflangs. Best one is in HTML code. 

> Can argue don’t need a site map at all as have links on the page but it’s best practice. As is this.

Sitemap has another role than the links on page. It's a way to tell Google what are the pages we want to make it discover / index. Also allows as to follow indexing metrics on Search Console.

bazzadp: > It is unnecessary to have both of the methods for hreflangs. Best one is in HTML code.

OK but is there any harm in having it as is?

I still go with, if we're going to have a site map, and we're going to have hreflang implemented on our site then surely it's better to use a hreflang sitemap?

Alternatively, if we do not have a hreflang sitemap do we still list all pages, for all languages in the sitemap but just not link them?

If no harm, but no real benefit as we have it in the HTML, then the only reason to remove it, is if we want to reduce code complexity. Will let @mikegeyser make the call on that. Otherwise I'd say keep it, as technically it's the right way to have a multilingual sitemap.

bazzadp: OK I've found out the answer to this one:

>I can't see a way to add lastmod dates to alternative languages versions

I've found out is that the sitemap should be reciprocal as well. See [Example in Google Docs](https://support.google.com/webmasters/answer/189077?hl=en). So we should list **both** languages and cross reference them back. Like this:

```
  <url>
    <loc>https://almanac.httparchive.org/en/2019/caching</loc>
    <lastmod>2019-11-05</lastmod>
    <xhtml:link 
               rel="alternate"
               hreflang="de"
               href="https://almanac.httparchive.org/de/2019/caching"/>
    <xhtml:link 
               rel="alternate"
               hreflang="en"
               href="https://almanac.httparchive.org/en/2019/caching"/>
  </url>
  <url>
    <loc>https://almanac.httparchive.org/de/2019/caching</loc>
    <lastmod>2019-11-05</lastmod>
    <xhtml:link 
               rel="alternate"
               hreflang="de"
               href="https://almanac.httparchive.org/de/2019/caching"/>
    <xhtml:link 
               rel="alternate"
               hreflang="en"
               href="https://almanac.httparchive.org/en/2019/caching"/>
  </url>
```

Currently I think we are only listing the English pages with the alternatives and not vice versa (though difficult to tell as no alternative languages currently).

OK for that reason then our Sitemap is going to grow exponentially with each language we add. Plus @mikegeyser said the code [was more complex](https://github.com/HTTPArchive/almanac.httparchive.org/pull/345#issuecomment-551094102):

>@bazzadp The implementation is a lot more complicated, unfortunately

So I think given all this, and the fact we have them listed on the pages themselves, let's revert that last commit and go with the simple sitemap without hreflangs.

Apologies for wasting your time @mikegeyser 😢

bazzadp:
OK they is good to merge now if you're happy @rviscomi ?

PR #355, opened 2019-11-07
Lazy loading images

### LINKED ISSUE #351: Add native lazy loading for chapter images (inc iframes)
Some of the chapters might have a lot of graphics. If possible we should add [native lazy loading](https://web.dev/native-lazy-loading/) for the chapter graphics (except for the chapter heading and the logo).

I'm not sure that is possible as we are writing in markdown and using [showdown](https://github.com/showdownjs/showdown/wiki) to convert to HTML and it doesn't look to have an option to add attributes to image tags.

Could do a bulk edit for image tags in body section at end of chapter generation but that sounds kind of hacky...

Easiest option might be to change all images in chapter markdown to native HTML instead of markdown? Once-off pain for long term gain?

Thoughts @rviscomi , @mikegeyser ?

mikegeyser:
Closes #351.

Changed the generated script, as well as all of the static pages, to use loading='lazy' for all images. I hope I wasn't *too* heavy handed with the `loading="lazy"`, but it seems reasonable to me and doesn't affect the experience of the site in any way.


src/tools/generate/lazy_load_images.js
```diff
@@ -0,0 +1,9 @@
+const lazy_load_images = (html) => {
+  const re = /<img\s/gi;
```
rviscomi: My only concern is that this is so broad that it'll pick up things like:

```html
<pre>
  <img href="example.jpg">
</pre>
```

where the image HTML is part of the actual content.

One way to mitigate this problem is to limit it to `` tag.

mikegeyser: I know what you mean, I was also worried about that. It doesn't seem to have picked any up, yet (I did a spot check on the generated content).

I think it would be better to explicitly exclude `` inside a `<pre>` tag rather than only look for them inside a `<figure>` tag. I know it would be possible with regex, but that may be beyond my skill level to do. If we try and do it via JSDOM, we will run into the same serialization issues that plagued the first attempt at `generate_figure_id`.

mikegeyser: But if we do only look for images inside `<figure>`, do you think that [this regex would suffice](https://regex101.com/r/F0zR4w/2)?

bazzadp: I much prefer this - means the chapter images and other logos will not be lazy loaded.

Just tried that regex with 3 figure images and it seems to only pick up the last one. Will have a play with it and see what I can come up with.

bazzadp: This one seems to be better: https://regex101.com/r/RAGFDE/1

mikegeyser: Oh cool! Thanks! I'll incorporate that one.

bazzadp: Added a test case for tables to prove they don't get picked up: https://regex101.com/r/RAGFDE/3

mikegeyser: Hm. The only problem with being more specific, is that not every chapter has put the images inside figures. I honestly believe that it may be better to be overly enthusiastic about putting `loading="lazy"` everywhere, than missing out most of the `` tags - but I stand to be corrected.

I'm committing both regexes and we can swap between them.

bazzadp: Every chapter should have images in figures and will have soon as part of copy editing. And if they don't then they just miss out on lazy loading - so not worst thing in the world.

Not sure on impact of being overly enthusiastic. Will the main chapter image, plus any logos of the like (which should not be lazy loaded) load slightly later as the browser takes a bit longer to figure out whether they are needed or not? Micro seconds extra maybe, but I still like trying to do it right.

bazzadp: Plus, we forgot about the main reason we did this more specific call was to address @rviscomi's concern:

> My only concern is that this is so broad that it'll pick up things like:
> 
> ```
> <pre>
>   
> </pre>
> ```
> 
> where the image HTML is part of the actual content.
> 
> One way to mitigate this problem is to limit it to `` tag.

src/templates/en/2019/chapter.html
```diff
@@ -112,7 +112,7 @@ <h2 class="header">Index</h2>
             <h1 class="title">
                 {{ metadata.get('title') }}
             </h1>
-            <img src="/static/images/2019/{{ get_chapter_image_dir(metadata) }}/hero_lg.jpg" class="content-banner">
+            <img loading="lazy" src="/static/images/2019/{{ get_chapter_image_dir(metadata) }}/hero_lg.jpg" class="content-banner">
```
bazzadp: Main image should not be lazy-loaded IMHO

```suggestion
            <img src="/static/images/2019/{{ get_chapter_image_dir(metadata) }}/hero_lg.jpg" class="content-banner">
```

src/templates/en/2019/contributors.html
```diff
@@ -297,12 +297,12 @@ <h1 class="title">
     Contributors
   </h1>
   <div class="team">
-      <img src="/static/images/avatars/1.jpg">
-      <img src="/static/images/avatars/8.jpg">
-      <img src="/static/images/avatars/4.jpg">
-      <img src="/static/images/avatars/0.jpg">
-      <img src="/static/images/avatars/11.jpg">
-      <img src="/static/images/avatars/9.jpg">
+      <img loading="lazy" src="/static/images/avatars/1.jpg">
```
bazzadp: Wouldn't bother lazy loading these:

```suggestion
      <img src="/static/images/avatars/1.jpg">
```

src/templates/en/2019/contributors.html
```diff
@@ -297,12 +297,12 @@ <h1 class="title">
...
+      <img loading="lazy" src="/static/images/avatars/8.jpg">
```
bazzadp: Wouldn't bother lazy loading these:

```suggestion
      <img src="/static/images/avatars/8.jpg">
```

src/templates/en/2019/contributors.html
```diff
@@ -297,12 +297,12 @@ <h1 class="title">
...
+      <img loading="lazy" src="/static/images/avatars/4.jpg">
```
bazzadp: Wouldn't bother lazy loading these:
```suggestion
      <img src="/static/images/avatars/4.jpg">
```

src/templates/en/2019/contributors.html
```diff
@@ -297,12 +297,12 @@ <h1 class="title">
...
+      <img loading="lazy" src="/static/images/avatars/0.jpg">
```
bazzadp: Wouldn't bother lazy loading these:
```suggestion
      <img src="/static/images/avatars/0.jpg">
```

src/templates/en/2019/contributors.html
```diff
@@ -297,12 +297,12 @@ <h1 class="title">
...
+      <img loading="lazy" src="/static/images/avatars/11.jpg">
```
bazzadp: Wouldn't bother lazy loading these:
```suggestion
      <img src="/static/images/avatars/11.jpg">
```

src/templates/en/2019/contributors.html
```diff
@@ -297,12 +297,12 @@ <h1 class="title">
...
+      <img loading="lazy" src="/static/images/avatars/9.jpg">
```
bazzadp: Wouldn't bother lazy loading these:
```suggestion
      <img src="/static/images/avatars/9.jpg">
```

src/templates/en/2019/index.html
```diff
@@ -105,8 +105,8 @@ <h2>
         Start exploring
       </a>
     </div>
-    <img class="intro-image-2019" src="/static/images/home-hero-2019.png" alt="The Web Almanac 2019 edition">
-    <img class="intro-image" src="/static/images/home-hero.png" alt="The Web Almanac 2019 edition">
+    <img loading="lazy" class="intro-image-2019" src="/static/images/home-hero-2019.png" alt="The Web Almanac 2019 edition">
```
bazzadp: Wouldn't bother lazy loading these - though we should do better here to only load once. See https://github.com/HTTPArchive/almanac.httparchive.org/issues/361

```suggestion
    <img class="intro-image-2019" src="/static/images/home-hero-2019.png" alt="The Web Almanac 2019 edition">
```

src/templates/en/2019/index.html
```diff
@@ -105,8 +105,8 @@ <h2>
...
+    <img loading="lazy" class="intro-image" src="/static/images/home-hero.png" alt="The Web Almanac 2019 edition">
```
bazzadp: Wouldn't bother lazy loading these - though we should do better here to only load once. See https://github.com/HTTPArchive/almanac.httparchive.org/issues/361
```suggestion
    <img class="intro-image" src="/static/images/home-hero.png" alt="The Web Almanac 2019 edition">
```

src/templates/en/2019/methodology.html
```diff
@@ -124,7 +124,7 @@ <h2 class="header">Index</h2>
     <h1 class="title">Methodology</h1>
 
     <section>
-      <img class="content-banner" src="/static/images/methodology-banner.png" alt="">
+      <img loading="lazy" class="content-banner" src="/static/images/methodology-banner.png" alt="">
```
bazzadp: Wouldn't bother lazy loading these:
```suggestion
      <img class="content-banner" src="/static/images/methodology-banner.png" alt="">
```

src/templates/en/2019/table_of_contents.html
```diff
@@ -213,10 +213,10 @@ <h2 class="part-name">Appendix</h2>
     </div>
 
     <section class="characters">
-      <img class="character" src="/static/images/character-painter.png" alt=""/>
-      <img class="character" src="/static/images/character-file.png" alt=""/>
-      <img class="character" src="/static/images/character-star.png" alt=""/>
-      <img class="character" src="/static/images/character-hat.png" alt=""/>
+      <img loading="lazy" class="character" src="/static/images/character-painter.png" alt=""/>
```
bazzadp: Wouldn't bother lazy loading these:
```suggestion
      <img class="character" src="/static/images/character-painter.png" alt=""/>
```

src/templates/en/2019/table_of_contents.html
```diff
@@ -213,10 +213,10 @@ <h2 class="part-name">Appendix</h2>
...
+      <img loading="lazy" class="character" src="/static/images/character-file.png" alt=""/>
```
bazzadp: Wouldn't bother lazy loading these:
```suggestion
      <img class="character" src="/static/images/character-file.png" alt=""/>
```

src/templates/en/2019/table_of_contents.html
```diff
@@ -213,10 +213,10 @@ <h2 class="part-name">Appendix</h2>
...
+      <img loading="lazy" class="character" src="/static/images/character-star.png" alt=""/>
```
bazzadp: Wouldn't bother lazy loading these:
```suggestion
      <img class="character" src="/static/images/character-star.png" alt=""/>
```

src/templates/en/2019/table_of_contents.html
```diff
@@ -213,10 +213,10 @@ <h2 class="part-name">Appendix</h2>
...
+      <img loading="lazy" class="character" src="/static/images/character-hat.png" alt=""/>
```
bazzadp: Wouldn't bother lazy loading these:
```suggestion
      <img class="character" src="/static/images/character-hat.png" alt=""/>
```

ymschaap:
Didn't see this ticket, but I also added lazyload on the files in here:
> https://github.com/HTTPArchive/almanac.httparchive.org/tree/tables-lazyload-fonts

Let me know what works better (and I will remove).

bazzadp:
> Didn't see this ticket, but I also added lazyload on the files in here:
> 
> > https://github.com/HTTPArchive/almanac.httparchive.org/tree/tables-lazyload-fonts
> 
> Let me know what works better (and I will remove).

Well as per above conversation, I think your version is overly aggressive and prefer to be selective on which images are lazy-loaded for reasons discussed above.

Does anyone know if it's bad to over use lazy-loading or not? Raised this query to ask Google for clarification: https://github.com/GoogleChrome/web.dev/issues/1850

AymenLoukil:
> s are lazy-loaded for reasons discussed abov

We could use eager for the hero images as described in the documentation of native lazy loading.

bazzadp:
> > s are lazy-loaded for reasons discussed abov
> 
> We could use eager for the hero images as described in the documentation of native lazy loading.

Changed my review above to do that!

src/tools/generate/lazy_load_images.js
mikegeyser: I think this is working as expected, now.

rviscomi:
Getting `SyntaxError: Invalid regular expression flags` when running this

mikegeyser:
Hi @rviscomi, I'm not sure why that would be, I didn't think we were using anything esoteric. Could it me the `/s` (dotall) flag that requires a later node version? It [seems pretty new](https://2ality.com/2017/07/regexp-dotall-flag.html).

This functionality seems to be causing quite a few problems, perhaps we shouldn't be doing this right now? We can abandon this PR if you think that would be better.

bazzadp:
> Hi @rviscomi, I'm not sure why that would be, I didn't think we were using anything esoteric. Could it be the `/s` (dotall) flag that requires a later node version? It [seems pretty new](https://2ality.com/2017/07/regexp-dotall-flag.html).

Works for me with node v12.13.0

> 
> This functionality seems to be causing quite a few problems, perhaps we shouldn't be doing this right now? We can abandon this PR if you think that would be better.

What's the problem with this functionality (other than this)? You're not getting it confused with the timestamp functionality are you?

rviscomi:
Hmmm ok. If it works for everyone else it must be me. @mikegeyser could you resolve the merge conflicts and then we can ship this?

bazzadp:
> Hmmm ok. If it works for everyone else it must be me. @mikegeyser could you resolve the merge conflicts and then we can ship this?

Isn't it kind of important for you to be able to build chapters? Did you try removing the `s` from `/s` to confirm if that was it? And what version of node are you on?

rviscomi:
Here's the full error:

```
$ npm run generate
...
almanac.httparchive.org/src/tools/generate/lazy_load_images.js:5
const images_inside_figures = /<figure(?: [^<>]+)?>(?:(?:(?!<figure).)*)(<img)(?:(?:(?!<figure).)*)<\/figure>/gis;
                              ^

SyntaxError: Invalid regular expression flags
```

I'm using node v8.7.0.

bazzadp:
> I'm using node v8.7.0.

That explains it! Apparently you have to use the [`--harmony` flag](https://node.green/#ES2015-misc-Proxy--internal--get--calls-RegExp-prototype-flags) in Node v8 for dotall support. So like this I guess?

`npm --harmony run generate`

Or upgrade to a decent version of node and stop living in the past 😀

mikegeyser:
Can we perhaps park this change for the moment? I don’t feel entirely comfortable with the PR, it feels like something isn’t quite right. 

It’s not actually essential for the release, but could risk destabilization of the generation script over the weekend. I honestly feel like the most responsible thing to do would be to revisit it next week, with fresh eyes, when the rush is off.

rviscomi:
@mikegeyser +1

bazzadp:
> I've updated to v12.13.0 and now the script runs but generates errors having to do with the HTML, for example:
> 
> ```
> Generating chapter: en, 2019, markup
> SyntaxError: Unexpected closing tag "figure". It may happen when the tag has already been closed by another tag. For more info see https://www.w3.org/TR/html5/syntax.html#closing-elements-that-have-implied-end-tags (249:1)
>   247 |     <iframe width="600" height="371" seamless="" frameborder="0" scrolling="no" src="https://docs.google.com/spreadsheets/d/e/2PACX-1vTbHgqcSepZye6DrCTpifFAUYxKT1hEO56585awyMips8oiPMLYu20GETuIE8mALkm814ObJyktEe2P/pubchart?oid=2141583176&amp;format=interactive"></iframe>
>   248 |   <figcaption>Figure 3. Element frequencies as of 2019.</figcaption>
> > 249 | </figure>
>       | ^
>   250 | <p>Comparing the latest data in Figure 3 to that of Hixie's report from 2005 in Figure 2, we can see that the average size of DOM trees has gotten bigger.</p>
>   251 |    252 | <figure id="fig5">
> ```
> 
> Note the weird stray `<img loading="lazy"` on line 251. I do think something is broken with the script.

I can't repeat this on this branch.

Issue #315, opened 2019-11-04
Next/Prev chapter broken styles


rviscomi:
When there are no previous chapters, the next/prev widget at the bottom of the chapter page appears to lose its vertical centering and spacing:

![image](https://user-images.githubusercontent.com/1120896/68149196-4d10d400-fef2-11e9-8d5b-8c9e96d9e326.png)

It should look consistent with this:

![image](https://user-images.githubusercontent.com/1120896/68149240-63b72b00-fef2-11e9-9a49-af7e45fed0ef.png)

@tyohan can you take a look at this or unassign yourself if not?

rviscomi:
No word from @tyohan, unassigning.

Any @HTTPArchive/developers available to take this?

catalinred:
> No word from @tyohan, unassigning.
> 
> Any @HTTPArchive/developers available to take this?

I can do it.

rviscomi:
Thanks @catalinred!

PR #356, opened 2019-11-07
Horizontal scrollbar

### LINKED ISSUE #352: Horizontal sliding bar
In chapter template, in the bottom and at the next / previous chapter narrows there is an horizontal sliding bar 
image

mikegeyser:
This closes #352.

Removed the `overflow:auto`  because it was causing a horizontal scrollbar on chrome on a mac.

### DIFF
--- a/src/static/css/page.css
+++ b/src/static/css/page.css
@@ -197,6 +197,7 @@
   padding:50px 36px;
   border:1px solid #1A2B490A;
   margin: 20px 0;
+  overflow: hidden;
 }
 
 #chapter-navigation>a{
--- a/src/static/css/page.css
+++ b/src/static/css/page.css
@@ -197,7 +197,6 @@
   padding:50px 36px;
   border:1px solid #1A2B490A;
   margin: 20px 0;
-  overflow: auto;
 }
 
 #chapter-navigation>a{


mikegeyser:
Just repeating [my comment](https://github.com/HTTPArchive/almanac.httparchive.org/issues/352#issuecomment-550839138) for context:

> It seems to be `overflow: auto;` set on the `#chapter-navigation` style. Removing it fixes the issue for me, but I'm not sure if it was put there for a reason? Is that an acceptable fix?

catalinred:
> Just repeating [my comment](https://github.com/HTTPArchive/almanac.httparchive.org/issues/352#issuecomment-550839138) for context:
> 
> > It seems to be `overflow: auto;` set on the `#chapter-navigation` style. Removing it fixes the issue for me, but I'm not sure if it was put there for a reason? Is that an acceptable fix?

That `overflow: auto;`  was added in order to create a block formatting context due to containing next-chapter `float`. Please see #315

I have tested on macOS FF, Chrome and Safari and didn't notice that horizontal scroll although I should have guessed `overflow: hidden` is the safer choice... 😞

So, in this case,  IMO `overflow: hidden;`  should be used instead `overflow:auto` to solve #352.

src/static/css/page.css
```diff
@@ -197,7 +197,6 @@
   padding:50px 36px;
   border:1px solid #1A2B490A;
   margin: 20px 0;
-  overflow: auto;
```
catalinred: overflow: hidden;

mikegeyser: Cool, done!

PR #334, opened 2019-11-05
Copyediting of the HTTP/2 Chapter

rachellcostello:
I've made some amends of the copy in this chapter, mainly for readability.

Issue #237, opened 2019-10-25
Embed data visualizations in chapters


rviscomi:
- [x] 	1. JavaScript
- [x] 	2. CSS
- [x] 	3. Markup
- [x] 	4. Media
- [ ] 	5. Third Parties
- [x] 	6. Fonts
- [x] 	7. Performance
- [x] 	8. Security
- [x] 	9. Accessibility
- [x] 	10. SEO
- [x] 	11. PWA
- [x] 	12. Mobile web
- [x] 	13. Ecommerce
- [x] 	14. CMS
- [x] 	15. Compression
- [x] 	16. Caching
- [x] 	17. CDN
- [x] 	18. Page Weight
- [x] 	19. Resource Hints
- [x] 	20. HTTP/2

1. Find a given metric in the chapter's [public results spreadsheet](https://github.com/HTTPArchive/almanac.httparchive.org/issues/237) and use pivot tables to get the data ready for visualizing. Usually this means adding desktop/mobile as separate columns.
2. Select the appropriate chart from the [Data Viz Style Guide](https://docs.google.com/spreadsheets/d/1vmi8BJ7JarHhDPYJEzZkb1EcRcYUEYzIOv2iE_YcN5Y/edit?usp=sharing) sheet and copy it into the metric's sheet.
3. Update the chart's data source and labels accordingly. Resize the chart so it is no more than 600px wide, using existing charts of known width as a guide if needed. (600px is the default width for new charts.)
4. From the chart's three-dot menu, select "Publish chart" and copy the HTML.
5. Paste the HTML into the markdown

All graphics in a chapter (charts, images, etc) should be manually wrapped in a `figure` element with a unique ID that is sequentially numbered. All `figure` elements have a `figcaption` that describes the contents of the figure, prefixed by the figure number. 

### Interactive figures (embedded charts from Sheets)
Interactive figures should have a `visually-hidden` element for screen reader users. Level of detail here should be to convey the main visual takeaways of the chart but not overly descriptive. Add the `aria-labelledby="fig1-caption"` attribute to the `iframe` element and create a corresponding element to describe the figure.

These figures must also have a corresponding anchor/image linking to a static version of the figure. They can reuse the same `visually-hidden` element for screen reader users. Images should have an explicit width equivalent to the iframe width.

CSS will be used to ensure the interactive figure is shown to desktop users and the static figure is shown to mobile users.

```html
<figure>
[...]

PR #359, opened 2019-11-07
HTTP/2 chapter edits

rviscomi:
This is a branch of @rachellcostello's PR #334 including all of my copyedits and data viz changes.

### DIFF
--- a/src/content/en/2019/http2.md
+++ b/src/content/en/2019/http2.md
@@ -255,7 +255,7 @@ link: </assets/jquery.js>; rel=preload; as=script; nopush
 
 5% of preload HTTP headers do make use of this attribute, which is higher than I would have expected as I would have considered this a niche optimization. Then again, so is the use of preload HTTP headers and/or HTTP/2 push itself!
 
-## Issues
+## HTTP/2 Issues
 HTTP/2 is mostly a seamless upgrade that, once your server supports it, you can switch on with no need to change your website or application. Of course, you can optimize for HTTP/2 or stop using HTTP/1.1 workarounds as much, but in general, a site will usually work without needing any changes-it will just be faster. There are a couple of gotchas to be aware of, however, that can impact any upgrade, and some sites have found these out the hard way.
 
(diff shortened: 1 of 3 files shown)


rviscomi:
@bazzadp I want to reiterate that this is an awesome chapter! I love the level of depth and passion you bring to the content. Great work!!

bazzadp:
> @bazzadp I want to reiterate that this is an awesome chapter! I love the level of depth and passion you bring to the content. Great work!!

That kind of buttering up can only means you’ve ripped it to shreds in your editing 😀

Will take a look.

src/content/en/2019/http2.md
```diff
@@ -165,85 +175,96 @@ Merging these two stats together, we can see the percentage of installs, of each
...
+However, the relatively low usage of fonts and images may mean that risk is not being seen as much as was feared. `<link rel="preload" ... >` tags are often used in the HTML rather than HTTP `link` headers and the meta tags are not a signal to push. Statistics in the [Resource Hints](./resource-hints) chapter show that fewer than 1% of sites use the preload HTTP `link` header, and about the same amount use preconnect which has no meaning in HTTP/2, so this would suggest this is not so much of an issue. Though there are a number of fonts and other assets being pushed, which may be a signal of this.
 
 To get around this unintended push, you can provide the `nopush` attribute in your preload header:
 
-    link: </assets/jquery.js>; rel=preload; as=script; nopush
+```
+link: </assets/jquery.js>; rel=preload; as=script; nopush
```
bazzadp: Is this all that's needed here? thought I had to do more as per https://github.com/HTTPArchive/almanac.httparchive.org/issues/237#issuecomment-550476367

src/content/en/2019/http2.md
```diff
@@ -165,85 +175,96 @@ Merging these two stats together, we can see the percentage of installs, of each
...
+```
 
-It looks like 5% of preload HTTP headers do make use of this attribute, which is higher than I would have expected as I would have considered this a niche optimization. Then again, so is the use of preload HTTP headers and/or HTTP/2 push itself!
+5% of preload HTTP headers do make use of this attribute, which is higher than I would have expected as I would have considered this a niche optimization. Then again, so is the use of preload HTTP headers and/or HTTP/2 push itself!
 
 ## Issues
```
bazzadp: Pull back a change lost in https://github.com/HTTPArchive/almanac.httparchive.org/pull/308 to make this headingless generic and more inline with other headings.
```suggestion
## HTTP/2 Issues
```

src/content/en/2019/http2.md
rviscomi: It _should_ be all that's needed. I can confirm that it doesn't work without the class name. But then, why aren't `code-block` styles the default for _all_ `pre` elements?

bazzadp: Missing a comma

https://github.com/HTTPArchive/almanac.httparchive.org/blob/f48f1eb5ab2ae8e2a2c1477d0f286dca5472016b/src/static/css/page.css#L29-L28

Should this:

```
.code-block pre {
  overflow-y: hidden;
  overflow-x: scroll;
}
```

Be this:

```
.code-block, pre {
  overflow-y: hidden;
  overflow-x: scroll;
}
```

rviscomi: Yes! Fixed.

PR #362, opened 2019-11-07
Tables, styling

### LINKED ISSUE #353: Data tables should be shrink-wrapped and centered
Current appearance:
image

Expected appearance:
image

@ymschaap can you take a look?

ymschaap:
This fixed #353


bazzadp:
@ymschaap can you pull the latest from master into this branch as looks like a lot more changes than there should be there?

Also should we change the name of this PR to remove the refernece to lazyload? Fine leaving the branch name as is, but let's avoid confusion with this PR name which is simple to edit.

bazzadp:
But other than that, and assuming it's just CSS changes, probably don't need me as a reviewer so drive on without waiting for me.

PR #363, opened 2019-11-07
Generate last_updated

### LINKED ISSUE #317: Autogenerate timestamps

mikegeyser:
This closes #317.

This PR adds a generate script for the `last_updated` timestamp on every content file. It is not tightly coupled to the `generate_chapters` functionality, because we might want it to be a pre-commit hook instead. At the moment it will run first and uses the date of the latest commit on the file to generate the `last_modified`.

I'm a little nervous of the updating of source files, but I do think it's better than having to do it by hand.

### DIFF
--- a/src/tools/generate/generate_last_updated.js
+++ b/src/tools/generate/generate_last_updated.js
@@ -1,0 +1,37 @@
+const fs = require('fs-extra');
+const { execSync } = require('child_process');
+const { find_files } = require('./shared');
+
+const generate_last_updated = async () => {
+  for (const file of await find_files()) {
+    console.log(`\n Setting the last_updated field on ${file}`);
+
+    // Fetch the last modified date, according to the git log.
+    const date = get_last_updated_date(file);
+    console.log(`  last_updated: ${date}`);
+
+    // Read the content of the file
+    let content = await fs.readFile(file, 'utf-8');
+
+    // Replace the frontmatter last_updated field. This is not
+    // greedy, so it will match the first instance (which must be
+    // in the frontmatter).
+    content = content.replace(/last_updated:\s*(\S*)/, `last_updated: ${date}`);
+
+    // Overwrite the file with the updated date.
+    await fs.outputFile(file, content, 'utf8');
+  }
+};
+
+const get_last_updated_date = (path) => {
+  const command = `git log -1 --date=iso-strict-local ${path} | cat`;
+  const stdout = execSync(command).toString();
+  const date_string = /Date:\s+(\S*)/g.exec(stdout)[1];
+  const date = new Date(date_string);
+
+  return date.toISOString();
+};
+
+module.exports = {
+  generate_last_updated
+};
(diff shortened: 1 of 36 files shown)


src/tools/generate/generate_last_updated.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+};
+
+const get_last_updated_date = (path) => {
+  const command = `git log -1 --date=iso-strict-local ${path} | cat`;
+  const stdout = execSync(command).toString();
+  const date_string = /Date:\s+(\S*)/g.exec(stdout)[1];
+  const date = new Date(date_string);
+
+  return date.toISOString();
```
mikegeyser: This uses the ISO_8601 format of representing UTC timezone with a `Z`, instead of `+00:00`. I can change this if needed.

bazzadp: As per https://github.com/HTTPArchive/almanac.httparchive.org/issues/317#issuecomment-551154505 think that is fine.

Issue #221, opened 2019-10-22
Give readers the ability to discover queries, results, and comments for each chapter


rviscomi:
Each chapter has a subdir in [sql/2019](https://github.com/HTTPArchive/almanac.httparchive.org/tree/master/sql/2019) with all of its queries and a corresponding spreadsheet with the results.

We should make these resources available to readers as they're on a chapter page. These aren't included in the design so we'll need to create it from scratch using the existing visual language of the site.

In the chapter designs (#130) we have a section for info about each author at the bottom of the page. We can use this space to also call attention to the chapter's SQL subdir and results sheet.

Additionally, I'd like to provide readers with an easy way to react to, report issues with, and comment on chapters. I'm not really sure what the best approach is, so I'm open to ideas. One idea is to create a GitHub issue for each chapter devoted to reader feedback. I don't really like keeping issues open perpetually, though, and GitHub isn't meant to be a discussion forum. We do have the HTTP Archive discussion forum at discuss.httparchive.org where we can do something similar. Thoughts?

---

- [x] Add a new field to each chapter's markdown metadata `discuss` whose value is the four digit thread ID corresponding to the chapter's discussion post. For example `1756` is the ID for the JavaScript thread: https://discuss.httparchive.org/t/chapter-1-javascript/1756
- [x] Add a UI component to the end of the chapter template to encourage readers to discuss their thoughts. It could be something simple like a link button with a chat icon and the label "Comments".
- [x] Generate the link href value based on the chapter's metadata.

Commenting is the biggest priority for launch. We can follow up with UI to discover the queries and results later.

@HTTPArchive/developers

catalinred:
The HTTP Archive discussion forum would be a good place to share and gather feedback for the almanac chapters. It is its purpose (being a forum), after all :)

As you said above, each chapter's bottom could contain some call to actions or sections like the following:

- Discuss on HTTP Archive forum (report issues, comment and send feedback)
- View chapter's code (SQL subdir and results sheet)

Also, in order to avoid a potential three-way GitHub/almanac.httparchive.org/discuss.httparchive.org feedback sharing/gathering chaos, I wouldn't add a comments section on the almanac chapters and instead, rely and focus on the above existing actions/sections for each chapter. 

Hope this helps!

rviscomi:
Thanks for your feedback @catalinred.

[...]

PR #382, opened 2019-11-08
Add a discuss button to each chapter

rviscomi:
Progress on #221 
image


src/templates/en/2019/base_chapter.html
```diff
@@ -17,9 +17,33 @@
...
+  var el = document.getElementById('num_comments');
+  el.innerText = comments + ' ' + (comments == 1 ? 'comment' : 'comments');
+});
 </script>
 {% endblock %}
 
+{# Calls to action for readers who want to engage more with this chapter. #}
+{% macro render_actions() %}
+  <div class="discuss">
+    <a class="alt btn" href="https://discuss.httparchive.org/t/{{ metadata.get('discuss') }}">
```
bazzadp: Open in new tab so you can leave the chapter you want to discuss still open?

```suggestion
    <a class="alt btn" target="_blank" href="https://discuss.httparchive.org/t/{{ metadata.get('discuss') }}">
```

rviscomi: I've heard of some [arguments like these](https://css-tricks.com/use-target_blank/) that target=_blank is usually not a good idea. Our audience is tech-savvy so I trust them to Cmd+Click or use the back button if they want to retain the chapter page.

PR #384, opened 2019-11-09
Accessible data viz for Performance chapter

rviscomi:
Progress on #237


src/templates/sitemap.xml
```diff
@@ -86,7 +97,7 @@
 
     <url>
         <loc>
-            https://almanac.httparchive.org/en/2019/compression
+            https://almanac.httparchive.org/en/2019/ecommerce
```
rviscomi: @bazzadp is it normal for this file to churn so much like this?

rviscomi:
Good bot.

src/templates/sitemap.xml
bazzadp: It basically just scans the directory for .md files:
https://github.com/HTTPArchive/almanac.httparchive.org/blob/80716288ec85058879a3313eff7e9999305dc619/src/tools/generate/generate_chapters.js#L19-L21
And then processes them one by one.

I copied this from `generate_chapter.js` code which does the same thing.

So it's basically doing them in alphabetical order. Which is fine - until you add a chapter (which causes everything to move along) or you remove a chapter (because your branch doesn't have that chapter yet) and everything shifts backwards.

In this case, master has ecommerce and security so it had those in sitemap. Then your branch does not have those so your sitemap generates without them. So a difference.

Answers I can see are one of these:

1. Pull from master before generating sitemaps to make sure you have same chapters (except ones you're genuinely adding in your PR)
2. Removing Sitemap from `npm run generate` and have another option (`npm run generate -withsitemap`?) which we use before deploying.
3. Remove sitemap file from your commit and just ignore the fact it's changed.
4. Live with it until all chapters in there. It's a short term pain cause we're in so much flux.

rviscomi: Yeah I can live with it. It seemed like something was non-deterministic but you could be right that I had new chapters.

PR #387, opened 2019-11-09
Table of Contents Translation

c-torres:
- Adding the table_of_contents.html file with the content translated into Spanish.

Please notice that because the outline information comes directly from the config/2019.json file it is not possible to translate the actual chapter's names. The current version only translates the Chapter and Part keywords from the table of contents.

I am OK with keeping the table of content this way as would say many chapters are _universal_ keywords used to build the web, but would like to hear from @JMPerez, @taytus, and @rviscomi.

### DIFF
--- a/src/templates/es/2019/table_of_contents.html
+++ b/src/templates/es/2019/table_of_contents.html
@@ -1,0 +1,228 @@
+{% extends "es/2019/base.html" %}
+
+{% block title %}Tabla de Contenidos | Web Almanac 2019{% endblock %}
+
+{% block description %}Tabla de contenido para el Web Almanac 2019 de HTTP Archive, que enumera cada sección: contenido de la página, experiencia del usuario, publicación de contenido, distribución de contenido{% endblock %}
+
+{% block meta %}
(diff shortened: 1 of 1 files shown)


src/templates/es/2019/table_of_contents.html
```diff
@@ -0,0 +1,228 @@
+{% extends "es/2019/base.html" %}
+
+{% block title %}Tabla de Contenidos | Web Almanac 2019{% endblock %}
+
+{% block description %}Tabla de contenido para el Web Almanac 2019 de HTTP Archive, que enumera cada sección: contenido de la página, experiencia del usuario, publicación de contenido, distribución de contenido{% endblock %}
```
JMPerez:
```suggestion
{% block description %}Tabla de contenidos del Web Almanac 2019 de HTTP Archive, que enumera cada sección: contenido de la página, experiencia del usuario, publicación de contenido, distribución de contenido{% endblock %}
```

rviscomi:
Filed #398 so we can resolve the English TOC issue

Issue #346, opened 2019-11-06
Randomize order of contributors


rviscomi:
Low priority but I'd like to see the Contributors page sorted randomly as opposed to alphabetically. Gives everyone a chance to be featured at the top :)

cc @HTTPArchive/developers

c-torres:
Can this be closed? Not sure what's the workflow.

bazzadp:
I think yes. Next time if you say. "Closes #346" or "Fixes #346" as the first line in your PR it will be closed automatically.

PR #390, opened 2019-11-09
Randomize order of contributors

c-torres:
Randomizing the order of the contributors on the contributors page as requested in #346


rviscomi:
Can you also go into `src/config/2019.json` and add yourself as a contributor to the developers and translators teams? 🎉

Issue #368, opened 2019-11-08
Broken external links


AymenLoukil:
We have these broken external links

In Mobile web chapter :  
https://almanac.httparchive.org/en/2019/%5D(https://gs.statcounter.com/)

In Ecommerce chapter : 
https://almanac.httparchive.org/en/2019/amazon.com
https://almanac.httparchive.org/en/2019/ebay.com

in third parties and http ones : 

https://almanac.httparchive.org/en/2019/security [it is ok since the chapters are not yet published... what would be better to give a 404 and not HTTP 500].

in HTTP/2 chapter:

https://almanac.httparchive.org/en/2019/cdn [it is ok since the chapters are not yet published... what would be better to give a 404 and not HTTP 500].

In JavaScript chapter  :  

https://almanac.httparchive.org/methodology
https://almanac.httparchive.org/compression
[it is ok since the chapters are not yet published... what would be better to give a 404 and not HTTP 500].

rviscomi:
Great catches. All external links must be prefixed with `http[s]://` otherwise the browser assumes they're internal. We should add this prefix to all external links.

Internal links to other chapters should be prefixed with `./` (current subdirectory) instead of `/` (root directory). This allows language/year preferences to be preserved without having to use template code to generate the URL (`url_for`).

The Mobile Web chapter needs the `http[s]://` prefix.

The Third Parties broken links are actually correct, but the chapters just don't exist yet. So no changes needed.

Which chapter links to CDN? It can't be the CDN chapter itself because that one doesn't exist.

The JS chapter needs to be edited to use `./`.

bazzadp:
Should we raise a separate issue to stop responding with a 500 when a chapter doesn’t exist?

Thinking especially if all chapters don’t make launch and as translations become unavailable (or will we only launch each language when it’s available in full)?

rviscomi:
I think we can catch this special case on the server when we try to render a chapter that is marked TODO and return a custom error message like "This chapter isn't available yet, come back tomorrow" or something.

bazzadp:
Ok but you’ll also get a 500 for non-existent chapters and typos. E.g. /markupp. Should be 404 for those if possible.

rviscomi:
Yes, we've had a long-outstanding TODO to validate the chapter. We can use this issue to track that:

[...]

PR #392, opened 2019-11-09
Validate chapters URLs and add ability to have redirects for common typos

bazzadp:
Addresses some of https://github.com/HTTPArchive/almanac.httparchive.org/issues/368

Realistically I'm not going to get a chance to work on this any more this weekend so opening this in case you want to accept it as is for now. Won't be offended if you would rather wait for the full thing, in which case feel free to close this.

Adds the following:

- validates chapters and 404s when a non-valid chapter is entered, rather than a 500
- adds redirect array/list to allow us to remap URLs incase any one links us with a typo

To Do:

- Currently hardcoding the list of chapter. Should load from JSON and list by year and language based on todos.
- Currently doesn't handle a missing or broken chapter that's expected to be there.


src/validate.py
```diff
@@ -2,25 +2,68 @@
...
+        chapter = kwargs.get('chapter')
 
         accepted_args = inspect.getargspec(func).args
 
+        if chapter:
+
+            validated_chapter = validate_chapter(chapter)
+
+            if chapter != validated_chapter:
+                return redirect("/" + lang + "/" + year + "/" + validated_chapter, code=301)
```
rviscomi:
```suggestion
                return redirect('/%s/%s/%s' % (lang, year, validated_chapter), code=301)
```

Issue #218, opened 2019-10-22
Copyedit completed contents


rviscomi:
Go through each chapter and "contentful" page to ensure that it fits our "work fun" voice, fix spelling and grammar, and finalize formatting. See the [Editors' Wiki](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Wiki) for an explicit list of best practices.

Whenever possible, also try to interlink chapters (#253).

- [x]	1. JavaScript	#131 @obto
- [x]	2. CSS	#132 @rachellcostello 
- [x]	3. Markup	#133 @rviscomi
- [ ]	4. Media	#170 @rachellcostello 
- [x]	5. Third Parties	#134 @bazzadp 
- [x]	6. Fonts	#143 @bazzadp
- [x]	7. Performance	#171 @rachellcostello 
- [x]	8. Security	#158 @bazzadp for first round + someone else
- [x]	9. Accessibility	#155 @obto
- [x]	10. SEO	#163 @rachellcostello 
- [x]	11. PWA	#164 @bazzadp 
- [x]	12. Mobile web	#147 @rviscomi 
- [x]	13. Ecommerce	#135 @bazzadp 
- [x]	14. CMS	#169
- [x]	15. Compression	#145 @bazzadp 
- [x]	16. Caching	#172 @bazzadp
- [x]	17. CDN	#149
- [x]	18. Page Weight	#144 @obto 
- [x]	19. Resource Hints	#136 @rviscomi 
- [x]	20. HTTP/2	#173 @rachellcostello 
- [x] Methodology #202 @bazzadp 
- [x] Home #212 @bazzadp 
- [x] Foreword #285 @bazzadp

rviscomi:
@ekharvey @Meggin @rachellcostello IIRC you've all expressed interest in helping out with copyediting. If you still have time over the next 3 weeks to review content, it'd be a huge help! 🙏

rviscomi:
The following chapters are merged and ready for editing:

3. Markup
5. Third Parties
10. SEO
11. PWA
20. HTTP/2

Markdown files are available at [src/content/en/2019](https://github.com/HTTPArchive/almanac.httparchive.org/tree/master/src/content/en/2019).

Most of them still need data viz embedded but that shouldn't block.

rachellcostello:
[...]

PR #397, opened 2019-11-09
CMS edits and data viz

rviscomi:
Progress on #237 and #218 

Adding reviewers but forcing this through to merge anyway. Just FYI. This will be deployed live shortly!


rviscomi:
Good bot.

Issue #407, opened 2019-11-10
Broken image in Perf chapter


rviscomi:
The mobile image for Figure 9 is broken.

PR #415, opened 2019-11-10
Added a regex script that removes unnecessary markup

mikegeyser:
Addresses #412, but is an interim fix.


src/tools/generate/remove_unnecessary_markup.js
```diff
@@ -0,0 +1,10 @@
+/* 
+  TODO: This is a workaround for the unnecessary markup that's being generated 
+  by JSDOM. Once that has been fixed, this can hopefully be removed. 
+  (See #412) 
+*/
+const remove_unnecessary_markup = (html) => html.replace(/<p><\/p>/g, '');
```
bazzadp: Should we make this more generic under the assumption there might be other hacks to come and explain each hack with a comment? Or is this overkill and can be done on the next hack (if there is one)?

```suggestion
const remove_unnecessary_markup = (html) => {
  // Remove pointless paragraph tags added by JSON in wrap_tables
  html = html.replace(/<p><\/p>/g, '');
  return html
}
```

mikegeyser: We definitely should. :P

bazzadp:
LGTM - merge away if done.

BTW on another note why is that image being compressed? How can that happen as surely should have been compressed automatically by calibre when the image was aded to the repo?

rviscomi:
> BTW on another note why is that image being compressed? How can that happen as surely should have been compressed automatically by calibre when the image was aded to the repo?

That was my fault. I snuck the image into a manual commit (not via PR) to fix #407.

bazzadp:
> That was my fault. I snuck the image into a manual commit (not via PR) to fix #407.

Cheeky!

BTW was the protocol here? Last approver merges? Or PR author merges on receiving all approvals? Or @rviscomi always approves if on it?

rviscomi:
Either way. As long as we don't merge bad code 😄

Issue #418, opened 2019-11-10
Chapter generation script broken


rviscomi:
There are some Spanish translation specific errors thrown when running `npm run generate`. This is preventing the sitemap from being generated.

PR #419, opened 2019-11-10
Sitemap date fix

bazzadp:
Closes https://github.com/HTTPArchive/almanac.httparchive.org/issues/418
Also solves https://github.com/HTTPArchive/almanac.httparchive.org/issues/286#issuecomment-552181436

Changes:
- Fixes date format for sitemap for one missed case (https://github.com/HTTPArchive/almanac.httparchive.org/issues/286#issuecomment-552181436)
- Makes `rel="alternative"` and `rel="canonical"` absolute (https://github.com/HTTPArchive/almanac.httparchive.org/issues/286#issuecomment-552181436)
- Handles when templates are missing, like they currently are for Spanish chapters (https://github.com/HTTPArchive/almanac.httparchive.org/issues/418)
- Add a `sitemap_languages` array in case a chapter exists but is not ready to be launched to Google
- Sets min publish date to launch date (2019-11-11)

### DIFF
--- a/src/tools/generate/generate_chapters.js
+++ b/src/tools/generate/generate_chapters.js
@@ -66,14 +66,16 @@ const write_template = async (language, year, chapter, metadata, body, toc) => {
   const template = `templates/${language}/${year}/chapter.html`;
   const path = `templates/${language}/${year}/chapters/${chapter}.html`;
 
-  let html = await ejs.renderFile(template, { metadata, body, toc });
-  let fomatted_html = prettier.format(html, {
-    parser: 'html',
-    printWidth: Number.MAX_SAFE_INTEGER
-  });
+  if (fs.existsSync(template)) {
+    let html = await ejs.renderFile(template, { metadata, body, toc });
+    let fomatted_html = prettier.format(html, {
+      parser: 'html',
+      printWidth: Number.MAX_SAFE_INTEGER
+    });
 
-  await fs.outputFile(path, fomatted_html, 'utf8');
-  await size_of(path);
+    await fs.outputFile(path, fomatted_html, 'utf8');
+    await size_of(path);
+  }
 };
 
 module.exports = {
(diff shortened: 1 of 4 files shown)


src/tools/generate/generate_chapters.js
```diff
@@ -3,6 +3,9 @@ const showdown = require('showdown');
 const ejs = require('ejs');
 const prettier = require('prettier');
 
+//Chapters may exist but not be ready to be launched so do not include in sitemap
```
rviscomi: In validate.py, `SUPPORTED_YEARS` maps years to supported languages. We should import that to avoid two sources of truth.

src/tools/generate/generate_sitemap.js
```diff
@@ -78,7 +81,7 @@ const set_min_date = (date) => {
     if (date < min_publish_date) {
       return min_publish_date.toISOString().substr(0, 10);
     } else {
-      return date;
+      return date.toISOString().substr(0, 10);
```
rviscomi: Just to confirm, last modified can be a date and not necessarily datetime?

bazzadp: It has to be a date: https://www.sitemaps.org/protocol.html#lastmoddef

src/tools/generate/generate_chapters.js
bazzadp: Agreed, but didn't know how to do that since this is just JS and not Python.

Any pointers or similar variable that is being pulled in from Python to JS?

src/tools/generate/generate_sitemap.js
bazzadp: Though I've only ever seen dates in the wild. E.g. https://web.dev/sitemap.xml

src/tools/generate/generate_chapters.js
rviscomi: Oh true. Hmmm... this isn't a launch blocker so let's keep as-is and merge.

Issue #149, opened 2019-09-08
Write content: Chapter 17. CDN


rviscomi:
Part | Chapter | Authors | Reviewers | Analysts | Metrics | Analysis
-- | -- | -- | -- | -- | -- | --
IV. Content Distribution | 17. CDN | @andydavies @colinbendell | @yoavweiss @paulcalvano @pmeenan | @raghuramakrishnan71 @rviscomi | #19 | #98

### Due by: October 7

- [x] Review the [results spreadsheet](https://docs.google.com/spreadsheets/d/1Y7kAxjxUl8puuTToe6rL3kqJLX1ftOb0nCcD8m3lZBw/edit?usp=sharing) and work with your analyst(s) (@raghuramakrishnan71 @rviscomi) if you have any questions about the data.
- [x] Write the contents of your chapter and get it peer reviewed (@yoavweiss @paulcalvano @pmeenan).
- [ ] File a PR to merge your written contents in markdown format to the [src/content/en/2019](https://github.com/HTTPArchive/almanac.httparchive.org/tree/master/src/content/en/2019) directory.

Refer to the [Authors' Guide](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide) for writing advice.

rviscomi:
👋 Greetings! Just checking in to make sure this is progressing and remind you of the October 7 deadline timed so we can launch at Chrome Dev Summit.

The checklist above indicates that you still need to review the results, write content, and get it reviewed. Feel free to reach out if you have any questions about the process.

rviscomi:
🛎 just a ping to see if any progress has been made on this. Let me know!

andydavies:
I'm on it…

How long do you expect the chapter to be?

rviscomi:
Thanks @andydavies! The [Authors' Guide](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide) has some advice about chapter length, to summarize feel free to make it as long or short as needed to get your point across. This chapter has ~25 metrics so if you discuss all of them it could be long, but you can pick and choose to discuss only the most interesting results. Up to you.

andydavies:
Just noticed that 17_02d queries sites that load 3rd-party content from a CDN rather than uses a public library CDN

Was trying to drive down the library route so we have some idea how much effect cache partioning is going to have

rviscomi:
@andydavies per our chat offline:

[...]

PR #430, opened 2019-11-10
Editing and data viz for CDN

rviscomi:
Progress on #218 and #237


src/templates/sitemap.xml
```diff
@@ -116,7 +123,7 @@
         </loc>
         
         <lastmod>
-            2019-11-11
+            2019-11-12
```
rviscomi: @bazzadp Today is the 10th. The md metadata has the 11th. How did the sitemap get the 12th?

bazzadp: Well that's interesting!

However the md does not have the 11th. It is missing as it has two `published` dates and no `last_updated` (which is [my fault](https://github.com/HTTPArchive/almanac.httparchive.org/issues/149#issuecomment-552193999)). Can you correct that?

And will see if I can figure this out. But not able to repeat it now and it's only the JavaScript chapter that it's doing that for... which is weird. You defo don't have a weird edit in that locally do you?

rviscomi: Aha! Fixed.

bazzadp: I still don't understand how https://github.com/HTTPArchive/almanac.httparchive.org/pull/430/commits/2b68fc7f199298cd4843ec8ba069250a2b5ad60f caused it to be the 12? And still can't repeat it.

rviscomi:
I'm going to merge these changes and deploy the chapter to the production website. I think that'll make it easier to review the changes. If there are any comments we can resolve them in a new PR.

PR #437, opened 2019-11-10
Updated links for TLS visualizations

colinbendell:


rviscomi:
LGTM. Leaving this open so I can update the screenshots.

rviscomi:
@colinbendell I've taken new screenshots but in the visual diff it doesn't look like anything has actually changed? In fact, the bot just removed the images from the commit because they're identical to what's already in the repo. :/

rviscomi:
@colinbendell can you confirm that the new data viz URLs correspond to updated charts? It seems to me that they're identical to the previous version.

Issue #452, opened 2019-11-11
Investigate 404 errors


rviscomi:
In the production server logs I'm seeing lots of ambiguous error messages like this:

```
werkzeug.exceptions.NotFound: 404 Not Found: The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.
at match (/env/lib/python3.7/site-packages/werkzeug/routing.py:1799)
at match_request (/env/lib/python3.7/site-packages/flask/ctx.py:336)
at raise_routing_exception (/env/lib/python3.7/site-packages/flask/app.py:1774)
at dispatch_request (/env/lib/python3.7/site-packages/flask/app.py:1791)
at full_dispatch_request (/env/lib/python3.7/site-packages/flask/app.py:1813)
```

At times the server is spiking at 200 404s per minute. (This is suspiciously high)

Sometimes this happens when a site doesn't have a favicon or something innocuous, but I can't imagine why we'd be having this many 404s unless there's a broken link somewhere.

Two things:

- [ ] Improve error logging so we know what the broken link is and where it's coming from (cc @mikegeyser)
- [ ] Rerun the SEO-style audit of the website so that we can more easily/proactively find broken links (#286 cc @AymenLoukil @catalinred @rachellcostello)

rviscomi:
This error may be related:

```
AttributeError: 'NoneType' object has no attribute 'get'
at render_template (/srv/main.py:17)
at page_not_found (/srv/main.py:136)
at handle_http_exception (/env/lib/python3.7/site-packages/flask/app.py:1644)
at handle_user_exception (/env/lib/python3.7/site-packages/flask/app.py:1713)
at full_dispatch_request (/env/lib/python3.7/site-packages/flask/app.py:1815)
at wsgi_app (/env/lib/python3.7/site-packages/flask/app.py:2292)
```

But it's similarly ambiguous.

bazzadp:
I added the favicon with https://github.com/HTTPArchive/almanac.httparchive.org/pull/438 btw.

Also see this with `<link rel="apple-touch-icon"` so could be that - and we probably should add anyway even if not that this time.

bazzadp:
PWA has a link to `./mobile` instead of `./mobile-web` but doubt that's the cause.

AymenLoukil:
I run a crawl and here are the links generating errors : 

[...]

PR #457, opened 2019-11-11
Additional logging for 404s

mikegeyser:
This is related to #452.

Logging the url as a part of the 404 error message. We can expand upon this, but it seems like a start.

Please check this? It does seem to work on my machine, and it's how [flask docs indicate](https://flask.palletsprojects.com/en/1.1.x/reqcontext/) you should work with the current http request context.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -141,7 +141,9 @@ def bad_request(e):
 
 @app.errorhandler(404)
 def page_not_found(e):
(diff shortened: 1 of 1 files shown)


bazzadp:
@mikegeyser I already did this as part of https://github.com/HTTPArchive/almanac.httparchive.org/pull/455

PR #458, opened 2019-11-11
Fix missing closing tag `third-parties`

ymschaap:
* Fix missing closing tag `third-parties`

(it caused all the text below shown in _cursive_)

(npm run generate had an issue, so couldn't commit that)


rviscomi:
Good catch! Can you also run `npm run generate` to add the corresponding html changes?

PR #469, opened 2019-11-11
More robust 404 handling

bazzadp:
Fixes https://github.com/HTTPArchive/almanac.httparchive.org/issues/452


rviscomi:
I'll deploy this shortly. We should do a thorough QA test of the website to make sure this doesn't blow anything up. Changing serving logic could be tricky.

bazzadp:
Did you send this live? Seems to be on production now.

Issue #439, opened 2019-11-10
HTML validation errors on Home Page


bazzadp:
Similar to https://github.com/HTTPArchive/almanac.httparchive.org/issues/309 index.html has the following:

- line 167 - `<h1><div class="alt-bg decorative-line"></div>`
- line 184 - `<h2><div class="decorative-line-center"></div>`
- line 221 - `<h2><div class="decorative-line"></div>`
- line 228 - `<h2><div class="decorative-line-center"></div>`

Divs are not allowed in heading tags. Tried a few fixes but decided to leave to someone with better CSS than me!

Also the following `<img>` tags could do with `alt=""` properties (assuming we're happy with that as these are just decorative): 

- line 221 - character-markup.png
- line 222 - character-star.png
- line 223 - character-hat.png

catalinred:
@bazzadp An easy validation fix would be to transform `<div class="decorative-line-center">` into `<span class="alt-bg decorative-line"></span>` and then add `display: block` to the `selector`. This way we could avoid any potential new issue without rewriting the  `.decorative-line`.

Alternatively, we could rewrite the `.decorative-line` to use CSS pseudo-elements instead. Either way, I could help with this.

rviscomi:
The decorative line is a vestige of an earlier style implementation. I rewrote it on the chapter/methodology/TOC pages but didn't realize we're still using it on the home page. We should be able to do this style entirely in CSS without the need for additional markup.

catalinred:
> The decorative line is a vestige of an earlier style implementation. I rewrote it on the chapter/methodology/TOC pages but didn't realize we're still using it on the home page. We should be able to do this style entirely in CSS without the need for additional markup.

If that's ok, you could assign me to merge the chapter/methodology/TOC  `.title` and index/homepage's `.decorative-line`.

I'll also add `alt=""` as @bazzadp suggested above for the decorative images/

PR #471, opened 2019-11-11
Consistent titles and subtitles

catalinred:
This fixes the HTML validation errors for #439

- Added alt="" for the decorative images.
- Everything related to `decorative-line`, `decorative-line-center`, `.alt-bg.decorative-line or `.alt-bg .decorative-line` was removed.
- All the main headings are now `.title`, with modifiers like `title-lg`, `title-alt` or `title-center`.
- The `subtitle`, seen within chapters as an adjacent to `<h1 class="title">`, share most of styles with `title`. I propose to remove the bold styling for `subtitle` as `title` is the `h1` in those chapter pages, also two "bold" lines is too much. Also, whenever `title` comes within the DOM after the `subtitle`, removed its `margin-top` for a better typography vertical rhythm.
- Remove legacy/obsolete rules like `.intro h1, .contributors h1,.methodology h1 {...}`, `.decorative-line-mobile {...}` or `.content .intro p {...}`
- The main menu pages: Table of Contents, Contributors & Methodology now have a consistent top margin, compared to the main header.
- Last but not least, I tested a lot :)


src/static/css/page.css
```diff
@@ -44,7 +43,7 @@
   grid-area: index;
 }
 .index .index-box {
-  margin: 0 8px 8px;
+  margin: 20px 8px 8px;
```
catalinred: This top 20px matches the pages' `h1 class="title"`, so consistent top margin everywhere.

rviscomi: Not sure about this, I think we want all titles to be on the top of the container. I'll need to take a closer look.

catalinred: This comment I added is related to the main menu pages: Table of Contents, Contributors & Methodology. now have a consistent top margin, compared to the main header.

Sure, any feedback is appreciated. Let me know if you have any questions.

catalinred:
title-border-alignment-mobile

index-alignment-and-bold-subtitle

I made the fixes, pls check out the screenshots above. Thanks for the feedback!

Also, FWIW methodology.css is not used (anymore) so we should remove it. 
Seen lots of duplicates in both page.css and methodology.css e.g. on `.index .index-box` styles.

PR #474, opened 2019-11-11
Hide rel="alternative" xreflang if only one language exists

bazzadp:
As discussed here: https://github.com/HTTPArchive/almanac.httparchive.org/issues/286#issuecomment-552645368


src/language.py
```diff
@@ -18,22 +18,25 @@ def __repr__(self):
 
   @property
   def lang_attribute(self):
-    # Returns the language attribute (eg "en-US") to be used in HTML.
-    return '%s-%s' % (self._lang_code, self._region_code)
+    # Returns the language attribute (eg "en" or "en-US") to be used in HTML.
```
rviscomi: Let's keep the region in the `Language` instantiation but since we only want to remove it from the `lang` attribute we can simplify this method to only return the `lang_code`:

```py
return self._lang_code
```

bazzadp: To be clear you mean on the last line, like this?:

```python
  def lang_attribute(self):
    # Returns the language attribute (eg "en" or "en-US") to be used in HTML.
    if (self._region_code):
      return '%s-%s' % (self._lang_code, self._region_code)
    return self._lang_code
```

I do wonder if we should just rip out any references to `region` completely from this file?

If you want a new region, could just set up a new language called 'en-US' for example, cause the current chapter folder structure doesn't support regions so, it would basically need to be set up as a language anyway to work.

So why bother having regions at all?

WDYT?

rviscomi: Sent you #482 with my suggestions. If it LGTY we can merge it into this PR.

bazzadp: OK accepted that and then updated the comments to explain why we have two functions the same.

Also did a slight change:

```python
    return '%s' % (self._lang_code)
```

to

```python
    return self._lang_code
```

Merge away if happy!

PR #476, opened 2019-11-12
fix: typo on JavaScript chapter

ftonato:


bazzadp:
Hi @ftonato - thanks for these. Can you run `npm run generate` in the src directory to also generate the equivalent HTML?

rviscomi:
No worries I'll generate the markup

PR #477, opened 2019-11-12
fix: markdown typo on Third Parties chapter

ftonato:


bazzadp:
Hi @ftonato - thanks for these. Can you run `npm run generate` in the src directory to also generate the equivalent HTML?

PR #484, opened 2019-11-13
Lazy loading images iframes

bazzadp:
Fixes https://github.com/HTTPArchive/almanac.httparchive.org/issues/351.

Guys, I've had another go at this now the dust has slightly settled. I've used JSDOM as that regex thing we tried before just wasn't working.

I've also added some height and width attributes as part of this as advised by Dev Tools console.

Looks to be working well now.

Let me know your thoughts.


src/templates/en/2019/chapter.html
```diff
@@ -115,7 +115,7 @@ <h1 class="title">
                 <span class="chapter-unedited">[Unedited]</span>
                 {% endif %}
             </h1>
-            <img src="/static/images/{{ year }}/{{ get_chapter_image_dir(metadata) }}/hero_lg.jpg" class="content-banner" alt="">
+            <img src="/static/images/{{ year }}/{{ get_chapter_image_dir(metadata) }}/hero_lg.jpg" class="content-banner" alt="" loading ="eager" />
```
rviscomi:
```suggestion
            <img src="/static/images/{{ year }}/{{ get_chapter_image_dir(metadata) }}/hero_lg.jpg" class="content-banner" alt="" loading="eager" />
```

Issue #413, opened 2019-11-10
Audit the website for performance issues


rviscomi:
Related: #286 and #303. Make sure we're not doing anything silly to hurt performance.

I'm aware that embedding 25 Sheets iframes is probably not a great idea for performance. We can look into preconnecting to the origin and lazy loading the iframes.

The audit should ideally happen before launch just so we know if we're doing anything egregious. Any P0 performance issues should be fixed prior to launch (within 24 hours). Everything else can be resolved after launch (apres ski).

cc performance experts @andydavies @paulcalvano 

---

### Performance issues

- [x] **P1** home page header background image https://github.com/HTTPArchive/almanac.httparchive.org/issues/413#issuecomment-552175467
- [x] **P1** load appropriate data viz https://github.com/HTTPArchive/almanac.httparchive.org/issues/413#issuecomment-552176411
- [x] **P2** lazy load footer and mobile images https://github.com/HTTPArchive/almanac.httparchive.org/issues/413#issuecomment-552175796
- [ ] **P2** set a long TTL for static images https://github.com/HTTPArchive/almanac.httparchive.org/issues/413#issuecomment-552176079
- [x] **P2** lazy load data viz iframes/images and author avatars https://github.com/HTTPArchive/almanac.httparchive.org/issues/413#issuecomment-552176643

bazzadp:
> I'm aware that embedding 25 Sheets iframes is probably not a great idea for performance. We can look into preconnecting to the origin and lazy loading the iframes.

Just noting that lazy loading images (and possibly iframes) is being discussed in https://github.com/HTTPArchive/almanac.httparchive.org/issues/351

rviscomi:
Yes I definitely think #351 would be helpful.

This issue can tell us how significant the perf impact of not lazy loading is.

bazzadp:
Any ideas why Lato font loads so many times?:

![image](https://user-images.githubusercontent.com/10931297/68541381-89cb4800-0396-11ea-884a-fa9087173bc8.png)

rviscomi:
In base.html we load the fonts here: `<link href="https://fonts.googleapis.com/css?family=Lato:400,400i,700,700i, 900|Poppins:300,400,700,900&display=swap" rel="stylesheet">`.

I think it's loading one font file for each weight/style combo.

rviscomi:
[...]

PR #485, opened 2019-11-13
Lazy loading the background image

mikegeyser:
This is a part of #413 

Setting the background header of the home page to a solid colour, and then lazily loading the background image using javascript and swapping it out. This was the only way I could find to do this reliably, given the background image being set on the body of the page and falling across the first two sections. The next step would be to remake the background using SVG, so that it's also lighter across the wire.

What do you think? If we don't want to do it this way, I don't mind. I did test it on 'slow 3g' and it works ok.

### DIFF
--- a/src/static/css/index.css
+++ b/src/static/css/index.css
@@ -1,8 +1,14 @@
 body {
-  background-image: url(/static/images/intro-background-fit.png);
   background-repeat: no-repeat;
   background-size: 100% 910px;
 }
+
+body.lazy {
+  /* This is a 1px solid #677486 image. */
+  background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNML2n7DwAFDQJi8nQACgAAAABJRU5ErkJggg==);
+  background-size: 100% 700px;
+}
+
 header.alt-bg {
   background-color: initial;
 }
--- a/src/templates/en/2019/index.html
+++ b/src/templates/en/2019/index.html
@@ -97,7 +97,7 @@
       reader.onload = function() {
         var body = document.getElementsByTagName('body')[0];
         body.setAttribute('style', 'background-image: url(' + reader.result + ');');
-        body.removeAttribute('class');
+        body.classList.remove('lazy');
       };
 
       reader.readAsDataURL(blob);
--- a/src/templates/en/2019/index.html
+++ b/src/templates/en/2019/index.html
@@ -107,13 +107,6 @@
       fetch(url)
         .then(function(response) { return response.blob(); })
         .then(function(blob) { setImage(blob); });
-    } else {
-      // Internet explorer.. :(
-      var request = new XMLHttpRequest();
-      request.open('GET', url);
-      request.responseType = 'blob';
-      request.onload = function () { setImage(request.response); };
-      request.send();
     }
   </script>
 {% endblock %}
--- a/src/templates/en/2019/index.html
+++ b/src/templates/en/2019/index.html
@@ -89,6 +89,35 @@
   <link rel="stylesheet" href="/static/css/index.css">
 {% endblock %}
 
+{% block scripts %}
+  <script defer="defer" nonce="{{ csp_nonce() }}">
+    var url = '/static/images/intro-background-fit.png';
+    var setImage = function(blob) {
+      var reader = new FileReader;
+      reader.onload = function() {
+        var body = document.getElementsByTagName('body')[0];
+        body.setAttribute('style', 'background-image: url(' + reader.result + ');');
+        body.removeAttribute('class');
+      };
+
+      reader.readAsDataURL(blob);
+    };
+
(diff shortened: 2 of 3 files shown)


src/static/css/index.css
```diff
@@ -1,8 +1,14 @@
 body {
-  background-image: url(/static/images/intro-background-fit.png);
   background-repeat: no-repeat;
   background-size: 100% 910px;
 }
+
+body.lazy {
+  /* This is a 1px solid #677486 image. */
+  background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNML2n7DwAFDQJi8nQACgAAAABJRU5ErkJggg==);
```
mikegeyser: This is not `background-color` because then it ignores the `background-size` attribute, and stretches the full length of the page and looks bad.

src/templates/en/2019/index.html
```diff
@@ -89,6 +89,35 @@
...
+      reader.onload = function() {
+        var body = document.getElementsByTagName('body')[0];
+        body.setAttribute('style', 'background-image: url(' + reader.result + ');');
+        body.removeAttribute('class');
+      };
+
+      reader.readAsDataURL(blob);
+    };
+
+    if (window.fetch) {
```
mikegeyser: We could also just not do the `fetch` part, because XHR works everywhere. It just made me sad to have to write it. :P

rviscomi:
How compatible is this PR with @catalinred's [SVG approach](https://github.com/HTTPArchive/almanac.httparchive.org/issues/413#issuecomment-553371144)? Would we still want to explore that option or does this solve the underlying issue?

mikegeyser:
I rambled a bit on #413 about it, but we can do a bit of both. I don't know if we should embed the SVG data URI inside the index.css file, there are some tradeoffs. Having the file as SVG would be great, but I also think having it separately cacheable is valuable. I dunno, what do you think?

catalinred:
@mikegeyser @rviscomi Regarding https://github.com/HTTPArchive/almanac.httparchive.org/issues/413#issuecomment-553627869 I double-checked and the gzip is already enabled on https://almanac.httparchive.org/ and that's good news, as base64 compresses very well under gzip, so CSS file size shouldn't be a problem because of base64 encoding.

On the solution above, in my opinion, JavaScript isn't the best and recommended approach. I believe we should focus instead on providing a good fallback using CSS background-color plus pseudo selectors if needed (for the bottom skewed part maybe) or just encode that `intro-background-fit` SVG as suggested earlier https://github.com/HTTPArchive/almanac.httparchive.org/issues/413#issuecomment-553371144

Hope this helps!

P.S. Where can I get that vector image `intro-background-fit` from? I'd love to give it a try tomorrow if you both agree.

rviscomi:
@mikegeyser thanks for clarifying. I'm ok going forward with this approach and as we investigate SVG we can continue to iterate.

@catalinred we don't have a vector version of that image AFAIK. I was given a PSD from our designer so it might be possible to generate SVG from that, but I'm not familiar with it. Would it be helpful if I sent you the PSD?

catalinred:
> @mikegeyser thanks for clarifying. I'm ok going forward with this approach and as we investigate SVG we can continue to iterate.
> 
> @catalinred we don't have a vector version of that image AFAIK. I was given a PSD from our designer so it might be possible to generate SVG from that, but I'm not familiar with it. Would it be helpful if I sent you the PSD?

PSD would do it, thanks.  I'll dig through it and see how the layers are composed and will figure it out from there.

rviscomi:
@catalinred https://drive.google.com/open?id=13vQ82LYw5eNx4gb0y7AdXntnFpIsxzKS

src/templates/en/2019/index.html
```diff
@@ -89,6 +89,35 @@
...
 
+{% block scripts %}
+  <script defer="defer" nonce="{{ csp_nonce() }}">
+    var url = '/static/images/intro-background-fit.png';
+    var setImage = function(blob) {
+      var reader = new FileReader;
+      reader.onload = function() {
+        var body = document.getElementsByTagName('body')[0];
+        body.setAttribute('style', 'background-image: url(' + reader.result + ');');
+        body.removeAttribute('class');
```
rviscomi:
```suggestion
        body.classList.remove('lazy');
```

src/templates/en/2019/index.html
rviscomi: We use `fetch` without fallbacks in other places. Not a great idea, but there's precedent if you wanted to ignore IE.

mikegeyser: I don't know, it's up to you. It's a non-critical part of the application, so IE users would just see a plain grey background. Nothing would 'break' per-se.

rviscomi: Let's remove it :)

image

IE is .3% of traffic and deprecated.

mikegeyser: image

PR #489, opened 2019-11-14
Update css.md

apalm:
Split into two commits so that the diff for the typo fixes (second commit) is less noisy.


rviscomi:
Thanks for these fixes @apalm!

Issue #432, opened 2019-11-10
Fallback Data Vis images don’t display on tablet


bazzadp:
On iPad we don’t see the Google Sheets images (presumably because of same reason as iPhone: https://github.com/HTTPArchive/almanac.httparchive.org/issues/369) but the fall back images do not show either whereas they do on iPhone.

CSS need a little tweaking to change the breakpoint?

rviscomi:
IIUC, this is working as intended. We need images to appear on small viewports because the 600px interactive charts would otherwise overflow. On a tablet with a wider screen, the wide interactive charts still fit. WDYT?

bazzadp:
Except that iPadOS at least doesn't render them. Not sure about Android.

rviscomi:
Because of this issue "Canvas area exceeds the maximum limit (width * height > 16777216)"?

Really surprised.

bazzadp:
> Because of this issue "Canvas area exceeds the maximum limit (width * height > 16777216)"?

Yup confirmed by hooking iPad up to Mac and using Web Inspector. Same error.

> Is this an issue for all charts or only very large ones? The default dimensions are `width="600" height="371"`for an area of 222,600. So I'm confused why those would be blocked.

All charts. Even my simple Pie Chart in [HTTP/2](https://almanac.httparchive.org/en/2019/http2#http2-push) doesn't render. Your Performance chapter is looking pretty bare...

rviscomi:
Ok in that case this is a launch blocker. Super strange. The canvas size limit is 75x larger than the area for most of these charts so I have no idea why they'd be blocked.

If we can't figure this out we can `display: none` all iframes and show the static images to everyone.

bazzadp:
> If we can't figure this out we can `display: none` all iframes and show the static images to everyone.

That would be a shame as they are nicer. However maybe it's better to do that until we fix https://github.com/HTTPArchive/almanac.httparchive.org/issues/428 anyway?

rviscomi:
Oof this is a really tough call. I'll think about it. Open to other suggestions.

bkardell:
Android, chrome, Moto g7, I can not see any of them, fallback even on the markup chapter. The only ones that show up are the static images, most are missing

bazzadp:
> Android, chrome, Moto g7, I can not see any of them, fallback even on the markup chapter. The only ones that show up are the static images, most are missing

[...]

Issue #428, opened 2019-11-10
Fallback when Google Sheets is not available


bazzadp:
Some companies (including my employer!) block Google Sheets for various reasons. Not sure if some countries do this similarly.

As we have fall back graphics for mobile, it would be nice if we could run a bit of JavaScript on a page to check for that blocking, and then hide the iframes and unhide the images.

rviscomi:
Yeah if the iframes take too long to load for any reason (blocked, latency, outage) we should fall back to the static images.

bazzadp:
BTW I think this should be in [SHIP-IT] milestone (https://github.com/HTTPArchive/almanac.httparchive.org/milestone/5) and not [Apres-Ski](https://github.com/HTTPArchive/almanac.httparchive.org/milestone/6). A non-trivial number of users will not be getting visuals. Hence why I've prioritised this.

Issue #379, opened 2019-11-08
Images missing alt text


AymenLoukil:
https://almanac.httparchive.org/static/images/character-markup.png
https://almanac.httparchive.org/static/images/methodology-characters.png
https://almanac.httparchive.org/static/images/character-star.png
https://almanac.httparchive.org/static/images/character-hat.png
https://almanac.httparchive.org/static/images/menu.png
https://almanac.httparchive.org/static/images/twitter.png
https://almanac.httparchive.org/static/images/methodology-banner.png
https://almanac.httparchive.org/static/images/character-file.png
https://almanac.httparchive.org/static/images/character-painter.png
https://almanac.httparchive.org/static/images/avatars/1.jpg
https://almanac.httparchive.org/static/images/avatars/8.jpg
https://almanac.httparchive.org/static/images/avatars/4.jpg
https://almanac.httparchive.org/static/images/avatars/9.jpg
https://almanac.httparchive.org/static/images/avatars/0.jpg
https://almanac.httparchive.org/static/images/avatars/11.jpg
https://almanac.httparchive.org/static/images/2019/05_Third_Parties/hero_lg.jpg
https://almanac.httparchive.org/static/images/2019/03_Markup/hero_lg.jpg
https://almanac.httparchive.org/static/images/2019/02_CSS/hero_lg.jpg
https://almanac.httparchive.org/static/images/2019/14_CMS/hero_lg.jpg
https://almanac.httparchive.org/static/images/2019/13_Ecommerce/hero_lg.jpg
https://almanac.httparchive.org/static/images/2019/19_Resource_Hints/hero_lg.jpg
https://almanac.httparchive.org/static/images/2019/09_Accessibility/hero_lg.jpg
https://almanac.httparchive.org/static/images/2019/07_Performance/hero_lg.jpg
https://almanac.httparchive.org/static/images/2019/07_Performance/fig7.png
https://almanac.httparchive.org/static/images/2019/07_Performance/fig11.png
https://almanac.httparchive.org/static/images/2019/07_Performance/fig2.png
https://almanac.httparchive.org/static/images/2019/07_Performance/fig10.png
https://almanac.httparchive.org/static/images/2019/07_Performance/fig12.png
https://almanac.httparchive.org/static/images/2019/07_Performance/fig16.png
https://almanac.httparchive.org/static/images/2019/07_Performance/fig1.png
https://almanac.httparchive.org/static/images/2019/07_Performance/fig6.png
[...]

Issue #303, opened 2019-11-02
Audit the website for accessibility issues


rviscomi:
Related to #286, we should do an audit of the website for accessibility issues before launch.

cc @obto @kleinab	@rachellcostello @ljme

### Outstanding issues:

- [x] **P0**: (Chapter) Graphs don't have descriptive alt text nor a link to the data, so screen reader users don't have access to that information
- [x] **P1**: (All pages) There are many images of cartoons that need to be decided to be either purely decorative (and get alt="") or have meaning and need alt text
- [x] **P1**: (Chapter) `div.big-number` has insufficient contrast
- [x] **P2**: (Contributors) The entire square that contains the images/links should not be in the tab order since there is no action that happens upon invoking them - this makes navigation hard and it's too verbose for screen reader users 
- [ ] **P2**: (Home) 
White text on gray background "Table of Contents" 
Foreground:#F2F2F2
Background:#778192
The contrast ratio is: 3.5:1
Expected contrast ratio: 4.5:1 or greater 

Yellow text in buttons on gray background such as "Start exploring" 
Foreground:#F7F779
Background:#697587
The contrast ratio is: 4.1:1
Expected contrast ratio: 4.5:1 or greater   

Twitter and Github logo in the lower right corner
Foreground:#8E99A3
Background:#5C687D
The contrast ratio is: 1.9:1
Expected contrast ratio: 4.5:1 or greater   

cc @HTTPArchive/developers can anyone help resolve these a11y issues?

rviscomi:
From an email thread it sounds like @ljme may be able to start on an accessibility audit. We'll prioritize the major findings to get them resolved before launch.

ljme:
Yes! Let's do this!

ljme:
I've found some bugs and will take time tomorrow to write them up individually with repro steps and check them all to ensure they are cross-platform. Where is a good spot to put them? 

The most pressing issues I see tonight are:

Inside the report page: 
**graphs don't have descriptive alt text nor a link to the data, so screen reader users don't have access to that information - most important bug** 
some text in the graphs has very very low contrast - I can calculate ratios tomorrow
tables don't have table headers marked up which makes them hard to navigate for screen reader users

Everywhere: 
**there are many images of cartoons that need to be decided to be either purely decorative (and get alt="") or have meaning and need alt text - second most important bug** 
The focus indication ring can be hard to see against some backgrounds I can calculate ratios tomorrow

Contributors page:
[...]

PR #491, opened 2019-11-14
Detect whether Google Sheets will work

bazzadp:
Makes progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/432
Makes progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/428
Makes progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/413#issuecomment-552176411
Makes progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/379

I present this PR for your consideration and for discussion.

Basically what I did was:
- Remove the `iframe`s. The `img` tags load by default and will be progressively enhanced with Sheets, which are added later with JavaScript if supported.
- Add `data-` elements to the `img` tag with the information needed to recreate the `iframe`.
- Added a `js/chapter.js` script which a couple of checks and then enables Sheets visuals if they pass. Quite a bit of JavaScript so decided to pull it out into a separate sheet rather than inline it. Edit: also moved HTTP Archive discussion counter check into that too.
- This has added benefit that those that can't use them, don't wastefully load Sheets as discussed on https://github.com/HTTPArchive/almanac.httparchive.org/issues/413#issuecomment-552176411
- Desktop _will_ still load `img`'s and then `iframe`'s but I can live with that for now.
- Add a test pixel for Google Docs favicon with an error function to revert the Sheets back to Images when Google Docs is blocked (e.g. in some corporate environments).
- Added an `alt` tag to `img`s, and a `title` tags to `iframe`s and changed `aria-labelledby` to point to figure (to avoid duplicating alt text to assistive technologies) and also added an `aria-describedby` to point to more descriptive text, as discussed in https://github.com/HTTPArchive/almanac.httparchive.org/issues/379#issuecomment-554036966

The HTTP/2 and Accessibility chapters are the only ones on new format as part of this PR. Will convert others with a follow up PR if this is approved. It's backwards compatible but thought I'd get your thoughts before doing the other chapters in a similar manner.

This seems to work on iPads now AND in my work which blocks Google Sheets.


src/static/js/chapter.js
```diff
@@ -0,0 +1,127 @@
...
+
+  return largeCanvasSupported;
+
+};
+
+//Google Sheets is sometimes blocked by companies/proxies so let's check for that
+//Unfortunately no easy way to check this as cannot make non-CORS request to
+//Another domain and can't check status response when CORS domain :-(
+//Below seems to work for my corporate proxy anyway
+const googleSheetsAllowed = async () => {
```
rviscomi: What's the browser support for async/await?

src/static/js/chapter.js
```diff
@@ -0,0 +1,127 @@
...
+          iframe.setAttribute('title', fig_img.getAttribute('alt'));
+          iframe.setAttribute('width', "600");
+          iframe.setAttribute('height', '371');
+          iframe.setAttribute('seamless', '');
+          iframe.setAttribute('frameborder', '0');
+          iframe.setAttribute('scrolling', 'no');
+          iframe.setAttribute('loading', 'lazy');
+
+          //Copy the atttributes from the data- attributes of the img
+          for (var i = 0, atts = fig_img.attributes, n = atts.length, arr = []; i < n; i++) {
```
rviscomi: `arr` seems to be unused

src/static/js/chapter.js
```diff
@@ -0,0 +1,127 @@
...
+            const att_name = atts[i].nodeName;
+            if (att_name.substring(0, 5) == 'data-') {
```
rviscomi: Can't we use the dataset API directly? There are only a few attributes we would use so we can look them up individually (in a loop) as needed.

src/templates/en/2019/base_chapter.html
```diff
@@ -19,20 +19,25 @@
 });
 
 fetch(`https://discuss.httparchive.org/t/{{ metadata.get('discuss') }}.json`)
```
rviscomi: Yeah as long as we're creating a chapter-specific script file, let's keep all chapter scripts together.

src/static/js/chapter.js
```diff
@@ -0,0 +1,127 @@
+//Data Save can be set to on, so let's check it
+function dataSaverEnabled() {
+  let dataSaver = false;
+  if ('connection' in navigator) {
+    dataSaver = navigator.connection.saveData;
+    if (dataSaver) {
+      console.log('DataSaver is enabled');
```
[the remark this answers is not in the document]
bazzadp: Good idea. Will have a look.

src/static/js/chapter.js
bazzadp: [Pretty good](https://caniuse.com/#feat=async-functions) and I tested this in IE11 (which doesn't support it) and it doesn't log any errors, but just ignores it. Which is fine as they will stay with images (which are default).

src/static/js/chapter.js
bazzadp: TIL! Never heard of that before. It is cleaner - though does mean we've a hardcoded list but, as you say, shouldn't change.

src/static/js/chapter.js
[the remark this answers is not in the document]
bazzadp: Since GA is loaded via GTM you OK if I just push an event via GTM data layer? Then will need to trigger a GA call from GTM based on this trigger.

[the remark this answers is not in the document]
bazzadp: I've commited that. It's better as queues the events. If you want to give me GTM access will set up the triggers to push them through to GA events?

src/static/js/chapter.js
bazzadp: @mikegeyser have you any thoughts on this? Am I right to say that this will just be ignored by incompatible browsers? Or is there a better way to write this?

src/static/js/chapter.js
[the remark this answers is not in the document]
rviscomi: I think you can call `gtag` directly:

```js
gtag('event', 'user', {
  'event_category' : 'data-saver',
  'event_label' : 'enabled'
});
```

https://developers.google.com/analytics/devguides/collection/gtagjs/events

src/static/js/chapter.js
rviscomi: AFAIK it will cause syntax errors and break parsing the file. So any other scripts we put in this file would fail to execute. I'm ok with ES6 but worried that someone someday might put something critical in this file and not realize that it would never work on browsers that don't support newer syntax. Maybe we just rename the file to be data viz specific?

I think we should also remove the discussion comments from this file for that reason. Sorry I think I asked you to add it.

src/static/js/chapter.js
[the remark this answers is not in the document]
bazzadp: Lovely. Tested and works a treat!

Going with the following - let me know if you want any changes

|Action|Category|Label|Value|
|---|---|---|---|
|'data-saver'|'user'|'Enabled'|1|
|'data-saver'|'user'|'Not Enabled'|0|
|'Hi-res Canvas'|'user'|'Supported'|1|
|'Hi-res Canvas'|'user'|'Not Supported'|0|
|'Sheets Access'|'user'|'Successful'|1|
|'Sheets Access'|'user'|'Blocked'|0|
|'Interactive Figures'|'user'|'Enabled'|1|
|'Interactive Figures'|'user'|'Not Enabled'|0|

src/static/js/chapter.js
bazzadp: OK you've persuaded me. I've removed async/await call instead and therefore left the discussion comments in this file.

Could you give it another once over?

bazzadp:
BTW once we're happy it would be good to ship this to prod, so can test the HTTP/2 chapter again in live before applying the similar changes to the other chapters.

src/static/js/chapter.js
```diff
@@ -0,0 +1,154 @@
...
+
+      let sheetsAllowed = true;
+
+      const sheets_url = 'https://docs.google.com/spreadsheets/';
+
+      //Google Sheets is sometimes blocked by companies/proxies so let's check for that
+      //Unfortunately no easy way to check this as cannot make non-CORS request to
+      //Another domain and can't check status response when CORS domain :-(
+      //Below seems to work for my corporate proxy anyway
+      fetch(window.sheets_url, { method: 'HEAD', mode: 'no-cors' })
```
rviscomi: I think this is causing an error in the console:

`HEAD http://127.0.0.1:8080/en/2019/undefined net::ERR_ABORTED 404 (NOT FOUND)`

Why would it appear as `/undefined`? And is there any way to prevent this from causing console errors?

src/static/js/chapter.js
```diff
@@ -0,0 +1,154 @@
...
+                  //Add the fig-mobile class to hide the img in desktop view
+                  parentLink.classList.add("fig-mobile");
+                }
+
+              }
+            });
+            gtag('event', 'interactive-figures', { 'event_category': 'user', 'event_label': 'enabled', 'value': 1 });
+          }
+        }).catch(function (err) {
+          console.error('Error' + e);
```
```suggestion
          console.error('Error' + err);
```

src/static/js/chapter.js
bazzadp: Typo, but it doesn't matter anyway. 😔See below.

bazzadp:
OK retesting on the work computer since I did the change shows this isn't going to work for companies that block access to sheets 😞

Earlier versions of this change I got working at work checked the call failed, but when checking the same code outside of work I had to tweak it to get Sheets to work when you did have access, and that then broke it for the block use case. As far as I can see there is simply no way for the browser to see whether a URL on a third-party domain is successful or not - probably for good security reasons.

I still think this is a good change to sort out iPads, and to save data for those on mobile or with `data-saver` enabled, albeit not as comprehensive as I thought. Plus it fixes some of the other SEO/Accessibility/Alt-tag things.

I've updated the initial description. Can you look over it again?

src/static/js/chapter.js
```diff
@@ -0,0 +1,134 @@
+//Data Save can be set to on, so let's check it
+function dataSaverEnabled() {
+  let dataSaver = false;
```
rviscomi: FWIW let and cost have the same syntax issues as async/await. I think we should be entirely ES6 or not at all.

bazzadp: OK fixed that nite... but also came up with the solution to my problem!!

bazzadp:
OK I found another way around it. See this commit: https://github.com/HTTPArchive/almanac.httparchive.org/pull/491/commits/3024b8ca1e05b401cf897bec80df62c962a9c529

So back in action for the full fix!

I also added one more chapter (Accessibility) with multiple images just to check my for loops.

Let me know your thoughts.

bazzadp:
@rviscomi are you happy to progress this now and publish those two pages to production so can do test for real on production site? Or is there anything holding this up (other than your own time not being unlimited! 😊). I asked for some a11y help at https://github.com/HTTPArchive/almanac.httparchive.org/issues/303#issuecomment-554786160 as discussed but not sure this should hold that up as can always iterate that later.

@mikegeyser if you're back off your travels and have any time, it would be good to have your opinion on this?

src/static/js/chapter.js
```diff
@@ -0,0 +1,189 @@
...
+
+  var largeCanvasSupported = false;
+
+  try {
+    // Set large sized canvas dimensions and draw test rectangle
+    var cvs = document ? document.createElement('canvas') : null;
+    var ctx = cvs && cvs.getContext ? cvs.getContext('2d') : null;
+    cvs.width = 6000;
+    cvs.height = 3700;
+    ctx.fillRect(5999, 3699, 1, 1);
```
mikegeyser: I think we may need to test this on low- or mid-range devices? It may be very processor intensive. I don't know, I'm just nervous.

src/static/js/chapter.js
```diff
@@ -0,0 +1,189 @@
...
+  gtag('event', 'interactive-figures', { 'event_category': 'user', 'event_label': 'not-enabled', 'value': 1 });
+
+}
+
+//We use Google Sheets for detailed visualisations
+//Check for support and switch out images if supported
+function upgradeInteractiveFigures() {
+
+  try {
+    if (bigEnoughForInteractiveFigures() && highResolutionCanvasSupported() && !dataSaverEnabled()) {
```
mikegeyser: Maybe do the `dataSaverEnabled()` check first, so that the condition can short-circuit quickly of it is enabled? If someone has dataSaver enabled, chances are that they may not be running on the best hardware, and the rest of the checks will likely return false. (This is pure conjecture.)

src/static/js/chapter.js
```diff
@@ -0,0 +1,189 @@
...
+  try {
+    if (window.discussion_url) {
+      fetch(window.discussion_url)
+        .then(function (response) { return response.json(); })
+        .then(function (response) {
+          if (!response) {
+            return;
+          }
+
+          var comments = + response.posts_count - 1;
```
mikegeyser: ~Is the `= +` deliberate?~ Oh, I see it was inherited from the previous code.

src/static/js/chapter.js
bazzadp: Great idea - done!

src/static/js/chapter.js
bazzadp: It's a fair point. Especially as Google Sheets appears to do some device detection and in Chrome only returns 1200 by 742 canvas sheets even on large screens, as opposed to iOS when it does this 6000 by 3700 sheet. Wish they handled this better for iOS as the canvas they return for it is not handled by it!

Anyway, I changed it to this:

```javascript
// Set large sized canvas dimensions and draw test rectangle
var cvs = document ? document.createElement('canvas') : null;
var ctx = cvs && cvs.getContext ? cvs.getContext('2d') : null;
var scale = window.devicePixelRatio || 1;
if (scale > 1) {
  cvs.width = 6000;
  cvs.height = 3700;
  ctx.fillRect(5999, 3699, 1, 1);
} else {
  cvs.width = 1200;
  cvs.height = 742;
  ctx.fillRect(1199, 741, 1, 1);
}
largeCanvasSupported = Boolean(ctx.getImageData(5999, 3699, 1, 1).data[3]) == 1;
```

And confirmed that still fixed it in iOS (which has a device pixel ratio of 2).

With that, and your `data-saver` re-order suggestion below I think that should address this concern. Plus I think overall we will be returning a lot less data for low end devices thanks to this change as we won't load sheets by default and will only do above test once per page at the start.

You agree?

src/static/js/chapter.js
bazzadp: Yup not mine. @rviscomi not sure if there is a reason for this? Resolving it for this PR anyway.

src/static/js/chapter.js
mikegeyser: I do agree. As long as we test it. I have a Nokia 2 at work (for testing purposes), I can try and test it once this goes live?

bazzadp: Good stuff. Any ideas what the low end devices are on Webpagetest? Will run a before and after on there too.

bazzadp: Moto G1 has been suggested on the webperformance slack. Pages don't seem to render on that. Maybe that'll change after this PR :-)

rviscomi:
Let me stew in this for at most a couple more days. These are big changes and I want to make sure I think through the implications.

bazzadp:
> Let me stew in this for at most a couple more days. These are big changes and I want to make sure I think through the implications.

Fair enough and don't want to rush you if we do think there is any risk. Just with the translators doing more pages, plus the occasional chapter edits we see, we're just giving more clean-up work for ourselves the longer we leave it to make a call on this. Plus, the fact it is actually broken on iPads right now (and also in some corporate environments like my own - but guess that's our own fault!). And finally I'm out later in the week at PerfNow (though will bring my laptop then so maybe can discuss there as think you're there too!).

Still, those are not good enough reasons to push something in until you have comfort in it. Let me know if there's anything more I can do to give you that comfort.

bazzadp:
@rviscomi / @mikegeyser I made one small, extra change to further de-risk this.

There is no point in calling the new chapter.js file for all the other chapters when we are only moving Accessibility and HTTP/2 chapters to the new format as part of this, for now.

Therefore I've temporarily moved that script tag out of base_chapter.hmtl and into the markdown for both those chapters (and reverted the discussion count URL back to being inline temporarily).

This means if we merge this, then only those two chapters will be affected. We can then check it is all working in production correctly, and monitor Google Analytics to see the events being fired and then make a decision whether to do this for the rest of the chapters as part of https://github.com/HTTPArchive/almanac.httparchive.org/pull/511 (including moving the script tag call back to base_chapter.html, along with the discussion count).

WDTY?

src/static/js/chapter.js
bazzadp: @mikegeyser if you could baseline production on your Nokia 2 that would be good to know so we have something to compare this too if we do go ahead with this. Try it on HTTP/2 chapter (only 1 Sheet chart), Accessibility (7 Sheets charts) and CSS (29 Sheets charts!!).

rviscomi:
@bazzadp I appreciate your work to minimize the risk of this issue. I'm comfortable moving this forward to a 1 or 2 chapter experiment. I'll merge and deploy the changes now.

bazzadp:
Thanks mate! Good timing as was just about to leave the office but will hang round to test the fix for our corproate proxy blocking Google Sheets first! 😊

rviscomi:
The change is live! 🚀

bazzadp:
And I can see visuals at work finally!! 😊

bazzadp:
It does try to load some of the charts, until the pixel check kicks in and sees Sheets is blocked.

This is not unexpected - my pixel deliberately code kicks in AFTER all the iframes have been added and then reverts back to images when it see it's blocked. This is because I didn't want to delay loading the Interactive visuals, waiting to see whether that pixel test would load correctly or not, as this would delay the majority of cases when Sheets is _not_ blocked just to the benefit of the minority where it _is_ blocked. Corporates that block can suffer the extra download attempts as they will be blocked anyway. And lazy loaded so will only attempt first few charts anyway.

For other cases (e.g. on mobile screen width, on data-saver, or with not enough Canvas support) the Sheets will not be loaded first so still a good network saving for those users. Will test that when I get home.

bazzadp:
Does exactly what's expected on iPads too. No Google Sheets loaded (except test pixel) and PNGs are shown. Confirmed other chapters still broken as expected.

Seeing some good hits on GA already:
- 63/91 or 69.23% of visitors do NOT enabled interactive visualisations. So we were needless loading them for those users.
- 52/83 or 62.65% of visitors have sheets blocked (some of them will be my tests - but not all of them).
- 21/91 or 23.08% people have data-saver enabled (far higher than I would have thought!).

Did spot one typo where I call the event `google-sheets` for fails and `sheets-access` for successes. Will consolidate that.

I'm happy this is working as expected so will submit the full PR and you can approve whenever you're happy.

rviscomi:
> 62.65% of visitors have sheets blocked

This doesn't sound right.

bazzadp:
I’d take it with a pinch of salt. GA can be slow to update and I ran a few tests in various browsers (maybe 20 max?) when blocked so invariably skewed the results. We’ll see over next few hours/days what the real stats are.

Wouldn’t be surprised if it was double digits though. Pesky corporates!

bazzadp:
> * 21/91 or 23.08% people have data-saver enabled (far higher than I would have thought!).

Ignore this. That's how many people reported a value for data-saver (does only Chrome report this?). They all reported the value `not-enabled` however to say data-saving was not in effect! 😬

Issue #470, opened 2019-11-11
Home page text contrast issue on mobile when zoomed


rviscomi:
Got this UX issue in a DM:

![image](https://user-images.githubusercontent.com/1120896/68620289-06872080-0482-11ea-9681-8bdf69f62c4a.png)

The user is on a mobile device with the page zoomed in.

cc #303

catalinred:
cc https://github.com/HTTPArchive/almanac.httparchive.org/issues/413#issuecomment-552175467 as well

Curious about how to reproduce the above zoomed in behavior.

rviscomi:
Yeah I wasn't able to reproduce this in Dev Tools.

catalinred:
> Yeah I wasn't able to reproduce this in Dev Tools.

Me neither, I tried on my iPhone as well. At first sight, I thought it might be an iOS text size setting but no.

rviscomi:
It seems the OS-level magnification settings need to be configured to 160% to trigger this bug.

![image](https://user-images.githubusercontent.com/1120896/68812874-fbbac000-0628-11ea-847f-b1699f9dce36.png)

catalinred:
In case it helps, I did try text scaling on iOS and wasn't able to reproduce this behavior, the above screenshot looks like is taken on an Android phone.

Either way, IMO this issue is due to the fact that the main background image `intro-background-fit.png` is set on the body and not on the HTML element that wraps the above text and button, e.g. `<section class="intro-container">`.  Related to #485

rviscomi:
Yeah, I think one of the reasons for doing it this way is because it also needs to appear behind the `<header>`.

mikegeyser:
They are definitely related, but even if it was set on the `intro-container` it would still cut off the text, wouldn't it? Because the header image is set to a fixed `background-height`.

catalinred:
@rviscomi yes, technically the current header content and intro-container should be wrapped together in HTML to apply that background somehow, therefore to avoid zooming issues.

@mikegeyser those ‘background-size’ hardcoded px unit values are indeed part of the issue. 

Will try to work on a solution tomorrow (I’m on EET timezone) and keep you updated.

rviscomi:
Related #512

PR #492, opened 2019-11-14
Optimize performance of home page header

catalinred:
So I managed to open a pull request on #470 and #413 

#470
- The `header` element now wraps both logo+navigation and `intro-container` on the homepage, thus zooming issues will be avoided as the background will scale along with the text.
- I removed all those ‘background-size’ hardcoded px unit values.

#413 P1 home page header background image
- With the .psd file, I exported the existing `intro-background-fit` as an SVG, cleaned it up with [SVGO](https://github.com/svg/svgo), then converted to base64 as discussed with @rviscomi and @mikegeyser 
- I had to remove the newly added #485 solution as it's not needed anymore @mikegeyser 

Please check out the filmstrip below, simulated on a 3G network. 

devtools

Let me know what you think!


src/templates/en/2019/base.html
```diff
@@ -48,7 +49,28 @@
           {{ language_switcher() }}
         </div>
       </nav>
-    </header>
+    </div>
+    {% if request.endpoint == 'home' %}
```
rviscomi: Rather than conditionally rendering content specific to the home page, could we overwrite the header block in the index/home template?

src/static/images/intro-background-fit.svg
```diff
@@ -0,0 +1 @@
+<svg id="BG" xmlns="http://www.w3.org/2000/svg" width="1920" height="1250"><defs><style>.cls-1{fill:#a8caba;fill-rule:evenodd;opacity:.05}</style></defs><path id="Rectangle_5_copy_4" data-name="Rectangle 5 copy 4" class="cls-1" d="M792 202h2088v651L792 1213V202z" transform="translate(-1023 -113)"/><path id="Rectangle_5_copy_5" data-name="Rectangle 5 copy 5" class="cls-1" d="M553 352h2088v651L553 1363V352z" transform="translate(-1023 -113)"/><path id="Rectangle_5_copy" data-name="Rectangle 5 copy" d="M879 94h2088v651L879 1105V94z" transform="translate(-1023 -113)" fill="#1a2b49" opacity=".7" fill-rule="evenodd"/><path class="cls-1" d="M553 30h2088v651L553 1041V30z" transform="translate(-1023 -113)"/><path id="Rectangle_5_copy_2" data-name="Rectangle 5 copy 2" d="M791 65h2088v651L791 1076V65z" transform="translate(-1023 -113)" fill="#f2f2f2" opacity=".05" fill-rule="evenodd"/><path id="Rectangle_5_copy_3" data-name="Rectangle 5 copy 3" d="M-403-258h2088v651L-403 753V-258z" transform="translate(-1023 -113)" opacity=".1" fill="#f2f2f2" fill-rule="evenodd"/></svg>
```
rviscomi: Nit: Could this be pretty printed? It'd be easier to spot ways to optimize it.

src/templates/en/2019/index.html
```diff
@@ -170,9 +130,9 @@ <h2 class="title title-alt">Contributors</h2>
     </div>
     <div class="people">
       <div class="people-number">{{ contributors }}</div>
-      <img id="character-markup" class="character" src="/static/images/character-markup.png" alt="" loading="lazy" />
-      <img id="character-star" class="character" src="/static/images/character-star.png" alt="" loading="lazy" />
-      <img id="character-hat" class="character" src="/static/images/character-hat.png" alt="" loading="lazy" />
+      <img id="character-markup" class="character" src="/static/images/character-markup.png" alt=""/>
```
rviscomi: Was the lazy loading removed on purpose?

catalinred: Due to rebase & conflicts, will add again. Sorry.

src/static/images/intro-background-fit.svg
catalinred: This is already optimized using SVGO https://jakearchibald.github.io/svgomg/ but sure, we could pretty-print it if necessary. We're using its base64 version anyway, but I've added in the images folder as a reference.

rviscomi: Yeah I'm just curious if we can remove some ineffective attributes like `data-name`, `id`, or `full-rule`.

src/templates/en/2019/base.html
catalinred: I guess we could, I must dig a bit more to find out how.

Saw ` if request.endpoint == 'home'` already used for the "Start exploring" button and I thought that is considered good practice in there :)

src/static/images/intro-background-fit.svg
catalinred: You were right, I was able to further optimize it. Thanks for the heads up!

src/templates/en/2019/base.html
catalinred: Tried it, apparently, we'd end up with a big chunk of duplicated markup wrapped by `<div class="top-header">`. Unless we isolate it. 

IMHO we could go with the current conditionally rendering for this. In the future, we will reorganize the header with an updated templating logic.

src/static/css/index.css
```diff
@@ -1,19 +1,19 @@
...
+  background-position: bottom left;  
   background-repeat: no-repeat;
-  background-size: 100% 910px;
-}
-
-body.lazy {
-  /* This is a 1px solid #677486 image. */
-  background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNML2n7DwAFDQJi8nQACgAAAABJRU5ErkJggg==);
-  background-size: 100% 700px;
+  background-size: 100% 100%;
```
rviscomi: I think this is causing the graphic to look distorted/stretched on mobile:

image

Live version for comparison:

image

Could you see if you can size the graphic similar to how it appears in production for mobile?

catalinred: I used `cover` for the background-size on displays lower than 600px. 

Also, as you may have noticed already, the `/static/images/home-hero.png` is now visually wrapped by the `intro-container` and from a design perspective that is good thing - having it look like it's standing on the edge of the skewed background.

Let me know what you think!

src/static/images/intro-background-fit.svg
```diff
@@ -0,0 +1,13 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="1920" height="1250">
+  <defs>
+    <style>
+      .cls-1{fill:#a8caba;fill-rule:evenodd;opacity:.05}
```
rviscomi: Curious why this is styled in CSS as opposed to attributes like the other elements? Can we style using elements for consistency?

rviscomi:
One small comment. This looks great!

src/static/images/intro-background-fit.svg
catalinred: If I'm not mistaken, this is an optimization made by the editing tools like Adobe Illustrator, or by SVGO due to the face that in this case `.cls-1` is used by three SVG `path`s and in terms of bytes, it is more efficient to have a CSS rule defined once and then apply it thrice. I didn't do the no of chars/bytes math though.

Either way, looks clean and DRY to me :)

mikegeyser:
I'm sorry I missed reviewing this PR because I was travelling. Just wanted to say that it's the right change, and far better than the solution it's replacing. Thanks for sticking with it @catalinred!

PR #504, opened 2019-11-17
Copyediting Media chapter

bazzadp:
Closes https://github.com/HTTPArchive/almanac.httparchive.org/issues/218.

Copy edits to bring chapter inline with other chapters. Mostly:

- Minor typos/style issues
- Enhanced aria descriptions
- Reordered figures (inc renaming fig files).

This should allow us to remove the `[unedited]` warning since @colinbendell already did the visualisations himself (excellent job btw!).

image


src/content/en/2019/media.md
```diff
@@ -8,316 +7,276 @@ authors: [colinbendell, dougsillars]
...
 
 <figure>
-  <iframe aria-labelledby="fig1-caption" width="600" height="371" seamless frameborder="0" scrolling="no" src="https://docs.google.com/spreadsheets/d/e/2PACX-1vSViHIntdF6-bHAI0cl1HelY_X8rR4lf0P3W2Y8I5SyVMxG-ptggTHfWA0qrrU47RvuAydLE6Zex6L3/pubchart?oid=1189524305&format=interactive"></iframe>
-  <a href="/static/images/2019/04_Media/fig1.png" class="fig-mobile">
-    <img src="/static/images/2019/04_Media/fig1.png" aria-labelledby="fig1-caption" width="600">
+  <iframe title="Figure 1. Web page bytes: image and video versus other." aria-labelledby="fig1-caption" aria-describedby="fig1-description" width="600" height="371" seamless frameborder="0" scrolling="no" src="https://docs.google.com/spreadsheets/d/e/2PACX-1vSViHIntdF6-bHAI0cl1HelY_X8rR4lf0P3W2Y8I5SyVMxG-ptggTHfWA0qrrU47RvuAydLE6Zex6L3/pubchart?oid=1189524305&format=interactive"></iframe>
```
rviscomi: Is `title` necessary?

Do we need both `labelledby` and `describedby`?

src/content/en/2019/media.md
```diff
@@ -8,316 +7,276 @@ authors: [colinbendell, dougsillars]
...
+  <a href="/static/images/2019/04_Media/fig1_bytes_images_and_video_versus_other.png" class="fig-mobile">
+    <img src="/static/images/2019/04_Media/fig1_bytes_images_and_video_versus_other.png" alt="" alt="Figure 1. Web page bytes: image and video versus other." aria-labelledby="fig1-caption" aria-describedby="fig1-description" width="600">
```
rviscomi: Ditto, we should only need one labelled/describedby.

There are also two alt attributes on this.

If we really need all of these redundant attributes it might be best to add them in the chapter generation script. That way we only write things once.

src/content/en/2019/media.md
bazzadp: It fails Firefox’s accessibility test without it. Title kind of acts like an alt tag for iframes. In my auto generation script as part of https://github.com/HTTPArchive/almanac.httparchive.org/pull/491 I set the title to the img alt text automatically.

Here was my thinking:

- alt/title - there for SEO/validation purposes. Can be different but currently just set to figcaption in most places.
- labelledby - there to link figcaption to prevent screen readers reading same text twice (once for alt, once for figcaption). Overrides alt/title is this takes priority for screen readers compared to those.
- describedby - deeper description of the image than just the alt/label/caption.

WDYT?

src/content/en/2019/media.md
bazzadp: Good spot - fixed.

See other comment in regards to the "duplication" and the fact that at least some of them are auto generated.

src/content/en/2019/media.md
rviscomi: It probably can't hurt to be _more_ accessible but I was just thinking that we should keep all data viz markup consistent and if/when we optimize it we can do everything at once. I'd also like to hear from one of the a11y experts whether these redundancies are helpful.

These changes are ok for now if you're sure they make the experience better. We should revisit this in #491 or in a follow-up issue.

rviscomi:
Sent you #506 with some copyediting suggestions.

src/content/en/2019/media.md
bazzadp: Not sure at all - like a lot of a11y it does seem there's a bit of guess work, but been reading up on it and do think this is the right way to use them. Would be nice if an a11y expert could give their input. Plus SOE and a11y requirements seem to clash sometimes (don't have an alt tag/complain alt-tag is missing) 🤨

Also I blame you for having such detailed descriptions in the chapters you edited - on top of fig captions! 😉

I already have a branch with the other chapters ready to go: https://github.com/HTTPArchive/almanac.httparchive.org/tree/figures_new_format/src/content/en/2019 but I want to push #491 to prod first and check it definitely fixes the issues it's supposed to on a few devices with the two example chapters, before doing the other chapters so will do a follow up PR after that testing is complete.

bazzadp:
OK merged. @dougsillars / @colinbendell feel free to raise separate PRs or issues if you disagree with any of these edits but keen to get the `[unedited]` flag removed so will drive on in the meantime!

PR #511, opened 2019-11-18
Migrate English chapter visualisations to new format

bazzadp:
Follow on from https://github.com/HTTPArchive/almanac.httparchive.org/pull/491 for the rest of the chapters.

Fixes https://github.com/HTTPArchive/almanac.httparchive.org/issues/432
Fixes https://github.com/HTTPArchive/almanac.httparchive.org/issues/428
Makes progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/413#issuecomment-552176411
Makes progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/379


bazzadp:
As discussed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/491#issuecomment-555715699 I think we're good to submit this for other chapters once you are.

bazzadp:
> Need to resolve a few merge conflicts. Let me know if this is ok to review after that.

Done!

bazzadp:
@rviscomi since we turned the JavaScript chapter on last night as a third chapter on the new format, we've already seen a bigger increase in GA data, which is shown below:

image

A few things to note:

- Low enough visitors who have sheets-access blocked (0.26%). This will be lower at weekends as suspect it'll be corporates normally blocking, but even so it's very low. It appears my company is an outlier!
- Low enough visitors who have canvas issues (0.47%). I checked and these are all iPads btw and iPads are approximately 2% of our visitors according to GA.
- However what we do see, which is more interesting for this change, is that the majority do not enable the interactive figures - and this is because they are on mobile. The new logic therefore be a data saver and performance boast for these users since currently the sheets are being loaded even with `display` style set to `none`. I'd be interested in seeing if the new report in Google Search Console leveraging CRuX data show any difference but that'll probably take a few days/weeks to feed through I presume?

Anyway, haven't seen or heard about any issues so I think we're ready to go live with the rest of the chapters if you are comfortable with this and get a chance to review them in this PR?

rviscomi:
Your GA analysis SGTM so let's ship it. I didn't go through each line in the PR, I trust everything works.

bazzadp:
Thanks! That's merged to master and production now. Going to spend a bit testing it all again in production branch to make sure nothing missed and then will push out shortly.

Issue #512, opened 2019-11-18
Home page broken on Safari


bazzadp:
Seeing this on my iPad and MacOS Safari:

![image](https://user-images.githubusercontent.com/10931297/69080686-5b9fd500-0a34-11ea-9ed9-8359dd4dc34c.png)

Note the blank background for the top of the header where the menu items are. Same (but even worse!) in Portrait:

![image](https://user-images.githubusercontent.com/10931297/69080823-93a71800-0a34-11ea-8a29-f7c4f7ff924c.png)

Same on Safari on MacOS:

![image](https://user-images.githubusercontent.com/10931297/69080977-ec76b080-0a34-11ea-90ae-8035eb8bea67.png)

Can't repeat in Chrome nor Firefox on MacOS so it's Safari specific.

I'm guessing maybe part of https://github.com/HTTPArchive/almanac.httparchive.org/pull/492 @catalinred ?

bazzadp:
https://github.com/HTTPArchive/almanac.httparchive.org/blob/87970a7a63fe7fb28ec2beb7b9f4b0bf413441b8/src/static/css/index.css#L1-L8

Changing `background-size` to `cover` instead of `100% 100%` seems to fix it, but does move the image about a bit.

rviscomi:
@bazzadp let's make the `cover` fix even if temporary. If you've tested it and it improves the experience, let's go for it. Great catch, thanks for reporting it.

bazzadp:
BTW on Chrome it looks like this:

![image](https://user-images.githubusercontent.com/10931297/69090382-c6efa400-0a40-11ea-898a-092cb015342c.png)

With my change it looks like this on Safari:

![image](https://user-images.githubusercontent.com/10931297/69090503-d66eed00-0a40-11ea-975b-13b584fc3666.png)

So not quite the same, but better than whats in there now!

catalinred:
@bazzadp Thanks for fixing it for now. I'll dig more to find out why Safari behaves like that given the `background-size` value and the current encoded SVG.

We did optimize the SVG and then encoded it so maybe Safari needs more info than SVGO stripped it out. Will make some tests later today and keep you up-to-date.

bazzadp:
Let’s reopen the issue while @catalinred investigates further.

PR #514, opened 2019-11-18
Style fix for Home Page top of page on Safari

bazzadp:
Fixes https://github.com/HTTPArchive/almanac.httparchive.org/issues/512


rviscomi:
Could you also run `npm run generate`? I know you didn't change any chapters in this PR but I did see https://github.com/HTTPArchive/almanac.httparchive.org/commit/87970a7a63fe7fb28ec2beb7b9f4b0bf413441b8 fly by.

bazzadp:
> Could you also run `npm run generate`? I know you didn't change any chapters in this PR but I did see [87970a7](https://github.com/HTTPArchive/almanac.httparchive.org/commit/87970a7a63fe7fb28ec2beb7b9f4b0bf413441b8) fly by.

Done! Also generates a weird Japanese performance chapter that's been annoying me for other commits. It's not right but might as well add it for now until we get round to fixing it.

catalinred:
@bazzadp Thanks for fixing it for now. I'll dig more to find out why Safari behaves like that given the background-size value and the current encoded SVG.

We did optimize the SVG and then encoded it so maybe Safari needs more info than SVGO stripped it out. Will make some tests later today and keep you up-to-date.

PR #515, opened 2019-11-18
Fix "Q" unit in CSS chapter

### LINKED ISSUE #513: No, Q is not the most popular physical CSS unit of length
Regarding https://almanac.httparchive.org/en/2019/css#units @zcorpan on Twitter asked:

> What is using the Q unit? Is it Japanese sites, or some common third-party CSS?

I started looking into this and I'm seeing many cases where Q was detected on stylesheets with base64-encoded data. I think these are false positives triggered by a weak regex I wrote for [02_07.sql](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/sql/2019/02_CSS/02_07.sql):

https://github.com/HTTPArchive/almanac.httparchive.org/blob/87970a7a63fe7fb28ec2beb7b9f4b0bf413441b8/sql/2019/02_CSS/02_07.sql#L6-L12

I reran the query with one small change: ignore any value longer than 20 characters. Most values should be small like `250px` or `1.6Q !important`. The ones that are longer than 20 characters are probably made up of mostly base64 crap. Just a heuristic but it seems to have made a big difference:
 
image

Q is much further down the list where we'd expect it. And the major units are mostly unchanged.

I'll coordinate with @argyleink and @una to rewrite this section with the revised data.

rviscomi:
Closes #513

I replaced the discussion of the Q unit with this aside:

> An earlier version of this chapter discussed the unexpected popularity of the `Q` unit. Thanks to the [community discussion](https://discuss.httparchive.org/t/chapter-2-css/1757/6) surrounding this chapter, we've identified that this was a bug in our analysis and have updated Figure 5 accordingly.

@una @argyleink let me know if this works for you.


una:
Happy to throw this in as a bandaid, or we can re-write that part of the chapter:

Current:
```
When it comes to units based on physical space, the q unit (1 quarter of a millimeter, or 1/40 of 1cm) is the most commonly used by far, with 22.46% on mobile and 15.13% on desktop. We knew about these types of units (e.q., mm, in, cm), which are specifically useful for print stylesheets, but didn't even know the q unit existed until this survey! Did you?
```

Update:
```
When it comes to units based on physical space, the `cm` (or centimeter) unit is the most popular by far, followed by `in` (inches), and then `Q`. We know these types of units are specifically useful for print stylesheets, but we didn't even know the q unit existed until this survey! Did you?

<aside>An [earlier version](<Link to earlier version>) of this chapter discussed the unexpected popularity of the <code>Q</code> unit. Thanks to the <a href="https://discuss.httparchive.org/t/chapter-2-css/1757/6">community discussion</a> surrounding this chapter, we've identified that this was a bug in our analysis and have updated Figure 5 accordingly.</aside>

```

rviscomi:
Yeah let's go with your suggestion.

As for linking to the earlier version, this will overwrite the chapter so we won't have a permalink except for the commit snapshot in GitHub.

PR #523, opened 2019-11-19
Use preserveAspectRatio="none" to avoid browsers inconsistencies

### LINKED ISSUE #512: Home page broken on Safari

catalinred:
Fixes #512 and #524

Use `preserveAspectRatio="none"` to avoid inconsistent behavior among browsers when it comes to stretching the header SVG background. Not really a bug, but more like inconsistent defaults among browsers implementations when one of the `viewBox`, `width & height` or `preserveAspectRatio` is not added.

That being said, I've also added `viewBox` and `width & height` for the background SVG, tested in lots of browsers and with the above added attributes I confirm the header background looks consistent across all the browsers I tested (macOS). 

browsers

### Small resolution (<= 600px)
sm

### Medium resolution (<=900px)
md

### Large resolution
lg


catalinred:
I’ll have a look at #524 later today and maybe we can wrap it in this PR too before merging? 

Will let you know as soon as I’ll have some findings on that.

Issue #530, opened 2019-11-21
Prevent dates flipping on commit


bazzadp:
The date stamps flip depending who commits and what time zone they are in. See https://github.com/HTTPArchive/almanac.httparchive.org/pull/529 for example.

We should change from using `toISOString` to `toUTCString` (see [this Stackoverflow answer](https://stackoverflow.com/questions/17545708/parse-date-without-timezone-javascript)) to prevent this.

PR #533, opened 2019-11-22
Set Javascript chapter to new format and fix dates

bazzadp:
Fixes https://github.com/HTTPArchive/almanac.httparchive.org/issues/530
Move Javascript to new date format as an extended trial as discussed with @rviscomi  before we move all chapters in https://github.com/HTTPArchive/almanac.httparchive.org/pull/511


rviscomi:
Feel free to deploy when ready!

bazzadp:
Deploy!

Note there appears to a problem with gcloud and had to downgrade: https://stackoverflow.com/questions/59000291/cant-deploy-to-app-engine-standard-with-gcloud-components-272-0-0.

Lets see how that goes and then upgrade rest of the chapters soon!

PR #537, opened 2019-11-24
Take Effective Connection Type into account when upgrading to interactive visuals

bazzadp:
Look at [Effective Connection Type](https://developer.mozilla.org/en-US/docs/Glossary/Effective_connection_type) and use that as part of the decision making as to whether to upgrade to Sheets interactive visuals or stay on images.

At the moment I say `slow-2g`, `2g` or `3g` is low bandwidth and everything else (including when ECT is not reported) is high bandwidth. That seem about right or should we include `3g` as high bandwidth too?

Also includes a small bug fix for IE11 that currently bails on the fetch call (even though fetch is not supported) and so fails to upgrade to Sheets visuals. This won't be needed once https://github.com/HTTPArchive/almanac.httparchive.org/pull/511 is merged but putting it in there for now anyway.

### DIFF
--- a/src/static/js/chapter.js
+++ b/src/static/js/chapter.js
@@ -28,7 +28,7 @@ function dataSaverEnabled() {
   return dataSaver;
 }
 
-//Check if networh API states this is a high bandwidth connection
+//Check if network API states this is a high bandwidth connection
 //Assume it is for those browsers who do not support this (e.g. Safari and IE)
 function highBandwidthConnection() {
   var highBandwidth = true;
@@ -217,4 +217,3 @@ function setDiscussionCount() {
 
 upgradeInteractiveFigures();
 //setDiscussionCount();
-
--- a/src/static/js/chapter.js
+++ b/src/static/js/chapter.js
@@ -21,9 +21,31 @@ function dataSaverEnabled() {
     } else {
       gtag('event', 'data-saver', { 'event_category': 'user', 'event_label': 'not-enabled', 'value': 0 });
     }
+  } else {
+    gtag('event', 'data-saver', { 'event_category': 'user', 'event_label': 'not-reported', 'value': 0 });
   }
 
   return dataSaver;
+}
+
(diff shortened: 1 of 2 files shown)


src/static/js/chapter.js
```diff
@@ -21,11 +21,33 @@ function dataSaverEnabled() {
...
       gtag('event', 'data-saver', { 'event_category': 'user', 'event_label': 'not-enabled', 'value': 0 });
     }
+  } else {
+    gtag('event', 'data-saver', { 'event_category': 'user', 'event_label': 'not-reported', 'value': 0 });
   }
 
   return dataSaver;
 }
 
+//Check if networh API states this is a high bandwidth connection
```
```suggestion
//Check if network API states this is a high bandwidth connection
```

src/templates/en/2019/base_chapter.html
```diff
@@ -18,7 +18,7 @@
   indexBox.classList.toggle('show');
 });
 
-fetch(`https://discuss.httparchive.org/t/{{ metadata.get('discuss') }}.json`)
+fetch("https://discuss.httparchive.org/t/{{ metadata.get('discuss') }}.json")
```
rviscomi: The string template `{{ ... }}` only works with backticks. Either revert this or use string concatenation. 
```suggestion
fetch(`https://discuss.httparchive.org/t/{{ metadata.get('discuss') }}.json`)
```

bazzadp: Are you sure? Just tested again and definitely seems to work...

rviscomi: Oh no you're right. I was confusing this with the `${var}` syntax in JS. The `{{ ... }}` are replaced by the Jinja renderer in Python.

bazzadp: Phewww! Thought you'd caught be out pushing untested code again - but not this time!! ☺️

Issue #350, opened 2019-11-06
What about having a bio for every contributor ? 


AymenLoukil:
What about having a bio for every contributor ?

rviscomi:
IMO it only makes sense for authors because their credentials add authority to their chapters. In the Contributors page designs we had a line for everyone to have their own tagline; just a few words about themselves like "SEO expert" or "World-class accessibility speaker". We can look into bringing that back if there's interest.

bazzadp:
I'm going to close this. I think now we've launched and we're getting less interaction from the wider team this would be a huge effort to implement on most of the contributors and, like @rviscomi , I'm not convinced of the need. Maybe one for the 2020 edition.

Feel free to reopen if you strongly disagree.

PR #541, opened 2019-11-25
Add keyboard shortcuts for previous/next chapter #377

### LINKED ISSUE #377: Add keyboard shortcuts for previous/next chapter
Not launch blocking.

It would be nice to be able to "page" through the chapters using keyboard shortcuts and not have to scroll to the bottom of the page for the prev/next links.

I was thinking of using either the "p" and "n" keys or "," and "." (< and >). We should also make these shortcuts discoverable somehow, ideally, for both accessibility and convenience.

arsenicraghav:
Navigate through chapters using **p**(previous) **n**(next) and **,**(previous)  **.**(next)

Closes #377


src/config/2019.json
```diff
@@ -746,6 +746,16 @@
       "avatar_url": "https:\/\/avatars2.githubusercontent.com\/u\/24447527?v=4&s=200",
       "github": "raghuramakrishnan71"
     },
+    "raghav": {
+      "name": "Raghvendra Kumar",
+      "bio":"Tech Enthusiast and Developer",
```
bazzadp: At the moment only Author's bios are used (see #350) so other contributors don't get a bio - sorry! 😊

```suggestion
```

src/templates/en/2019/base_chapter.html
```diff
@@ -119,6 +127,7 @@
 
 {% macro render_prevnext() %}
   {% if prev_chapter %}
+test
```
bazzadp: This needed?

```suggestion
```

src/templates/en/2019/base_chapter.html
```diff
@@ -119,6 +127,7 @@
...
       <a id="previous-chapter" href="/{{lang}}/{{year}}/{{prev_chapter['title'].lower().replace(' ', '-').replace('/', '')}}">
```
bazzadp: Could we add a title to show the shortcuts on hover? And similarly on next button?

```suggestion
      <a title="Previous Chapter (press 'p' or ',')" id="previous-chapter" href="/{{lang}}/{{year}}/{{prev_chapter['title'].lower().replace(' ', '-').replace('/', '')}}">
```

src/templates/en/2019/base_chapter.html
```diff
@@ -32,6 +32,14 @@
   var el = document.getElementById('num_comments');
   el.innerText = comments + ' ' + (comments == 1 ? 'comment' : 'comments');
 });
+document.addEventListener("keypress", function onPress(event) {
+ 	if (event.key === 'p' || event.key === ',') {
```
bazzadp: Should we cater for other versions of these keypresses?

```suggestion
 	if (event.key === 'p' || event.key === 'P' || event.key === ',' || event.key === '<') {
```

src/templates/en/2019/base_chapter.html
```diff
@@ -32,6 +32,14 @@
...
+ 		document.getElementById('previous-chapter').click();
+ 	}
+ 	if (event.key === 'n' || event.key === '.') {
```
bazzadp: Should we cater for other versions of these keypresses?

```suggestion
 	if (event.key === 'n' || event.key === 'N' || event.key === '.' || event.key === '>') {
```

src/templates/en/2019/base_chapter.html
arsenicraghav: We could include check for the uppercase but I think we should not include the arrow("<", ">") keys for this as it's a bit counter-intuitive ( cause arrows are also used for vertical and horizontal scrolling).

src/templates/en/2019/base_chapter.html
```diff
@@ -32,6 +32,14 @@
   var el = document.getElementById('num_comments');
   el.innerText = comments + ' ' + (comments == 1 ? 'comment' : 'comments');
 });
+document.addEventListener("keypress", function onPress(event) {
+ 	if (event.key === 'p' || event.key === ',') {
+ 		document.getElementById('previous-chapter').click();
```
rviscomi: Need to handle the edge cases for the first and last chapters, for example:

```suggestion
 		var previous = document.getElementById('previous-chapter').click();
 		if (previous) {
 		  previous.click();
 		}
```

src/templates/en/2019/base_chapter.html
rviscomi: In this case we're only targeting angled brackets, not the actual arrow keys. I think it's ok to handle `Shift` key versions of other supported shortcut keys.

bazzadp:
@arsenicraghav thanks for your work on this and for the fixes you've put. Any thoughts on my `title` suggestions? Not sure if you've missed that as now outdated.

Also just done a merge to master so you'll need to fix conflicts. Gimme a shout if you need a hand with this.

Not sure if the non-Querty keyboard thing is an issue - anyone got any opinions on this?

@rviscomi any thoughts on accessibility concerns? I checked a few common tools (JAWS, NVDA) and these particular keys don't seem mapped to anything by default but not sure how many people add custom keys and what usual practice is to avoid any issues. Quick Google shows nothing.

bazzadp:
Also updated the first comment so it will auto-close the corresponding issue then merged.

arsenicraghav:
> Should it be easy to switch these off in case they override screen-readers use of these keys?

Yeah, event listener for 'keypress' can be turned off. 

 >Also just done a merge to master so you'll need to fix conflicts.

have resolved the conflicts

>thoughts on my title suggestions

have added 'title' attribute on both the anchor tag

> non-Querty keyboard thing is an issue ?

detecting keyboard layout or type using javascript does not seem possible so not sure how to handle that.  

Ready for review again

src/templates/en/2019/base_chapter.html
```diff
@@ -17,6 +17,36 @@
...
+    return;
+  }
+
+  var comments = +r.posts_count - 1;
+  if (isNaN(comments)) {
+    return;
+  }
+  var el = document.getElementById('num_comments');
+  el.innerText = comments + ' ' + (comments == 1 ? 'comment' : 'comments');
+});
```
bazzadp: This whole section has been moved to [static/js/chapter.js](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/static/js/chapter.js) so should not be here any longer.

src/config/2019.json
```diff
@@ -746,6 +746,15 @@
       "avatar_url": "https:\/\/avatars2.githubusercontent.com\/u\/24447527?v=4&s=200",
       "github": "raghuramakrishnan71"
     },
+    "raghav": {
+      "name": "Raghvendra Kumar",
+      "teams": [
+        "developers"
+      ],
+      "avatar_url": "https:\/\/avatars3.githubusercontent.com\/u\/9475742?s=200&v=4",
```
bazzadp: Minor nit for consistency:

```suggestion
      "avatar_url": "https:\/\/avatars3.githubusercontent.com\/u\/9475742?v=4&s=200",
```

Issue #532, opened 2019-11-22
Add Apache license info to page footer


rviscomi:
Change

> Copyright © 2019 Web Almanac. All rights reserved.

to

> Copyright © 2019 Web Almanac. Licensed under [Apache 2.0](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/LICENSE).

cc @brucelawson

bazzadp:
This has just been deployed to production and was going to close this but looks like some issues:

Desktop is missing a space:
![image](https://user-images.githubusercontent.com/10931297/69576921-31ff2480-0fc5-11ea-9eea-8135cd97d16a.png)

Mobile has the licence over to the right:
![image](https://user-images.githubusercontent.com/10931297/69576961-46432180-0fc5-11ea-85ec-55f815e61d65.png)

Both seem to be due to the use of `display: flex` for `.copyright` class and removing that fixes it so gonna go ahead and do that. Shout if you see a better way of fixing this.

PR #543, opened 2019-11-25
Fix display issues with Copyright logo

bazzadp:
Closes: #532


src/static/css/2019.css
```diff
@@ -333,7 +333,6 @@ footer .language-switcher {
 }
 footer .copyright {
   grid-area: copyright;
-  display: flex;
```
rviscomi: AFAICT this changes the alignment of the copyright with the social links. I think the correct fix is to ensure there's a space (not a newline) between "under" and the link in the template.

One other thing I'd ask you to fix is to ensure the color of the link is white by default and yellow on hover, like other footer links.

bazzadp: > AFAICT this changes the alignment of the copyright with the social links.

I don't see it. Can you explain more what you mean?

> I think the correct fix is to ensure there's a space (not a newline) between "under" and the link in the template.

That new line is not there in the source code:

image

Chrome dev tools is adding that automatically, if you saw it in Element inspector.

BTW it also works adding a `&nbsp` instead of the space. Basically seems the browser is treating these two items as separate items where flat space between them can be collapsed because of the `flex` property.

> One other thing I'd ask you to fix is to ensure the color of the link is white by default and yellow on hover, like other footer links.

That's a fair ask - added now!

What do you think about the underline? I think keep it even though other nav items on home page don't have it cause otherwise not obvious it's a link without it.

Also we have "Copyright ©" and I think it's redundant to have both the text and the symbol. It also makes the line stretch to two lines on most mobiles which looks ugly:

image

Let me know if you're cool removing the text and just using the © symbol (on both Desktop and Mobile):

image

rviscomi: It's vertically aligned to the top:

image

Originally it was vertically centered:

image

bazzadp: OK went with [a different fix](https://stackoverflow.com/questions/44432447/displayflex-with-text-and-tag-inside-removes-spaces) in the [latest commit](https://github.com/HTTPArchive/almanac.httparchive.org/pull/543/commits/0cf3811aef5369e94ca6d4fa97f51693ee3831db). Think that is a bit more a hack but seems to work fine.

bazzadp: Actually I had to do the `<span>` thing as otherwise it still put the link to the right on mobile:

image

Issue #531, opened 2019-11-22
Logo tweak


rviscomi:
![Screenshot_20191122-092256](https://user-images.githubusercontent.com/1120896/69409722-c9d0ea00-0d09-11ea-83c3-c07521adfd72.png)

Extend the yellow line of the logo to extend fully to the right edge of the word "almanac". It occurred to me right now that it appears to underline the letters "man" which looks weird.

catalinred:
I'll extend that yellow line, convert this to an SVG and get rid of the obsolete 2x image approach.

bazzadp:
While we're on Logo tweaks the HTTPArchive logo rudely cutting off the 'a' has been bugging me for a while. 

![image](https://user-images.githubusercontent.com/10931297/69676841-8542a700-1099-11ea-8d9d-ef401071d548.png)

Do we have a better version of this @rviscomi ? Or @catalinred could you create one?

catalinred:
> While we're on Logo tweaks the HTTPArchive logo rudely cutting off the 'a' has been bugging me for a while.
> 
> ![image](https://user-images.githubusercontent.com/10931297/69676841-8542a700-1099-11ea-8d9d-ef401071d548.png)
> 
> Do we have a better version of this @rviscomi ? Or @catalinred could you create one?

I'll check this as well and let you know @bazzadp ;)

rviscomi:
We have an SVG version of the HTTP Archive logo here: https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/images/http-archive-logo-color-nogradient.svg

Related #262

catalinred:
> We have an SVG version of the HTTP Archive logo here: https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/images/http-archive-logo-color-nogradient.svg
> 
> Related #262

Awesome, because I didn't find it within the .psd file you sent me a while ago. Thank you!

PR #549, opened 2019-11-26
Update the main logo to SVG

catalinred:
- Minor update to the logo - extend the yellow line #531
- Replace the PNG to SVG and get rid of the obsolete 2x image approach.
- Update the alt attribute values to match the brand names.
- Remove loading=lazy from the footer's logo as the image has been already loaded to the header.
- Considering the [distances threshold](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/frame/settings.json5?l=971-1003&rcl=e8f3cf0bbe085fee0d1b468e84395aad3ebb2cad) I believe loading=lazy is not necessary for the top part (header) of the site as well, so I removed that too. Please correct me if I'm wrong.
- Fix cropping for the .png HTTP Archive logo for now. Will update later to SVG as there are optimizations to be made to the [current SVG](https://raw.githubusercontent.com/HTTPArchive/legacy.httparchive.org/master/images/http-archive-logo-color-nogradient.svg) and I need more time to do that carefully.


bazzadp:
Awesome work!

> * Remove loading=lazy from the footer's logo as the image has been already loaded to the header.

This is only loaded in the mobile menu, so may not be loaded in desktop view. So that was my thinking to add lazy-loading attribute to that. But from my testing [hiding an image with CSS doesn't avoid it being lazy-loaded](https://github.com/GoogleChrome/web.dev/issues/1850#issuecomment-554126355) so yeah agree to rip it out.

> * Considering the [distances threshold](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/frame/settings.json5?l=971-1003&rcl=e8f3cf0bbe085fee0d1b468e84395aad3ebb2cad) I believe loading=lazy is not necessary for the top part (header) of the site as well, so I removed that too. Please correct me if I'm wrong.

Agree. Again this was a desperate attempt to use lazy-loading for images that only appeared on one form factor, but it didn't make a difference so happy to rip out.

> Will update later to SVG as there are optimizations to be made to the current SVG and I need more time to do that carefully.

That would be great! As its better with this improvement but still a little fuzzy.

catalinred:
> Awesome work!

Thank you, Barry! ❤️

I'll open a new pull request soon with the new SVG HTTP Archive logo.

In the future, I'm planning maybe to iterate on an SVG icons system with using `symbol` e.g. for HTTP Archive and Web Almanac logos, twitter or Github icons.

IMO there's lots of room to improve when it comes to "svgify"/vectorize the images we're using on the site. For example, `static/images/rectangles.png` could be implemented using the same technique I used for `/static/images/intro-background-fit.svg` and so on.

rviscomi:
> IMO there's lots of room to improve when it comes to "svgify"/vectorize the images we're using on the site. For example, static/images/rectangles.png could be implemented using the same technique I used for /static/images/intro-background-fit.svg and so on.

Strongly agree! Thanks for your work on this @catalinred!

Issue #540, opened 2019-11-25
Translate content to Arabic


SilentJMA:
These are the chapters to be translated, and they exist in the [content directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/master/src/content):
- [ ] accessibility
- [ ] caching
- [ ] cdn
- [ ] cms
- [ ] compression
- [ ] css
- [ ] ecommerce
- [ ] fonts
- [ ] http2
- [ ] javascript
- [ ] markup
- [ ] media
- [ ] mobile-web
- [ ] page-weight
- [ ] performance
- [ ] pwa
- [ ] resource-hints
- [ ] security
- [ ] seo
- [ ] third-parties

Additionally the following pages need translated too in the [language specific templates directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/master/src/templates):
- [ ] base.html
- [ ] base_chapter.html
- [ ] contributors.html
- [ ] error.html
- [ ] index.html
- [ ] featured_chapters.html
- [ ] methodology.html
- [ ] table_of_contents.html
- [ ] accessibility_statement.html

There is no need to translate the chapters HTML pages are they are generated off the markdown combined with the above templates.

Common notes for writing consistency are here: https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Translators'-Guide. Feel free to edit that and/or add Arabic-specific extras by editing this comment.

Arabic-specific extra advice:
- None yet

bazzadp:
Greats stuff @SilentJMA - I've updated your first comment showing what needs to be done. Feel free to stick your name against an item, clone the repo, and submit pull requests (PRs) with your translations.

Note a lot of the visualisation graphics are due to change in https://github.com/HTTPArchive/almanac.httparchive.org/pull/511 so might want to hold off for that for some of the chapters, but the JavaScript, Accessibility and HTML chapters are good to go now.

Unfortunately my Arabic is non-existent so would be good to have someone else fluent in the language to review your PRs. I'm happy to give PRs a once over for other technical issues though.

SilentJMA:
> Greats stuff @SilentJMA - I've updated your first comment showing what needs to be done. Feel free to stick your name against an item, clone the repo, and submit pull requests (PRs) with your translations.
> 
> Note a lot of the visualisation graphics are due to change in #511 so might want to hold off for that for some of the chapters, but the JavaScript, Accessibility and HTTP/2 chapters are good to go now.
> 
[...]

PR #551, opened 2019-11-27
Add ability to overwrite HTML tag (e.g. for dir="rtl" for Arabic)

bazzadp:
As discussed in https://github.com/HTTPArchive/almanac.httparchive.org/issues/540#issuecomment-559025999


src/templates/base.html
```diff
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="{% block lang %}{{ language.lang_attribute }}{% endblock %}">
+{% block htmltag %}<html lang="{{ language.lang_attribute }}">{% endblock %}
```
rviscomi: Very minor but I'd recommend doing this in a way that doesn't interfere with the lang attribute.
```suggestion
<html lang="{% block lang %}{{ language.lang_attribute }}{% endblock %}" {% block dir_attribute %}{% endblock %}>
```

And in the Arabic template you can specify:

```html
{% block dir_attribute %}dir="rtl"{% endblock %}
```

So it's unset by default but overridable in child templates.

bazzadp: I'm in two minds about this.

I did try to go similar to what you did but ended up changing to what I submitted for a few reasons:

- It's not a `dir_attribute`, it's an `html_tag_other_attributes`, which got a little long-winded.
- I quite liked the ability to override the tag completely, so you could specify the order for example. Don't think it really matters here, but [examples online](https://www.w3.org/International/questions/qa-html-dir) have `dir` before `lang` for example.
- It could be in future certain languages might want to override the language, or include the region as well as the language (e.g. if we have an en-GB).

Anyway don't think it's a big deal whichever way we go but if we are going your way then think we should change the block name from `dir_attribute` but surely there's a better name than `html_tag_other_attributes`!!

bazzadp: Any thoughts or preferences here @mikegeyser?

If not will just go with what @rviscomi suggests some we can close this out.

mikegeyser: I really like @rviscomi's suggestion.

bazzadp: Done! Decision made.

Issue #400, opened 2019-11-09
Validate chapters by year


rviscomi:
As a follow-up to #392, we shouldn't hardcode the list of chapters. It should support any given year and pull the chapter IDs from the config.

arsenicraghav:
Is this being worked on by someone ?

bazzadp:
> Is this being worked on by someone ?

Nope, not had a chance to look at it yet - go for it!

arsenicraghav:
Sure! I will give it a shot. can you assign it ?

bazzadp:
> can you assign it ?

Are you not able to grab it yourself by clicking on the "assign yourself" link?

![image](https://user-images.githubusercontent.com/10931297/69609023-8aafdb00-1020-11ea-8cce-1ff7ff728681.png)

arsenicraghav:
No, I don't think so. I don't see an option.

rviscomi:
Need to add @arsenicraghav to the GitHub Almanac team for them to be assignable

bazzadp:
> Need to add @arsenicraghav to the GitHub Almanac team for them to be assignable

Well we should add him to there anyway. And to developers. Can you permission me to be able to do that?

However I am able to assign him. So not convinced he _needs_ to be in there. But then that could be cause of my permissions and I can assign anyone? So not sure if it's a permission thing or a UX thing - I think assigning issues in GitHub isn't the most obvious UX and is easily missed!

arsenicraghav:
>pull the chapter IDs from the config.

@rviscomi  do you mean to say from an existing config file ?

rviscomi:
Yes, see 2019.json https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/config/2019.json

bazzadp:
> Yes, see 2019.json https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/config/2019.json

He's already done that as part of #541 .

Think he's asking how he can get access to assign items himself. Just confirmed people in the team can't assign issues to themselves. Can we permission him to be able to do that without full permissions for the repo if we want to keep that locked down a bit for now?

rviscomi:
I've just given @bazzadp @obto and @mikegeyser permission to add members to the [Almanac Contributors team](https://github.com/orgs/HTTPArchive/teams/almanac-contributors) and sent @arsenicraghav an invitation to join the Developers subteam.

bazzadp:
[...]

PR #557, opened 2019-11-27
Validate chapters by year #400

arsenicraghav:
load and maintain chapters dynamically from json file for each year

Please have a look and let me know your feedback

### DIFF
--- a/src/validate.py
+++ b/src/validate.py
@@ -59,8 +59,8 @@ def validate(func):
 
 def validate_chapter(chapter,year):
 
-    CHAPTERS_FOR_YEAR = CHAPTERS.get(year)
-    if chapter not in CHAPTERS_FOR_YEAR:
+    chapters_for_year = CHAPTERS.get(year)
+    if chapter not in chapters_for_year:
         if chapter in TYPO_CHAPTERS:
             logging.debug('Typo chapter requested: %s, redirecting to %s' % (chapter, TYPO_CHAPTERS.get(chapter)))
             return TYPO_CHAPTERS.get(chapter)
@@ -160,9 +160,12 @@ def get_chapters(file):
 
     return chapters
 
-json_files = get_json_files(CONFIG_DIR)
+def init():
+    year_config_files = get_json_files(CONFIG_DIR)
 
-for file in json_files:
-    CHAPTERS.update({file[9:-5] : set(get_chapters(file))})
+    for file in year_config_files:
+        CHAPTERS.update({file[9:-5] : set(get_chapters(file))})
 
-DEFAULT_YEAR = sorted(CHAPTERS.keys())[-1]
+    return sorted(CHAPTERS.keys())[-1]
+
+DEFAULT_YEAR = init()
--- a/src/validate.py
+++ b/src/validate.py
@@ -13,14 +13,13 @@ import config as config_util
 
 CONFIG_DIR = './config'
 
-DEFAULT_YEAR = str(date.today().year)
 SUPPORTED_YEARS = {
     # When there is one supported language, it must have a trailing comma.
     '2019': (Language.ENGLISH,)
 }
 
 #TO DO - Stop Hardcoding these
-CHAPTERS = set();
+CHAPTERS = {}
 
 TYPO_CHAPTERS = {
     'http-2': 'http2',
@@ -41,7 +40,7 @@ def validate(func):
 
         if chapter:
 
-            validated_chapter = validate_chapter(chapter)
+            validated_chapter = validate_chapter(chapter,year)
 
             if chapter != validated_chapter:
                 return redirect('/%s/%s/%s' % (lang, year, validated_chapter), code=301)
@@ -59,9 +58,10 @@ def validate(func):
     return decorated_function
 
 
-def validate_chapter(chapter):
+def validate_chapter(chapter,year):
 
-    if chapter not in CHAPTERS:
+    CHAPTERS_FOR_YEAR = CHAPTERS.get(year)
+    if chapter not in CHAPTERS_FOR_YEAR:
         if chapter in TYPO_CHAPTERS:
             logging.debug('Typo chapter requested: %s, redirecting to %s' % (chapter, TYPO_CHAPTERS.get(chapter)))
             return TYPO_CHAPTERS.get(chapter)
@@ -127,7 +127,7 @@ def parse_accept_language(header, supported_langs):
     # If all else fails, default the language.
     return DEFAULT_LANGUAGE.lang_code
 
-def get_json_files_from_dir(path):
+def get_json_files(path):
 
     files_found = []
     for root, directories, files in os.walk(path):
@@ -150,7 +150,7 @@ def get_entries_from_json(path, p_key, s_key):
 
     return entries
 
-def get_chapters_for_year(file):
+def get_chapters(file):
 
     chapters = []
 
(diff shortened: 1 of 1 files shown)


src/validate.py
```diff
@@ -1,43 +1,26 @@
...
 from functools import wraps
 from language import Language, DEFAULT_LANGUAGE
+from datetime import date
 
 import config as config_util
 
-DEFAULT_YEAR = '2019'
+CONFIG_DIR = './config'
+
+DEFAULT_YEAR = str(date.today().year)
```
bazzadp: Bit concerned about this. If next Almanac doesn't come out until Nov next year, will this stop working for first 11 months of 2020?

Also DEFAULT_YEAR is also used in main.py to handle redirect from / to /en/2019/.

src/validate.py
```diff
@@ -144,3 +126,53 @@ def parse_accept_language(header, supported_langs):
...
+    return entries
+
+def get_chapters_for_year(file):
+
+    chapters = []
+
+    data = get_entries_from_json(file,'outline','chapters')
+    for list in data:
+        for entry in list:
+            chapters.append(entry.get('title').lower().replace(" ", "-").replace("/",""))
```
bazzadp: Seems a repeat of `get_chapter_slug` from main.py. Should this be generalised into one function that both can call?

src/validate.py
```diff
@@ -144,3 +126,53 @@ def parse_accept_language(header, supported_langs):
...
+
+    return parts
+
+json_files_present = get_json_files_from_dir(CONFIG_DIR)
+
+
+#TO DO - Lazy load previous year chapters
+
+for file in json_files_present:
+    if DEFAULT_YEAR in file:
```
bazzadp: Don't understand this. So it will ONLY load the default year JSON file? That's not a default then (which implies a value to fall back to if nothing is specified/found) - but a hardcoded value. Which, given my other concerns seems like a problem.

src/validate.py
```diff
@@ -144,3 +126,53 @@ def parse_accept_language(header, supported_langs):
...
+def get_parts_for_year(file):
+
+    parts = get_entries_from_json(file,'outline','part')
+
+    return parts
+
+json_files_present = get_json_files_from_dir(CONFIG_DIR)
+
+
+#TO DO - Lazy load previous year chapters
```
bazzadp: What do you mean "lazy load"? Don't we need all valid chapters at each call to be able to validate any URL?

src/validate.py
```diff
@@ -1,43 +1,26 @@
...
-DEFAULT_YEAR = '2019'
+CONFIG_DIR = './config'
+
+DEFAULT_YEAR = str(date.today().year)
 SUPPORTED_YEARS = {
     # When there is one supported language, it must have a trailing comma.
     '2019': (Language.ENGLISH,)
 }
 
 #TO DO - Stop Hardcoding these
```
bazzadp: Shouldn't you remove this now you've done this TO DO 😀
```suggestion
```

src/validate.py
```diff
@@ -144,3 +126,53 @@ def parse_accept_language(header, supported_langs):
...
+
+    entries = []
+
+    if p_key in data:
+        for values in data.get(p_key):
+             entries.append(values.get(s_key))
+
+    return entries
+
+def get_chapters_for_year(file):
```
bazzadp: Not loving the naming of these functions. We say `_for_year` but then pass a file in. I know each year has a single file, but still it seems weird. Rename to `get_chapters_for_file(file)` or simply `get_chapters(file)`?

src/validate.py
arsenicraghav: I mean we can load all the chapters from the current year when the server starts up and for previous years we delay loading of chapters until we receive the first request. Or should we load everything right from the start ?

src/validate.py
arsenicraghav: thanks for pointing to the edge case here, will cover that 

so what should be DEFAULT_YEAR, current year ?

src/validate.py
bazzadp: OK. Not sure how gcloud works to be honest - is the python loaded once per request? Or initialised once and then can call extra things only when needed as you suggest? One for @rviscomi / @mikegeyser to answer I think.

src/validate.py
bazzadp: I think it should probably stay hardcoded. Even when 2020 is started to be released we probably want to continue defaulting to 2019 until launch day.

src/validate.py
arsenicraghav: right.. so I wanted to confirm that.  will be having multiple json files for each year ?

src/validate.py
arsenicraghav: Let me have a look at main.py

src/validate.py
bazzadp: I think it'll be one per year.

src/validate.py
arsenicraghav: have renamed them

src/validate.py
arsenicraghav: have refactored the code to load everything upfront.

src/validate.py
arsenicraghav: It will default to the latest year among the list of json files.

src/validate.py
arsenicraghav: It will default to the latest year among the list of json files.

mikegeyser: I'm not sure, I share @bazzadp's concerns with this implementation.

src/validate.py
```diff
@@ -77,9 +57,10 @@ def decorated_function(*args, **kwargs):
     return decorated_function
 
 
-def validate_chapter(chapter):
+def validate_chapter(chapter,year):
 
-    if chapter not in CHAPTERS:
+    CHAPTERS_FOR_YEAR = CHAPTERS.get(year)
```
mikegeyser: If I'm unmistaken the `ALL_CAPS` convention is for module scoped variables, such as `CHAPTER` and `CONFIG_DIR`. This is method scoped and should probably be `chapters_for_year` to avoid confusion? (I was trying to figure out where else it was being used.)

src/validate.py
```diff
@@ -144,3 +125,44 @@ def parse_accept_language(header, supported_langs):
...
+    chapters = []
+
+    data = get_entries_from_json(file,'outline','chapters')
+    for list in data:
+        for entry in list:
+            chapters.append(entry.get('title').lower().replace(" ", "-").replace("/",""))
+
+    return chapters
+
+json_files = get_json_files(CONFIG_DIR)
```
mikegeyser: Similarly, could this be capitalised as it is module scoped or perhaps named a little more meaningfully, such as `YEAR_CONFIG_FILES` or something.

src/validate.py
```diff
@@ -144,3 +125,44 @@ def parse_accept_language(header, supported_langs):
...
+
+for file in json_files:
```
mikegeyser: Can we perhaps do this work in the `get_json_files` or `get_chapters` method, instead of spreading it between both there and the module scope? The methods are only used in the one place anyway?

src/validate.py
arsenicraghav: sure

src/validate.py
arsenicraghav: sure

src/validate.py
arsenicraghav: have separated the functions based on the unit and type of work so that they remain testable. 

`get_json_files`  walks through the directory structure and provides the list of json files.

since the json file has many segments I tried to keep their retrieval and processing separate as different segments need different processing hence `get_chapters`

let me know if that's desirable.

bazzadp:
Should we pause any work on this until we decide on https://github.com/HTTPArchive/almanac.httparchive.org/issues/561 ? That will change the implementation here significantly and seems like it is going to go ahead...

mikegeyser:
I agree with @bazzadp. This PR is great, but the validation functionality may be dramatically changed by what happens in #561, and we should perhaps avoid thrashing unnecessarily.

src/validate.py
arsenicraghav: sure, let me try to wrap them up.

rviscomi:
I agree with @bazzadp and @mikegeyser. Thank you for working on this @arsenicraghav but I think we should close this PR until #561 is resolved.

PR #562, opened 2019-11-27
Use the HTTP Archive logo as an SVG

catalinred:
Related to #531 and to already merged #549


bazzadp:
LGTM. Lovely and crisp. It does jump from 1.7 KB for PNG to 4.6 KB for SVG but worth it IMHO. And presume that'll be gzipped anyway so be smaller.

Any chance you could do a rebase with master on your clone? Always see your old commits so looks like much more change than there is!

catalinred:
Last time I tried the rebase I messed up things but I understood squash and merge is applied here so that's the reason I didn't rebase anymore. 

But sure, I'll try to rebase next time.

bazzadp:
Yeah it does so changes are small. Just the list of commits in this conversation tab starts to grow - even though it’s only last one that’s a real commit.

PR #566, opened 2019-11-28
Handle print mode

### LINKED ISSUE #520: Ability to force non-interactive data viz
In some cases, like printing a paper copy or saving to PDF, it's better to have static images than interactive iframes. The images fit the screen better and don't require interactivity.

I'd like to add support for something like a `?print` querystring parameter on chapter pages that forces all figures to use images. For now it must also disable lazy loading due to a [bug in Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=875403) that doesn't load images below the fold when printing.

This is a small step towards a "PDF-friendly" version of the Almanac per #37.

bazzadp:
Closes #520 

Note it does it in two ways:

1. if `print=` is provided as a URL param it avoids the upgrade to interactive JS (and also removes Lazy Loading).
2. if print event is called, even after the upgrade, then it removes lazy-loading in js, and hides some elements and switches visuals in print.css. No need for an `onafterprint` as no need to re-add lazyloading and only switch the visuals in print stylesheet.

So a bit of duplication and they aren't exactly the same, but think it's good.

### DIFF
--- a/src/static/js/chapter.js
+++ b/src/static/js/chapter.js
@@ -2,58 +2,39 @@
 //Useful for print view for example (https://bugs.chromium.org/p/chromium/issues/detail?id=875403)
 function removeLazyLoading() {
 
-  console.log("Removing lazy loading...");
-
-  var all_imgs = document.querySelectorAll('img');
-  for (index = 0; index < all_imgs.length; ++index) {
-    var img = all_imgs[index];
-    if (img.getAttribute('loading')) {
-      img.removeAttribute('loading');
-    }
-  }
-  
-  var all_iframes = document.querySelectorAll('iframe');
-  for (index = 0; index < all_iframes.length; ++index) {
-    var iframe = all_iframes[index];
-    if (iframe.getAttribute('loading')) {
-      iframe.removeAttribute('loading');
-    }
+  //If no Array.from then pretty sure there will be no native lazy-loading support to remove!
+  if (Array.from) {
+    console.log("Removing lazy loading...");
+    
+    Array.from(document.querySelectorAll('img[loading], iframe[loading]')).forEach(function(element) {
+      element.removeAttribute('loading');
+    });
   }
 }
 
 //Add an event handler to remove LazyLoading when entering print mode
 function removeLazyLoadingOnPrint() {
   if ("onbeforeprint" in window) {
-    window.onbeforeprint = function() {
-      removeLazyLoading();
-    }
+   window.onbeforeprint = removeLazyLoading;
   }
 
 }
 
 //Check if in print mode so we can remove lazy loading and block interactive visuals
-function printMode() {
+function isInPrintMode() {
   var printMode = false;
-  var field = 'print';
-  var url = window.location.href;
 
-  if(url.indexOf('?' + field + '=') != -1) {
-    printMode = true;
-  } else if(url.indexOf('&' + field + '=') != -1) {
-    printMode = true;
+  if (window.URL && window.URLSearchParams) {
+    var url = new URL(window.location);
+    printMode = url.searchParams.has('print');
   }
-
   if (printMode) {
-    console.log("Print mode");
-    gtag('event', 'print-mode', { 'event_category': 'user', 'event_label': 'true', 'value': 1 });
-    
+    console.log ("Print Mode");
     removeLazyLoading();
-
-    return true;
-  } else {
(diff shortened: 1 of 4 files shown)


src/static/js/chapter.js
```diff
@@ -1,3 +1,61 @@
+//This function removes the lazy-loading attributes from all img and iframe tags
+//Useful for print view for example (https://bugs.chromium.org/p/chromium/issues/detail?id=875403)
+function removeLazyLoading() {
+
+  console.log("Removing lazy loading...");
+
+  var all_imgs = document.querySelectorAll('img');
```
rviscomi: Combine img and iframe into one query, and limit to elements with loading attrs. You could also use forEach to simplify the loop.
```suggestion
  Array.from(document.querySelectorAll('img[loading], iframe[loading]')).forEach(...);
```

src/static/js/chapter.js
```diff
@@ -1,3 +1,61 @@
...
+    if (iframe.getAttribute('loading')) {
+      iframe.removeAttribute('loading');
+    }
+  }
+}
+
+//Add an event handler to remove LazyLoading when entering print mode
+function removeLazyLoadingOnPrint() {
+  if ("onbeforeprint" in window) {
+    window.onbeforeprint = function() {
```
rviscomi: Could be simplified
```suggestion
    window.onbeforeprint = removeLazyLoading;
```

src/static/js/chapter.js
```diff
@@ -1,3 +1,61 @@
...
+      removeLazyLoading();
+    }
+  }
+
+}
+
+//Check if in print mode so we can remove lazy loading and block interactive visuals
+function printMode() {
```
```suggestion
function isInPrintMode() {
```

src/static/js/chapter.js
```diff
@@ -1,3 +1,61 @@
...
+  var printMode = false;
+  var field = 'print';
+  var url = window.location.href;
+
+  if(url.indexOf('?' + field + '=') != -1) {
```
rviscomi: I think this could be simplified using the URL API:

```js
var url = new URL(window.location);
var printMode = url.searchParams.has('print');
gtag('event', 'print-mode', { 'event_category': 'user', 'event_label': '' + printMode, 'value': +printMode })
return printMode;
```

src/static/js/chapter.js
bazzadp: IE 11 doesn’t like `foreach` so have tried to avoid that.

src/static/js/chapter.js
bazzadp: Again not IE11 compatible. IE11 isn’t a huge amount of usage but at same time don’t think it’s a massive improvement so would prefer to keep IE11 support for this is we can.

WDYT?

src/static/js/chapter.js
bazzadp: Did try this but couldn’t get it to work. Will give it another go in case I was doing something stupid.

src/static/js/chapter.js
rviscomi: I think it's supported https://caniuse.com/#feat=mdn-javascript_builtins_array_foreach

src/static/js/chapter.js
rviscomi: You could be right. It should work in theory but IIRC you need to give it context, eg `removeLazyLoading.bind(window)` or something silly.

src/static/js/chapter.js
bazzadp: Ah maybe. Will give it a go. Did try it for something else and that’s when I notice it failed but that might have been using this: https://caniuse.com/#feat=mdn-api_nodelist_foreach

src/static/js/chapter.js
rviscomi:
```js
if (!('URL' in window)) {
  return false;
}
```

Gracefully degrade. IE11 users don't get print mode in that case. We should still be free to write clean, modern code whenever possible.

bazzadp: Fair enough.

And they will get normal print mode when actually printing - just not URL param print mode - as they do support `onbeforeprint` and print stylesheets.

src/static/js/chapter.js
bazzadp: Isn't supported in IE11. But then neither is native-lazy loading so doesn't matter here. And code is MUCH cleaner - so thanks!

src/static/js/chapter.js
bazzadp: No this works fine. Must have been doing something stupid.

PR #569, opened 2019-11-29
Localized Team Names

### LINKED ISSUE #568: i18n - Add a way to localize teams' names.
Current team names (Analysts, Authors, Brainstormers, Designers…) are use on [the Contributors page](https://almanac.httparchive.org/en/2019/contributors) to filter people. We should think about a way to localize theses teams' names.

borisschapira:
With fallback to English team names if the dictionary is not populated.

Fixes #568

### DIFF
--- a/src/templates/en/2019/contributors.html
+++ b/src/templates/en/2019/contributors.html
@@ -302,9 +302,23 @@
     <div class="contributors-count"><span id="filtered-contributors">{{ config.contributors.items() | length }}</span> contributors</div>
     <div class="filter-container">
       <p>Filter by team:</p>
-    {% for id, team in config.teams.items() %}
-      <button class="{{ id }}" type="button">{{ team.name }}</button>
-    {% endfor %}
+      {% set localizedTeamNames = {
+        "analysts": "Analysts",
+        "authors": "Authors",
+        "brainstormers": "Brainstormers",
+        "designers": "Designers",
+        "developers": "Developers",
+        "editors": "Editors",
+        "reviewers": "Reviewers",
+        "translators": "Translators", 
+        } %}
+        {% for id, team in config.teams.items() %}
+          {% if localizedTeamNames[id]|length %}
+            <button class="{{ id }}" type="button">{{ localizedTeamNames[id] }}</button>
+          {% else %}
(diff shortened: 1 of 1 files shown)


src/templates/en/2019/contributors.html
```diff
@@ -302,9 +302,23 @@ <h1 class="title title-lg">
...
+        "authors": "Authors",
+        "brainstormers": "Brainstormers",
+        "designers": "Designers",
+        "developers": "Developers",
+        "editors": "Editors",
+        "reviewers": "Reviewers",
+        "translators": "Translators", 
+        } %}
+        {% for id, team in config.teams.items() %}
+          {% if localizedTeamNames[id]|length %}
```
mikegeyser: Does the null coalescing `or` not work in this situation (`localizedTeamNames[id] or team.name`)? If it doesn't perhaps set a variable once (inside the `if else`) instead of duplicating the markup for the button.

mikegeyser:
I'm happy with the chance, but it isn't strictly speaking necessary for `en/2019/contributors.html`. Should we not do this as a part of `fr/2019/contributors.html` instead?

bazzadp:
> I'm happy with the chance, but it isn't strictly speaking necessary for `en/2019/contributors.html`. Should we not do this as a part of `fr/2019/contributors.html` instead?

I think each translator will likely use `en` as the base, so personally I think we should do it here, rather than trying to explain to a Spanish speaker (for example) to mostly base it on the `en` template, but also checkout the `fr` template for some language-specific overrides.

WDYT?

bazzadp:
> > I'm happy with the chance, but it isn't strictly speaking necessary for `en/2019/contributors.html`. Should we not do this as a part of `fr/2019/contributors.html` instead?
> 
> I think each translator will likely use `en` as the base, so personally I think we should do it here, rather than trying to explain to a Spanish speaker (for example) to mostly base it on the `en` template, but also checkout the `fr` template for some language-specific overrides.
> 
> WDYT?

And another reason for doing it in English is consistency. So if we break it (e.g. by moving from JSON to Python) then we break it in all languages and so test it (to avoid breaking it in the first place) and if we miss it there we see it broken quicker.

mikegeyser:
I'm happy with that. 😊

Issue #548, opened 2019-11-26
Add Interest Form link to Contributors page


bazzadp:
I really think we’re missing a trick not having an interest form link for next year on the Contributors page.

Either directly or linking to the appropriate part of the Methodology page.

It’s the first place I went looking for it when I was reminded I hadn’t gotten round to signing up for next year yet.

Thoughts?

rviscomi:
I agree, I think the contributors page makes sense as a place to bring new people in. Not sure about the design aspect though, open to ideas.

catalinred:
Usually, you'd want to add a form at the end of the contributors page, but considering the current page design and height, that wouldn't be easy to complete and submit from a UX perspective.

Maybe an idea would be to add a "Become a contributor" call to action near to the main `h1` and link to the current Google Forms? This would be a good start IMO.

Alternatively, we could design/add an avatar/silhuette/+/join/etc... icon near to the existing `/static/images/avatars/number.jpg` ones and make it stand out from the rest - from a design perspective. Again, a link on it will go to the existing [Google form](https://forms.gle/Qyf3q5pKgdH1cBhq5).

bazzadp:
Are we not over thinking this? Wouldn’t an introductory paragraph with a link suffice? I think the page is a bit bare without this anyway.

I’m thinking something like this:

“The Web Almanac was built by a large team of contributors from analysts, authors...etc.

We are still continuing to develop the Web Almanac website including translate the chapters into other languages. We would invite people to comment on our [GitHub Issues list](https://github.com/HTTPArchive/almanac.httparchive.org/issues/) if they want to get involved.

We hope to make this an annual publication and have an [interest form](http://example.com) for those who wish to be involved in 2020.”

rviscomi:
No strong opinions. I kind of like the idea @catalinred was exploring with a silhouette, almost like an index-0 placeholder contributor in the list, which is styled similar to the other contributor items. It doesn't need much text, maybe just a "Join the team" button that links to the interest form or a special CONTRIBUTORS.md file here in the repo.

@bazzadp does that work for you? If you prefer to go with the paragraph approach that's fine too.

bazzadp:
I’m happy as long as it works! 😀

Will admit I’m struggling to visualise it and see how exactly it’ll be as obvious to people as a simple intro paragraph.

Wanna do a mock up @catalinred ?

catalinred:
[...]

Issue #556, opened 2019-11-27
Code block on methodology page lost syntax highlighting


rviscomi:
On https://almanac.httparchive.org/en/2019/methodology#dataset

![image](https://user-images.githubusercontent.com/1120896/69724838-b8615680-10ea-11ea-9492-5c9b102353d4.png)

I've manually applied syntax highlighting spans and classes to the query but the style selectors are not matching anymore. I think a parent class was renamed or the stylesheet moved or something.

catalinred:
I dug a bit into this because at first, I thought I messed this up with one of my old commits. 😱 

Even though I modified `methodology.html` and `methodology.css` in the past, the latter I only modified it for [merging CSS duplicated rules](https://github.com/HTTPArchive/almanac.httparchive.org/commit/71e0e3e1afbd629f2153da9d098f5eb15047f9a3#diff-99747f4a7bab79e7de9502f180a97932). Then I remembered [I also did notice](https://github.com/HTTPArchive/almanac.httparchive.org/pull/471#issuecomment-553340163) that `methodology.css` wasn't used/loaded anymore in any page.

So I went further and I played Sherlock a bit. 🕵 
I found that the `methodology.css` was added in `methodology.html` by @rviscomi on [Oct 18](https://github.com/HTTPArchive/almanac.httparchive.org/commit/6bb9efa64cd938d6fe7d56550d09a12b9d471401#diff-99747f4a7bab79e7de9502f180a97932) and seem to have been removed by @tyohan on [Oct 29](https://github.com/HTTPArchive/almanac.httparchive.org/commit/77946a9f82d08e689f65dd02f2e42c1d415efced#diff-99747f4a7bab79e7de9502f180a97932).

Anyways, I re-added the missing `.code-block` bits to currently in use `page.css` 
Fixed in #573

catalinred:
Fixed by #573

Issue #46, opened 2019-06-13
Design the home page


rviscomi:
- [x] list all of the information that will need to be included on the home page
- [x] design the UX of the home page, establishing the identity and style of the Almanac website

I'll get started on the first one to unblock the designer who will work on the second one.

rviscomi:
Outline of the information we'll need on the home page, actual copy to be written later:

- overview of the Almanac project
  - mission statement
  - facts and stats (eg 90 contributors from the web community, ~5M websites tested, 20 TB of data processed, etc)
  - should be obvious that this is the first of an annual series, make "2019" prominent 
- primary call to action: start reading the report
- secondary flows
  - find out how it works in our methodology
  - see who made this possible on our contributors page
  - select alternate language in a footer language picker

rviscomi:
Home page is ready!

![home-001c](https://user-images.githubusercontent.com/1120896/62047735-32cfa780-b1d9-11e9-9f18-bc3a0f9db6e1.jpg)

PR #573, opened 2019-11-29
Consistent body and header width

### LINKED ISSUE #480: Fix Table of Contents margin
The margin on the TOC page is not consistent with other pages. The main section of the page should always be as wide as the header.

image

catalinred:
This fixes #480

- The body is always as wide as the header due to the new `.container`class
- Within the contributors page, I've moved the counter to the left, as specified [here](https://github.com/HTTPArchive/almanac.httparchive.org/issues/548#issuecomment-559602998). Also, due to the new consistent `.container`, the contributors page is now supporting 5 columns of cards.


catalinred:
@bazzadp I did a rebase and `git push -f` but still wasn't able to get rid of those extra old commits already merged. I guess I'm missing something.  :sigh:

bazzadp:
Well at least you tried! 😀

Never worry for this PR. Will review as is.

Here’s what I normally do btw: https://stackoverflow.com/questions/7929369/how-to-rebase-local-branch-with-remote-master/18442755#18442755

But I’m no git expert myself to be honest.

catalinred:
I'm working on #556 as well and would like to fix that too with this PR, if you agree!

catalinred:
> Well at least you tried! 😀
> 
> Never worry for this PR. Will review as is.
> 
> Here’s what I normally do btw: https://stackoverflow.com/questions/7929369/how-to-rebase-local-branch-with-remote-master/18442755#18442755
> 
> But I’m no git expert myself to be honest.

Thank you for the link. Actually I am/supposed to be very familiar with `pull --rebase` but...

src/templates/fr/2019/contributors.html
```diff
@@ -299,9 +294,11 @@ <h1 class="title title-lg">
       <img src="/static/images/avatars/9.jpg" height="100" width="100" loading="lazy" />
   </div>
   <div class="contributors {{ ' '.join(config.teams.keys()) }}">
-    <div class="contributors-count"><span id="filtered-contributors">{{ config.contributors.items() | length }}</span> personnes</div>
     <div class="filter-container">
-      <p>Filtrer par équipe&nbsp;:</p>
+      <p>
+        Filtrer par équipe&nbsp;:
```
bazzadp: Any reason for this `&nbsp;` @borisschapira ?

src/static/css/2019.css
```diff
@@ -30,6 +30,14 @@ a:focus, a:hover {
   text-decoration: underline;
 }
 
+/* The main wrapper, center aligned */
+.container {
+  max-width: 1280px;
```
bazzadp: Why 1280 when previously it was 1324? Means all the chapters are a little bit thinner and longer.

```suggestion
  max-width: 1324px;
```

src/static/css/2019.css
```diff
@@ -30,6 +30,14 @@ a:focus, a:hover {
...
+  padding-left: 20px;
+  padding-right: 20px;
```
bazzadp: Any reason we need this padding? Wasn't there previously.

src/templates/en/2019/base.html
```diff
@@ -73,7 +75,9 @@ <h2>
   </header>
   {% endblock %}
 
-  {% block main %}{% endblock %}
+  <div class="container">
+    {% block main %}{% endblock %}
+  </div>
 
   {% block footer %}
     <footer class="alt-bg">
```
bazzadp: Why is the footer not in a container div? Currently it's very wide compared to everything else on the page.

src/static/css/index.css
```diff
@@ -17,6 +17,19 @@ header.alt-bg {
   background-color: transparent;
 }
 
+header .container {
+  max-width: none;
```
bazzadp: Why no max-width? I agree that is the current design but seems weird that Home page is very wide, but rest are constrained. Maybe not part of this PR and maybe one to raise in an issue and discuss, but looks weird to me none-the-less so thought I'd ask.

src/static/css/index.css
```diff
@@ -17,6 +17,19 @@ header.alt-bg {
...
+  padding-left: 60px;
+  padding-right: 60px;
+}
+
+@media (max-width: 600px) {
+  header .container {
+    padding-left: 30px;
+    padding-right: 30px;
```
bazzadp: Why 30px instead of 20px? It makes the menu appear weird on mobile for index page only:

image

src/static/css/2019.css
catalinred: As I wrote [here](https://github.com/HTTPArchive/almanac.httparchive.org/issues/480#issuecomment-556009375), the Methodology page was relying on a `max-width: 1324px` while Table of Contents & Contributors pages on `max-width: 1024px`. 

While I thought 1324px is an arbitrary value chosen for a breakpoint and 1024px way too small, considering a 1200px is recommended value for desktop-and-up displays when it comes to how [frameworks structure their breakpoints](https://getbootstrap.com/docs/4.3/layout/overview/#responsive-breakpoints) I decided to go with a median to say so between the 1200 which I well guessed you will consider being too low and the actual 1324.

I'm happy to update this to 1324 if that's what you want but I still consider this value is "hardcoded" and too big.

src/static/css/2019.css
catalinred: When defining a container element, you want to set left and right padding "bleeds" which are both visible and effective on mobile displays. 

Also, I could have used a shorthand like `padding: 0 20px`to get rid of an extra line but this could have caused issues when combining .e.g. with other padding helpers - helpers which we don't have now but we might in the future.

src/templates/fr/2019/contributors.html
catalinred: Noticed that too, wanted to change but as this got to master, I thought there might be a reason for that.

src/static/css/2019.css
bazzadp: OK I hadn't considered the chapters changing. I'm OK with 1200.

src/static/css/2019.css
bazzadp: Then is it needed on desktop? Just with the reduction from 1324 -> 1200 and then losing another 40px here, we've effectively gone from 1324 to 1160. Not that that is necessarily bad, but thought I'd ask the question.

src/templates/fr/2019/contributors.html
bazzadp: I saw remove. Unless @borisschapira gives good reason to keep it.

borisschapira: French microtypography. We have non-break spaces before ":", ";", "!", "?", between numbers and units ("14&nbsp;%")…

src/templates/en/2019/base.html
catalinred: I decided to let the footer that wide on purpose, as I thought it looks pretty good to me - from a design perspective - considering its current content: logo, menu, social, icons, copyright info.

Also, having a full-width footer is most of the time a common practice in web design/development.

src/templates/fr/2019/contributors.html
bazzadp: OK. As long as there's a reason. Resolving this with no change then.

src/templates/en/2019/base.html
bazzadp: Yeah seen it before but not a fan of that design principal. Especially when not a huge amount in our footer to need that. But hey that's a personal opinion and sure there's plenty of others who prefer this design.

The [original design](https://github.com/HTTPArchive/almanac.httparchive.org/issues/46#issuecomment-515966879) has narrower footer, then again it has wider header - that we are changing with this PR - so not sure that's helping my argument! 😀

Happy to resolve this with no change if you think it's working as intended.

src/static/css/index.css
catalinred: Sure, that's a good question to ask.

I didn't ask about the current design decisions either, as I wasn't involved from the beginning and that wouldn't have been fair coming from me. 

It would be a good idea to raise this as part of a different issue and besides that, we should take into account the fact that the header "jumps" on the index due to the "Start exploring" right upper button. I'm pretty sure that wasn't on purpose and we should address that too maybe.

bazzadp: Raised #574 . Let's resolve this for this PR.

src/static/css/index.css
catalinred: That is correct, thanks for the heads up. 

`20` is the value I was aiming at, see the existing comment for the responsive nav: `/* overcome the parent container's 20px side padding */  `. I guess I got pretty tired at the end. Fixed it.

src/templates/en/2019/base.html
catalinred: TBH lots of things are a bit different compared to the original design, and that's not necessarily a bad thing, because this is what it happens usually when you implement a design :)

Yes, in my opinion, the footer looks and works as intended. Thanks!

src/static/css/2019.css
catalinred: Having set 1280px for the `.container`, the total "remaining" available space for the content is 1240px and not 1160px.

bazzadp:
@HTTPArchive/developers anyone free to review this change? If not will merge this PR this weekend.

PR #579, opened 2019-12-02
Single-line paragraphs for en Methodology template

borisschapira:
Use single-line paragraphs to simplify translations (the original text and the translated text may be on the same line, facilitating discussions).


bazzadp:
LGTM. Thanks for doing this and good idea to do as a separate PR rather than part of the French one.

PR #584, opened 2019-12-04
FR - "’" instead of "'" (French microtypography)

borisschapira:
@nico3333fr made me realize an issue with template currently PR'ed. The issue is already in merged FR templates, so here's a fix.

### DIFF
--- a/src/templates/fr/2019/base.html
+++ b/src/templates/fr/2019/base.html
@@ -22,7 +22,7 @@
       <div class="cta">
         {% if request.endpoint == 'home' %}
           <a href="{{ url_for('table_of_contents', year=year, lang=lang) }}" class="alt btn">
-            Commencer l'exploration
+            Commencer l’exploration
           </a>
         {% endif %}
         {{ language_switcher() }}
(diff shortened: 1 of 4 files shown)


bazzadp:
We’ve avoided uses smart quotes in the English versions.

Not sure what the reason for that is @rviscomi - was it just for consistency?

And not sure if that still applies for the translated versions where it might be more important? I’ll be honest and don’t fully get what microtypography is in French so maybe it is more important there so should allow it?

borisschapira:
I'm not sure it's more important in French that it is in French, in the case of quotes. For other rules, it is, because of the spaces before some symbols that can cause orphan signs on some lines.

If you don't want smart quotes, I'll add this to the list of rules to be respected for consistent writing throughout the French corpus.

bazzadp:
> I'm not sure it's more important in French that it is in French, in the case of quotes. For other rules, it is, because of the spaces before some symbols that can cause orphan signs on some lines.

Thanks for explaining. Good to know and happy to approve this PR based on that. Will merge later.

> If you don't want smart quotes, I'll add this to the list of rules to be respected for consistent writing throughout the French corpus, and make another PR to remove the ones that were contributed before.

I’m not sure to be honest. It does introduce a potential consistency issue (which I suspect is why we brought it it) so may be for the best. Let’s see if @rviscomi sees this and can give more background and/or an opinion.

PR #592, opened 2019-12-10
Removes framework usage section

housseindjirdeh:
A number of these frameworks aren't being detected 100% correctly and many missed detections are skewing the results.

I'm currently working on making sure the detections in place are accurate so will bring this section back once that's complete.


rviscomi:
Could you elaborate on which detections are incorrect? We should also open issues against the Wappalyzer repo to get this fixed upstream.

I'm hesitant to cut out content before understanding what's wrong.

housseindjirdeh:
@bazzadp thanks, updated 🚀

And definitely not critical! This can easily go in after the change freeze is done :)

housseindjirdeh:
@rviscomi Most detections in wappalyzer (React, Vue, etc...) rely on global vars which mean bundled apps that don't expose them aren't detected correctly. There's an [open issue](https://github.com/AliasIO/Wappalyzer/issues/2450) for this.

Switching wappalyzer's detection model to relying on DOM checks (via a Treewalker for example) would probably take some time which is why I've begun relying on the [Library Detector](https://github.com/johnmichel/Library-Detector-for-Chrome/) extension since we have admin access to that. 

I can definitely push for updating how frameworks get detected in Wappalyzer and hopefully we can see that movement there. In the meantime, would you be okay if we rely on querying against the library detector extension for the results in this chapter?

src/content/en/2019/javascript.md
```diff
@@ -287,26 +287,9 @@ There are a number of possible reasons:
...
-- A component-based architecture
-
-Although there has been a shift towards a component-based model, many older frameworks that follow the MVC paradigm ([AngularJS](https://angularjs.org/), [Backbone.js](https://backbonejs.org/), [Ember](https://emberjs.com/)) are still being used in thousands of pages. However, [React](https://reactjs.org/), [Vue](https://vuejs.org/) and [Angular](https://angular.io/) are the most popular component-based frameworks ([Zone.js](https://github.com/angular/zone.js) is a package that is now part of Angular core).
-
-Although this analysis is interesting, it's important to keep in mind that these results rely on a third-party detection library - [Wappalyzer](https://www.wappalyzer.com). All these usage numbers depend on the accuracy of each [detection mechanism](https://github.com/AliasIO/Wappalyzer/blob/master/src/apps.json) in place.
+Note: A previous version of this chapter discussed the usage of frameworks in this section. However, we are investigating queries around the accuracy of these statistics and have decided to remove this section temporarily while we complete this review.
```
bazzadp:
```suggestion
<aside class="note">Note: A previous version of this chapter discussed the usage of frameworks in this section. However, we are investigating queries around the accuracy of these statistics and have decided to remove this section temporarily while we complete this review.</aside>

<!-- Empty figure to keep figure numbers the same
<figure>
</figure>
-->
```

housseindjirdeh:
@bazzadp thank you! And good catch on keeping the figure numbers consistent.

@rviscomi Will also defer to you. If you would still like to keep the current text instead of removing it altogether, then that's fine! This section seems to be the most popular section on the chapter hence my hesitation to keep things as they are :)

rviscomi:
I'd propose the following:

1. Add a paragraph to the [Wappalyzer](https://almanac.httparchive.org/en/2019/methodology#wappalyzer) section of the Methodology describing the limitations of library detection.
2. Add a note to the JS chapter linking to the Wappalyzer section of the Methodology and summarizing how the limitations may affect our analysis.
3. Continue to push for a solution in https://github.com/AliasIO/Wappalyzer/issues/2450

As I mentioned to @housseindjirdeh in a DM: "We do our best to report on the state of the web given the tools available to us. If they have blind spots, we should note them but not necessarily discount their findings."

housseindjirdeh:
Sounds good @rviscomi! Added a bolder note at the beginning of the section to emphasize these results aren't entirely accurate

CC @StephenFluin

src/content/en/2019/javascript.md
```diff
@@ -289,24 +289,24 @@ Other top used JavaScript libraries include jQuery variants (jQuery migrate, jQu
 
 ### Frameworks and UI libraries
 
+<aside class="note">Although interesting, the results in this section are not accurate. As mentioned in our <a href="./methodology#wappalyzer">methodology</a>, the third-party detection library used in HTTP Archive (Wappalyzer) has a number of limitations with regards to how it detects certain tools. For JavaScript libraries and frameworks, the limitations are described in more detail <a href="https://github.com/AliasIO/wappalyzer/issues/2450">here.</a></aside>
```
rviscomi: Rather than suggesting that the results are inaccurate, which I would dispute, I think all that needs to be acknowledged is that there's a known limitation or blind spot in the Wappalyzer library and these numbers should be taken as a lower bound.

All of our tools have blind spots: HA only measures home pages, CrUX excludes SPAs, Lighthouse only measures on mobile, WebPageTest only runs in Chrome, Wappalyzer only looks at global JS, etc. We acknowledge these limitations on the Methodology page but we haven't dissuaded readers from trusting the results.

housseindjirdeh: I decided to mention that because for this specific scenario, the blind spot may actually be quite large (which can affect the end result of "which framework is more popular than the rest").

But good point, there are many limitations with the tools used in the entire almanac. Removed the entire sentence 👍.

src/content/en/2019/javascript.md
```diff
@@ -289,24 +289,24 @@ Other top used JavaScript libraries include jQuery variants (jQuery migrate, jQu
 
 ### Frameworks and UI libraries
 
+<aside class="note">As mentioned in our <a href="./methodology#wappalyzer">methodology</a>, the third-party detection library used in HTTP Archive (Wappalyzer) has a number of limitations with regards to how it detects certain tools. For JavaScript libraries and frameworks, the limitations are described in more detail <a href="https://github.com/AliasIO/wappalyzer/issues/2450">here.</a></aside>
```
bazzadp: Nit: Can we change links to Markdown?

Also can we have a better link than "here" (plus full stop was in wrong place).

What about something like this?:

```suggestion
<aside class="note">As mentioned in our (methodology)[./methodology#wappalyzer"], the third-party detection library used in HTTP Archive (Wappalyzer) has a number of limitations with regards to how it detects certain tools. There is an open issue [to improve detection of JavaScript libraries and frameworks](https://github.com/AliasIO/wappalyzer/issues/2450), which will have impacted the results presented here.</aside>
```

bazzadp:
Ah too late. Ah well never mind.

rviscomi:
Sorry. Let's fix in another PR.

housseindjirdeh:
Thanks @bazzadp, PR #624 to update the second link 

> Nit: Can we change links to Markdown?

I don't think markdown links are supported within HTML elements. Also noticed that there are some chapters that are doing this and aren't [rendering links properly](https://almanac.httparchive.org/en/2019/ecommerce#popular-image-formats).

I can try to update all the broken links in the almanac when I get the chance 🙏

bazzadp:
> > Nit: Can we change links to Markdown?
> 
> I don't think markdown links are supported within HTML elements. Also noticed that there are some chapters that are doing this and aren't [rendering links properly](https://almanac.httparchive.org/en/2019/ecommerce#popular-image-formats).

Oh good point! Forgot I asked you to put is in an `<aside>`. Thanks for all the changes - that other one's merged now.

> I can try to update all the broken links in the almanac when I get the chance 🙏

Oh oh. Going to have to do a grep for this and correct ASAP before my OCD explodes!!! 😊

bazzadp:
> > I can try to update all the broken links in the almanac when I get the chance 🙏
> 
> Oh oh. Going to have to do a grep for this and correct ASAP before my OCD explodes!!! 😊

Did a grep and that's the only one I can find wrong so submitted #627 for this.

PR #598, opened 2019-12-23
Show translators even if there is only one

borisschapira:


bazzadp:
Thanks! One of the [two hardest things in computer science](https://twitter.com/codinghorror/status/506010907021828096?lang=en) 😞

PR #600, opened 2019-12-23
Encode Unicode bullet character in before content

ibnesayeed:
Fixes the issue illustrated in the attachment below:

unicode-bullet-in-css-content

### DIFF
--- a/src/static/css/page.css
+++ b/src/static/css/page.css
@@ -118,7 +118,7 @@
 }
 .content ul li::before {
   position: absolute;
-  content: "•";
+  content: "\2022";
   font-size: 16px;
   color: #a9cbbb;
   display: inline-block;


bazzadp:
Thanks! Agree we should do this (and probably should also change content-type to `text/css; charset=utf-8`).

ibnesayeed:
> and probably should also change `content-type` to `text/css; charset=utf-8`

I don't think we have control over the HTTP headers returned by GitHub Pages. That said, while we can add `charset` attribute in the `<link rel="stylesheet" charset="utf-8">` to act as a fallback, it is actually deprecated in HTML5. I think, it would be safer and more portable to encode Unicode characters that are outside of the ASCII range.

bazzadp:
Agreed a have merged this. Will include your change in next release in new year. Just saying no harm to also do that change as well if possible for when served from live site (not GitHub).

PR #603, opened 2019-12-27
Use appropriate chapter hero images

### LINKED ISSUE #141: Optimize and serve chapter/page illustrations
The raw illustration JPEGs are 2000 by 1000 pixels and up to 7 MB in size.

- [x] upload raw illustrations
- [x] figure out which image dimensions we'll actually need
- [x] generate optimized versions for each size
- [x] write descriptive alt text for non-visual users
- [ ] write the markup and styles to serve the optimized version depending on the viewport size and capabilities
- [x] set the chapter illustration to be the default social media cover image

cc @HTTPArchive/developers

bazzadp:
Closes #141 

Implement `<Picture>` tag to load large or small hero images for chapters depending on viewport width, pixel density and webp support.


src/templates/en/2019/chapter.html
```diff
@@ -115,7 +115,16 @@ <h1 class="title title-lg">
                 <span class="chapter-unedited">[Unedited]</span>
                 {% endif %}
             </h1>
-            <img src="/static/images/{{ year }}/{{ get_chapter_image_dir(metadata) }}/hero_lg.jpg" class="content-banner" alt="" loading="eager" />
+            <!-- Show large image for large screens and high density screens and use webp when supported -->
+            <picture>
+              <source media="(min-width: 327px)" type="image/webp" srcset="/static/images/{{ year }}/{{ get_chapter_image_dir(metadata) }}/hero_lg.webp" />
```
mikegeyser: Considering this is already in the `2019` directory, do we need to be dynamic about the year?

We also have several duplicate uses of `get_chapter_image_dir(metadata)`, could we not simplify this by setting a variable?

Sort of like: `{% set chapter_image_dir =  "/static/images/2019/%s % (get_chapter_image_dir(metadata))}`

And then you can simplify all of the markup below like: 
` <source media="(min-width: 327px)" type="image/jpeg" srcset="{{ chapter_image_dir }}/hero_lg.jpg" />`

What do you think? It's not important, but might be a little neater (especially considering that it's a template for every generated chapter).

bazzadp: Good idea. Pushed a commit with that. Can you re-review?

PR #604, opened 2019-12-28
Home page Hero Image fix

### LINKED ISSUE #361: Only load mobile and desktop images appropriately
Currently we have things like this:

```
<img class="intro-image-2019" src="/static/images/home-hero-2019.png" alt="The Web Almanac 2019 edition">
<img class="intro-image" src="/static/images/home-hero.png" alt="The Web Almanac 2019 edition">
```

And then we show and hide the appropriate image on desktop and mobile using CSS.

However this causes both images to be load, even though one will not be used.

This can be mitigated by using background-images for these images and hiding the parent element appropriately. This causes the browser to realise whether the image will be displayed or not after evaluating the CSS and not load it unless it's needed.

So I think we should change this to something like this:

```
<span class="desktop-image">
    <span class="intro-image-2019" style="background-image: '/static/images/home-hero-2019.png'">
</span>
<span class="mobile-image">
    <span class="intro-image" style="background-image: '/static/images/home-hero.png'" alt="The Web Almanac 2019>
</span>
```
(though with the `background-image` set in CSS and not inline like this which is just done for illustrative purposes).

And then set the `display:none` option on the `desktop-image` and `mobile-image` styles as appropriate.

Alternatively a potentially better, more accessible option is to use the `picture` and `srcset` elements, but I'm less familair with them to be honest...

See here for more info: http://designingforperformance.com/responsive-web-design/#images

bazzadp:
Stop loading both desktop and mobile image on home page and only load appropriate image.

Fixes #361 

I went with the wrapper `display:none` of a background-image fix rather than `<picture>` or `` method because [they no not allow you to hide an image based on media queries](https://github.com/ResponsiveImagesCG/picture-element/issues/243) which means you can't have two fall backs. And since some clients (e.g. IE11, Android 4.4) [do not support `<picture>`](https://caniuse.com/#feat=picture), I thought best to implement the other way.

### DIFF
--- a/src/static/css/index.css
+++ b/src/static/css/index.css
@@ -98,32 +98,28 @@ p {
 .intro-image-wrapper {
   grid-area: image;
   display: none;
-  width: 100%;
-  padding-bottom: 78.6%;
-  position: relative;
+  width:100%;
 }
 
 .intro-image {
   background-image: url(/static/images/home-hero.png);
   background-size: cover;
-  position: absolute;
-  top: 0; bottom: 0; left: 0; right: 0;
+  height: 67.5vw;
+  width: 86vw;
 }
 
 .intro-image-2019-wrapper {
   grid-area: image;
-  max-height: 470px;
-  max-width: 100%;
-  height: auto;
-  padding-bottom: 103%;
-  margin-bottom: -90px;
+  width:100%;
 }
 
 .intro-image-2019 {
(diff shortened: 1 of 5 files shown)


bazzadp:
> The thing is that right now, these two images switched on small and large resolutions are almost identical. So hopefully, in the future, we'll be able to improve that by merging them into a unique image.

Good call!

> Happy new year! :)

And to you!

PR #606, opened 2019-12-28
Remove flags from language menu

### LINKED ISSUE #565: Get a source of country flags
What's the source of the country flag pngs? Or were they created from scratch?

For example: https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/static/images/es.png

Need a French one and am sure that won't be the last!

bazzadp:
Closes #565 

Note I've also fixed the translation menu location in the footer as part of this and avoided some double divs!


borisschapira:
I'm really not competent in art direction and/or US flag, so I'd rather not comment on this.

AymenLoukil:
Hi!
Adding Countries flags is not coherent since we are not targeting specific regions. All the contents on Web Almanac are targeting language only and not language-country.

bazzadp:
> Adding Countries flags is not coherent since we are not targeting specific regions. All the contents on Web Almanac are targeting language only and not language-country.

@AymenLoukil let’s take [that discussion to this issue](https://github.com/HTTPArchive/almanac.httparchive.org/issues/565#issuecomment-569424076) and not make this PR dependent on resolving that since a flag is part of the current design.

rviscomi:
I'd suggest we close this PR per my comment in https://github.com/HTTPArchive/almanac.httparchive.org/issues/565#issuecomment-571279247

bazzadp:
> I'd suggest we close this PR per my comment in [#565 (comment)](https://github.com/HTTPArchive/almanac.httparchive.org/issues/565#issuecomment-571279247)

Leave it open and I'll update it to remove flags instead. Almost done on that now...

bazzadp:
Should be good to review and merge now. Updated title and initial comment too.

Issue #423, opened 2019-11-10
Cache TTL for static content is too low


paulcalvano:
All of the static content is served with a low cache TTL:

`cache-control: public, max-age=600`

Should we increase this to something higher before this goes live?  It's kind of silly for the caching chapter to fail so badly on caching :) 

<img width="962" alt="image" src="https://user-images.githubusercontent.com/7459458/68546633-7d4eeb80-03a6-11ea-9616-e4af1bae93ea.png">

<img width="671" alt="image" src="https://user-images.githubusercontent.com/7459458/68546635-85a72680-03a6-11ea-8463-b7abd69484f9.png">

paulcalvano:
Nevermind - just realized it's already being tracked here https://github.com/HTTPArchive/almanac.httparchive.org/issues/413

PR #608, opened 2019-12-29
Add better default caching

### LINKED ISSUE #413: Audit the website for performance issues

bazzadp:
Closes #413 

This sets the following cache control settings:
- Fonts - 1 year (currently not served locally but may be added by #607 )
- Other static files (JS, CSS, SVG, JPEG, PNG) - 3 hours
- Dynamic HTML - 3 hours

Note the static files need to be set in two different places depending whether they are served by Google App Engine (production) or Python/Flask (for local development) and I set in both for consistency. And I've not been able to test the App Engine changes so will need to double check it when deployed.

Background:

In #423 @paulcalvano noted the Google App engine default of 600 seconds (10 mins) for static content and no caching for dynamic content was too low but we didn't really agree what the higher limit should be.

Some advocate for extremely large times (e.g. 1 year) but this depends on immutable resources, which we don't currently have as we don't currently set a unique hash value in our static resource file names. The longer we make it, the more risk of showing broken content or not being in control to change it. I'm also sceptical as to the benefit of longer term caching to be honest: unless you are a very popular site visited frequently by the users, I'd imagine most users won't have your assets cached between sessions and caching is really for within a session when the user is browsing about the site so short term caching is actually fine in my opinion. However I agree 10 mins is too short - as some of our articles could easily take over 10 mins to read.

Therefore I've gone with a relatively small 3 hours for most resources. So if someone is browsing around the site then they should not have to reload caching - unless here for an extremely long time, in which case they might take a small hit every 3 hours! I could even be persuaded to move this to 1 hour. I'm also a fan of having a bit of caching for HTML since it shouldn't change much either and it makes browsing around the site a lot quicker.

This should also move us from the frankly embarrassing F we get in the caching when you run webpagetest on our site:

image

However will likely still give us a lower ranking (my own blog follows the same rules and gets a D rating). Saying that, it does set us up for how to set this in future if we do want to make longer caching, for example if we move to immutable static resources with unique filenames.

Thoughts on this strategy - and on the PR in general - greatly appreciated!

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -7,12 +7,35 @@ import logging
 import random
 from validate import validate, SUPPORTED_YEARS, DEFAULT_YEAR
 
-app = Flask(__name__)
+# Set WOFF and WOFF2 caching to return 1 year as they should never change
+# Note this requires similar set up in app.yaml for Google App Engine
+class MyFlask(Flask):
+    def get_send_file_max_age(self, name):
+        if name.lower().endswith('.woff') or name.lower().endswith('.woff2'):
+            return 31536000
+        return Flask.get_send_file_max_age(self, name)
+
+app = MyFlask(__name__)
+# Cache static resources for 10800 secs (3 hrs) with SEND_FILE_MAX_AGE_DEFAULT.
+# Flask default if not set is 12 hours but we want to match app.yaml
+# which is used by Google App Engine as it serves static files directly
+app.config['SEND_FILE_MAX_AGE_DEFAULT'] = 10800
 Talisman(app,
          content_security_policy=csp,
          content_security_policy_nonce_in=['script-src'])
 logging.basicConfig(level=logging.DEBUG)
 
+@app.after_request
+def add_header(response):
+    # Cache responses for 3 hours if no other Cache-Control header set
+    # This is used for the dynamically generated files (e.g. the HTML)
+    # (currently don't use unique filenames so cannot use long caches and
+    # some say they are overrated anyway as caches smaller than we think).
+    # Note this IS used by Google App Engine as dynamic content.
+    if 'Cache-Control' not in response.headers:
+        response.cache_control.public = True
+        response.cache_control.max_age = 10800
+    return response
 
 def render_template(template, *args, **kwargs):
     year = request.view_args.get('year', DEFAULT_YEAR)
@@ -104,7 +127,6 @@ def sitemap():
 @app.route('/<lang>/<year>/<chapter>')
 @validate
 def chapter(lang, year, chapter):
-    # TODO: Validate the chapter.
     config = config_util.get_config(year)
     (prev_chapter, next_chapter) = get_chapter_nextprev(config, chapter)
     return render_template('%s/%s/chapters/%s.html' % (lang, year, chapter), config=config, prev_chapter=prev_chapter, next_chapter=next_chapter)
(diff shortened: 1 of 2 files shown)


src/main.py
```diff
@@ -7,12 +7,35 @@
 import random
 from validate import validate, SUPPORTED_YEARS, DEFAULT_YEAR
 
-app = Flask(__name__)
+# Set WOFF and WOFF2 caching to return 1 year as they should never change
+# Note this requires similar set up in app.yaml for Google App Engine
+class MyFlask(Flask):
```
mikegeyser: We could maybe move the `MyFlask` class into a separate file and import it here? Is there no way to do the `app.config['SEND_FILE_MAX_AGE_DEFAULT'] = 10800` within the custom flask class, rather than external to it? (Just asking, I'm not sure.) 

Otherwise, it looks like a good change.

bazzadp: Not sure of benefit of externalising MyFlask, when it's only there to override one function? Would external files for such a simple use case not just cause more readability issues than help?

To be honest I'd rather not have this MyFlask class at all, but was the only way I could see to override the `get_send_file_max_age` function to allow per extension defaults. Presume there's no other way but to instantiate another class (never used Python before tbh!)?

And, it's maybe even overkill to have this as it's not strictly necessary, as could just live with the 3 hours for fonts too in dev, but found it confusing already to have different values in production (app.yaml) and dev (main.py) so decided to add this here to avoid that confusion (somewhat!) in future.

I tried again to set that config on the custom class but couldn't. Open to suggestions here but gain not sure what gains there are from doing that? Isn't the current way of doing it with `app.config` similar to how CSP is set on the `app` instance in the following lines?

```python
app.config['SEND_FILE_MAX_AGE_DEFAULT'] = 10800
Talisman(app,
         content_security_policy=csp,
         content_security_policy_nonce_in=['script-src'])
```

mikegeyser: Ok cool. I think consistency between dev and prod is important, so we should do it.

The rest was just a thought. With any MVC-styled app (regardless of language), I just try and keep things as tidy as possible. Single file, single responsibility as far as is reasonable (but I agree there can be diminishing value for such small things). I would personally prefer to separate any 'bootstrapping' concerns such as the CSP from the controller logic, but again - it may be overkill in this case.

Otherwise, it's a good change, feel free to resolve and merge. :)

bazzadp: OK will leave as is for now and merge. You can do separate PR to tidy up both this, and CSP (and anything else) if you want.

PR #611, opened 2019-12-30
Fix typo in en 2019 Javascript, 'cam' to 'can'

danstarns:


bazzadp:
Thanks @danstarns ! Will publish in next release in the New Year.

Btw if you run `npm run generate` from the src directory and it will generate the HTML version of the markdown too. But I took care of this for this one in #613

PR #615, opened 2020-01-03
Using the safe filter to include content that may contain HTML

borisschapira:


src/templates/en/2019/index.html
```diff
@@ -93,7 +93,7 @@
...
+  {%- set featured_chapter_quote = "The open web is vast, linkable, and interoperable by design. The ability to grab someone else’s complex library and use it on your site with a single <code>&lt;link&gt;</code> o&gt;&gt;<code>&lt;script&gt;</code> element has supercharged developers’ productivity and enabled awesome new web experiences; on the flip side, the immense popularity of a select few third-party providers raises important performance and privacy concerns. This chapter examines the prevalence and impact of third-party code on the web in 2019, the web usage patterns that lead to the popularity of third-party solutions, and potential repercussions for the future of web performance and privacy." %}
```
bazzadp:
```suggestion
  {%- set featured_chapter_quote = "The open web is vast, linkable, and interoperable by design. The ability to grab someone else’s complex library and use it on your site with a single <code>&lt;link&gt;</code> or <code>&lt;script&gt;</code> element has supercharged developers’ productivity and enabled awesome new web experiences; on the flip side, the immense popularity of a select few third-party providers raises important performance and privacy concerns. This chapter examines the prevalence and impact of third-party code on the web in 2019, the web usage patterns that lead to the popularity of third-party solutions, and potential repercussions for the future of web performance and privacy." %}
```

Looks like you made a typo here?

borisschapira: Damn!

borisschapira:
> On that point should we use nbsp in English versions as well to prevent % going to second line?

Without a space between the number and the `%`, it's never going to happen, so no, I don't think so.

bazzadp:
OK so confused why it’s needed in the French version then? I thought main reason was to avoid an unwanted line break? Or is there another reason?

bazzadp:
And btw you’re happy with the change from:

> Lire le chapitre sur les Tierces Parties

to:

> Lire le chapitre: Tierces Parties

?

Trying to make this generic so it works for all chapter names to removed “sur les”.

Also removed the “(3P)” from heading bane for same reason.

borisschapira:
> OK so confused why it’s needed in the French version then? I thought main reason was to avoid an unwanted line break? Or is there another reason?

In french you have a space between a number and what follows (symbol, measurement unit…), so orphan `%` is a possibility, that's why we use non-break spaces to avoid this. No space, no risk.

bazzadp:
> > OK so confused why it’s needed in the French version then? I thought main reason was to avoid an unwanted line break? Or is there another reason?
> 
> In french you have a space between a number and what follows (symbol, measurement unit…), so orphan `%` is a possibility, that's why we use non-break spaces to avoid this. No space, no risk.

Ah gotcha! Missed that.

PR #623, opened 2020-01-09
Making sure that the chapter generation works on windows

### LINKED ISSUE #622: Generating chapters does not work on Windows
Currently line 26 of generate_chapter.js to the following:

```
    const re = /content\/(.*)\/(.*)\/(.*).md/;
```
which only works on linux/mac based systems.

For windows this needs to be:

```
    const re = /content\\(.*)\\(.*)\\(.*).md/;
```

Ideally we should auto detect this and use the appropriate regex.

mikegeyser:
Fixes #622

NB: I don't have access to a windows machine at the moment, could someone please ensure that the regex works as expected?


bazzadp:
LGTM - and tested on windows machine - thanks! @mikegeyser !

PR #636, opened 2020-01-15
Add a "Join the team" card

### LINKED ISSUE #548: Add Interest Form link to Contributors page

catalinred:
This fixes #548

The "Join the team!" card:
- is always in the first position
- is a hoverable and actionable link
- The "Join the team!" is a `<text>` within an inline SVG so it can be easily translated

P.S. There's [a single commit](https://github.com/HTTPArchive/almanac.httparchive.org/pull/636/commits/baf86bbb5e37819ddb9761656801deab52515147) involved here, but I guess I need to figure out how to merge/squash all these already-on-master commits...

image


src/templates/en/2019/contributors.html
```diff
@@ -315,6 +331,15 @@ <h1 class="title title-lg">
     </div>
 
     <div class="contributor-grid">
+      <a href="https://forms.gle/Qyf3q5pKgdH1cBhq5" class="contributor contributor-join analysts authors brainstormers designers developers editors reviewers translators">
```
rviscomi: By listing every team name on this card it will be included in `getFilteredContributorCount` and we'd be showing N+1 contributors per team.

One way around this is to remove the team names and `contributor` class name from this element and add the `.contributor-join` selector to the `.contributor` style rule on line 243.

```suggestion
      <a href="https://forms.gle/Qyf3q5pKgdH1cBhq5" class="contributor-join">
```

```
  .contributor,
  .contributor-join {
    flex: 1;
    ...
}
```

src/templates/en/2019/contributors.html
```diff
@@ -315,6 +331,15 @@ <h1 class="title title-lg">
...
+        <svg xmlns="http://www.w3.org/2000/svg" width="174" height="100" viewBox="0 0 203 117">
+          <path fill="#5e687c" fill-rule="evenodd" d="M0 0l11 89 71 4.4L65.9 117l40.7-22.4L201 89l2-79z" clip-rule="evenodd"/>
+          <text x="50%" y="45%" fill="#fff" font-weight="bold" text-anchor="middle">
```
rviscomi: Can these styles be set in CSS?

catalinred: Good idea, done that!

src/templates/en/2019/contributors.html
catalinred: I made this change too, thanks for the heads up! I didn't notice the n+1 count before.

src/templates/en/2019/contributors.html
```diff
@@ -315,6 +349,15 @@ <h1 class="title title-lg">
     </div>
 
     <div class="contributor-grid">
+      <a href="https://forms.gle/Qyf3q5pKgdH1cBhq5" class="contributor-join">
+        <svg xmlns="http://www.w3.org/2000/svg" width="174" height="100" viewBox="0 0 203 117">
+          <path d="M0 0l11 89 71 4.4L65.9 117l40.7-22.4L201 89l2-79z"/>
+          <text x="50%" y="47%">
+            Join the team!
```
rviscomi: One last question. Is this accessible for assistive technology users? For example, does it know to read this text to describe the anchor? If not we'll need to add some aria attributes.

catalinred: `aria-label="Join the Web Almanac team"` for the anchor or `title="Join the Web Almanac team"` for the inline SVG?  🤔

rviscomi: Let's label the anchor. 👍

PR #550, opened 2019-11-27
FR - Markup chapter

SilentJMA:
Part of #539

PR #637, opened 2020-01-15
Fix blank "Translated By" lines

### LINKED ISSUE #635: Remove empty "Translated by" in chapters

bazzadp:
Fixes #635 

Dunno how I missed that 😞


rviscomi:
So it's an array with one item, an empty string? Could we fix this by changing the yaml parser to produce an empty array as expected instead? Or should the translators property be omitted from configs in which there are no translators?

bazzadp:
> So it's an array with one item, an empty string?

Yes [it would appear so](https://github.com/HTTPArchive/almanac.httparchive.org/blob/cc0abdbf85c8916c445e2e32e69d879876ed38b8/src/templates/en/2019/chapters/accessibility.html#L13): 

```
{% set metadata = {..."reviewers":["ljme"],"translators":[""],"discuss":"1764... %}
```

> Could we fix this by changing the yaml parser to produce an empty array as expected instead?

Potentially. Looks like we'd need to change the `parse_array` function to cater for this. But I think this function is the easier, simpler fix to be honest.

> Or should the translators property be omitted from configs in which there are no translators?

No. This is a placeholder specifically added in #550 to avoid gitlocalise issues and to make it easier for translators to know where to add their names (as they should all be translated from English).

bazzadp:
> > Could we fix this by changing the yaml parser to produce an empty array as expected instead?
> 
> Potentially. Looks like we'd need to change the `parse_array` function to cater for this. But I think this function is the easier, simpler fix to be honest.

Was a fairly easy fix after all so did it that way.

rviscomi:
Great thanks for looking into that

PR #642, opened 2020-01-17
Only compress images if including images in PR

bazzadp:
Currently we run the calibre app every time we commit - even if no images to compress. Bit of a waste so let's limit it to only when we include image files.


JMPerez:
Well done, @bazzadp!

PR #645, opened 2020-01-19
Large batch of Accessibility and HTML improvements

bazzadp:
Apologies in advance for the size of this, definitely should have split it up more in hindsight but all related to Accessibility and HTML validation so thought I'd put them all in one PR. Let me know if you'd really prefer me to split out to make review easy (wouldn't blame you!).

Was inspired by [this Tweet](https://twitter.com/SassCSS/status/1214345728580972544?s=20) and what they did in that, to have another look at the accessibility on our site and think it leaves a lot of be desired to be honest, despite what we did for #303 😞

This PR addresses most it what I spotted by making the following changes:
- Adds a [skiplink/skipnav](https://webaim.org/techniques/skipnav/)
- Makes both the main menu, and the article menus keyboard accessible
- Fixes some colour contrast issues
- Fixes some alt texts.
- Fixes some [accessibility issues with SVG images](https://css-tricks.com/accessible-svgs/)
- Fixes out of sequence tabbing in the footer with the HA logo (mostly).

It also includes some structural HTML changes, for better accessibility and HTML validation:
- Stops using the `<aside>` element [which has special meaning for accessibility](https://dequeuniversity.com/rules/axe/3.4/landmark-complementary-is-top-level?application=AxeChrome) and most of our asides are notes, rather than true asides.
- Uses the `<article>` element better - for the main article and not the article and the menu
- Adds a `<main>` element
- Fixes overuse of the `<section>` element when it's not really a section but a grouping of different things with no heading [ref: MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section), plus [ref: Bruce's recent article on this](https://www.smashingmagazine.com/2020/01/html5-article-section/)
- Moves the Author section title from a `<h4>` to `<h2>` but keeps the styling the same
- Adds some more heights and widths to images (more still to do)
- Fixes an HTML issue with the `<aside>` element being inside a `<p>` element
- Adds the join now to French contributors page - @borisschapira if you get a chance can you review the French template files for some translations I've done with the help of Google translate! And if you've any comments on the rest of it then happy to hear that too!

We already get 100% on Lighthouse for Accessibility, before this, but it's quite a low bar to be honest and we didn't score so well on some of the other tools:

- All pages, bar the index page, should now pass the [WebAim WAVE accessibility testing tool](https://wave.webaim.org/) - apart from some warnings about a few repeated links (that I think we should keep despite being annoying to screen readers). The index page however still has some colour contrast issues though as per https://github.com/HTTPArchive/almanac.httparchive.org/issues/303#issuecomment-569547149 we're pretty close and these are for non-body text items.

- This also solves most of the issues I spotted using the [Firefox Accessibility Inspector](https://hacks.mozilla.org/2019/10/auditing-for-accessibility-problems-with-firefox-developer-tools/) though the Google Sheets still causes issues (which we have no real control over unless we stop using them), it doesn't seem to recognise some of the keyboard events (looks like a bug in their tool to me), and again the colour contrast issues on home page.

- [AXE](https://chrome.google.com/webstore/detail/axe-web-accessibility-tes/lhdoppojpmngadmnindnejefpokejbdd) also is pretty much clean, though again it's complaining about colour contrast issues on home page. It would also like elements that scroll off screen (e.g. wide code blocks and tables) to be keyboard focusable so they can be scrolled with keyboard - which seems fair enough and a good idea - but when I do that other tools complain of a focusable element with no click action. Can't make both sides happy, so leaving as is for now with no focus. It's rare for items to scroll off screen in desktop size which I image most keyboard users will be in.

With this PR all pages now also pass the [W3 HTML validator](https://validator.w3.org/nu/) except for warnings about `loading` attribute (for lazy loading images and frames) that they don't support yet.


bazzadp:
Friendly ping!

@rviscomi any thoughts on this? Looks like a lot but if you ignore the chapter `<asides>`->`<p class="note">` and the generated chapter templates it becomes a lot more manageable to review.

@borisschapira any feedback on the French translations I did in the /fr/2019/templates/*.html files?

rviscomi:
Thanks for the ping. @bazzadp this is epic, THANK YOU for working on all of these improvements!

PR #664, opened 2020-02-02
Home page accessibility fixes

### LINKED ISSUE #663: Accessibility issues with the home page
So I've been continuing my obsession with Accessibility and decided to tackle the home page.

The home page has the following accessibility issues:

**Issue 1: The background image is too light, causing slight color contrast issues with the text particularly in the top left but also in the larger middle section**

Home page banner color contrast issues

***Recommendation**: remove the top left light-grey section and slightly darken the middle grey section (this is also more inline with the other pages which have a smaller, darker heading)*

home page darker banner

**Issue 2: There is no fallback if the image fails to load (e.g. if some of the SVG syntax is not understood or we're in high contrast mode).**

If this fails to load we end up with a transparent background, with white text and the text is unreadable. This is also why [WAVE highlights this](https://wave.webaim.org/report#/https://almanac.httparchive.org/en/), even if we fix the contrast:

> A CSS background color that provides sufficient contrast must be defined when a background image is in place. This provides fall-back contrast in case images or CSS do not display.

We can add some fallbacks so if CSS syntax is not recognised at all, but it won't help if the syntax is recognised but the SVG doesn't load or render for whatever reason (no SVG support, unrecognised SVG syntax, browser bug... whatever).

```css
header.alt-bg {
  background: #5C687D;
  background: transparent linear-gradient(#5C687D 70%, transparent 30%);
  /* background: transparent url(/static/images/intro-background-fit.svg) bottom left / 100% 100% no-repeat; */
  background: transparent url(data:image/svg+xml;base64,PHN...
```

On the one hand, the background image is inlined as Base 64 to the CSS so won't fail to load (unless the CSS also fails in which case text is not white!). I also tested it in Windows IE11 with high contrast mode and in that case it did remove the image, but also changed the background color anyway. So doesn't look to be an issue there. It's only WAVE that complains (AXE also complains but says it doesn't support background images so may be a false positive).

However if we add a background color behind the header div (using the `::after` pseudo header) and then set a white background on the SVG itself so this backup background color isn’t ever seen. WAVE and AXE don’t seem clever enough to recognise this but I’m happy with this as a solution.

***Recommendation**: Add some background fallbacks and a fallback background on a div with negative z-index.*

**Issue 3: The intro text is part of the Header not the Main content**

I think this is because it makes it easier to ensure the header background image covers this text.

The accessibility issue is where does the "main" part of the page start? In #645 when I added the skip link, I linked it to the intro paragraph even though its in the header. All in all it's a bit weird, though none of the tools complain about it.

Howeve,r in Safari reader mode it loses the intro H1 and text and thinks the featured chapter is the main content. We really should show an H1 and the intro text in reader mode. And I'm also using Safari reader mode as a proxy for how other non-browsers (e.g. AT tools) might render this.

***Recommendation**: Fix this by bringing the home page intro content out of base.html to index.html.*

**Issue 4: The other parts of the page (featured quote, contributors...etc.) are not part of a landmark element**

AXE complains about this, and it's related to Issue 3 above.

Answer is to wrap them in a `<main>` element, and include the H1 and intro paragraph in that too as per issue 3 above.

***Recommendation**: Wrap them in a `<main>` element and fix 3 above to solve the mismatch

**Issue 5: The background quote behind the featured chapter**

We have the following style in 2019.css:

```css
blockquote::before {
  content: '"';
```
This gives the very feint quote marks in the background used on home page (and also for a quote in the markup chapter):

Background quote marks example

Firefox accessibility tester sees this as text, and with low contrast, and unfortunately there is no way to mark aria attributes on pseudo-classes to tell it to ignore these.

We've a few options:

1. Change it to a span and set `aria-hidden="true"`. Seems to work but makes the HTML a little more verbose (`<blockquote><span class="background_quotes" aria-hidden="true"></span>The open web...`)

2. Remove the quote. It's probably so feint people aren't even noticing it to be honest.

3. Ignore it, it's not a real issue as we don't care that it has low contrast as just presentational.

4. Add [alt text to the pseudo element with the following syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/content) (which sets a blank alt text in this case):

```css
content: '"' / "";
```

***Recommendation**: Option 4 - add the alt text which now passes the Firefox accessibility tester now*

**Issue 6: Contributor count color contrast issue with yellow character**

Contributors section with large 89 number slightly obscured by characters

This is not a real issue. while there is a color contrast issue between white and yellow, the 89 figure is more than able to standout even with this, so I'm considering this a false positive. The characters also already have blank alt texts so are presentational. In fact so is the 89 number since it's on the text on the left.

***Recommendation**: Mark this whole number with `aria-hidden="true"` to prevent it from being seen by assistive technologies and remove the false positive.*

I'll submit a PR for these fixes but wanted to open the issue for discussion as well as there were a few ways to tackle some of these.

bazzadp:
Fixes #663


src/static/css/index.css
```diff
@@ -1,20 +1,26 @@
-header {
-  /* background-image: url(/static/images/intro-background-fit.svg); */
-  background-image: url(data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSIxOTIwIiBoZWlnaHQ9IjEyNTAiIHZpZXdCb3g9IjAgMCAxOTIwIDEyNTAiIHByZXNlcnZlQXNwZWN0UmF0aW89Im5vbmUiPjxkZWZzPjxzdHlsZT4uY2xzLTF7ZmlsbDojYThjYWJhO2ZpbGwtcnVsZTpldmVub2RkO29wYWNpdHk6LjA1fTwvc3R5bGU+PC9kZWZzPjxwYXRoIGNsYXNzPSJjbHMtMSIgZD0iTS0yMzEgODloMjA4OHY2NTFsLTIwODggMzYwVjg5eiIvPjxwYXRoIGNsYXNzPSJjbHMtMSIgZD0iTS00NzAgMjM5aDIwODh2NjUxbC0yMDg4IDM2MFYyMzl6Ii8+PHBhdGggZD0iTS0xNDQtMTloMjA4OHY2NTFMLTE0NCA5OTJWLTE5eiIgZmlsbD0iIzFhMmI0OSIgb3BhY2l0eT0iLjciIGZpbGwtcnVsZT0iZXZlbm9kZCIvPjxwYXRoIGNsYXNzPSJjbHMtMSIgZD0iTS00NzAtODNoMjA4OHY2NTFMLTQ3MCA5MjhWLTgzeiIvPjxwYXRoIGQ9Ik0tMjMyLTQ4aDIwODh2NjUxTC0yMzIgOTYzVi00OHoiIGZpbGw9IiNmMmYyZjIiIG9wYWNpdHk9Ii4wNSIgZmlsbC1ydWxlPSJldmVub2RkIi8+PHBhdGggZD0iTS0xNDI2LTM3MUg2NjJ2NjUxbC0yMDg4IDM2MFYtMzcxeiIgb3BhY2l0eT0iLjEiIGZpbGw9IiNmMmYyZjIiIGZpbGwtcnVsZT0iZXZlbm9kZCIvPjwvc3ZnPg==);
+header.alt-bg {
```
catalinred: Usually, when adding background fallbacks you add `background-color: ...`  to render a bg color until the HTTP request is made and proper bg image loads, next a  `background: ...` shorthand or the safer `background-image`.

I am not sure I understand the fallback order in this CSS rule but I am sure I might missing out bits here. The way I see and understand these lines, please correct me if I'm wrong:

`background-color: #5C687D;`
The classic bg color fallback.

`background: transparent linear-gradient(#5C687D 70%, transparent 30%);` 
This shorthand overrides the above completely and usually, the gradient is seen as a progressive enhancement when the browser has better support than the simple color background. I feel that's not really the case here, as we're not supporting old browsers at all, at this time (e.g. by using CSS grid and flexbox for the site's structure without any fallbacks).

`background: transparent url(/static/images/intro-background-fit.svg) bottom left / 100% 100% no-repeat;`
Isn't this completely overridden by the next background Base64 declaration? Also, I didn't checkout this branch yet, but, if I'm not mistaken, this allows an extra  `/static/images/intro-background-fit.svg`, HTTP request which we've been getting rid of it previously.

src/static/css/index.css
```diff
@@ -335,6 +332,12 @@ p {
 }
 
 @media (max-width: 900px) and (min-width: 601px) {
+
+  header.alt-bg {
+    padding-bottom: 1800px;
```
catalinred: Previously, we had a single "magic number" e.g. `400px`  which we used to accommodate the slanted SVG background header. As we already chatted, that was implemented in #492 and helped us getting rid of some hardcoded/magic numbers.

It seems that now we have `900px`, `1650px` and `1800px`.

src/static/images/intro-background-fit.svg
```diff
@@ -1,13 +1,12 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="1920" height="1250" viewBox="0 0 1920 1250" preserveAspectRatio="none">
+<svg xmlns="http://www.w3.org/2000/svg" width="1920" height="1250" viewBox="0 0 1920 1250" preserveAspectRatio="none" style="background:white;">
```
catalinred: For consistency,. maybe move the inline style within the inner SVG's style .e.g. `svg {background: #fff}`?

src/static/css/index.css
```diff
@@ -1,20 +1,26 @@
...
-  padding-bottom: 400px;
-}
-
-@media (max-width: 600px) {
-  header {
-    background-size: cover;
-  }
+  padding-bottom: 900px;
+  margin-bottom: -900px;
 }
```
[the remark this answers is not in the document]
bazzadp: My thinking is more if the SVG fails because it is not support. I'm ot sure if any Assistive technology (AT) outside of the mainsteam browsers do not render SVGs. Or if SVG backgrounds are ignored for any particular accessibilty settings (e.g. high contrast). See next comment too.

src/static/css/index.css
bazzadp: So my thinking here is:
`background-color: #5C687D;`
Fall back when all else fails. However it fills up too much of the screen and bleeds into the Featured Chapter section so not ideal. Which leads on to:
`background: transparent linear-gradient(#5C687D 70%, transparent 30%);`
A slightly nicer fall back which only covers 70% of the screen when `linear-gradient` is supported. It's not actually a gradient as it has hard cut offs.
Then the SVGs.

> Isn't this completely overridden by the next background Base64 declaration?

Yes. The URL is commented out so shouldnt cause a download. This is also how it exists as present so think you added it? I'm seeing it as a bit of hint/comment as to where that base64 is derviced from so if anyone (like me in this PR!) wants to change it, then they know were to get the source from.

src/static/css/index.css
bazzadp: Yes we do have more magic numbers in this one. Personally I think that's an acceptable trade off to fix the HTML.

As I mentioned in our chat, currently in Safari reader mode (on both desktop on MacOS and mobile in iOS) the home page does not render correctly and drops the intro text that I think is important for the home page.

I'm also using Safari reader mode as a bit of proxy for how other non-browsers will render the page.

src/static/css/index.css
bazzadp: > I feel that's not really the case here, as we're not supporting old browsers at all, at this time (e.g. by using CSS grid and flexbox for the site's structure without any fallbacks).

I think this is a big part of what I'm trying to solve here. It's find not supporting grid if the cotent is still readable. For example, IE11 disables the home page, but without laying it out properly and without the fonts but it's still readable. That's fine and the page is progressively enhanced for other browsers. We don't not support IE11 - we just have a better experience when using a better browser.

However if the SVG does not load (because AT doesn't support it, or you're in reader mode, or just due to a bug in a browsers SVG rendering engine) then the text is white on white and completely unreadable. Yes it may be unlikely but I don't know all the AT technologies out there, or any future bugs, and the fallbacks work, and are not hacky so why not have them?

So we should have progressive enhancement, but not cut off older browsers or other non-browsers if they don't support the code we use. That's the beauty of the web which degrades gracefully and this PR allows us to do this.

src/static/images/intro-background-fit.svg
bazzadp: Cool. Will have a look at this. Not overly familair with SVGs so great feedback!

bazzadp:
>Great work, Barry! I've added some comments here and there, let me know what you think!

Thanks. Replied to most of them. Hopefully explains my thinking a bit better. Let me know if that makes sense or if you've any more comments on then.

> Also, with this PR, I've noticed once again it's a bit difficult to keep all the multilingual templates markup in sync. In the future, it would be great if all the pages inherit the same parametrized template, as it's the same markup after all.

Yeah it's a bit annoying alright. We probably can do more here to generalise things. I'm honestly hoping to calm down these big PRs soon though... but I've promised that before and then found something else I've wanted to fix! 😉

src/static/css/index.css
bazzadp: Ah I see I forgot to comment this out again. Fixed now!

src/static/images/intro-background-fit.svg
bazzadp: Fixed!

PR #665, opened 2020-02-03
Add Accessibility Statement

### LINKED ISSUE #486: Add an accessibility statement
https://www.w3.org/WAI/planning/statements/

bazzadp:
Fixes #486 

Note will need to add to translator's issues list, once accepted.

I've said we're aiming for the top WCAG 2.1 AAA rating. Do we all agree to that? It's a tough bar for most sites to reach (see all the [AAA sections notes in the WCAG quick ref](https://www.w3.org/WAI/WCAG21/quickref/?currentsidebar=%23col_customize&levels=a%2Caa#top)) and may mean extra work to maintain that, but being a static site (and new) it's a good bit easier for us. It does also depend on #664 (since merged) and #666 and if we want to scale those back, then should probably drop back to AA aim, “with many pages reaching AAA standard.”


rviscomi:
I think we should commit to the most realistic rating rather than an aspirational one. Probably AA.

src/templates/ar/2019/base.html
```diff
@@ -65,6 +65,7 @@
         {{ language_switcher() }}
       </nav>
       <p class="copyright"><span>© 2019 Web Almanac. Licensed under <a href="https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/LICENSE">Apache 2.0</a>.</span></p>
```
rviscomi: WDYT of something like this instead?

```suggestion
      <p class="copyright"><span>© 2019 Web Almanac. Licensed under <a href="https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/LICENSE">Apache 2.0</a>.</span><br>
      Read our <a>Accessibility Statement</a>.</p>
```

src/templates/ar/2019/base.html
```diff
@@ -65,6 +65,7 @@
...
+      <p class="accessibility"><a href="{{ url_for('accessibility_statement', year=year, lang=lang) }}">Accessibility</a></p>
```
rviscomi: It's not clear what this link means. We have an accessibility chapter and this might cause some confusion with that. I proposed a change above to merge it with the copyright notice paragraph. That may also help simplify the CSS changes.

src/templates/en/2019/accessibility_statement.html
```diff
@@ -0,0 +1,203 @@
+{% extends "en/2019/base.html" %}
+
+{% block title %}Accessibility Statement | The Web Almanac by HTTP Archive{% endblock %}
+
+{% block description %}Describes the accessibilty goals of the HTTP Archive Web Almanac{% endblock %}
```
rviscomi:
```suggestion
{% block description %}Describes the accessibility goals of the Web Almanac{% endblock %}
```

src/templates/en/2019/accessibility_statement.html
```diff
@@ -0,0 +1,203 @@
...
+
+{% block meta %}
```
rviscomi: Do we need to redeclare all of this metadata or is the default sufficient?

src/templates/en/2019/accessibility_statement.html
```diff
@@ -0,0 +1,203 @@
...
+    </div>
+  </nav>
+
+  <article id="maincontent" class="content">
+    <h1 class="title title-lg">Accessibility Statement for the Web Almanac</h1>
+      
+        <h2 id="overview">Overview</h2>
+
+        <p>
+          The <a href="https://httparchive.org">HTTP Archive</a> Web Almanac is committed to ensuring digital accessibility for all people. We are continually improving the user experience for everyone and applying the relevant accessibility standards.
```
rviscomi:
```suggestion
          The Web Almanac is committed to ensuring digital accessibility for all people. We are continually improving the user experience for everyone and applying the relevant accessibility standards.
```

src/templates/en/2019/accessibility_statement.html
```diff
@@ -0,0 +1,203 @@
...
+        <h2 id="assessment-approach">Assessment Approach</h2>
+        
+        <p>
+          Self-evaluation: the content was evaluated by the Web Almanac team using their own expertise and a number of tools including the <a href="https://wave.webaim.org/">WAVE Accessibility Tool</a>, the <a href="https://hacks.mozilla.org/2019/10/auditing-for-accessibility-problems-with-firefox-developer-tools/">Firefox Accessibility Inspector</a> and the <a href="https://chrome.google.com/webstore/detail/axe-web-accessibility-tes/lhdoppojpmngadmnindnejefpokejbdd" rel="nofollow">AXE Chrome plugin</a>.
+        </p>
+      
+        <h2 id="feedback">Feedback</h2>
+
+        <p>
+          We welcome feedback on the accessibility of this website. The best way of doing this is by raising an issue on <a href="https://github.com/HTTPArchive/almanac.httparchive.org/issues">GitHub issues page</a>. We would also welcome pull requests for those that wish to contribute directly. The Web Almanac is a community based project, managed by volunteers, and so we cannot guarantee a timeframe for responding to feedback but will aim to respond within 5 working days if not sooner.
```
rviscomi:
```suggestion
          We welcome feedback on the accessibility of this website. The best ways to reach us would be to <a href="https://github.com/HTTPArchive/almanac.httparchive.org/issues">raise an issue on GitHub</a> or email us at <a href="mailto:team@httparchive.org">team@httparchive.org</a>. We would also welcome pull requests for those that wish to contribute directly. The Web Almanac is a community based project, managed by volunteers, and so we cannot guarantee a timeframe for responding to feedback but will respond as soon as possible.
```

src/main.py
```diff
@@ -115,6 +115,12 @@ def methodology(lang, year):
     return render_template('%s/%s/methodology.html' % (lang, year))
 
 
+@app.route('/<lang>/<year>/accessibility-statement')
```
rviscomi: Is it possible to route this in a way that only varies by language and not year? Does it make sense for this statement to be site-wide as opposed to year-specific?

src/templates/ar/2019/base.html
bazzadp: I originally had "Accessibility Statement" and this is what the W3C has.

However just "Accessibility" seems much more common as I discovered when I looked around to find the French and Spanish translation of this. For example [Apple](https://www.apple.com/), [Microsoft](https://www.microsoft.com/en-us/) and [your own, little known, company](https://about.google/?) 😀

rviscomi: Sure, but in the context of the Almanac, "Accessibility" is also the title of a chapter so I think it's ambiguous.

bazzadp: Fair point!

src/templates/en/2019/accessibility_statement.html
bazzadp: We don't have a default one, except a very light one in src/templates/base.html so I say we need this in line with all the other pages.

src/main.py
bazzadp: Ideally, but this is complicated as it means we'd lose the styling, navigation...etc.

I think ignore this for now and revisit when we look at 2020 and how that will interact with 2019 site - will it have the same styling? Different one? Will we have more generic pages?

Alternatively we could host this statement in our GitHub repository like we do our licence instead of directly on our site?

src/templates/ar/2019/base.html
bazzadp: I like it!

Fixed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/665/commits/1638e53c39d5a178076de0e81a8e981a6f29bd9e

src/templates/ar/2019/base.html
bazzadp: Fixed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/665/commits/1638e53c39d5a178076de0e81a8e981a6f29bd9e

src/main.py
rviscomi: Maybe change the route to exclude year and hardcode 2019 in the `render_template` call for now?

```suggestion
@app.route('/<lang>/accessibility-statement')
```

src/templates/en/2019/accessibility_statement.html
bazzadp: We should try to tackle this as part of #673 . Resolving this for now.

src/main.py
bazzadp: Fixed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/665/commits/b2cbdab0a95ef731ecfe9781d842c10263e6145c

bazzadp:
Made that generic now @rviscomi . Think that's addressed all the feedback so good to merge if you're happy?

PR #666, opened 2020-02-08
Make menus and nav more accessible

### LINKED ISSUE #668: Home page background color glitching
I think this is related to #664. Playing with the `header.alt-bg` style rule's `background` style seems to toggle this glitch.

It's very fickle: it appears on page load, disappears when I open dev tools at the bottom of the screen, and reappears if I move dev tools to a separate window. If I had to guess there's something weird going on related to these styles:

```css
padding-bottom: 900px;
margin-bottom: -900px;
```

image

bazzadp:
Makes the following changes:

- [Menus are now enclosed in `ul` and `li` elements](https://www.w3.org/WAI/tutorials/menus/structure/) (note [some have tried to argue against this](https://css-tricks.com/wrapup-of-navigation-in-lists/) but general consensus is this is what should be done)
- [Stops hiding the `nav` element on mobile when closed](https://www.a11ymatters.com/pattern/mobile-nav/#3--dont-hide-the-nav-element)
- Add `aria-expanded` tags and better `aria-labels`
- Remove opacity on menu as difficult to read on mobile
- Add escape key to close menu/ToC
- Add facility to click outside of menu to close menu
- Fixes some discrepancies between translated chapters
- Also fixes #668

A couple of notes on this:
- I did not use the `role="menu"` as [apparently that is wrong for navigation menus](https://www.smashingmagazine.com/2017/11/building-accessible-menu-systems/)  which are lists of links and should only be used for true application menus
- I bumped version on `2019.css` to `2019.css?v=2` as these are breaking changes.


rviscomi:
Not sure if it's this PR but the layout appears broken on the home page. Could you look into this? 
image

src/templates/ar/2019/base.html
```diff
@@ -3,79 +3,128 @@
...
+        </nav>
         {% if request.endpoint == 'home' %}
+        <div class="cta">
           <a href="{{ url_for('table_of_contents', year=year, lang=lang) }}" class="alt btn">
             Start Exploring
           </a>
+        </div>
         {% endif %}
-        {{ language_switcher() }}
+        <nav title="Page menu" id="Menu" aria-labelledby="MenuBtn">
```
rviscomi: This is introducing a new naming convention. Throughout the site we use lowercase and hyphens.
```suggestion
        <nav title="Page menu" id="menu" aria-labelledby="menu-btn">
```

src/templates/ar/2019/base.html
```diff
@@ -3,79 +3,128 @@
...
+          <button type="button" id="MenuBtn" class="menu-btn" aria-label="Open the menu" aria-expanded="false">
```
rviscomi: Should this label be more descriptive? And is "menu" the right word for this UI?

```suggestion
          <button type="button" id="MenuBtn" class="menu-btn" aria-label="Expand the list of navigation links" aria-expanded="false">
```

src/templates/ar/2019/base.html
```diff
@@ -3,79 +3,128 @@
...
 
   {% block footer %}
     <footer class="alt-bg">
       <a class="navigation-logo home-logo" href="{{ url_for('home', year=year, lang=lang) }}"><img src="/static/images/logo.svg" alt="Web Almanac by HTTP Archive" height="39" width="202"></a>
-      <nav title="Footer menu" class="nav-items">
-        <a href="{{ url_for('table_of_contents', year=year, lang=lang) }}">Table of Contents</a>
-        <a href="{{ url_for('contributors', year=year, lang=lang) }}">Contributors</a>
-        <a href="{{ url_for('methodology', year=year, lang=lang) }}">Methodology</a>
-        {{ language_switcher() }}
+      <nav title="Footer navigation" class="nav-items">
```
rviscomi: Do you know if there's any harm in having two nav sections on the page?

And if an AT user wants to jump to the first nav section, would it actually better for them to arrive at the footer, where there is no menu to open? 🤔

src/templates/ar/2019/base.html
```diff
@@ -97,10 +146,45 @@
...
+
+        function triggerMenu() {
+          menuBtn.click();
+        }
 
-        document.querySelector('.menu-btn').addEventListener('click', function() {
+        menuBtn.addEventListener('click', function() {
           var menuOpen = document.body.classList.toggle('menu-open');
           menuIcon.src = menuOpen ? menuIconSrcClose : menuIconInitialSrc;
+          menuBtn.setAttribute('aria-expanded',menuOpen);
```
rviscomi:
```suggestion
          menuBtn.setAttribute('aria-expanded', menuOpen);
```

src/templates/ar/2019/base.html
```diff
@@ -97,10 +146,45 @@
...
+          if (menuBtn.getAttribute('aria-expanded') === 'true') {
+            document.body.addEventListener('click', triggerMenu, false);
+          } else {
+            document.body.removeEventListener('click', triggerMenu, false);
+          }
+
+        });
+        
+        /* Add a click listener to menu so when it's open it swallows click to avoid above click closing it */
+        menu.addEventListener("click", function () {
```
rviscomi: Single quotes for consistency

```suggestion
        menu.addEventListener('click', function () {
```

src/templates/ar/2019/base.html
```diff
@@ -97,10 +146,45 @@
...
+          event.stopPropagation();
+        });
+        
+        
+        menuNav.addEventListener("keydown", function onPress(event) {
```
rviscomi: Any reason not to make this function anonymous?

```suggestion
        menuNav.addEventListener('keydown', function (event) {
```

src/templates/ar/2019/base.html
```diff
@@ -97,10 +146,45 @@
         var menuIcon = document.querySelector('.menu-icon');
         var menuIconInitialSrc = document.querySelector('.menu-icon').src;
         var menuIconSrcClose = document.querySelector('.menu-icon').dataset.srcClose;
+        var menuBtn = document.querySelector('.menu-btn');
+        var menuNav = document.querySelector('#Menu');
+        var menu = document.querySelector('#Menu');
+
+        function triggerMenu() {
+          menuBtn.click();
```
rviscomi: I think this is a case for naming the click event handler and invoking that function directly, rather than going through the event loop.

src/templates/ar/2019/base.html
```diff
@@ -97,10 +146,45 @@
...
+        }
 
-        document.querySelector('.menu-btn').addEventListener('click', function() {
+        menuBtn.addEventListener('click', function() {
           var menuOpen = document.body.classList.toggle('menu-open');
           menuIcon.src = menuOpen ? menuIconSrcClose : menuIconInitialSrc;
+          menuBtn.setAttribute('aria-expanded',menuOpen);
+          var ariaLabel = menuOpen ?  "Close the menu" : "Open the menu";
```
rviscomi: I think we should avoid setting text in JS if possible. Is there a way to have open and close messages as data attributes on the element instead? I think it'd be easier for translators to only have to worry about translating markup and not have to comb through JS looking for strings.

Having a language-agnostic script would also enable us to pull this block into a common template and avoid duplicating the code for every language. Could you also do that in this PR?

bazzadp:
Not seeing this. Can you give more more details (Browser, OS, screensize)?

And definitely cleared your cache when you loaded this?

src/templates/ar/2019/base.html
bazzadp: I think it's fine: https://www.a11ymatters.com/pattern/mobile-nav/#52-always-include-menu-as-text

src/templates/ar/2019/base.html
bazzadp: > Do you know if there's any harm in having two nav sections on the page?

[Absolutely fine](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav#Usage_notes):

> A document may have several `<nav>` elements, for example, one for site navigation and one for intra-page navigation. `aria-labelledby` can be used in such case to promote accessibility.

> And if an AT user wants to jump to the first nav section, would it actually better for them to arrive at the footer, where there is no menu to open? 🤔

Nah I think this is fine. AT users should navigate to `main` content if they want the main content and `nav` if they want navigation.

src/templates/ar/2019/base.html
bazzadp: Sorry I'm not sure what this means? Can you explain to someone who's JavaScript is less than stellar?

rviscomi: Sure, for example we'd bind the click event listener like this:

```js
function toggleNavMenu() {
  // implement the existing handler code here...
}

menuBtn.addEventListener('click', toggleNavMenu);
```

Then in places where you're calling `menuBtn.click()` or `triggerMenu()` instead you'd do `toggleNavMenu()`.

src/templates/ar/2019/base.html
bazzadp: Fixed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/666/commits/1d5172131674bfb0a4e5f21ed6431d4a09e56eef

src/templates/ar/2019/base.html
bazzadp: Fixed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/666/commits/1d5172131674bfb0a4e5f21ed6431d4a09e56eef

src/templates/ar/2019/base.html
bazzadp: Fixed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/666/commits/1d5172131674bfb0a4e5f21ed6431d4a09e56eef

src/templates/ar/2019/base.html
bazzadp: None at all!

Fixed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/666/commits/1d5172131674bfb0a4e5f21ed6431d4a09e56eef

src/templates/ar/2019/base.html
bazzadp: Fixed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/666/commits/1d5172131674bfb0a4e5f21ed6431d4a09e56eef

src/templates/ar/2019/base.html
bazzadp: Fixed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/666/commits/1d5172131674bfb0a4e5f21ed6431d4a09e56eef

bazzadp:
Anything holding up merging this @rviscomi ? Think I answered all your questions but let me know if you still have questions.

src/templates/ar/2019/base.html
rviscomi: The script still appears to be duplicated for every language. Could we try to reduce those redundancies?

bazzadp: I think that would be a good idea. However I think that’s separate to this PR to be honest because we should generalise all scripts (and potentially other common things).

Appreciate this extra JS may be the straw that’s broken the camel’s back but still feel it should be a separate PR to restructure the templates and build process. Happy to raise an issue for that so we don’t forget?

rviscomi: Ok SGTM

bazzadp: Raised #673 for that work.

PR #676, opened 2020-02-24
Refactor chapters to remove duplication

### LINKED ISSUE #673: Refactor templates to reduce duplication and assist with Translations
It's been noted a few times (e.g. [here](https://github.com/HTTPArchive/almanac.httparchive.org/pull/664#pullrequestreview-356514443),  [here](https://github.com/HTTPArchive/almanac.httparchive.org/pull/666#discussion_r382729585) and [here](https://github.com/HTTPArchive/almanac.httparchive.org/pull/665#discussion_r379587777)) that there is a lot of duplication of code, especially with the translations and it would be good to refactor these to reduce that and keep the code more maintainable in future.

We should probably have a `templates/base/2019/` folder or the like and each translation (including English) can inherit from that and only have the translatable text in each language specific directory rather than the current mix of HTML, JS and Text.

bazzadp:
Fixes #673 

This is a large refactor of all the template files to try to separate content from code. This should make translations easier (as more obvious what to translate, rather than hunting for English text in amongst code) as well as making any future development effort easier too (as less places to change the code).


src/templates/ar/2019/base.html
```diff
@@ -1,226 +1,97 @@
-{% extends "base.html" %}
+{% extends "base.html" {% extends "base/2019/base.html" %}
```
rviscomi:
```suggestion
{% extends "base/2019/base.html" %}
```

src/templates/base/2019/accessibility_statement.html
```diff
@@ -0,0 +1,50 @@
+{% extends "%s/2019/base.html" % lang %}
```
rviscomi: Why does this extend the language-specific base as opposed to the 2019 base?

Won't this overwrite all language-specific blocks into English?

bazzadp: This is so it picks up the language specific text. There’s a lot of common, language-specific-text in that needed by **all** templates.

`”%s/2019/base.html”` actually extends `”base/2019/base.html”` but overrides the base text with language-specific base text.

Then base Accessibility extends that and overrides with Accessibility-specific code.

Then the language-specific Accessibility extends that and overrides with language-specific Accessibility-specific text.

It sounds very convoluted, but it’s the best way I could see to get a language specific base and then extend it with page-specific code, and then use that as template for language-specific, page-specific code. Wow that’s a mouthful!

src/templates/ar/2019/base.html
bazzadp: Thanks! Typo. Was fine in other languages but this one slipped the net cause I couldn't test it.

bazzadp:
Merging after discussion over Slack with @rviscomi

Issue #539, opened 2019-11-25
Translate 2019 content to French


SilentJMA:
These are the core templates - without which we cannot release any translated chapters. They are in the [language specific templates directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/templates):
- [x] base.html + authors bio - @borisschapira 
- [x] base_chapter.html - @borisschapira 
- [x] contributors.html - @borisschapira 
- [x] contributors.html 2020 - @borisschapira 
- [x] error.html - @borisschapira 
- [x] index.html - @borisschapira 
- [x] index.html 2020 - @borisschapira 
- [ ] featured_chapters.html (partially done)
- [x] table_of_contents.html @borisschapira 

These are the chapters to be translated, in rough order of popularity. They exist in the [content directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content):
- [x] javascript - @borisschapira 
- [ ] css - @nico3333fr 
- [x] seo - @AymenLoukil 
- [ ] performance -  
- [ ] http2 - 
- [x] markup -  @SilentJMA & @borisschapira
- [x] third-parties - Partie Tiers @borisschapira
- [ ] fonts - 
- [x] accessibility - @nico3333fr 
- [ ] pwa - 
- [x] cms - @JustinyAhin
- [ ] security - @nico3333fr 
- [ ] media - @borisschapira 
- [x] mobile-web - @borisschapira 
- [ ] cdn - 
- [ ] ecommerce - 
- [ ] page-weight - 
- [x] caching - @allemas
- [x] compression - @allemas
- [x] resource-hints - @borisschapira 

Additionally the following pages need translated too in the [language specific templates directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/templates):
- [x] methodology.html @borisschapira 
- [x] accessibility_statement.html @nico3333fr 
- [ ] base_ebook.html

There is no need to translate the chapters HTML pages are they are generated off the markdown combined with the above templates.

Please include "Makes progress on #539" in all pull requests so a link is created from the PR to this issue.

---

Common notes for writing consistency are here: https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Translators'-Guide. Feel free to edit that and/or add French-specific extras by editing this comment.

French specific extra advice:

[...]

PR #669, opened 2020-02-20
Translation of caching chapter to French 

allemas:
Makes progress on #539

It took a while, but it's finally over. :) 
FR Translations of caching chapter and it's open for review.

Hope that master branch synch is ok.

PR #679, opened 2020-03-06
Translation of Caching chapter to French 

allemas:
New pull request with reviews and page generation :) 

Makes progress on #539

### DIFF
--- a/src/content/fr/2019/caching.md
+++ b/src/content/fr/2019/caching.md
@@ -576,8 +576,7 @@ En fait, l'un des [rapports de tendance des archives HTTP montre l'adoption des 
   <a href="/static/images/2019/16_Caching/ch16_fig14_service_worker_adoption.jpg">
     <img src="/static/images/2019/16_Caching/ch16_fig14_service_worker_adoption.jpg" alt="Figure 17. Série chronologique de pages contrôlées par des service worker." aria-labelledby="fig17-caption" aria-describedby="fig17-description" width="600" height="311">
     </a>
-  <div id="fig17-description" class="visually-hidden">
-  Un graphique de séries chronologiques montre l'utilisation des sites contrôlés par les service worker d'octobre 2016 à juillet 2019. L'utilisation a augmenté régulièrement au fil des ans, tant pour les téléphones portables que pour les ordinateurs de bureau, mais reste inférieure à 0,6&nbsp;% pour les deux.</div>
+  <div id="fig17-description" class="visually-hidden">Un graphique de séries chronologiques montre l'utilisation des sites contrôlés par les service worker d'octobre 2016 à juillet 2019. L'utilisation a augmenté régulièrement au fil des ans, tant pour les téléphones portables que pour les ordinateurs de bureau, mais reste inférieure à 0,6&nbsp;% pour les deux.</div>
   <figcaption id="fig17-caption">Figure 17. Série chronologique de pages contrôlées par des service workers. (Source&nbsp;: <a href="https://httparchive.org/reports/progressive-web-apps#swControlledPages">HTTP Archive</a>)</figcaption>
 </figure>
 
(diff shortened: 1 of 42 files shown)


src/content/fr/2019/caching.md
```diff
@@ -0,0 +1,691 @@
...
+Pour qu'un client HTTP mette en cache une ressource, il doit répondre a deux questions&nbsp;:
+
+*   "Combien de temps dois-je mettre en cache&nbsp;?"
+*   "Comment puis-je valider que le contenu est encore frais&nbsp;?"
+
+Lorsqu'un navigateur Web envoie une réponse à un client, il inclut généralement dans sa réponse des en-têtes qui indiquent si la ressource peut être mise en cache, pour combien de temps et quel est son âge. La RFC 7234 traite plus en détail de ce point dans la section [4.2 (Freshness)](https://tools.ietf.org/html/rfc7234#section-4.2) et [4.3 (Validation)](https://tools.ietf.org/html/rfc7234#section-4.3).
+
+Les en-têtes de réponse HTTP généralement utilisées pour transmettre la durée de vie sont&nbsp;:
+
+*   <i lang="en">`Cache-Control`<span> vous permet de configurer la durée de vie du cache (c'est-à-dire sa durée de validité).
```
bazzadp: Typo here (think it's my fault btw!)

```suggestion
*   <i lang="en">`Cache-Control`</i> vous permet de configurer la durée de vie du cache (c'est-à-dire sa durée de validité).
```

src/content/fr/2019/caching.md
```diff
@@ -0,0 +1,691 @@
...
+  </a>
+  <div id="fig3-description" class="visually-hidden">Un graphique à barres superposées montrant que 20&nbsp;% des réponses pour ordinateurs de bureau ne peuvent être mises en cache, 47&nbsp;% ont un cache supérieur à zéro, 27&nbsp;% sont mises en cache de manière heuristique et 6&nbsp;% ont un TTL de 0. Les statistiques pour les mobiles sont très similaires (19&nbsp;%, 47&nbsp;%, 27&nbsp;% et 7&nbsp;%)</div>
+  <figcaption id="fig3-caption">Figure 3. Distribution des réponses pouvant être mises en cache.</figcaption>
+</figure>
+
+Le tableau ci-dessous détaille les [TTL](https://fr.wikipedia.org/wiki/Time_to_Live) du cache pour les requêtes en provenance d'ordinateurs de bureau. La plupart des types de contenu sont mis en cache, mais les ressources CSS semblent toujours être mises en cache à des valeurs TTL élevées.
```
bazzadp: Doesn't look like my previous comment was addressed here:

I see you've included a wikipedia reference here thats not in the English version. If you think things like this are valuable then I would encourage you to also include such changes to the English version i) to keep them in synch and ii) to benefit English speakers. Similarly if you spot typos while translating.

However in this specific case I would question its value. The [French version](https://fr.wikipedia.org/wiki/Time_to_Live) talks about IP TTL and DNS TTL, but not HTTP TTL which we are discussing here. Funnily enough, [the English version of that page](https://en.wikipedia.org/wiki/Time_to_live#HTTP) _does_ talk about HTTP TTL but even then only covers it very briefly so not sure if that link is more confusing than beneficial?

To be honest I'd remove this. However if you really want it, then a better link might be the MDN docs (in [French](https://developer.mozilla.org/fr/docs/Glossaire/TTL#Mise_en_cache) and [English](https://developer.mozilla.org/en-US/docs/Glossary/TTL#Caching))?

```suggestion
Le tableau ci-dessous détaille les TTL du cache pour les requêtes en provenance d'ordinateurs de bureau. La plupart des types de contenu sont mis en cache, mais les ressources CSS semblent toujours être mises en cache à des valeurs TTL élevées.
```

allemas:
I didn't find the last change requested :(

bazzadp:
Yeah was all hidden away in the review: https://github.com/HTTPArchive/almanac.httparchive.org/pull/669#discussion_r382925557

LGTM so let's merge!

PR #683, opened 2020-03-15
Better image directory folders

### LINKED ISSUE #654: Chapter slugs and image directories based on title so don't work for Translations
main.py has the following:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/7015c954088b525b95a2ee0179c94ea56d92e67c/src/main.py#L60-L68

This means we use the chapter title from the meta data to get the chapter slug (for the ToC) and the chapter image directory (for the chapter hero images and figures).

This does not work if we translate the title as @borisschapira has done for #652 and @SilentJMA did for #539 

I think we should change these two functions to be based on either:
1. The md or html filename (ideally at chapter generation so we're not dependent on doing this at runtime and being dependent on the URL which is user supplied data so shouldn't be trusted - at least without verification).
2. The config - but not sure how to map this (based on the part and chapter number from the meta data?).

I know we've talked about a larger restructure of the config in #561 and even changing how the chapters are generated completely in #634 but we don't seem to be making progress on those so think we need to do something in the short term to address this issue as it prevents viewing the chapters with images in another language (for checking PRs in dev) and will obviously prevent going live with any translations in prod.

@mikegeyser any thoughts?

bazzadp:
Fixes #654 

This moves the chapter images from folders like this:
```/static/images/2019/05_Third_Parties/fig7.png```
to folders like this
```/static/images/2019/third-parties/fig7.png```

This is more in-line with other folder names and works better for translations as some chapter titles do not match English (this is the issue #654 )

I've also added the chapter slug to the meta-data for ease of look up. This is automatically added to the metadata as part of the chapter generation.

I've also added redirects for anyone still referencing the old image URLs. Will need to test this as part of next deploy as difficult to test this. @rviscomi is it possible to deploy this elsewhere? You did that once before when test Google Fonts.

### CODE REVIEW

**Comment by rviscomi** [comment]
FYI image directory naming was based on the query directory naming scheme discussed by @KJLarson in https://github.com/HTTPArchive/almanac.httparchive.org/issues/62#issuecomment-505588971. One benefit of keeping the chapter number in the name is so each chapter's directory appears in the source code in order of appearance rather than alphabetical order. I'll defer to whatever you think is best.

What I did to test the Google Fonts change was run a special App Engine deploy command:

```sh
gcloud app deploy --project webalmanac --no-promote
```

The [no-promote](https://cloud.google.com/sdk/gcloud/reference/app/deploy) ensures that it doesn't receive any traffic. From the GCP console, navigate to the [App Engine version page](https://console.cloud.google.com/appengine/versions?folder=&organizationId=&project=webalmanac&serviceId=default&versionssize=50), and you can click on the version name to get its unique URL. You could also go in there periodically and delete older versions, since we'll never be rolling back to them.

bazzadp:
Thanks for the background. That seems in reference to the SQL folder structure and then seems to have been carried over to the images.

I’m actually OK with the SQL folders staying as is and can (sorta!) understand the argument to make them easier to reference since people do come to GitHub to look them up.

However for the images I think they should be closer aligned to the chapter content which doesn’t follow this naming structure.

Renaming image folders makes it a lot easier to look them up for the templates. We could add a look up table (I did this originally but never committed that code to this branch) or refer back to English name to construct the folder name (kind of like we do currently but extending it so it works for non-English languages too) but all seems a bit overly complex. As I say I think the site (in src folder) should have consistent folder URL structure and at present it doesn’t.

Thanks for info on how to do a test deploy. Will give it a go to see if my code actually works, give it some more thought, and then decide one way or another.

Let me know if you (or anyone else) has any more thoughts or opinions here to help decide but if no objections, then I’m still heavily veering towards renaming the image folders.

rviscomi:
That's fine, your approach SGTM.

rviscomi:
@bazzadp I just got pinged with this error:

```
AttributeError
'str' object has no attribute 'repkace'
```

Is this PR related?

bazzadp:
Yes but only the no-promote version so won't be affecting the real site – that's why testing is good! 😀

Hope I didn't wake you!!

BTW when it's a more sensible hour for you I can't delete old versions:

> You do not have permission to view activities in project ‘Web Almanac’. Ask a project administrator to give you an Owner, Editor or Viewer role.

rviscomi:
Gave you admin privileges for App Engine, try now.

bazzadp:
> Gave you admin privileges for App Engine, try now.

Same thing I'm afraid.

rviscomi:
I would expect "Service Admin" to be a subset of "Admin" privileges, but you have both now, so it should do the trick.

PR #685, opened 2020-03-17
Enable first translations (French and Spanish)

bazzadp:
This PR enables the French and Spanish languages in both the sitemap and the python script.

Only translated chapters (and non-chapters) will be available. The rest will be redirected back to English if they exist.

This depends on a few things:
- Successfully merging #684 (Edit - since done!)
- One of the French speakers doing a review of the base templates in templates/fr/2019 as I auto translated some content. Note these have changed quite a bit with #676 so should be easier as only contains the text rather than code as well.
- One of the Spanish speakers translating the base templates (Done - thanks @c-torres !)
- Testing!!!

And of course if we could translation more chapters and templates, then that would be great!


bazzadp:
Updated PR to include Spanish now!

borisschapira:
Thanks a lot for this work, I'm going to review the French part.

First discovery (but I don't know how to fix):

In the `contributors_description` block of the index template (all languages except EN), the `{{ contributors }}` renders nothing, probably because this line:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/enable-first-translations/src/templates/base/2019/index.html#L51

does not pass the variable to the block.

bazzadp:
> In the `contributors_description` block of the index template (all languages except EN), the `{{ contributors }}` renders nothing, probably because this line:

It actually did affect the EN version too. Fixed in 9f34695

src/templates/base/2019/index.html
```diff
@@ -24,6 +24,8 @@
   <link rel="stylesheet" href="/static/css/index.css">
 {% endblock %}
 
+{% block contributors %}{{ config.contributors.keys() | length }} {% endblock %}
```
borisschapira: Creating a new block to pass the variable through. Elegant solution there. Well done!

bazzadp: It's a little convoluted but works a treat and I think it makes the templates easier to translate - even if the logic is more difficult to understand.

Going to do same to fix your contributors issue to avoid reusing the same capitalised title.

bazzadp: Done in https://github.com/HTTPArchive/almanac.httparchive.org/pull/685/commits/4064c1d1a633074c5ce5877fbad322b50b32b067. Let me know if I got the translation right.

Issue #498, opened 2019-11-16
Translate content to Spanish


JMPerez:
- [ ] accessibility
- [ ] caching
- [ ] cdn
- [ ] cms
- [ ] compressino
- [ ] css
These are the chapters to be translated:

- [ ] ecommerce
- [ ] fonts
- [ ] http2
- [ ] javascript
- [X] markup
- [ ] media
- [ ] mobile-web
- [ ] page-weight
- [X] performance
- [ ] pwa
- [ ] resource-hints
- [ ] security
- [ ] seo
- [ ] third-parties

c-torres:
Should we include the specific pages? Like Table of Contents, Methodology, etc.

rviscomi:
@c-torres there are still some changes happening to those pages so let's leave those for last.

c-torres:
Working on the JavaScript chapter.

c-torres:
Working on the CSS chapter.

bazzadp:
> Working on the CSS chapter.

Hey @c-torres just so you're aware, very shortly (I hope - nudge, nudge @rviscomi !!😉 ) we will be merging #511 which contains a change of format for the diagrams and more descriptive text for them.

We already did the JavaScript chapter and I checked and you got the latest version so that's cool. However unfortunately we will need you guys to revist the other chapters you've already done.

So before you submit the CSS chapter you might be best to double check master to ensure you've got the latest diagram markdown. Or alternatively take the source from the [`figures_new_format` branch](https://github.com/HTTPArchive/almanac.httparchive.org/blob/figures_new_format/src/content/en/2019/css.md) under the assumption it will be merged.

Sorry about this change, but I thought it would be worthwhile giving you the heads up to this as know there are a lot of figures in the CSS chapter!

c-torres:
No worries, thanks for the heads up, really appreciated.

bazzadp:
That's now merged to master @c-torres and will shortly be updated in prod. So you're good to grab the latest version. It was the same that was in the branch if you've already started on that. Will raise a separate issue for the other chapters already done to have another look over.

c-torres:
Submitted a PR for CSS translation. Will continue working on Media Chapter.

bazzadp:
Hey @c-torres @JMPerez @taytus !

We’re now in the position to launch chapters that have been translated without having to wait for all chapters to be translated.

[...]

PR #687, opened 2020-03-17
Base chapter translations to Spanish

c-torres:
Makes progress on #498  

Translating base_chapter HTML to Spanish and fixing typos on Spanish base HTML. 

Please review as I might be missing something from the translation process.


bazzadp:
LGTM! Thanks for speedy response. Will add Spanish to the initial language release in #685 !

PR #688, opened 2020-03-18
Enable first translations boris review

borisschapira:
Some modifications to improve the French translation because making it public.

I'm not sure about the usage of `{{ self.contributors_title() }}` in the contributors.html base template, but I wanted to reuse a preexisting block rather than create another one. WDYT?


bazzadp:
Thanks!

PR #709, opened 2020-04-05
Translation of security chapter to Japanese

ksakae1216:


bazzadp:
Weird. Not heard of that before and can't see any settings to enable that. Will do some tests.

Anyway that's released now: http://127.0.0.1:8080/ja/2019/security

Thanks for your help again!

PR #711, opened 2020-04-05
Correct filter on load and add total

### LINKED ISSUE #704: Number of per-team contributors not set correctly on page load
For example, when visiting https://almanac.httparchive.org/en/2019/contributors?teams=editors it shows "92 contributors" even though this team only has 5 members.

If we start from https://almanac.httparchive.org/en/2019/contributors and select "Editors", it shows the correct number of contributors.

So there's a bug in the initial count of contributors on page load and we should take team selection(s) from the URL into account.

bazzadp:
Fixes #704.

Also adds a total when filtered.

Unfiltered:

Filtered image

Filtered:

Unfiltered image

### DIFF
--- a/src/templates/en/2019/contributors.html
+++ b/src/templates/en/2019/contributors.html
@@ -9,7 +9,7 @@
 
 {% block breadcrumb_name_2 %}{{ year }} Contributors{% endblock %}
 
-{% block filter_by_team %}Filter by team: <span id="filtered-contributors">{{ self.contributors() }}</span> contributors{% endblock %}
+{% block filter_by_team %}Filter by team: <span id="filtered-contributors">{{ self.contributors() }}</span> contributors<span id="contributors-total-text" class="hidden"> out of <span id="contributors-total">{{ config.contributors.items() | length }}</span> in total</span>.{% endblock %}
 
 {% block join_the_team_title%}Join the Web Almanac team{% endblock %}
 {% block join_the_team_text%}Join the team!{% endblock %}
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -819,6 +819,10 @@ p.copyright a {
   white-space: nowrap; /* added line */
 }
 
+.hidden {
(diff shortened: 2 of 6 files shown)


bazzadp:
Can you have a quick look at this one too if you’re in a reviewing mood @rviscomi ?

src/templates/en/2019/contributors.html
```diff
@@ -9,7 +9,7 @@
 
 {% block breadcrumb_name_2 %}{{ year }} Contributors{% endblock %}
 
-{% block filter_by_team %}Filter by team: <span id="filtered-contributors">{{ self.contributors() }}</span> contributors{% endblock %}
+{% block filter_by_team %}Filter by team: <span id="filtered-contributors">{{ self.contributors() }}</span> contributors<span id="contributors-total-text" class="hidden"> out of <span id="contributors-total">{{ config.contributors.items() | length }}</span> in total</span>.{% endblock %}
```
rviscomi, line 12: This sounds verbose to me. How much do you like it? 😁

bazzadp: Well obviously I do or I wouldn’t have implemented it 😀

I do like showing the total contributors as much as possible (we do it twice in home screen!) and feel only showing filtered number (especially if coming straight from a link) on this — the contributors page — underplays the total contributors figure.

For some small groups (e.g. translators) it may be obvious for for others (e.g. brainstormers) it could easily be assumed to be the total.

I think we have the space, even on mobile, so why not? Though admittedly in some of the languages it goes to two lines but for some (French) it does that anyway now so still only two lines.

bazzadp: Btw you did spot that I don’t show this total when the first number is the total already?

rviscomi: I like the idea behind it. I was getting at the verbosity, for example I'd suggest rewording it to something more concise like "46 of 92 contributors"

Issue #686, opened 2020-03-17
Move Supported Languages to year config


bazzadp:
Currently Support Languages is defined in two places:
1. In [validate.py](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/validate.py) for the Python scripts.
2. In [generate_chapters.js](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/tools/generate/generate_chapters.js) for generating the sitemap.

This also means it’s set once at project level when we might have different languages supported in different years.

We should move this to the year-specific [JSON config](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/config/2019.json) so we once config isn’t forgotten and so can support different languages per year.

bazzadp:
This has been moved to JSON config now as part of #712 :

```
{
  "settings": [
    {
      "isLive": true,
      "supported_languages": ["en","es","fr","ja"]
    }
  ],
```

However we still need to update the [`generate_chapter.js`](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/tools/generate/generate_chapters.js) to use this instead of hardcoding this:

```
//Chapters may exist but not be ready to be launched so do not include in sitemap
const sitemap_languages = ['en','es','fr','ja'];
```

We also need to make it year specific so need to update [`generate_sitemap.js`](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/tools/generate/generate_sitemap.js) for this. Still think we should do the load in `generate_chapter.js` (as the config might be used for other things in future) but it should be set in a variable that is then used in `generate_sitemap.js`.

PR #712, opened 2020-04-05
Validate chapters, languages and years

### LINKED ISSUE #400: Validate chapters by year

### LINKED ISSUE #713: Support multiple year's JSON config
I presume we're going to start on 2020 edition soon. On the most we're set up well for it with folder structures and therefore templating set up for years.

However we have a few things to update including hardcoding 2019 in [config.py](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/config.py)

```
  # TODO(rviscomi): Dynamically handle multiple annual configs.
  with open('config/2019.json', 'r') as config_file:
    config_json['2019'] = json.load(config_file)
```

bazzadp:
Fixes #400 
Fixes #713 
Makes progress on #686

This implements @arsenicraghav 's fix for hardcoded chapters in validation (#400) from https://github.com/HTTPArchive/almanac.httparchive.org/pull/557 (though in config.py instead of validate.py to avoid multiple loads of JSON file). We had delayed this before but we're not making any progress on #561 and not entirely convinced we should do that any more anyway!

It also allows loading of multiple year config JSONs (#713 ) including the ability to exclude some years in JSON config (while we work on them and before they are ready for release).

It also makes languages configureable per year and in JSON as part of #686. The JavaScript side still needs to be done.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -16,8 +16,8 @@ DEFAULT_AVATAR_FOLDER_PATH = '/static/images/avatars/'
 AVATAR_SIZE = 200
 AVATARS_NUMBER = 15
 
-CHAPTERS = {}
-LANGUAGES = {}
+SUPPORTED_CHAPTERS = {}
+SUPPORTED_LANGUAGES = {}
 
 def get_config(year):
   global config_json
@@ -52,8 +52,8 @@ def get_languages(json_config):
 
 def update_config():
   global SUPPORTED_YEARS
-  global CHAPTERS
-  global LANGUAGES
+  global SUPPORTED_CHAPTERS
+  global SUPPORTED_LANGUAGES
   global config_json
 
   for root, directories, files in os.walk('config'):
@@ -69,8 +69,8 @@ def update_config():
       json_config = json.load(config_file)
       config_json[year] = json_config
 
-      LANGUAGES.update({year : get_languages(json_config)})
-      CHAPTERS.update({year : set(get_chapters(json_config))})
+      SUPPORTED_LANGUAGES.update({year : get_languages(json_config)})
+      SUPPORTED_CHAPTERS.update({year : set(get_chapters(json_config))})
 
       for contributor_id, contributor in json_config['contributors'].items():
         if 'avatar_url' not in contributor:
--- a/src/config.py
+++ b/src/config.py
@@ -8,9 +8,38 @@ DEFAULT_AVATAR_FOLDER_PATH = '/static/images/avatars/'
 AVATAR_SIZE = 200
 AVATARS_NUMBER = 15
 
+CHAPTERS = {}
+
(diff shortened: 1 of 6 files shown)


src/config.py
```diff
@@ -1,31 +1,93 @@
...
+  languages = []
+  data = get_entries_from_json(json_config,'settings','supported_languages')
+  for list in data:
+    for entry in list:
+      languages.append(getattr(Language,entry))
+  return languages
+
+
+def get_live(json_config):
+  isLive = False
```
rviscomi, line 52: Nit: Python style is to use lowercase and underscores for variable names.
```suggestion
  is_live = False
```

src/language.py
```diff
@@ -27,12 +27,12 @@ def lang_code(self):
 
 # Currently we are only supporting languages and not regions
 class Language(object):
-  JAPANESE = _Language('日本語', 'ja', 'JP')
-  ENGLISH = _Language('English', 'en', 'US')
-  SPANISH = _Language('Español', 'es', 'ES')
-  FRENCH = _Language('Français', 'fr', 'FR')
+  ja = _Language('日本語', 'ja', 'JP')
```
rviscomi, line 30: I had used uppercase here as an enum value. Is this not really an enum?

src/config.py
bazzadp: Cool. Will fix. First dabble into Python here.

src/language.py
bazzadp: It is an enum and probably should be in caps for that reason assuming that’s python convention.

However the reason I changed it was to allow the JSON config to be used for #686 and the two digit language code (in lowercase) will make that easier.

I could make it uppercase here and keep the JSON lowercase and just uppercase it when I read it convert it to a language object. WDYT?

rviscomi: SGTM

src/validate_test.py
```diff
@@ -1,9 +1,10 @@
+from config import DEFAULT_YEAR, SUPPORTED_YEARS
 from language import Language, DEFAULT_LANGUAGE
-from validate import parse_accept_language, DEFAULT_YEAR, SUPPORTED_YEARS
+from validate import parse_accept_language
 
-SUPPORTED_LANGUAGES = (Language.ENGLISH.lang_code, Language.JAPANESE.lang_code)
+SUPPORTED_LANGUAGES = (Language.en.lang_code, Language.ja.lang_code)
```
rviscomi, line 5: Should these be EN, JA?

bazzadp: Yup. Fixed.

We really should write more of these test cases. Only discovered this file recently!

PR #717, opened 2020-04-07
Add vary header for base URL

### LINKED ISSUE #716: Base redirect does not set Vary header so may redirect to wrong language
When you enter the base url: https://almanac.httparchive.org/ then main.py looks at your `Accept-Language` header and redirects you to the appropriate language if we support it.

However, we are not setting the `Vary` header so this response is cached for every subsequent visitor for the next 3 hours!

bazzadp:
Fixes #716

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -107,7 +107,9 @@ def lang_only(lang):
 @app.route('/')
 @validate
 def root(lang):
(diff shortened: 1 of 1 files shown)


bazzadp:
Tested fix on test deployment (https://20200407t191033-dot-webalmanac.uk.r.appspot.com/) and looks good.

PR #723, opened 2020-04-11
Language and year specific error pages

bazzadp:
This creates language-specific error pages.

At present if you get a 404 for a French page, for example, you get the English error page:

Generic English 404 error page

This is not the most friendly to our foreign readers. Also notice that the Language drop down isn't rendered correctly (as there are no translations of this page).

This PR changes it to the following (and similarly for Japanese and Spanish):

Language specific 404 error page

This will also become more important when we have different years, which might have completely different styles. This will allow us to keep the user in the same design and flow rather than a jarring move to another year's design on error.

It drops back to default year (currently 2019), and then default lang and year (currently en and 2019) if an error page doesn't exist.

@borisschapira / @ksakae1216 / @c-torres - it would be good if you could give your respective language error pages a once over if you get a chance. Mostly done with Google Translate.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -40,10 +40,11 @@ def add_header(response):
     return response
 
 def render_template(template, *args, **kwargs):
-    year = request.view_args.get('year', DEFAULT_YEAR)
+    year = kwargs.get('year',request.view_args.get('year', DEFAULT_YEAR))
     supported_languages = SUPPORTED_LANGUAGES.get(year, (DEFAULT_LANGUAGE,))
 
-    lang = request.view_args.get('lang') or ''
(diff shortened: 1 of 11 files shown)


borisschapira:
I'm too late for review but I still have time to thank you very much for this improvement, @bazzadp. French is 👍 .

bazzadp:
You snooze you lose round here coz I'm very inpatient 😊 Did some more machine French in #728 if you wanna cast your eyes over that?

```
{% block queries %}Afficher les requêtes{% endblock %}
{% block results %}Voir les résultats{% endblock %}
````

PR #724, opened 2020-04-11
Add link for full image for markup fig 7

### LINKED ISSUE #599: Top HTML elements plot is hiding y-axis labels
In the [Markup chapter](https://almanac.httparchive.org/en/2019/markup) under the "Perspective on value and usage" section, the "Top HTML elements" plot hides important y-axis labels. These labels are not sequential that can be inferred from the values above and below. While the the plot on the page is interactive and one can hover over any bar to see the corresponding label, I still think it is important to make them all visible in `src/static/images/2019/03_Markup/fig7.png` at the cost of smaller font size of labels and taller graph.

/cc @rviscomi @bkardell

bazzadp:
Fixes #599

Trying to clean-up the issues queue so finally got round to tackling this one. I went with the link idea in as suggested in #599. Still not 100% convinced on that, but do understand the issue with an extremely large image and think this is at least a bit better.

It's currently in two places:

1. If showing fallback images, then clicking on the figure now opens the full image. Interactive google sheets can't be clicked on, but they can see full detail by hovering over it anyway and we also have the second place for them, which we’ll discuss next 😀
2. Added a link in the Fig Caption, to make this more obvious. Note this only opens the full fallback image and not a larger interactive Google Sheets as think that's overkill when you already have an interactive sheet on this page.

Screenshot of how the Fig Caption looks:

Link to full detail in fig caption

CC: @ibnesayeed


ibnesayeed:
LGTM! Thanks!

PR #726, opened 2020-04-12
Automatically remove any trailing slashes

bazzadp:
Happened to notice that the following doesn't work: /en/2019/css/ (i.e. URLs with a trailing slash)
Best practice is to handle this, so this PR 301s to /en/2019/css. It also works for the other static pages (e.g. methodology, ToC...etc.).

@rviscomi on a separate note can you look at my GCloud permissions? Get the following when trying to deploy:

```
MaxRetrialsException: last_result=(None, (<class 'googlecloudsdk.api_lib.storage.storage_api.UploadError'>, 
UploadError(u'403 Could not upload file [/Users/barry/almanac.httparchive.org/src/content/ja/2019/compressio
n.md] to [staging.webalmanac.appspot.com/03836fa1b7206614bdaa2154f6b610712bfe4e93]: barry@tunetheweb.com doe
s not have storage.objects.delete access to staging.webalmanac.appspot.com/03836fa1b7206614bdaa2154f6b610712
bfe4e93.',), <traceback object at 0x1072d9f38>)), last_retrial=3, time_passed_ms=1837,time_to_wait=0
```

Weirdly I am still able to delete old version from https://console.cloud.google.com/appengine/versions?project=webalmanac&serviceId=default&versionssize=50 and tested that by cleaning up the old versions. So a bit confused why I have access to do that from the web console, but not deploy from the command line?

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -160,6 +160,13 @@ def accessibility_statement(lang):
     return render_template('%s/2019/accessibility_statement.html' % (lang))
 
 
+# Handle trailing slashes for accessibility statement (as it's a special case)
+@app.route('/<lang>/accessibility-statement/')
+@validate
+def accessibility_statement_with_trailing_slash(lang):
+    return redirect("/%s/accessibility-statement" % (lang)), 301
+
+
 @app.route('/sitemap.xml')
 def sitemap():
     xml = render_template('sitemap.xml')
--- a/src/main.py
+++ b/src/main.py
@@ -43,12 +43,12 @@ def add_header(response):
 def render_template(template, *args, **kwargs):
     # If the year has already been set (e.g. for error pages) then use that
     # Otherwise the requested year, otherwise the default year
-    year = kwargs.get('year',request.view_args.get('year', DEFAULT_YEAR))
+    year = kwargs.get('year', request.view_args.get('year', DEFAULT_YEAR))
     supported_languages = SUPPORTED_LANGUAGES.get(year, (DEFAULT_LANGUAGE,))
 
     # If the lang has already been set (e.g. for error pages) then use that
     # Otherwise the requested lang, otherwise the default lang
-    lang = kwargs.get('lang',request.view_args.get('lang', DEFAULT_LANGUAGE))
+    lang = kwargs.get('lang', request.view_args.get('lang', DEFAULT_LANGUAGE.lang_code))
 
     language = get_language(lang)
     langcode_length = len(lang) + 1 # Probably always 2-character language codes but who knows!
@@ -162,7 +162,6 @@ def accessibility_statement(lang):
 
 
 @app.route('/sitemap.xml')
-@validate
 def sitemap():
     xml = render_template('sitemap.xml')
     resp = app.make_response(xml)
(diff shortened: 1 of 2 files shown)


rviscomi:
Try reauthenticating: `gcloud auth login`

bazzadp:
Nope same thing. It does recognise me:

`barry@tunetheweb.com does not have storage.objects.delete access to staging.webalmanac.appspot.com`

which surely it wouldn't have done if I hadn't authenticated?

But still doesn't think I have access...

bazzadp:
Seems to have fixed itself, so deployed what we have so far as been working on quite a few changes. Haven't included the open PRs as will wait to see if you have any comments first.

bazzadp:
Any thoughts on this one @rviscomi ? A few code clean ups make it look bigger than it really is, as it's really quite small.

Let me know if any major concerns otherwise will merge away.

src/validate.py
```diff
@@ -26,6 +26,11 @@ def decorated_function(*args, **kwargs):
         year = kwargs.get('year')
         chapter = kwargs.get('chapter')
 
+        # Handle the pages that don't follow /lang/year/page structure
+        # which can end up here if they have trailing slashes in URL
+        if year == "accessibility-statement":
+            return redirect('/%s/accessibility-statement' % lang, code=301)
```
rviscomi, line 32: Not sure that this is a job for the validator. Could this be avoided by adding

```
@app.route('/<lang>/accessibility-statement/')
```

to the AS renderer and redirecting as needed?

bazzadp: Done!

Note we can't do this for the chapters.

src/main.py
```diff
@@ -161,8 +160,14 @@ def accessibility_statement(lang):
     return render_template('%s/2019/accessibility_statement.html' % (lang))
 
 
-@app.route('/sitemap.xml')
+# Handle trailing slashes for accessibility statement (as it's a special case)
+@app.route('/<lang>/accessibility-statement/')
```
rviscomi, line 164: I think you should be able to decorate the existing `accessibility_statement` function with this `@app.route` and decide wither to redirect there based on trailing slash.

bazzadp: @rviscomi I cannot get this to work.

I can get it to work the other way around. So if the route is setup as `/<lang>/accessibility-statement/` then flask automatically redirects any request to `/en/accessibility-statement` (without the last slash) to include the slash (with a 308 redirect). It also serves `/en/accessibility-statement/` just fine without any redirect.

However our URLs have no trailing slash (except for home pages like /en/2091/) so it's `/en/accessibility-statement` or `/en/2019/methodoloy`. And the only two ways I can get this to redirect requests to `/en/accessibility-statement/` to `/en/accessibility-statement` is either 1)  the have two routes like I have currently. Or 2) have validate catch this and redirect.

This is also explained in the Flask documentation: https://flask.palletsprojects.com/en/1.1.x/quickstart/#unique-urls-redirection-behavior

As we've been launched for a while now, and linked from many places, I don't think it's worth changing our URLs now (even with a 301 redirect), so we're left with handling this accessibility-statement edge case in the ways described above.

@rviscomi to you want to give it a go and see if I'm missing something?

bazzadp: OK finally resolved this @rviscomi 
Can you check if happy now?

PR #728, opened 2020-04-12
Add Results and Queries buttons to each chapter

### LINKED ISSUE #221: Give readers the ability to discover queries, results, and comments for each chapter

bazzadp:
Fixes #221 . Should have done this ages ago as quick nice to have the links in each chapter!

Just knocked together some quick icons myself but let me know if you have better ones to use

**Desktop:**

desktop

**Mobile**:

mobile


bazzadp:
BTW could make this cleaner and remove one bit of chapter meta data I added in this PR (`queries` metadata) if we renamed the [sql directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/master/sql/2019) to be named by chapter slug (similarly to what I did with the image folders in #683). WDYT @rviscomi ?

rviscomi:
IIRC there are some external links floating around that point directly to the SQL directories. Also, I still like how they are prefixed by chapter number so the directories can be browsed in order. So my preference is to keep the naming as is.

My only comment for this PR is that my original intent was to link each figure with its corresponding tab in the spreadsheet so readers can go directly to the underlying data that powers the chart. Since the spreadsheets are indexed by metric ID and not figure number, there is no clear mapping for readers to reference data for a specific chart and it might be a bit disorienting. Overall, it's an improvement, so I'm happy to land this change. I don't think we need to itemize each chart with its corresponding data sheet for 2019, but something to consider moving forward.

bazzadp:
Fair points - though not sure I 100% agree the average reader would rather have folders in chapter order rather than alphabetical order but don't feel strongly about ti one way or the other and, unlike images, can't redirect if it has been linked so happy to leave as is.

Anyway, as you say, this is an improvement as points them in the right direction. And think the summary at the end (near the discussion link) is quite a good way of doing it, so even if we add figure-specific links, way may want this anyway.

Merging this in now anyway!

borisschapira:
French is right here too, thanks @bazzadp

PR #729, opened 2020-04-12
Resize dropdown arrow image

bazzadp:
The dropdown arrow (used for Language selection) is massive! 8192 x 8192 pixels and 326KB despite the fact it's only displayed as 20 x 20 pixels.

Resized to 40 x 40 pixels so still looks good on retina screen and now 273 bytes.

Probably should replace with an inline SVG as it's such a simple image but this'll do for now. Plus it is used in a few places.


rviscomi:
😱

Good find and a much needed fix!

PR #731, opened 2020-04-13
Allow up to 200% font zooming for Accessibility

bazzadp:
This switches are font sizes to `rem`s (with fallback to `px` for those that don't support it) which respects the browsers base font (normally 16px) to allow users to increase the default font size for accessibility.

There is some debate in the a11y community as to whether browser zooming is sufficient ([relevant text](https://www.w3.org/WAI/WCAG21/Understanding/resize-text.html)), but testing this showed it didn't work on iOS anyway and we've set a high bar in our [accessibility statement](https://almanac.httparchive.org/en/accessibility-statement) so would like to support both screen zooming, and font-only zooming.

This also fixes an issues with the tab order of the footer on mobile (as we reorder the footer content) for that.

Also fixes some long running `code` text in the resource hints chapter that caused scroll on zooming.


bazzadp:
@HTTPArchive/developers any comments on this? Been open for 2 weeks now with no comments so will merge later today unless I hear back from any of you. Thanks!

PR #732, opened 2020-04-13
Improved index sidebar

### LINKED ISSUE #454: Improve index sidebar styles
Carrying over a few feature requests from #224. For pages with an index sidebar (chapter and methodology templates):

- [ ] make the index sticky on desktop so it's visible while scrolling (max-height 100% and overflow-y scroll)
- [ ] as the page is scrolled, update the "active" index item with a green highlight (using IntersectionObserver)

The "active" style is already implemented, we just have to apply the class name at the right time using JS.

cc @HTTPArchive/developers

bazzadp:
Fixes #454


src/static/js/chapter.js
```diff
@@ -256,6 +256,102 @@ function setDiscussionCount() {
...
+          indexScroller.classList.add('page-height');
+
+          var indexItems = {};
+          var currentItem;
+
+          // Create a function to handle highlighting a new index item
+          // that will be called y bthe IntersectionObserver
+          function highlightIndexEntry(link) {
+            
+            var oldItem = currentItem;
```
rviscomi, line 288: A simpler way to add/remove the `active` class, without the bookkeeping:

```js
document.querySelector('.index-scroller .active').classList.remove('active');
var index = document.querySelector(`.index-scroller a[href="#${link}"]`).parentNode;
index.classList.add('active');
```

src/static/js/chapter.js
```diff
@@ -256,6 +256,102 @@ function setDiscussionCount() {
...
+            var newItem = indexItems[link];
+            if (newItem){
+              newItem.classList.add('active');
+              if (oldItem && oldItem !== newItem) {
+                oldItem.classList.remove('active');
+              }
+              currentItem = newItem;
+
+              // If the index is too large to display in full then might need to change scroll
+              if (indexScroller.scrollHeight > indexScroller.clientHeight) {
```
rviscomi, line 298: A simpler way to scroll to the active index item:

```js
index.scrollIntoView({behavior: 'smooth', block: 'center'})
```

We can do without the smooth scrolling if needed.

src/static/js/chapter.js
```diff
@@ -256,6 +256,102 @@ function setDiscussionCount() {
...
+            }
+          }
+
+          // Set up a new Interstection Observer for when the title is 80% from the bottom of the page
+          var options = {
+            root: null,
+            rootMargin: "0px 0px -80% 0px",
+            threshold: null
+          };
+          var observer = new IntersectionObserver(function(entries) {
```
rviscomi, line 320: I may be wrong but I think instead of one IntersectionObserver in which we handle the logic of determining the visible heading, we should create separate IOs for each heading so that whenever one becomes visible it triggers its callback function.

src/templates/base/2019/base_chapter.html
```diff
@@ -212,11 +212,13 @@
 <main id="chapter" class="main">
     <nav aria-label="{{ self.index_nav_title() }}" class="index">
         <div class="index-box floating-card">
-          <h2 class="header">
-            <button class="index-btn" aria-expanded="false" aria-label="{{ self.open_the_index() }}">{{ self.index_title() }}</button>
-            <span class="no-button">{{ self.index_title() }}</span>
-          </h2>
-          {{ self.index() }}
+          <div class="index-scroller">
```
rviscomi, line 215: Is this extra div needed? Could we add the `index-scroller` class to the `index-box` element?

bazzadp: Lost the rounded corners as scrollbar only has box corners.

src/static/js/chapter.js
bazzadp: I was trying to avoid the querySelector call for perf reasons. And you have two here. Think I'm over-worrying about that?

src/templates/base/2019/base_chapter.html
rviscomi: Ok good catch

src/static/js/chapter.js
rviscomi: I don't see any cause for concern with these selectors.

src/templates/base/2019/base_chapter.html
bazzadp: Chrome for example:

image

image

Though now I look at it again I think the straight edge is better as scrollbar is flush with right hand side. Will change it.

src/static/js/chapter.js
bazzadp: I don't think that's necessary: https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/observe but not an expert so will change if you really think this is needed. It does work as is in all browsers I've tested though (Chrome, Safari, Firefox, Edge, iOS Safari)

src/static/js/chapter.js
bazzadp: Unfortunately can;'t use this as it scrolls the current element - which is fun when you have two scrolling elements like we do here! :-(

So works fine when clicking on the links in ToC, but not if focus is in main chapter text.

bazzadp:
Thanks for the awesome feedback @rviscomi - much cleaner code now and easier to read!

Unfortunately couldn't use `scrollIntoView` as that appears to only work to scroll current element and not the ToC when focus is in chapter text. Pity as it was much cleaner that way!

src/static/js/chapter.js
```diff
@@ -256,6 +256,94 @@ function setDiscussionCount() {
...
+  chapterIndex.classList.add('page-height');
+
+  // Create a function to handle highlighting a new index item
+  // that will be called by the IntersectionObserver
+  function highlightIndexEntry(link) {
+    
+    var indexLink = document.querySelector('.index-box a[href="#' + link + '"]');
+    var oldIndexLink = document.querySelector('.index-box .active');
+
+    if (indexLink && indexLink != oldIndexLink) {
```
rviscomi, line 297: Invert and return to avoid nesting

Also using `isEqualNode`. Not sure if totally necessary but better to use builtins to be safe.

```suggestion
    if (!indexLink || indexLink. isEqualNode(oldIndexLink)) {
      return;
    }
```

src/static/js/chapter.js
```diff
@@ -256,6 +256,94 @@ function setDiscussionCount() {
...
+      if(oldIndexLink) {
+        oldIndexLink.classList.remove('active');
+      }
+      indexLink.parentNode.classList.add('active');
+
+      // If the index is too large to display in full then might need to change scroll
+      if (chapterIndex.scrollHeight > chapterIndex.clientHeight) {
```
rviscomi, line 304: Invert/return

src/static/js/chapter.js
```diff
@@ -256,6 +256,94 @@ function setDiscussionCount() {
...
+  }
+
+  // Check if 'position:sticky' is supported (as this is not great UX when not so don't bother)
+  // Add the sticky class (which sets 'position:sticky') and then test if that stuck :-)
+  // Also use endsWith to support vendor prefixes (Safari v12 needs this)
+  chapterIndex && chapterIndex.classList.add('sticky');
+  var chapterIndexStyles = getComputedStyle(chapterIndex);
+  if (!chapterIndexStyles || !chapterIndexStyles.position || !chapterIndexStyles.position.endsWith('sticky')) {
+    gtag('event', 'index-highlighter', { 'event_category': 'user', 'event_label': 'not-enabled', 'value': 0 });
+    return
```
rviscomi, line 284:
```suggestion
    return;
```

rviscomi:
I got `scrollIntoView` working locally but as you say that might just have to do with where focus is at the time. I think we could get it working by temporarily moving focus to the index, but your solution is already working fine so I'm ok to leave it.

bazzadp:
> I got `scrollIntoView` working locally but as you say that might just have to do with where focus is at the time. I think we could get it working by temporarily moving focus to the index, but your solution is already working fine so I'm ok to leave it.

OK found the answer [on Stackoverflow](https://stackoverflow.com/a/60555370/2144578). Basically any new scroll events that fire will interrupt `scrollIntoView` stop it from happening. So if you scroll using keyboard, or very slowly, it works. If you use a mouse scroll wheel (as I was doing!) or scroll fast it doesn't work. My solution doesn't suffer from this issue as doesn't use `scrollIntoView` so will stick with that.

Fixed those other comments so can you have another look and let me know if you spot anything else?

PR #735, opened 2020-04-15
Stop showing scrollbars when not needed for code blocks

bazzadp:
Not sure how I never noticed this before but code blocks have a scroll bar when not needed:

Scrollbars showing due to bad overflow setting

I presume this is not intentional @rviscomi ?


rviscomi:
Hadn't noticed that! Good find/fix

PR #737, opened 2020-04-18
Ref links for chapter headings and figures

### LINKED ISSUE #733: Make headings and figures more easily referenced with anchor link UI
We're intending the Web Almanac to be a a quotable resource. However we don't make it easy for people to link to specific parts of a chapter.

It's common to have links to chapter headings and we discussed this in #199 and resolved that by creating the links in #257, but didn't expose this to the user, as suggested in #199 .

GitHub uses this for markdown pages, with a link icon that appears before the chapter heading on hover (and appears all the time on mobile). For example I've hovered over Expensing queries in below example from our [Analysts Guide](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)

GitHub adds pre-anchor links to headings on hover

As does web.dev, with a # icon that appears before the chapter heading on hover (and on tap on mobile). for example I've hovered over Field Data in below example from their [How to Measure Speed article](https://web.dev/how-to-measure-speed/#field-data).

Web.Dev adds post-anchor links to headings on hover

Clicking on both of these simple anchor links, changes the URLs in the URL bar to the direct link to that anchor for easy copy and pasting. Not sure if this is the most obvious UX, but it is a fairly common pattern - though link icons like /GitHub uses rather than # icons like web.dev is more common (perhaps the # is more reflective of web/dev's technical readership?).

As an alternative, I was certain MDN used to do it to, but when you clicked on it on desktop it offered a "Copy to Clipboard", but can't reproduce this now so maybe I imagined it? Or maybe it was it another site? Probably simple anchor links are more widely understood despite not being the most obvious to those not familiar with this pattern.

Also not sure whether should show on hover or all the time - not the most obvious that this functionality exists when only done on hover or tap. Definitely need to think about accessibility here and make sure UX is keyboard accessible. Maybe if worried about it taking from the design, then we do only show on hover/tap/focus?

Now with the ToC (and especially if we land #732 ) maybe this isn't such a big deal - Wikipedia doesn't do this, but is a heavily cited source. However I think it's a good thing and think we should implement it before 2020 edition is released.

Also while the ToC is good for chapter headings, the figures also have anchor links that aren't exposed at all, except with someone inspecting the HMTL. We should expose them in similar way.

bazzadp:
Fixes #733 

Heading with no hover or focus:

image

Heading with hover or focus:

image

Figcaption with no hover or focus:

image

Figcaption with hover or focus:

image

Edit: @catalinred provided a nice scalable SVG icon! Screenshots updated.

Note I've only included one sample chapter to make the review eaiser. GitHub actions will generate the rest.


src/content/en/2019/markup.md
```diff
@@ -37,7 +37,7 @@ Instead, as in Hixie's original study,  what we will look at is how many sites i
 
 In 2005, Hixie's survey listed the top few most commonly used elements on pages.  The top 3 were `html`, `head` and `body` which he noted as interesting because they are optional and created by the parser if omitted.  Given that we use the post-parsed DOM,  they'll show up universally in our data.  Thus, we'll begin with the 4th most used element. Below is a comparison of the data from then to now (I've included the frequency comparison here as well just for fun).
 
-<figure id="fig1" data-markdown="1">
+<figure id="fig-1" data-markdown="1">
```
rviscomi, line 40: Why are the fig IDs changing? I'm concerned about breaking existing links.

src/static/css/page.css
```diff
@@ -295,6 +296,46 @@ img {
   max-width: 100%;
   height: auto;
 }
+
+article h2 a,
+article h3 a,
```
rviscomi, line 301: There's a lot of repetition. Could we target the anchors more directly using a class selector, so we don't need the ancestry tree?

src/tools/generate/generate_chapters.js
```diff
@@ -51,6 +51,9 @@ const parse_file = async (markdown,chapter) => {
   const html = converter.makeHtml(markdown);
   let body = html;
 
+  // Make headers links
+  const header_regex = /<h([2-6]) id="(.*?)">(.*?)<\/h[2-6]>/gi;
```
rviscomi, line 55: Will this work if there are other attributes in the header?

src/tools/generate/generate_figure_ids.js
```diff
@@ -1,8 +1,10 @@
 const generate_figure_ids = (html) => {
-  const re = /<figure>|<figure markdown>/gi;
+  const re = /<figure>|<figure data-markdown>/gi;
 
   let i = 1;
-  return html.replace(re, () => `<figure id='fig-${i++}'>`);
+  html = html.replace(re, () => `<figure id='fig-${i++}'>`);
+  const figcaption_regex = /<figcaption(.*?)>(.*?)([0-9]+)\./gi;
```
rviscomi, line 6: Haven't tested but should the .* be changed to [^>]*?

src/content/en/2019/markup.md
bazzadp: For two of the chapters the fig ids were written in the markdown. For the rest they were generated in the script/. Unfortunately the ones which were handwritten were of a slightly different format and I need them to be the same format as really don't want to do a multi-line regex.

Will add the old links on the img itself to preserve old links. Bit messy but only two chapters.

src/static/css/page.css
bazzadp: Good idea - can add the class when I add the link.

src/tools/generate/generate_chapters.js
bazzadp: The headers are generated from markdown so currently don't have any other attributes other than the id (which is auto generated by showndown).

In theory they could if someone uses HTML for the headers, but would rather keep the regex simple that complicate it with before and after bits for the id, so I say we leave as is and revisit it we need to.

WDYT?

src/tools/generate/generate_figure_ids.js
bazzadp: Which one? I've used non-greedy (`.*?`) and seems to work.

src/static/css/page.css
bazzadp: Fixed!

src/content/en/2019/markup.md
bazzadp: OK changed it to handle this:

```javascript
  //Some of our 2019 chapters (markdown and performance) had different fig ids
  const figcaption_regex = /<figcaption(.*?)>(.*?)([0-9]+)\./gi;
  if (html.includes('<figure id="fig1"')) {
    return html.replace(figcaption_regex, "<figcaption$1><a href=\"#fig$3\" class=\"anchor-link\">$2 $3.</a>");
  } else {
    return html.replace(figcaption_regex, "<figcaption$1><a href=\"#fig-$3\" class=\"anchor-link\">$2 $3.</a>");
  }
```

bazzadp:
@catalinred provided a nice scalable SVG icons! Screenshots in first comment updated.

src/static/css/page.css
```diff
@@ -133,6 +133,7 @@
 }
 .content > section, .content > article {
   margin-bottom: 32px;
+  margin-left: 10px;
```
rviscomi, line 136: What's the purpose of this style change?

src/static/css/page.css
```diff
@@ -295,6 +296,28 @@ img {
...
+  text-decoration: underline;
+}
+
+.anchor-link:hover::before,
+.anchor-link:focus::before {
+  content: url('/static/images/link.svg');
+  content: url('/static/images/link.svg') / '';
+  left: -30px;
+  position: relative;
+  float: left;
```
rviscomi, line 316: Does float apply here?

src/templates/base/2019/accessibility_statement.html
```diff
@@ -35,12 +35,14 @@ <h2 class="header">
     </div>
   </nav>
 
-  <article id="maincontent" class="content">
-    <h1 class="title title-lg">Accessibility Statement</h1>
-      
-    {{ self.main_content() }}
-
-  </article>
+  <div class="content">
```
rviscomi, line 38: This looks redundant, having two .content elements. Why do we need the extra div?

src/tools/generate/generate_chapters.js
```diff
@@ -51,6 +51,9 @@ const parse_file = async (markdown,chapter) => {
...
+  body = body.replace(header_regex, "<h$1 id=\"$2\"><a href=\"#$2\" class=\"anchor-link\">$3</a></h$1>");
```
rviscomi, line 56: Nit: "anchor-link" could be more descriptive, eg "heading-permalink"
Nit: mixed use of quotes in this file, use single quotes for consistency (and avoids the need to escape double quotes)

rviscomi, line 56: Implement this in an external function for consistency with other transformations

src/tools/generate/generate_figure_ids.js
```diff
@@ -1,8 +1,16 @@
...
+  const re = /<figure>|<figure data-markdown>/gi;
 
   let i = 1;
-  return html.replace(re, () => `<figure id='fig-${i++}'>`);
+  html = html.replace(re, () => `<figure id='fig-${i++}'>`);
+
+  //Some of our 2019 chapters (markdown and performance) had different fig ids
+  const figcaption_regex = /<figcaption(.*?)>(.*?)([0-9]+)\./gi;
+  if (html.includes('<figure id="fig1"')) {
+    return html.replace(figcaption_regex, "<figcaption$1><a href=\"#fig$3\" class=\"anchor-link\">$2 $3.</a>");
```
rviscomi, line 10: Nit: use single quotes for strings

src/static/css/page.css
bazzadp: Thought it looked a little close to the edge without it:

image

So this adds a little breathing space:

image

src/tools/generate/generate_chapters.js
bazzadp: I use the same class for figures. So heading-or-figure-permalink. Seems a bit long and not sure I agree that permalink is more descriptive (or accurate) than anchor-link tbh. So will leave as is unless you feel strongly about this.

Will fix the quotes.

src/static/css/page.css
bazzadp: Yes without this it reserves some space evem with the `width: 0` after this line - to be honest, I'm not sure why, but float fixes it:

image

src/templates/base/2019/accessibility_statement.html
bazzadp: Will remove the class from the article.

The extra div is needed to make it similar to chapters so it gains the 10px margin-left mentioned about:

```css
.content > section, .content > article {
  margin-bottom: 32px;
  margin-left: 10px;
}
```

Chapters need the two divs because they have the article, and then the extra info (discussion/query/results links, author links, prev/next). Technically the accessibility page doesn't need this but would prefer to add for consistency, rather than treat specially.

The Methodology page doesn't use this extra div, but works because it's divided into sections which aren't really needed (the chapters are not divided into `<sections>`), and are a bit of an anomaly compared to chapters. Now you've pointed this out I think I'd like to clean this up as part of this and add the extra div to methodology as well and remove the `<sections>` - again to be consistent with chapters. Plus `<section>` is fairly meaningless to be honest: https://www.smashingmagazine.com/2020/01/html5-article-section/

> In most screen readers, a `<section>` element isn’t announced unless it has an accessible name. In this case, the text of the aria label.
>
>...
>
>`<section`> can be used with aria-label to signal to a screen reader user where a particular sub-part of an article begins and ends. Otherwise, forget about it, or use another element, such as `<aside aria-label="quick summary">` or `<div role=”region” aria-label="quick summary">`.

src/static/css/page.css
bazzadp: Appreciate this does move the article in 10px but think a small price to pay for this useful functionality.

src/tools/generate/generate_chapters.js
bazzadp: I added a bit more to this to make sure we at least don't delete attributes:

```javascript
  const header_regex = /<h([2-6]) id="(.*?)"(.*?)>(.*?)<\/h[2-6]>/gi;
  return html.replace(header_regex, '<h$1 id="$2"$3><a href="#$2" class="anchor-link">$4</a></h$1>');
```

src/templates/base/2019/accessibility_statement.html
bazzadp: Changed my mind and added `<section>` elements to accessibility statement to bring it into line with Methodology. So have removed the extra div.

src/tools/generate/generate_chapters.js
bazzadp: Fixed

src/tools/generate/generate_figure_ids.js
bazzadp: Fixed. Force of habit due to my Perl days where single quotes and double quotes are different (variables are not evaluated in single quotes in Perl). Will try to remember JavaScript is different! 😀

src/static/css/page.css
bazzadp: Moved this to a better location (which also means it doesn't affect mobile) in https://github.com/HTTPArchive/almanac.httparchive.org/pull/737/commits/a947f817ee11bd3db15aa75abeaa750586ebe466

src/static/css/page.css
bazzadp: Ah figured it out!

The `::before` pseudo-element is an inline element by default, and I stupidly forgot the `width` is applied to the whole pseudo-element and not the `content` (which is an image). So, without setting `float: left;`, the `width` is ignored (as element is inline) and hence the extra spacing.

So we either need `float: left;` to implicitly covert into a block element or I can set `display: inline-block;` to so similar. The advantage of `float: left;` is that it takes it out of the flow completely, so it is ignored when the link image is slightly larger than the text (like it is for figcaptions), so it doesn't change the height of the link element and cause a slight movement on hover.

So yes you need `float:left;` and `left:-30px;` (which I since changed to `left: -25px;` btw as looks nicer).

bazzadp:
Any more thoughts on this @rviscomi ? Am I good to merge or do you want some more time to review?

PR #743, opened 2020-04-22
Fix line numbers

borisschapira:
@AymenLoukil In order to ease the comparison between EN and FR, here's a fix on the line numbers. I still have to proofread in detail, but at least it allows me to put the two versions side by side.


bazzadp:
Any chance we could leave out the chapter generation and sitemap @borisschapira ? Just it'll get out of sync unless @AymenLoukil also regenerates before we merge, which then means GitHub action then complains when subsequently tries to regenerate it when we merge his PR into master.

Not a massive deal if you'd prefer to leave it in, as that GitHub action will fail and it'll just be picked up on next one, but cleaner to either not have it at all, or to ensure it matches markdown when merged.

borisschapira:
@bazzadp Not a problem. Done.

bazzadp:
Thanks. @AymenLoukil you happy to merge this into your branch? Probably best to do that before you make any other changes that this might overwrite.

PR #753, opened 2020-04-24
Remove lastmod from sitemap

### LINKED ISSUE #752: Remove lastmod from SiteMap
As discussed on Slack with @rviscomi and @AymenLoukil , until we fix #317 the `lastmod` date in the Sitemap is basically unreliable. As it's optional we should just remove it for now.

BTW relevant: https://www.seroundtable.com/google-lastmod-xml-sitemap-20579.html

bazzadp:
Fixes #752

### DIFF
--- a/src/templates/sitemap.ejs.xml
+++ b/src/templates/sitemap.ejs.xml
@@ -3,7 +3,6 @@
 <% for (let url of urls ) { %>
     <url>
         <loc>https://almanac.httparchive.org/<%= url.url %></loc>
-        <% if (url.lastmod) { %><lastmod><%= url.lastmod %></lastmod><% } %>
     </url>
 <% } %>   
 </urlset>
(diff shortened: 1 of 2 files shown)


AymenLoukil:
Hi @bazzadp. 
LGTM, we should also see how to have the real dates for the structured data markup. 

For the new translated chapter we have : 
https://search.google.com/structured-data/testing-tool?hl=fr#url=https%3A%2F%2Falmanac.httparchive.org%2Ffr%2F2019%2Fseo
datePublished | 2019-11-11T00:00:00
dateModified | 2019-11-23T00:00:00

bazzadp:
I actually thought it was mandatory in there, so left it in, but looks like it's optional. Will add to this PR.

bazzadp:
Ah it does show a warning if not provided:

image

So I'd rather leave it in just to remove the noise of that warning.

PR #757, opened 2020-04-25
Translate "and" and show correct localized chapter name

bazzadp:
Cleans up a few places where we were still using English.

@ksakae1216 has Google Translate chosen the right Japanese word for "and" here or is there a better one?

```
{% block and %}そして{% endblock %}
```

It appears when we have multiple authors:

image

Currently that's incorrectly showing as "and":

image


ksakae1216:
@bazzadp This is no good, And this has 2 problems.

The following is correct.
↓↓↓
Una Kravets と Adam Argyle によって書かれた。

src/templates/ja/2019/base.html
```diff
@@ -41,6 +41,8 @@
 {% block open_the_index %}目次を開く{% endblock %}
 {% block close_the_index %}目次を閉じる{% endblock %}
 
+{% block and %}そして{% endblock %}
```
ksakae1216, line 44: `{% block and %} と {% endblock %}`

ksakae1216: This is correct.

bazzadp:
```suggestion
{% block and %} と {% endblock %}
```

bazzadp:
Like this?

image

bazzadp:
Or even this?
image

bazzadp:
OK will go with the second screenshot.

PR #768, opened 2020-05-02
Force push generate action

### LINKED ISSUE #767: Generate GitHub action broken due to permission changes
@rviscomi we now get this when generate GitHub action runs ([see here](https://github.com/HTTPArchive/almanac.httparchive.org/runs/638419637?check_suite_focus=true)):

```
Commiting changes
[master 177796e] Regenerated chapters
 3 files changed, 499 insertions(+), 1 deletion(-)
 create mode 100644 src/templates/ja/2019/chapters/media.html
remote: error: GH006: Protected branch update failed for refs/heads/master.        
remote: error: You're not authorized to push to this branch. Visit https://help.github.com/articles/about-protected-branches/ for more information.        
To https://github.com/HTTPArchive/almanac.httparchive.org
 ! [remote rejected] master -> master (protected branch hook declined)
error: failed to push some refs to 'https://github.com/HTTPArchive/almanac.httparchive.org'
##[error]Process completed with exit code 1.
```

This is even though this runs under my id.

This is because even those with write permissions to master (i.e. you and I) need to approve changes and that doesn't work with GitHub actions.

Can you and I be given permission to commit directly to master?

bazzadp:
Fixes #767 

@rviscomi I added a `--force` to the push and also [changed the rule setting to allow this](https://help.github.com/en/github/administering-a-repository/enabling-force-pushes-to-a-protected-branch).

It will still only work for those with push permissions to master (i.e. you and I) so think should be reasonably safe.

Feel free to close if you'd rather configure the rule to allow us to commit directly without force?


rviscomi:
Reluctant approval. I'd love to find a way to grant the GitHub action explicit permission to push.

What are the side-effects of `--force`? Fore example could it push something that's broken?

bazzadp:
On the one hand it's quite safe. We only allow these files to be committed in this action:

```
        git add src/templates/*/*/chapters/
        git add src/templates/sitemap.xml
```

These are generated files so even if we screwed them up completely we could generated.

However, there are two downsides:

1. We could overwrite any other changes to these files. For example if you merge two pull requests in quick succession and they both contained chapter template changes.

Now, we're often not bothering to include the chapter template files anymore (because of this action!) and in theory the second run of this action for the second merge would put back any changes but still - it's icky.

2. We could lose history of changes to above files when we do this. For same reasons as above.

So, it is kind of safe. But I agree not loving it.

However I think I've found a better way so going to close this and talk in the issue on that.

(Also looks like I created this branch from production branch rather than from master hence all the extra commits so don't really want to merge this anyway!)

PR #779, opened 2020-05-03
Don't cache redirects and errors

bazzadp:
Currently our 302 redirects (e.g. when a chapter doesn't exist) and other non-successfully error code (e.g. 404, 301...etc.) are cached for the same 3 hours that our main pages are cached for.

This changes it to not cache them so when a chapter is launched (inc Translations) it is available right away.

Also fixes a duplicate definition that VS Code flags as an error.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -255,7 +255,7 @@ def page_not_found(e):
 
 
 @app.errorhandler(500)
-def server_error_500(e):
+def handle_internal_server_error(e):
     logging.exception('An error occurred during a request due to internal server error: %s', request.path)
     return render_error_template(error=e, status_code=500)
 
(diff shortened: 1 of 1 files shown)


src/main.py
```diff
@@ -248,13 +255,13 @@ def page_not_found(e):
 
 
 @app.errorhandler(500)
-def server_error(e):
+def server_error_500(e):
```
rviscomi, line 258: While we're at it we can change the function name to be more descriptive.
```suggestion
def handle_internal_server_error(e):
```

src/main.py
```diff
@@ -248,13 +255,13 @@ def page_not_found(e):
...
     logging.exception('An error occurred during a request due to internal server error: %s', request.path)
     return render_error_template(error=e, status_code=500)
 
 
 @app.errorhandler(502)
-def server_error(e):
+def server_error_502(e):
```
rviscomi, line 264:
```suggestion
def handle_bad_gateway(e):
```

PR #781, opened 2020-05-03
SVG icons system #740

### LINKED ISSUE #740: Move simple PNG icons to SVG
As a follow on from #738, the below PNGs don't scale well on large screens, or when zoomed in. As they are all simple (or have Vector versions for GitHub/Twitter) we should move them to SVG format.

- [ ] discuss.png
- [ ] results.png
- [ ] queries.png
- [ ] dropdown-arrow.png
- [ ] close.png
- [ ] menu.png
- [ ] blog.png
- [ ] github.png - [Official link to Vector icons](https://github.com/logos)
- [ ] twitter.png - [Official link to SVG icons](https://github.com/logos).

Note some of these need a hover state version (the first 3 - but the dropdown-arrow, blog, github and twitter would also benefit from having this so maybe can solve at same time?).

catalinred:
Fixes #740 
SVG icons system with using `<symbol>`s:

- discuss, results, sql, also blog(globe), github and twitter are now inline SVG's
-  dropdown-arrow.png is replaced by a CSS pseudo elements that respects link `:hover` state
 queries.png
-  the close and menu PNG's were replaced by a `button` with animated vector bars/lines, inspired by one of my demos: https://github.com/catalinred/animenu
-  rectangles.png was replaced by a base64 SVG encoded background image.
- fix other inconsistencies on the fly.

Let me know if you have any questions, please.


src/templates/base/2019/base_chapter.html
```diff
@@ -61,13 +61,23 @@
 {# Calls to action for readers who want to engage more with this chapter. #}
 {% macro render_actions() %}
   <a class="alt btn" href="https://discuss.httparchive.org/t/{{ metadata.get('discuss') }}">
-    <img src="/static/images/discuss.png" alt="{{ self.discuss_this_chapter() }}" height="24" width="24" loading="lazy" /><span id="num_comments"></span> <span data-translation id="comment-singular">{{ self.comment() }}</span><span data-translation id="comment-plural">{{ self.comments() }}</span>
+    <svg width="18" height="18">
```
bazzadp, line 64: We've lost the alt tag here. Can we add it back with a title tag as otherwise VoiceOver reads the 4 digit link:
```suggestion
    <svg width="18" height="18">
    <title>{{ self.discuss_this_chapter() }}</title>
```

src/templates/base/2019/base_chapter.html
```diff
@@ -61,13 +61,23 @@
...
+      <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#comment"></use>
+    </svg>    
+    <span id="num_comments"></span> <span data-translation id="comment-singular">{{ self.comment() }}</span>
+    <span data-translation id="comment-plural">{{ self.comments() }}</span>
   </a>
   <a class="alt btn" href="{{ metadata.get('results') }}/">
-    <img src="/static/images/results.png" alt="" height="24" width="24" loading="lazy" />{{ self.results() }}
+    <svg width="18" height="18">
```
bazzadp, line 71: As this image is decorative we need to hide it since we are not giving a description.
```suggestion
    <svg width="18" height="18" aria-hidden="true">
```

src/templates/base/2019/base_chapter.html
```diff
@@ -61,13 +61,23 @@
...
+      <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#bar-chart"></use>
+    </svg>       
+    {{ self.results() }}
   </a>
   <a class="alt btn" href="https://github.com/HTTPArchive/almanac.httparchive.org/tree/master/sql/{{ year }}/{{ metadata.get('queries')  }}/">
-    <img src="/static/images/queries.png" alt="" height="24" width="24" loading="lazy" />{{ self.queries() }}
+    <svg width="18" height="18">
```
bazzadp, line 77: As this image is decorative we need to hide it since we are not giving a description.
```suggestion
    <svg width="18" height="18" aria-hidden="true">
```

src/templates/base/2019/base.html
```diff
@@ -213,17 +227,23 @@
         <a href="{{ url_for('accessibility_statement', lang=lang) }}">{{ self.accessibility_statement() }}</a>
       </p>
       <a class="navigation-logo ha-logo not-mobile" href="https://httparchive.org/">
-        <img src="/static/images/ha-home.svg" alt="{{ self.http_archive_link() }}" height="35" width="70">
+        <svg width="70" height="35" aria-label="{{ self.http_archive_link() }}">
```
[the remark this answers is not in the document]
catalinred: I clearly missed this one. I'll fix it. 10x for the heads up here.

catalinred:
> Wow - that's some seriously impressive work @catalinred !!
> 
> Does make the HTML slightly bigger for every page, but with compression I don't think it will really be noticeable and the benefit of not waiting for images to download - particularly for the drop down arrow should compensate.
> 
> I've a couple of minor comments about some of the labelling of images for accessibility but overall am loving this!

Thanks! :)
I've added your suggestions and moved the `aria-label` to the above link.

bazzadp:
Thanks. Made a few more suggestions in this pull request: https://github.com/catalinred/almanac.httparchive.org/pull/1

bazzadp:
@catalinred I was waiting for you to accept or reject https://github.com/catalinred/almanac.httparchive.org/pull/1 before I re-reviewed this.

catalinred:
> @catalinred I was waiting for you to accept or reject [catalinred#1](https://github.com/catalinred/almanac.httparchive.org/pull/1) before I re-reviewed this.

Merged, thank you!

I was aware of this approach but I was never a big fan as it's too verbose if you ask me. I always considered `aria-label` is a good alternative to using the title/aria-labelledby pairs.

bazzadp:
Yeah it shouldn’t be needed I agree. But Firefox seems unhappy without it and trying to keep those accessibility audits as clean as possible to reduce risk of missing a real issue!

Though others say to avoid `aria-label` if possible, and to prefer `aria-labelledby` for translation reasons, which is another reason to use this (admittedly verbose) technique: https://adrianroselli.com/2019/11/aria-label-does-not-translate.html

Anyway thanks for merging so I’m happy with this now and will merge and release in a bit.

bazzadp:
That’s deployed. Add a random query param to the URL to cache bust if you’re still seeing the old page for next 3 hours.

Thanks so much for all the work on this! Really pleased with it and learned a few things! Will be interesting if our performance metrics improve because of this.

PR #782, opened 2020-05-03
Time to Interactive is NOT the time until a page is interactive

borisschapira:
The definition of the Time To Interactive is very complex and has been wrongly advertised in the Lighthouse documentation with the sentence:

> The Time to Interactive (TTI) metric measures how long it takes a page to become interactive.

I have said many times that this was completely wrong and the definition found nowadays on Web.dev is much more accurate:

> The TTI metric measures the time from when the page starts loading to when its main sub-resources have loaded and it is capable of reliably responding to user input quickly.

The subtlety here is to understand that a page can be interactive, even with nearly no freeze, before the TTI. The TTI only shows the moment when the last slowdown occurred. But we don't know if it's the first, or the hundredth, and the duration of these slowdowns.

I tried in this PR to modify the paragraph which relies on the TTI to deduce the non-usability of the page so that it is less clear-cut because in reality, a high TTI does not imply that the page is not usable.

I'll suggest @obto, @AymenLoukil and @logicalphase as reviewers because they wrote and reviewed this chapter.

See https://boris.schapira.dev/2019/05/measuring-interactivity-time-to-interactive/#disambiguation for a full explanation.


logicalphase:
I'll take a look and discuss. 👍

src/content/en/2019/mobile-web.md
```diff
@@ -55,7 +55,7 @@ I imagine some of you are surprised by these results. They may be far worse cond
...
 
-Why is this a problem? Because sites loading this much JS take upwards of [10 seconds](https://httparchive.org/reports/loading-speed?start=earliest&end=2019_07_01&view=list#ttci) to become interactive. Or in other words, your page may appear fully loaded, but when a user clicks any of your buttons or menus, nothing happens because the JavaScript hasn't finished executing. Users are forced to keep clicking the button for upwards of 10 seconds, just waiting for that magical moment where something actually happens. Think about how confusing and frustrating that can be.
+Why is this a problem? Because sites loading this much JS take upwards of [10 seconds](https://httparchive.org/reports/loading-speed?start=earliest&end=2019_07_01&view=list#ttci) to become consistenly interactive. Or in other words, your page may appear fully loaded, but when a user clicks any of your buttons or menus, the user may experience some junk because the JavaScript hasn't finished executing. Users may be forced to keep clicking the button for upwards of 10 seconds, just waiting for that magical moment where something actually happens. Think about how confusing and frustrating that can be.
```
bazzadp: Did you mean “junk” or jank? Assuming it’s jank then think we should have a link to a definition for this term?

bazzadp: Never mind. See you updated. Resolving this.

src/content/en/2019/mobile-web.md
```diff
@@ -55,7 +55,7 @@ I imagine some of you are surprised by these results. They may be far worse cond
...
+Why is this a problem? Because sites loading this much JS take upwards of [10 seconds](https://httparchive.org/reports/loading-speed?start=earliest&end=2019_07_01&view=list#ttci) to become consistenly interactive. Or in other words, your page may appear fully loaded, but when a user clicks any of your buttons or menus, the user may experience some slowdown because the JavaScript hasn't finished executing. In the worst case scenario (assuming that the slowdown is due to a single CPU-intensive task that would last up to the 10th second), users may be forced to keep clicking the button for upwards of 10 seconds, just waiting for that magical moment where something actually happens. Think about how confusing and frustrating that can be.
```
rviscomi, line 58: Let's work this edit down to the smallest footprint possible while still fixing the technical error. I'd like to keep as much of this in the authors' own words and minimize any extra translation overhead.

Is it sufficient to only change "interactive" to "consistently interactive"?

obto: First of all just want to say big thanks for the feedback. Really appreciate it :)

I like changing `interactive` to `consistently interactive`, but while the added parenthesis of information is technically true... I feel it makes the paragraph more confusing and harder to digest.

What about adding a link to more information on this particular metric down below?

borisschapira: There are 4 propositions in this paragraph. Assuming we only add "consistently" to the first one.

> this much JS take upwards of 10 secondes to become **consistently** interactive

Makes sense.

> when a user clicks any of your buttons or menus, nothing happens

You still can't infer that from a high TTI.

> Users are forced to keep clicking

You still can't infer that from a high TTI.

> Think about how confusing and frustrating that can be.

Or maybe it isn't. This is a page with a 21-second TTI and you can scroll and input text.

The TTI is a "Time To Last Long Task". You don't know what happened before, and can't say the page was not interactive.

borisschapira: The only way a user can't interact with a 10-second TTI page is if the page executes one 10-second JS long task (and as the TTI is 10 seconds, this would be the only long task). So we need more nuance here, like "interact satisfactorily" or "interact well", or "user may feel the urge to click"… because otherwise, what's written here is untrue.

borisschapira: And we could also switch from TTI to First CPU Idle ([TTFI](https://httparchive.org/reports/loading-speed?start=earliest&end=2019_07_01&view=list#ttfi)) which is stil discutable – the window of quiescence is large – but more adequate than TTI for this.

bazzadp: > I like changing `interactive` to `consistently interactive`, but while the added parenthesis of information is technically true... I feel it makes the paragraph more confusing and harder to digest.

What about removing the parenthesis bit? I agree it does make the sentence quite wordy and not sure it's needed to be honest – it does say in the worst case in the revised wording! So below suggestion would be a middle ground somewhere between @rviscomi 's minimal edit, and @borisschapira 's full one:

```suggestion
Why is this a problem? Because sites loading this much JS take upwards of [10 seconds](https://httparchive.org/reports/loading-speed?start=earliest&end=2019_07_01&view=list#ttci) to become consistently interactive. Or in other words, your page may appear fully loaded, but when a user clicks any of your buttons or menus, the user may experience some slowdown because the JavaScript hasn't finished executing. In the worst case scenario, users may be forced to keep clicking the button for upwards of 10 seconds, just waiting for that magical moment where something actually happens. Think about how confusing and frustrating that can be.
```

BTW this is not a sleight @borisschapira – overly wordy sentence is something I'm very, very guilty of myself!! 😀

> What about adding a link to more information on this particular metric down below?

We could, but my preference would be not to add more links that the author didn't include originally (even though I loved your piece @borisschapira !) unless really necessary. We're bordering on trying to keep this up to date, when I think that's what future year's editions are for.

borisschapira: > We're bordering on trying to keep this up to date

Actually, I was going to translate this part and I can't, because I can't translate s.t. I think is not true.

But I'm far from being the only person who can translate this to French, so I'll close this PR if you think I'm too demanging (I know I can be).

bazzadp: So you don’t think my compromise suggestion is sufficient to address your concerns? Especially with your “in the worst case” addition?

bazzadp: Reworded my last comment to be less confrontational! Wasn’t intended as such but in re-reading it, I thought better to reword it to make that clear.

borisschapira: Yes, your suggestion is perfect, I wanted only to comment this part.

bazzadp:
@obto let us know if that latest iteration is an acceptable compromise to you?

Will hold off merging for a few days to give you a chance to respond one way or another.

obto:
Sorry for the delay! The latest version looks good to me.

@borisschapira Thanks again for pointing this out and I think this kind of discussion could be really interesting to bring up in the mobile web chapter this year. Even a section talking about the updates that have been made to various metrics and which have been replaced as we've come up with better ones would be fantastic.

bazzadp:
Merged and will be included in the next release.

@ksakae1216 could you make the same change in the Japanese version?

PR #783, opened 2020-05-03
Generate chapters action fix

### LINKED ISSUE #317: Autogenerate timestamps

### LINKED ISSUE #767: Generate GitHub action broken due to permission changes

bazzadp:
Fixes #767
Fixes #317 

This changes the generate chapters github action to open a pull request, rather than try to merge directly to master.

This fixes the permission issue (#767) since we locked down the master branch recently.

It also allows us to take more risks with changes and add timestamps, since we will have a chance to review the changes before committing them.

We stick with day timestamps, to avoid a recursion (providing you merge the generated pull request on the same day - otherwise you'll just get a new pull request the next day).


bazzadp:
Seems to work well, except the project can't be set. Have raised an issue: https://github.com/peter-evans/create-pull-request/issues/243 but we can live without it and set it manually.

PR #784, opened 2020-05-03
Update lastmod dates

bazzadp:
Makes progress on #317 

Do a once off sync of all the last updated dates to master, before we fix this going forward in #783


bazzadp:
Going to merge this as just date updates (and a small change to code that is not used in case we ever want to generate dates from git in future).

PR #786, opened 2020-05-04
Add support for the old grid spec for IE

bazzadp:
I still test most releases in IE11 – mostly cause I figure if it works there, it'll work anywhere!

It's always bugged me that the layout is shockingly bad because IE does not support grid. The content (the important part!) is at least readable so we can think of grid as a progressive enhancement but it still annoys me. The home page graphics are in one column, and the chapters have the Index if full screen and taking up all the space at the top.

And, as I'm trying to move to CSS Grid at work (where we still have to support IE11 for... reasons), I've been looking into this and it's quite easy to fix with a few small fallbacks. So, even though our IE user base is very small (< 2%), this bugs me and the fix is also small so here it is.

Note this is not a commitment to supporting IE going forward, nor an expectation that other developers should have to support it. Just it bugged me, so I fixed it.

Relevant links:
- https://medium.com/@elad/supporting-css-grid-in-internet-explorer-b38669e75d66
- https://rachelandrew.co.uk/archives/2016/11/26/should-i-try-to-use-the-ie-implementation-of-css-grid-layout/

### DIFF
--- a/src/static/css/index.css
+++ b/src/static/css/index.css
@@ -33,6 +33,7 @@ header .container {
 
 #maincontent {
   overflow: hidden;
+  display: -ms-grid;
   display: grid;
   grid-template-areas:
     'intro'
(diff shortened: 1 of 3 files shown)


bazzadp:
Yeah we can revisit if we move to that. More an irritant than a real problem so won’t push for this as it causes other problems.

PR #800, opened 2020-05-05
Bump CSS version numbers

bazzadp:
FYI @catalinred need to remember to bump version numbers when making breaking changes to CSS in case browser (or more likely the CDN)  has old version cached as we cache for 3 hours.

### DIFF
--- a/src/templates/base/2019/accessibility_statement.html
+++ b/src/templates/base/2019/accessibility_statement.html
@@ -20,7 +20,7 @@
 
 {% block styles %}
 {{ super() }}
-<link rel="stylesheet" href="/static/css/page.css?v=2">
+<link rel="stylesheet" href="/static/css/page.css?v=4">
 {% endblock %}
 
 {% block main %}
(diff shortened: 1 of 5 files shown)


rviscomi:
@bazzadp do you have any ideas for ways that we can improve/automate this process?

bazzadp:
Funny was just thinking we could add a GitHub action for this when we commit a css file - similar to how we automatically update the last updated date when committing a page.

GitHub actions are my new favourite thing at the mo 😀

catalinred:
Why not using a timestamp when compiling e.g.:

`<link rel="stylesheet" href="/static/css/2019.css?v={{ currentSomethingHere }}">`

bazzadp:
That's exactly what I've done here: https://github.com/HTTPArchive/almanac.httparchive.org/pull/802

catalinred:
Oh, that's cool. I admit I lost myself in numbers: #800 #801 #802 :D

PR #802, opened 2020-05-06
Add CSS incrementer GitHub action

### LINKED ISSUE #801: Auto increment CSS version numbers
As discussed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/800

We currently implement a cache-busting query param to our CSS files to ensure the latest CSS is applied for breaking changes.

At present this must be manually incremented when you know you are making breaking changes. We should automate this (perhaps through a GitHub action?).

bazzadp:
Fixes #801.

Tested in my own forked repo and seems to work


.github/workflows/increment_css.yml
```diff
@@ -0,0 +1,51 @@
...
+      with:
+        fetch-depth: 0
+    - name: Update CSS version for modified CSS files
+      env:
+        COMMIT_SHA: ${{ github.event.commits[0].id }}
+      run: |
+        echo "files changed in commit $COMMIT_SHA:"
+        git diff-tree --no-commit-id --name-only -r $COMMIT_SHA
+        echo "Incrementing any references to changed CSS files"
+        git diff-tree --no-commit-id --name-only -r $COMMIT_SHA src/static/css | grep "\.css" | sed 's!.*/!!' | sed "s/\(.*\)/s\/\1?v=[0-9][0-9]*\/\1?v=$(date +%Y%m%d%H%M%S)\//" | sed -f - -i src/templates/base/*.html"
```
rviscomi, line 34: Any chance this regex is too broad? Looks like it doesn't care if the version string is on our CSS or an external link. An easy fix would be to make the param more project-specific, like `almanac-css-version` to be ultra unique.

bazzadp: I don't think so - but it is some complex `sed` syntax there!!!

Best demonstrated in pieces. The first command does this to get the CSS files in the commit:
```
% git diff-tree --no-commit-id --name-only -r 86877ea383bdffa0d889719b42fb4869c6d7b595 src/static/css
src/static/css/2019.css
src/static/css/index.css
src/static/css/page.css
```

We then filter to just the CSS files (this step isn't really necessary since I've already filtered to the CSS path above, but who knows if some crazy fool will add a non-CSS file to that folder!):
```
% git diff-tree --no-commit-id --name-only -r 86877ea383bdffa0d889719b42fb4869c6d7b595 src/static/css | grep "\.css"
src/static/css/2019.css
src/static/css/index.css
src/static/css/page.css
```

Then we strip it down to just the file names:
```
% git diff-tree --no-commit-id --name-only -r 86877ea383bdffa0d889719b42fb4869c6d7b595 src/static/css | grep "\.css" | sed 's!.*/!!'
2019.css
index.css
page.css
```

Then use those filenames to create the regex's:

```
% git diff-tree --no-commit-id --name-only -r 86877ea383bdffa0d889719b42fb4869c6d7b595 src/static/css | grep "\.css" | sed 's!.*/!!' | sed "s/\(.*\)/s\/\1?v=[0-9][0-9]*\/\1?v=$(date +%Y%m%d%H%M%S)\//"
s/2019.css?v=[0-9][0-9]*/2019.css?v=20200507163324/
s/index.css?v=[0-9][0-9]*/index.css?v=20200507163324/
s/page.css?v=[0-9][0-9]*/page.css?v=20200507163324/
```

Then those regex's are actually what is run in the final sed command. As you can see, the regex is very specific to the CSS file(s) committed (in this case three of them).

Complex, and maybe not the most readable (and hence maintainable!) in future but it is native linux syntax so doesn't require installing node in the Action, and most importantly it works from my testing! 😀

bazzadp: Oh and this is only applied to the `src/templates/base/*.html` file=, so unlikely we have an unrelated stylesheet in there with the same name:

```
<link rel="stylesheet" href="www.example.com/css/index.css?v=4">
```

If we wanted to we could change the second last sed to this:

from this:
```
sed "s/\(.*\)/s\/\1?v=[0-9][0-9]*\1?v=$(date +%Y%m%d%H%M%S)\//"
```
to this
```
sed "s/\(.*\)/s\/\1?v=[0-9][0-9]*\/\/static\/css\/\1?v=$(date +%Y%m%d%H%M%S)\//"
```

to scope to that path. Does mean if we later subfolder our CSS files then this wouldn't work. As we're limiting to the base files I don't think this is necessary.

rviscomi: Ahh ok, thanks for clarifying! The `\1` in the sed command makes all the difference.

One small nit: is `v=[0-9][0-9]*` equivalent to `v=[0-9]+`?

bazzadp: > One small nit: is `v=[0-9][0-9]*` == `v=[0-9]+` == `v=\d+`?

Not quite. In sed the '+' needs to be escaped (unless using [Extended Regular Expressions](https://www.gnu.org/software/sed/manual/html_node/Extended-regexps.html) which would require lots of other escapes) so the regex would need to change from this:
```
sed "s/\(.*\)/s\/\1?v=[0-9][0-9]*\1?v=$(date +%Y%m%d%H%M%S)\//"
```
to this:
```
sed "s/\(.*\)/s\/\1?v=[0-9]\\\+\1?v=$(date +%Y%m%d%H%M%S)\//"
```
to escape the escape and also the `+`, so I thought just using `[0-9][0-9]*` was cleaner.

bazzadp: Oh and [sed doesn't have `\d`](https://superuser.com/questions/513412/how-to-match-digits-followed-by-a-dot-using-sed/513415) so it would need to `[[:digit:]]` but I, again, find `[0-9]` clearer than that syntax 😀

PR #621, opened 2020-01-09
Ebook for the almanac

mikegeyser:
This addresses #37.

PR #817, opened 2020-05-10
Ebook for the Almanac - take 2

### LINKED ISSUE #37: Generate an ebook
@HTTPArchive/developers curious to hear thoughts from others about this, might be crazy.

I'd like to see the _entire contents of the Almanac_ on a single web page, formatted similar to a book. It would also have a print stylesheet to handle things like page breaks and page numbers, so one could print to PDF and it'd just work™️ as a fully formed e-book. It'd also be a PWA that could be added to home screen and read offline.

There are concerns like lazy loading, history state management, deep linking, etc but I think these are all solvable problems.

I'm excited about this idea because a report on the state of the web should ideally maximize the web's capabilities for a great UX.

WDYT?

### Structure:
- [x] Table of contents
- [x] Page numbers
- [x] Header/footer in the margins
- [x] Cover page
- [x] Methodology section
- [x] Contributors section

### Rendering:
- [x] Set page metadata correctly (title etc.)
- [ ] Mirror page margins - I say we skip this. No one's gonna print this.
- [ ] Solve the problem of urls when printed - I say we skip this. No one's gonna print this.
- [x] Table wrappers aren't rendering properly
- [x] Problem with 'that z-index figure' in the CSS chapter
- [x] Weird black chart in the markup chapter
- [x] Prevent breaking a figure over multiple pages (separating the caption from the image)
- [x] Figure out what to do with tables that span multiple pages (particularly font chapter)
- [x] Figure out what to do with charts that are too big for a page
- [x] Handle internal links in-chapter references, that are repeated (e.g. #fig1, #conclusion).
- [x] Handle internal links cross-chapter references).
- [x] Handle internal links (Author links top).
- [x] Handle internal links (Author links top).
- [x] Author Avatars missing in PDF.
- [x] Internationalise book name (e.g. ebook-en.pdf)
- [x] Internationalise CSS content (e.g. title, pg...etc.) - maybe with inline `<style>` tags in ebook template?
- [x] ToC page numbers are misaligned in the Japanese PDF.

### Tooling:
- [ ] Try and come up with a solution that doesn't need the html to be served from flask
- [ ] Integrate weasyprint into the generate script
- [ ] Make the rendering config more dynamic
- [ ] Currently have to call weasyprint per year and language - script this.

Please feel free to add any more, and we can see if they're ~possible~ *feasible*. :)

bazzadp:
Fixes #37

Steals all @mikegeyser 's hard work from #621 and brings it up to date with the changes to the Almanac Code since.

Suggest we merge this sooner, rather than later, to avoid having to do that again, and then iterate before releasing it to the wild if needs be.


bazzadp:
OK did some more work on this and believe it's done.

I think I solved all the launch-blocking problems and updated https://github.com/HTTPArchive/almanac.httparchive.org/issues/37#issue-453223338. A few minor problems I don't think we need to solve.

I had to move some of the template text into `base.html` to make it available to the ebook and not just the chapter, and also fix a number of broken anchor links (particularly in Japanese chapters where English anchors had been used, but Japanese Anchors were generated - weasyprint is quote good for spotting them!), so looks like there's quite a bit of change in here, because of that.

You can see the [English version here (487 pages!)](https://github.com/HTTPArchive/almanac.httparchive.org/blob/ebook2/src/static/pdfs/web_almanac_2019_en.pdf) and the [Japanese version here (a mere 474 pages)](https://github.com/HTTPArchive/almanac.httparchive.org/blob/ebook2/src/static/pdfs/web_almanac_2019_ja.pdf).

There's still a bit of clean up to do (particularly in generating the ebooks dynamically, rather than one command per language and year), but this PR is already big enough and going to be hard enough to review so suggest we leave those until later.

Let me know your thoughts.

src/templates/base/2019/base_ebook.html
```diff
@@ -0,0 +1,170 @@
+{% extends "%s/2019/base.html" % lang %}
+<!-- TODO: I'm not sure what meta tags would be appropriate for a single-page ebook -->
```
rviscomi, line 2: This HTML comment gets rendered to the client. I think this is the correct syntax for template-only comments.

```suggestion
{# TODO: I'm not sure what meta tags would be appropriate for a single-page ebook #}
```

src/templates/base/2019/base_ebook.html
```diff
@@ -0,0 +1,170 @@
...
+            </div>
+            <div class="chapter">
+              <a href="#chapter-contributors">{{ self.contributors_title() }}</a>
+            </div>
+          </div>
+        </section>
+      </section>
+
+      <section id="foreword">
+        {{ self.foreward() | replace('./contributors#', '#contributors-') }}
```
rviscomi, line 108: Nit: misspelling of `foreword`

src/main.py
```diff
@@ -111,10 +112,32 @@ def convertOldImagePath(folder):
     return '%s' % folder[3:].replace('HTTP_2','http2').replace('_', '-').lower()
 
 
+# Render the methodolgoy chapter and pull out the section. Also applies some
```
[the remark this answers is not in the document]
bazzadp: Fixed.

src/templates/base/2019/base_ebook.html
bazzadp: Removed this as no longer TODO anyway 😀

src/templates/base/2019/base_ebook.html
bazzadp: Fixed!

src/README.md
```diff
@@ -58,10 +58,29 @@ npm install
 npm run generate
 ```
 
+3. For generating PDFs of the ebook, WeasyPrint will need some additional libraries:
+
+```
+brew install cairo
+brew install pango
+brew install gdk-pixbuf
```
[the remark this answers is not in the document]
bazzadp: Yeah I dunno how to do that. This is only needed for the PDF generation and people can alway File->Print to generate this if they don't want to use command line.

src/README.md
```diff
@@ -58,10 +58,29 @@ npm install
...
+npm run ebook_2019_ja
+```
+
+(TODO: make this a script to handle all languages and years at some point)
+
 ## Deploying changes
 
 If you've been added to the "App Engine Deployers" role in the GCP project, you're able to push code changes to the production website.
 
+_Make sure you have generated the ebooks PDFs as that currently requires some extra steps that are not automated!_
```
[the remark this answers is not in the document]
bazzadp: eBook updates are fully automated - except for the PDF. Which comes back to your point above!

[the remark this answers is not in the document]
bazzadp: Oh and automating the PDF generation is one of the things left to do, as shown unticked here: https://github.com/HTTPArchive/almanac.httparchive.org/issues/37#issue-453223338

bazzadp:
> Gave it a very quick and incomplete pass. A few comments, but I love it. Great work @bazzadp and @mikegeyser!

Thanks! Committed a few extra changes there now - including knocking down the font-size in print mode. I knew #731 would serve me in good stead and not just be an OCD thing 😀

> One general comment: I thought a user could visit the ebook.html page and choose to print to PDF. It looks like we're making the PDF version a statically available resource on its own. Do we need both?

You can do both by visiting `/en/2019/ebook` for example. However I think it's nice to have a PDF ready to download. Also Chrome does not implement all of CSS4 print stuff like headers and footers so printing to PDF from Chrome loses you quite a bit. Hence why we use WeasyPrint. The down side is generating the PDFs via WeasyPrint is a little manual at the moment - though the HTML is fully automated into the build process (which GitHub actions also calls), so it's just a small step on release day to also generate the PDFs.

bazzadp:
Oh and Chrome doesn't download all the lazy-loaded images when you save to PDF 😔

For the HTML version should it be the blank page? Or should it be the standard page with the Chapter Titles as the ToC on the left hand page? Can implement that (as had that working earlier anyway!), however I'm not convinced about exposing this in HTML to be honest.

The best way of reading this in HTML is via the individual chapters with more manageable page length, chapter specific ToC, additional chapter info like links to forum, results and queries.

A single PDF makes sense to me, but not a single HMTL.

src/README.md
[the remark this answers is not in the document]
bazzadp: Oh and BTW we did add WeasyPrint to requirements.txt, but those three packages are not available through pip and must be installed separately and manually. Did try adding there but couldn't get it to work!

bazzadp:
@ksakae1216 would you mind reviewing the Japanese pieces?

Particularly the [`templates/ja/2019/base_ebook.html`](https://github.com/HTTPArchive/almanac.httparchive.org/blob/ebook2/src/templates/ja/2019/base_ebook.html) file and also the [`templates/ja/2019/base.html`](https://github.com/HTTPArchive/almanac.httparchive.org/blob/ebook2/src/templates/ja/2019/base.html) file as noticed some English still in there, even without the pieces I added in this PR. Ignore the [`templates/ja/2019/ebook.html`](https://github.com/HTTPArchive/almanac.httparchive.org/blob/ebook2/src/templates/ja/2019/ebook.html) file as that's automatically generated - and quite large! 😀

And if you could have a look over the [Japanese PDF](https://github.com/HTTPArchive/almanac.httparchive.org/blob/ebook2/src/static/pdfs/web_almanac_2019_ja.pdf) and give any feedback on that, that would be good too!

Thanks,
Barry

src/main.py
```diff
@@ -95,6 +96,13 @@ def chapter_lang_exists(lang, year, chapter):
         return False
 
 
+def ebook_exists(lang, year):
+    if os.path.isfile('templates/%s/%s/ebook.html' % (lang, year)):
+        return True
+    else:
+        return False
```
rviscomi, line 103:
```suggestion
    return os.path.isfile('templates/%s/%s/ebook.html' % (lang, year))
```

src/main.py
```diff
@@ -111,10 +119,32 @@ def convertOldImagePath(folder):
...
+    methodology_template = render_template('%s/%s/methodology.html' % (lang, year), config=config)
+    methodology_maincontent = re.search('<article id="maincontent" class="content">(.+?)</article>', methodology_template, re.DOTALL|re.MULTILINE)
+    if methodology_maincontent:
+        methodology_maincontent = methodology_maincontent.group(1)
+        methodology_maincontent = re.sub('href="#', 'href="#methodology-', methodology_maincontent)
+        methodology_maincontent = re.sub('<h([0-6]) id="', '<h\\1 id="methodology-', methodology_maincontent)
+        methodology_maincontent = re.sub('href="\.\/', 'href="#', methodology_maincontent)
+        methodology_maincontent = re.sub('href="#([a-z0-9-]*)#', 'href="#\\1-', methodology_maincontent)
+        return methodology_maincontent
+    return False
```
rviscomi, line 139:
```suggestion
    if not methodology_maincontent:
        return False
        
    methodology_maincontent = methodology_maincontent.group(1)
    methodology_maincontent = re.sub('href="#', 'href="#methodology-', methodology_maincontent)
    methodology_maincontent = re.sub('<h([0-6]) id="', '<h\\1 id="methodology-', methodology_maincontent)
    methodology_maincontent = re.sub('href="\.\/', 'href="#', methodology_maincontent)
    methodology_maincontent = re.sub('href="#([a-z0-9-]*)#', 'href="#\\1-', methodology_maincontent)
    return methodology_maincontent
```

src/templates/base/2019/table_of_contents.html
```diff
@@ -121,6 +121,12 @@ <h2 class="part-name">{{ self.appendix() }}</h2>
       </section>
     </div>
 
+    <section class="ebookd-download">
+      {% if ebook_exists(lang, year) %}
+      <p><a href="/static/pdfs/web_almanac_{{ year }}_{{ lang }}.pdf">{{ self.ebook_download() }}</a></p>
```
rviscomi, line 126: Should this use `url_for`?

src/templates/base/2019/table_of_contents.html
```diff
@@ -121,6 +121,12 @@ <h2 class="part-name">{{ self.appendix() }}</h2>
       </section>
     </div>
 
+    <section class="ebookd-download">
```
rviscomi, line 124:
```suggestion
    <section class="ebook-download">
```

src/static/css/ebook.css
```diff
@@ -0,0 +1,252 @@
...
+  margin: 2.5cm;
+  padding: 0;
+  font-family: 'Lato', sans-serif;
+  font-size: 10pt;
+
+  @bottom-left {
+    width: 100%;
+    margin: 10pt 0 30pt 0;
+    border-top: .25pt solid #666;
+    content: "The Web Almanac 2019 by HTTP Archive";
```
rviscomi, line 144: Should we have English-specific content in the base stylesheet? Or is this always overwritten by the templates?

src/templates/en/2019/methodology.html
```diff
@@ -310,7 +310,7 @@ <h3 id="wappalyzer"><a href="#wappalyzer" class="anchor-link">Wappalyzer</a></h3
       </p>
 
       <p>
-        Wappalyzer powers many chapters that analyze the popularity of developer tools like WordPress, Bootstrap, and jQuery. For example, the <a href="{{ url_for('chapter', year=year, lang=lang, chapter='ecommerce') }}">Ecommerce</a> and <a href="{{ url_for('chapter', year=year, lang=lang, chapter='cms') }}">CMS</a> chapters rely heavily on the respective <a href="https://www.wappalyzer.com/categories/ecommerce">Ecommerce</a> and <a href="https://www.wappalyzer.com/categories/cms">CMS</a> categories of technologies detected by Wappalyzer.
+        Wappalyzer powers many chapters that analyze the popularity of developer tools like WordPress, Bootstrap, and jQuery. For example, the <a href="./ecommerce">Ecommerce</a> and <a href="./cms">CMS</a> chapters rely heavily on the respective <a href="https://www.wappalyzer.com/categories/ecommerce">Ecommerce</a> and <a href="https://www.wappalyzer.com/categories/cms">CMS</a> categories of technologies detected by Wappalyzer.
```
rviscomi, line 313: Why is `url_for` replaced in this file? Hard-coding URLs could be brittle if things move.

src/templates/base/2019/table_of_contents.html
bazzadp: Can we use that for non-template files?

Which leads to another question for you - should this PDF live in static folder or templates folder? It's not dynamically generated as such which is why I put in static folder but @mikegeyser originally had it generating in templates folder. If it's in the static folder then don't need Python to serve it so seems faster to me. Especially as it's a big file so that's why I went with the static folder.

What's your opinion?

rviscomi: Yeah that can be used for anything with an `@app.route` decorator. So if we pass it the year and lang it will dynamically generate URL corresponding to this pattern `@app.route('/<lang>/<year>/ebook')`.

rviscomi: ebook.html should be a template
ebook.pdf should be a static file

src/static/css/ebook.css
bazzadp: It's overwritten by this inline style.

https://github.com/HTTPArchive/almanac.httparchive.org/blob/f3e54ef4d4dc4f5f730f1ffaca9b7df6fb13585b/src/templates/base/2019/base_ebook.html#L19-L25

Hence why it appears as "2019 Web Almanac by HTTP Archive" in English version. Probably should update this to match. Or could remove it completely from `ebook.css` but no harm to have English fallback.

src/templates/en/2019/methodology.html
bazzadp: I really don't see the point of `url_for`. We don't use it in the `.md` files so if thinks move then we're in trouble already.

We need to remove this or importing this for the ebook doesn't work. Can't remember why so can have a look but, as I say, I don't see the point of this function given we "hardcode" the URLs in the `.md` files.

src/static/css/ebook.css
rviscomi: Removing it SGTM. In 4 months if I wanted to edit this text I wouldn't remember which file to edit.

src/templates/base/2019/table_of_contents.html
bazzadp: So this line is a `.pdf` file, so in static folder, so not `@app.route` so fine to leave as is?

src/static/css/ebook.css
bazzadp: It's gone!

src/templates/en/2019/methodology.html
rviscomi: `url_for` is useful so we don't need the templates to care about how the URL are structured by the router in `main.py`. This abstraction gives us flexibility if needed.

I'm not strongly advocating for keeping it, just wanted to hear your reasoning for removing it when I think it provides value.

But as long as you're doing a regex replace of the href values anyway, why does it matter if they're `./` or the full URL?

bazzadp: Also we pass in the template name to `url_for` so if we move them about won't we need to change the inputs to the `url_for` function anyway? At which point why no just do a search for `./.*` and fix the "hardcoded" urls?

src/templates/base/2019/table_of_contents.html
bazzadp: And it is a link to the PDF and not the `ebook.html` as think the PDF is more useful as per my comments here: https://github.com/HTTPArchive/almanac.httparchive.org/pull/817#issuecomment-627028323

rviscomi: You're right, this is ok.

src/templates/en/2019/methodology.html
rviscomi: The name passed into `url_for` is the function name of the route defined in `main.py`.

bazzadp: OK retested and remember why I did this now. The regex in `main.py` is below:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/3db26fac48360d459dda542191da50173a09db64/src/main.py#L119-L137

so without this change to "hardcode" URLs we won't touch those reference, so that means when the ebook is generated the URL is:
```html
<a href="/en/2019/ecommerce">Ecommerce</a>
```
as opposed to:
```html
<a href="#ecommerce">Ecommerce</a>
```
(which is the internal link to that chapter in the ebook).

So the links don't work within the PDF.

Also I can't change this to:
```
{{ url_for('chapter', year=year, lang=lang, chapter='#ecommerce') }}
```
Or even:
```
{{ url_for('chapter', year=year, lang=lang, chapter='ecommerce',ebook=True) }}
```

Because that will break the real Methodology chapter and I'm reusing this.

I could just regex for this in the get_ebook_methodology function:
```
{{ url_for('chapter', year=year, lang=lang, chapter='#ecommerce') }}
```
and replace it with the hardcoded link in the above Python code, but why not use same hardcoded link for ebook and Methodology page to avoid having to do this? The hardcoding works for both.

So there is a good reason to do this change! 😀

rviscomi: My suggestion is to keep the existing `url_for` code throughout the methodology template and update `get_ebook_methodology` to map href URLs like `/en/2019/ecommerce` to `#ecommerce`. (Or maybe using a more unique ID like `#chapter-ecommerce` in case of heading collisions :))

Is that doable?

bazzadp: Or...

I can replace this in `main.py`:

```
    methodology_maincontent = re.sub('href="\.\/', 'href="#', methodology_maincontent)
```

to this:

```
    methodology_maincontent = re.sub('href="\/%s\/%s\/' % (lang, year), 'href="#', methodology_maincontent)
    methodology_maincontent = re.sub('href="\.\/', 'href="#', methodology_maincontent)
```

And then I can leave the `url_for` function. Probably no harm to do that anyway, just in case someone does add `url_for` back into the methodology page. Will do that!

And actually to handle other years and languages I should add one more row:

```
    methodology_maincontent = re.sub('href="\/%s\/%s\/' % (lang, year), 'href="#', methodology_maincontent)
    methodology_maincontent = re.sub('href="\/', 'href="https://almanac.httparchive.org/', methodology_maincontent)
    methodology_maincontent = re.sub('href="\.\/', 'href="#', methodology_maincontent)
```

bazzadp: > My suggestion is to keep the existing `url_for` code throughout the methodology template and update `get_ebook_methodology` to map href URLs like `/en/2019/ecommerce` to `#ecommerce`. (Or maybe using a more unique ID like `#chapter-ecommerce` in case of heading collisions :))
> 
> Is that doable?

Yup - just figured out how to do it.

And the chapters headers are given links like `#chapter-header` (for example `#performance-first-contentful-paint`) to avoid name clashes. The `#chapter` links are resevered for the chapter names themselves.

rviscomi: Perfect, thanks for thinking through this with me!

bazzadp: > Perfect, thanks for thinking through this with me!

No, no thank you for the great feedback - keep it coming! This makes the ebook more robust and better.

All the `url_for` have been reverted so going to mark this as resolved. Un-resolve if again if you spot one I missed.

PR #825, opened 2020-05-14
Track ebook link clicks

bazzadp:
Thanks to #817 we're about to launch the PDF ebook of the Almanac. There's a link on the ToC page. Let's track clicks on that link since the PDF page won't be track in Google Analytics.

Still won't track direct downloads of the PDF but, hey it's something.

### DIFF
--- a/src/templates/base/2019/table_of_contents.html
+++ b/src/templates/base/2019/table_of_contents.html
@@ -122,7 +122,7 @@
       </section>
     </div>
 
-    <div class="ebook-download">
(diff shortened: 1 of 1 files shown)


bazzadp:
It doesn’t. But can set it up in GTM easily enough if you prefer to do it that way?: https://michaelhoweely.com/2019/02/01/how-to-track-pdf-downloads-using-google-tag-manager/

rviscomi:
Oof that's not much easier than this, let's keep it the way you implemented!

PR #827, opened 2020-05-15
Stop links opening in a new tab

bazzadp:
Currently the Author links (Twitter, GitHub and Websites) open in new tabs which is generally recognised as bad UX and [doing it without telling the user is a failure under WCAG 3.2.1](https://www.w3.org/WAI/WCAG21/Understanding/on-focus.html).

I propose we just remove the `target="_blank"` code to stop them opening in a new window.

Alternatively, if we wanted to leave them, then we should add indication that they will open in a new tab. This can be with `aria-label`, though doesn't leave anything for sighted users (could use `title` attribute if didn't want to add more UI).

### DIFF
--- a/src/templates/base/2019/base_ebook.html
+++ b/src/templates/base/2019/base_ebook.html
@@ -155,7 +155,7 @@
           </a>
           {% endif %}
           {% if contributor.website %}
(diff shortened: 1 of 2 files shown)


bazzadp:
Oh and btw the links at the bottom of chapters don't open new windows - only the links in Contributor page. So removing them for there would be better for consistency too.

PR #833, opened 2020-05-16
Switch eBook PDF generation from WeasyPrint to Prince

bazzadp:
This moves us from using WeasyPrint to generate the PDF to [Prince](https://www.princexml.com/) with the following benefits:
- Allows us to create a fully Accessible, tagged, PDF that conforms to [PDF/UA-1](https://en.wikipedia.org/wiki/PDF/UA). See [king accessible tagged PDFs with Prince by Bruce Lawson](https://www.brucelawson.co.uk/2019/making-accessible-tagged-pdfs-with-prince/)
- Is easier to install, and cross platform - addressing https://github.com/HTTPArchive/almanac.httparchive.org/pull/817#discussion_r423365904
- Supports more CSS - including fonts (hadn't realised we were using our fallbacks until I ran this and saw the difference!) and better tables as supports `box-shadow`.
- Better support of Meta attributes.
- Is much faster to generate PDFs.

It is free for non-commercial use (which [they define as "fees are charged", "there is a money transaction", or "there is an income generated"), however they do stamp the Prince logo on first page in the top right hand corner:

Prince logo added to each PDF

The also insist on a link to their webpage in return:

Prince link

Don't see a problem with that and think that's fair enough. If it ever does become a problem we can revert back to WeasyPrint easily enough and hopefully it will have improved in meantime to support all the things that Prince does. @rviscomi would be good for you to confirm you're happy with that?

This PR also includes some other errors I spotted when reviewing the PDF, and moving more of the ebook template from EJS to Jinja2 for consistency wiht the pages they are based upon, and to address an issue with fallback Avatars not showing for those without an Avatar.

After this is merged I think we're good to launch the eBook for English and Japanese.


src/templates/en/2019/accessibility_statement.html
```diff
@@ -70,7 +70,11 @@ <h2 id="conformance-status"><a href="#conformance-status" class="anchor-link">Co
         <p>
           In some chapters, we also include other content on the site, not created by ourselves, including YouTube videos, PDF links or links to other articles, which do not meet the strict criteria required of WCAG 2.1 level AAA. See note below on external content.
         </p>
-      
+
+        <p>
+          The PDF version of the Web Almanac follows PDF accessibility best practices and meets the <a href="https://en.wikipedia.org/wiki/PDF/UA">PDF/UA-1</a> standard.
```
rviscomi, line 75: Maybe link "PDF version"

src/templates/base/2019/base_ebook.html
```diff
@@ -27,125 +32,127 @@
...
-          {{ self.methodology_description() }}
-          <a href="#methodology" class="alt">
-            {{ self.methodology_link() }}
-          </a>
-        </p>
-        <img class="methodology-characters" src="/static/images/methodology-characters.png" alt="" width="984" height="354" loading="lazy" />
+    <h1 class="title title-center">{{ self.intro_title() }}</h1>
+    <h2 class="title title-center">{{ self.intro_sub_title() }}</h2>
+
+    <img class="intro-image" src="/static/images/home-hero.png" alt="" width="984" height="354" />
```
rviscomi, line 41: This image should have a transparent background but appears with a gray background in the PDF. Not sure what we can do about that.
image

src/templates/base/2019/base_ebook.html
```diff
@@ -27,125 +32,127 @@
...
+      </div>
+    </div>
+  </section>
+
+  <section id="foreword">
+    <h1>{{ self.foreword_title() }}</h1>
+    {{ self.foreword() | replace('/%s/%s/contributors#' % (lang, year), '#contributors-') }}
+  </section>
+
+  {{ self.chapters() }}
```
rviscomi, line 125: The page numbers are misaligned with the chapter links
image

src/templates/base/2019/base_ebook.html
```diff
@@ -27,125 +32,127 @@
 {% endblock %}
 
 {% block content %}
-<section class="content">
+<div class="content">
```
rviscomi, line 35: I'd love to see the cover page stand out more (maybe darker background color, white/yellow text) and give "2019" more prominence, using the home page design as inspiration:

image
image

src/templates/base/2019/base_ebook.html
rviscomi: Upon closer inspection, the area where you'd expect the Chapter 15 link is still clickable but has no text. 🤔

src/templates/base/2019/base_ebook.html
```diff
@@ -27,125 +32,127 @@
...
-          </a>
-        </p>
-        <img class="methodology-characters" src="/static/images/methodology-characters.png" alt="" width="984" height="354" loading="lazy" />
+    <h1 class="title title-center">{{ self.intro_title() }}</h1>
+    <h2 class="title title-center">{{ self.intro_sub_title() }}</h2>
+
+    <img class="intro-image" src="/static/images/home-hero.png" alt="" width="984" height="354" />
+  </section>
+
+  <section>
```
rviscomi, line 44: I'd suggest removing this page. Feels out of place.

rviscomi:
Could you point me to the Prince docs that talk about watermark/backlinks? I'm not keen on adding their link from our website, maybe that's something we can do in the source code?

src/templates/base/2019/base_ebook.html
bazzadp: This was a deliberate choice and I added the grey with CSS:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/fe540e7f0f35a932d42a9c362bcc75e19bcb6053/src/static/css/ebook.css#L45-L50

Without this, the image blends into the background too much:

image

src/templates/en/2019/accessibility_statement.html
bazzadp: Good idea - done!

src/templates/base/2019/base_ebook.html
bazzadp: OK. Was trying to reuse the nice content from Home Page without dumping this on front cover. will remove then.

src/templates/base/2019/base_ebook.html
rviscomi: Ah ok. The transparency may not be an issue if we go with https://github.com/HTTPArchive/almanac.httparchive.org/pull/833#discussion_r426164222

bazzadp:
> Could you point me to the Prince docs that talk about watermark/backlinks? I'm not keen on adding their link from our website, maybe that's something we can do in the source code?

It's in their [licence](https://www.princexml.com/license/):

> **1.1 Non-commercial license**
> If this is a Non-commercial license, Licensee may download, install and use the Software for Non-commercial Purposes on a computer that is accessible to any number of end users. PDF documents generated by the Software include notices that identify the Software. **Licensee shall not change or remove these notices or assist or encourage third parties to remove or change such notices.**
> 
> When the Non-commercial license is used to routinely generate documents, **a prominent link to the www.princexml.com Web site shall be displayed on the pages from where the generated documents can be fetched**, and in a prominent public Web page where business partners are listed. If PDF documents are sent as email attachments by Licensee, all email messages must contain the www.princexml.com Web address in the message body.

I understand your hesitance, but it really is great software (especially from Accessibility point of view) so would really like to use it. Could also add to our methodology page (we mention lots of other tools there), and maybe we should add in there anyway, but seems from above they do want it where I have it.

src/templates/base/2019/base_ebook.html
bazzadp: Fixed

src/templates/base/2019/base_ebook.html
bazzadp: It's gone.

src/templates/base/2019/base_ebook.html
bazzadp: Uploaded a new version there. Let me know your thoughts.

Not done the diagonal stripe yet though, but gives some idea of what it might look like.

src/templates/base/2019/base_ebook.html
bazzadp: Resolving this as discussing below.

src/templates/base/2019/base_ebook.html
```diff
@@ -27,125 +32,101 @@
...
+      </div>
+    </div>
+  </section>
+
+  <section id="foreword">
+    <h1>{{ self.foreword_title() }}</h1>
+    {{ self.foreword() | replace('/%s/%s/contributors#' % (lang, year), '#contributors-') }}
+  </section>
+
+  {{ self.chapters() }}
```
rviscomi, line 99: Could we add a page break between chapters?

image

src/templates/base/2019/base_ebook.html
```diff
@@ -27,125 +32,101 @@
...
-        <ul class="contributors">
-        {% for id, contributor in config.contributors.items() %}
-        <li class="contributor" id="contributors-{{ id }}">
-          <img class="contributor-avatar" src="{{contributor.avatar_url}}" height="100" width="100" alt="{{ contributor.name }} avatar" />
-          <div>
+      </div>
+    </div>
+  </section>
+
+  <section id="foreword">
```
rviscomi, line 94: Looks like every page after the foreword has "Foreword" as the header text.
image

src/templates/base/2019/base_ebook.html
```diff
@@ -27,125 +32,101 @@
...
+        <img src="/static/images/avatars/8.jpg" alt="" height="100" width="100" />
+        <img src="/static/images/avatars/4.jpg" alt="" height="100" width="100" />
+        <img src="/static/images/avatars/0.jpg" alt="" height="100" width="100" />
+        <img src="/static/images/avatars/11.jpg" alt="" height="100" width="100" />
+        <img src="/static/images/avatars/9.jpg" alt="" height="100" width="100" />
+    </div>
+    <p>{{ self.contributors_description() }}</p>
+    <ul class="contributors">
+    {% for id, contributor in config.contributors.items() %}
+      <li class="contributor" id="contributors-{{ id }}">
```
rviscomi, line 123: A couple of suggestions for this section:
- remove the bullet symbol next to each contributor
- if possible, try to fit two per row

image

src/templates/base/2019/base_ebook.html
```diff
@@ -27,125 +32,101 @@
...
+    {{ get_ebook_methodology(lang,year) | safe }}
+  </section>
+
+  <section class="chapter" id="contributors">
+    <div class="subtitle">Appendix B</div>
+  
+    <h1 class="title title-lg">
+      {{ self.contributors_title() }}
+    </h1>
+    <div class="team">
```
rviscomi, line 112: Flexbox with `space-between` so these images spread out across the page width

src/templates/base/2019/base_ebook.html
bazzadp: Oops. Changing from `<section>` to `<article>` in last commit broke that. Reverted that back and re-uploaded.

src/templates/base/2019/base_ebook.html
bazzadp: > if possible, try to fit two per row

You don't know how many hours I spent on this in Weasyprint! With this it's as simple as `columns: 2`

Both done.

src/templates/base/2019/base_ebook.html
bazzadp: Again Flexbox not supported in WeasyPrint, but works in Prince! So fixed.
Please approve using this😀

src/templates/base/2019/base_ebook.html
bazzadp: Diagonal stripe added by re-using our SVG.

bazzadp:
The other option to avoid the watermark and back linking is to buy a licence of course. $495 per person or $3800 per server (and then $123 or $950 per year after if want updates). Buying myself (and perhaps you) a licence may be sufficient for now, but that would be restrictive.

I think the watermark and back link is small enough (and well deserved to be honest) that it’s fine. It is only near the PDF bits and not like we have it all over our site.

src/static/css/ebook.css
```diff
@@ -115,51 +153,102 @@ ul li.contributor::before {
...
 }
 
-.social-icon {
-  height: 20px;
-  width: 20px;
+@prince-pdf {
+  prince-pdf-display-doc-title: true;
 }
 
 @page {
```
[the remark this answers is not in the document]
bazzadp: Yes, that's exactly why:

```css
  @page :left {
    ...
    margin-right: 4cm;
  }

  @page :right {
    ...
    margin-left: 4cm;
  }
```


To be honest I was thinking of making this an optional parameter in the URL (like I've allowed you to override the A4 page side with a URL param if you want).

WDYT?

src/static/css/ebook.css
```diff
@@ -1,3 +1,51 @@
...
+  font-weight: 700;
+}
+
+#title h2.title {
+  font-size: 32px;
+  font-size: 2rem;
+  font-weight: 300;
+}
+
+.intro-image {
```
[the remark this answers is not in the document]
bazzadp: Could get tricky if image resizes (e.g. for different page sizes) but will have a look...

src/static/css/ebook.css
```diff
@@ -234,13 +357,63 @@ ul li.contributor::before {
...
+    page-break-before: avoid;
+  }
+
+  .authors li {
+    page-break-inside: avoid;
+  }
+
+  .fn {
+    display: inline;
+    float: footnote;
```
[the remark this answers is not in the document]
bazzadp: Yeah, they are nice aren't they!

Not intentional so good spot. Will be a little tricky and will probably need to write another Python function for this but should be doable...

src/static/css/ebook.css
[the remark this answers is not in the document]
rviscomi: Optional param SGTM. Most people will read as PDF so consistent margins are probably the general case and we can override for print as needed.

src/static/css/ebook.css
```diff
@@ -234,13 +357,63 @@ ul li.contributor::before {
...
+  }
+
+  .fn::footnote-marker {
+    text-align: left;
+    font-size: 10px;
+    font-size: 0.625rem;
+    font-weight: normal;
+    min-width: 18px;
+  }
+}
```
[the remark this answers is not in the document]
bazzadp: Fixed.

Not sure why this negative margin is here but overrode it for the ebook:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/f7f369ab45c4d14ad6c553eea8edf24533baae77/src/static/css/page.css#L330-L333

src/static/css/ebook.css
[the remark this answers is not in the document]
bazzadp: Fixed.

src/templates/base/2019/base_ebook.html
```diff
@@ -17,146 +22,194 @@
...
+  </section>
+
+  <section id="toc" class="contents-wrapper">
+    <h1>{{ self.table_of_contents_title() }}</h1>
+    <div class="parts">
+      <div class="part">
+        <h2 class="part-name">{{ self.introduction() }}</h2>
+        <div class="toc-chapters">
+          <div class="toc-chapter">
+            <a class="front-matter" href="#foreword">{{ self.foreword_title() }}</a>
```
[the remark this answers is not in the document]
bazzadp: Fixed.

src/static/css/ebook.css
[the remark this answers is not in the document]
bazzadp: Fixed

src/static/css/ebook.css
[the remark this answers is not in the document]
bazzadp: Fixed.

PR #839, opened 2020-05-20
Correct digit to respect figure

AdamDemirel:


bazzadp:
Thanks! Merged and will include in next release!

PR #863, opened 2020-05-24
Colour contrast image fixes

### LINKED ISSUE #855: Illustration image about contrasts are misleading users: recommended colours are not contrasted enough
Hello,

In these two articles, there is an image to illustrate what is a text with insufficient colour contrasts and versus what is a text with sufficient colour contrasts:

- https://almanac.httparchive.org/fr/2019/accessibility
- https://almanac.httparchive.org/fr/2019/mobile-web

The image is here: https://almanac.httparchive.org/static/images/2019/accessibility/example-of-good-and-bad-color-contrast-lookzook.png

The right column of the image is telling that the #F56905 orange with white text on it is a recommended colour contrast. But, the contrast ratio between the orange background colour and the white text is 3:1.

In the WCAG accessibility rules, contrast ratio are more complicated than that and this image is misleading users.

- You can have a 3:1 contrast ratio if:
    - Your non-bold text has a font-size of at least 24px;
    - Your bold text has a font-size of at least 18.5px.
- You must have a 4.5:1 contrast ratio if:
    - Your non-bold text has a font-size of less than 24px;
    - Your bold text has a font-size of less than 18.5px.

So, the image is incomplete. To prevent misleading users, I think you should use an image where the recommended colours have a contrast ratio of at least 4.5:1 (or you should explain the entire rule but it can be a little bit long).

I have fixed the image: https://www.lelutinduweb.fr/wp-content/uploads/example-of-good-and-bad-color-contrast.svg

Feel free to use it and modify it because it's a SVG so you can do what you want. To test contrast ratio of the orange, the [Tanaguru Contrast Finder tool](https://contrast-finder.tanaguru.com/result.html?foreground=%23FFFFFF&background=%23F56905&ratio=4.5&isBackgroundTested=true&algo=Rgb&distanceSort=asc) can be used.

Note: I added a text for the greyscales boxes because it was not clear, in my opinion - as I explained it here: #854

### LINKED ISSUE #856: Accessibility Chapter questions over Figure 1
Pulling this out from #854 raised by @juliemoynat

[Figure 1 in the Accessibility chapter](https://almanac.httparchive.org/en/2019/accessibility#fig-1) is the following:

image

And has the following in the (hidden) description (emphasis added by myself to one sentence):

> Four colored boxes of orange and gray shades with white text overlaid inside creating two columns. The left column says Too lightly colored and has the orange background color written as #FCA469. The right column says Recommended and the orange background color is written as #F56905. The top box in each column has an orange background with white text #FFFFFF and the bottom box has a gray background with white text #FFFFFF. **The grey background shows the orange color in greyscale**. Courtesy of LookZook.

It is not clear from the image nor the text that "The grey background shows the orange color in greyscale" so if that is the intention, then it should be made clear. If that is not the intention then the long description should be corrected.

bazzadp:
Fixes #855 
Fixes #856 
As raised by @juliemoynat

This replaces this image:

Old image using recommended colours that meet 3.0 colour contrast ratio but not 4.5

With this one:

New image that meets 4.5

It is also an SVG so scales better.

We also add some text to accessibility chapter, showing that that the greyscale is shown to to indicate what the colours would look like in grey. @borisschapira and @ksakae1216 can you review the translations here please?


obto:
@bazzadp I'm good with this. Just two things:

1. I have the source for the graphic in illustrator. I can just adjust the colors and export it to SVG if you'd like
2. The "Too lightly colored" text and bar should be in an orange or red to show its "not good". The preview image you posted has it in green

bazzadp:
Thanks @obto happy for you update the original image, export as an SVG and update this branch. Probably better to do that, rather than use my hobbled together version.

What do you think of the Grayscale header text? I quite like them so could you add to your version too?

Ahhh - I was **very** confused why the “not good” message was the same colour as the “good” too right box. Didn’t realise that was supposed to represent orange for bad and think reusing the colour was a point of confusion. Maybe now top right has moved to brown we can reuse orange there - though that also brings in the colour contrast issues with that too! Also I didn’t see the “Good” message as green (looks more blue-green to me) so also missed that too. Maybe we should just use red and green instead? I know red/green colour blindness is a thing, and these colours don’t always represent bad/good in all cultures but think the colour is extra to the text so think that would be OK. What do you think?

obto:
@bazzadp Here is the new SVG, how does it look (had to zip it so it'd upload)?

[example-of-good-and-bad-color-contrast-lookzook-v2.zip](https://github.com/HTTPArchive/almanac.httparchive.org/files/4685799/example-of-good-and-bad-color-contrast-lookzook-v2.zip)

bazzadp:
Hey @obto , it seems that illustrator has exported the text as images. This has two downsides:

- The file is a good bit bigger (58KB as opposed to 7 KB).
- The text (e.g. colour codes) can no longer be selected, copy and pasted. Maybe not a big deal, but since we have it, why not make it

So I'd prefer to use the existing SVG in the repo. I did update the colours and text to match yours though:

Updated Image of Fig1 from Accessibility chapter

Happy enough with this?

obto:
Aaah that's right. I can just turn that setting off and re-export. Forgot I
usually have that enabled haha

obto:
Here you go. Always nice to save ~60% on file size for free

[example-of-good-and-bad-color-contrast-lookzook-v2.zip](https://github.com/HTTPArchive/almanac.httparchive.org/files/4689658/example-of-good-and-bad-color-contrast-lookzook-v2.zip)

bazzadp:
Much better. Updated [here](https://github.com/HTTPArchive/almanac.httparchive.org/tree/colour_contrast_image_fixes/src/static/images) under Accessibility and Mobile Web chapters.

If you're happy then click that little Approve button and I'll merge.

PR #868, opened 2020-06-02
Accessibility statement improvements

bazzadp:
Just noticed the index in the accessibility statement doesn't use the dynamic sidebars introduced in #732 as we don't reference the chapter javascript so added this.

Also added a statement date to the accessibility statement. @ksakae1216 could you review the Japanese translation please?

PR #872, opened 2020-06-03
Fix colour on Language selector as causing Firefox issue

bazzadp:
Weirdly choosing any colour other than white (we added `#F2F2F2` in #868) causes a scrollbar to appear on Firefox:

Weird scrollbar on Firefox

So setting it to White, and also marking the presentational spacing bar as `aria-hidden`.

### DIFF
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -588,7 +588,7 @@ header .cta {
 .language-switcher option,
 .year-switcher option {
   color: #1A2B49;
-  background-color: #F2F2F2;
+  background-color: white;
 }
 
 .language-switcher:focus-within,
--- a/src/templates/base/2019/base.html
+++ b/src/templates/base/2019/base.html
@@ -330,7 +330,7 @@
       {{ l }}
     </option>
     {% endfor %}
-    <option disabled="disabled">
+    <option disabled="disabled" aria-hidden="true">
       ────
     </option>
     <option value="https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Translators'-Guide">


rviscomi:
Why does changing the bgcolor remediate a layout problem and is there anything we can do to make the layout more resilient?

I'd expect a robust fix to be something like dynamic height or overflow control.

bazzadp:
> Why does changing the bgcolor remediate a layout problem

I don’t know! It’s very odd. Try it in Firefox by overriding the `background-color` in dev tools. It’s very odd! There are [some similar issues in Bugzilla to this](https://bugzilla.mozilla.org/buglist.cgi?quicksearch=option+background-color) and suspect they are related.

>  is there anything we can do to make the layout more resilient?

Using `white` seems to work 😀 To be honest the only reason I added a `background-color` at all was because I occasionally run the [Firefox accessibility tester](https://developer.mozilla.org/en-US/docs/Tools/Accessibility_inspector#Check_for_accessibility_issues) on pages I change and was bored of seeing this false positive alert - it complains of a colour contrast issue as we don’t set a `background-color` so it presumes the grey from header and footer is the colour without realising all browsers style `select` boxes with a light background. So even with the white `background-color` it’s basically ignored. But stops the accessibility fail being flagged so less noise to check.

> I'd expect a robust fix to be something like dynamic height or overflow control.

[Browsers only allow very limited styling over `select` and `options` elements](https://stackoverflow.com/a/7208814) —hence why my `background-color` change doesn’t actually do anything in most browsers! Only Firefox and IE seem to actually take notice of `background-color` stylings for `option` elements - Chrome and Safari ignore them (though weirdly still show them as valid styles in dev tools and not crossed out). I did try adding `line-height` and even adjusting the `overflow` setting but both are ignored.

Ultimately this is not that important and because the browsers (or operating systems) handle the styling and even the scroll bar wasn’t the worst - but it bugged me and changing it to white seemed to fix it and also still avoid the accessibility errors from being flagged so think it’s robust enough.

rviscomi:
Gotcha, thanks for the context!

bazzadp:
Bug raised anyway: https://bugzilla.mozilla.org/show_bug.cgi?id=1643157 but can't imagine it'll be high priority (it certainly isn't for me as discussed!).

Looks like it's a MacOS specific issue as tested on windows and couldn't recreate. MacOS SELECT elements do look different.

Issue #815, opened 2020-05-09
Create a 2020 landing page


rviscomi:
Goals:

- start serving a 2020 page to test out our infrastructure and make any fixes to ensure that it can support multiple years
  - avoid duplicating code used by both years, reuse in year-agnostic files (templates, static resources)
 - work out any kinks in SEO
 - work out any design kinks in the year switcher UI
- inform readers about our plans for 2020
  - ways to contribute and how to join the project
  - set expectations about a 2020 publication date

We're on our own in terms of design, but we can borrow any needed UX patterns from the 2019 home page - it doesn't need to be completely reimagined with each new year.

UX goals:

- be clear that we're talking about 2020 which is unpublished as of yet
  - the current desktop home page design uses an image for the "2019" so we may need to get creative to update this
- provide a clear way to switch to and from 2019, similar to the language switcher

Besides direct links, the only entry point onto this page from the rest of the site will be via the year switcher on 2019 pages. We won't be routing top-level navigations to `https://almanac.httparchive.org` to this page until 2020 is officially published.

rviscomi:
@bazzadp any thoughts on the technical implementation, especially related to multi-year support on the backend? I think we've done a really good job of planning ahead and keeping things organized by year.

Also, any thoughts on translating the landing page? IMO translating it would be nice to have but somewhat of a lost cause since this page is temporary by design and will be overwritten when 2020 is published. One consideration is that users may be on a translated page when they use the year switcher to visit this landing page: should we have English-language content on each combination of `/<lang>/2020` pages? If we have translators willing to do the work, I don't see a reason not to do it?

bazzadp:
> @bazzadp any thoughts on the technical implementation, especially related to multi-year support on the backend? I think we've done a really good job of planning ahead and keeping things organized by year.

Don't think it should be too difficult to be honest, if we're keeping the design reasonably the same. That was my big question for you - how different is it going to be? And that's wider than the langing page, and more about the 2020 edition.

There's a balance to be struck here between keeping it very similar or not:

**Positives to keeping it similar:**

- Emphasises the Web Almanac brand
- Makes it easy to switch between years without a jarring move of year selectors about the page and the like.
- Less development required!
- More maintainable in future

**Negatives**

- Maybe gets a bit boring?
- How to differentiate the years?

[...]

Issue #634, opened 2020-01-15
Serve static HTML


rviscomi:
Our backend is currently hosted on Google App Engine running a Python server and rendering Jinja templates in Flask based on JSON config data.

For every URL, there is effectively one static HTML page. However, every time a user visits one of these pages, the Python server needs to render the templates using (in-memory) JSON config data. Because each type of page requires a different amount of rendering, performance varies across the site:

![image](https://user-images.githubusercontent.com/1120896/72463781-0c5ef200-37a2-11ea-8503-cba70cd0b2fb.png)

This chart tracks the 30-day trend of CrUX data for four of our pages: Home, CSS, JS, and Table of Contents. The ToC page has the highest % of fast FCP on mobile, at ~80%, although it's been dropping over time. The CSS page, which is quite large in terms of content and assets, has the fewest fast FCP experiences, at about 40%.

I'd argue that these pages should take more or less the same amount of time to download and render, and they should be closer to 80% fast than 40%. Ideally 100% fast!

The key to this, I think, is to stop server-side rendering pages. When we deploy the site, none of the HTML will change. So we should take a snapshot of the markup and serve those files statically as we would a JPEG. Sure, the caching headers can be different but the key is that the server doesn't have to think before it returns the contents. For larger documents like the CSS or HTTP/2 chapters, I'd love to explore HTTP chunking as a way to give clients a head start to load dependencies and render above-fold content early.

I think this can simplify our tech stack as well. For example, we can host our static assets (HTML, JS, CSS, images) on a host like Netlify or even directly from this repo using GitHub Pages. Without server-side code in production, we can more quickly deliver contents to users, which should improve our bottom-line metrics like engagement rates.

One very small exception to this is the Contributors page which shuffles the list of contributors on each request. The HTML is different each time and we can either shuffle in JS (bad because of a delay in rendering) or remove that feature entirely (bad because it unfairly features some contributors at the top more prominently). Open to discussion about how to resolve that.

@HTTPArchive/developers any thoughts on this proposal?

borisschapira:
As a strong JAMStack advocate, I'm in. I recommend using [11ty](https://www.11ty.dev/) which is a pretty straightforward node SSR. 11ty support nunjucks templates, and nunjucks is heavily inspired by jinja2 so I guess migrating the templates won't be long. And there's an easy way to deal with internationalization [through conventions and a content key](https://www.webstoemp.com/blog/multilingual-sites-eleventy/).

[...]

PR #893, opened 2020-06-22
2020 year support and basic 2020 landing page

bazzadp:
Makes progress on #815 

This adds the basic config and templates for 2020 support, enables the year drop down (if 2020 config is turned on - it's off by default) and adds a basic 2020 landing page with just a coming soon message and an updated 2020 hero image.

Every supported language needs an index.html page. And if chapter from that year is not supported in that language and is requested, it 302s back to the home page for that year.

If a completely unsupported chapter (i.e. not this not valid in the config) is requested it 404s.

On loading a page that is supported in other years the year drop down shows as you'd expect:

Year switcher showing 2019 and 2020

If only one year is supported for the current page then you get additional option to got to the home page of other years that are supported:

image

This works backwards also for new chapters added (e.g. JAMStack or WASM) and "2019 Home" will show for any such new chapters when looking at 2020 version.

I'm not 100% sure on UX of this, but didn't like the idea of the year selector not making it obvious there were other years - just not for this chapter. Suggestions welcome.

The 2020 style is currently exactly the same as the 2019 style - except it uses black instead of grey (I figured black was the colour of 2020, and with our yellow is in keeping with Black Lives Matter colour scheme!). This looks OK for the chapters but very dark for the home page and doesn't work for the hero image there as figures have black legs. I'm not a designer, so definitely open to other ideas here but will need help on this.

Because the styling is currently the same, everything is based off of 2019 templates/CSS but with some CSS style overrides for colour. This is implemented by adding the year to the body element to allow us to target years with CSS selectors:

```
<body class="year-2019">
```
or
```
<body class="year-2020">
```

You can play around with it here: https://20200622t113454-dot-webalmanac.uk.r.appspot.com/en/2020/ and I've copied the JavaScript chapter as an example chapter: https://20200622t113454-dot-webalmanac.uk.r.appspot.com/en/2020/javascript

Let me know your thoughts.


rviscomi:
Thanks @bazzadp! Eager to try it out.

Could you reenable 2020 for testing? Getting "Unsupported year requested" at the staging link.

bazzadp:
Updated the links after a few bug fixes. Try these links:

> You can play around with it here: https://20200622t113454-dot-webalmanac.uk.r.appspot.com/en/2020/ and I've copied the JavaScript chapter as an example chapter: https://20200622t113454-dot-webalmanac.uk.r.appspot.com/en/2020/javascript

rviscomi:
Regarding the bg color, IMO we should keep the styles identical to 2019, at least for this temp home page.

bazzadp:
> Regarding the bg color, IMO we should keep the styles identical to 2019, at least for this temp home page.

Cool. Can do that. What about home page? It's currently too small for the SVG background so solid grey for now?

rviscomi:
I was thinking about adding content to this page about how to contribute, so it may grow to fit the svg. If we can clip the svg overflow, let's keep it, otherwise solid bg color is fine.

bazzadp:
Updated:
- Home 2019 https://20200622t142821-dot-webalmanac.uk.r.appspot.com/en/2019/
- Home 2020 https://20200622t142821-dot-webalmanac.uk.r.appspot.com/en/2020/
- JS 2020 https://20200622t142821-dot-webalmanac.uk.r.appspot.com/en/2020/javascript

bazzadp:
Hero image needs some work - just bashed it together real quick. We really should reuse the mobile version and make the text actual text. Will raise and issue and maybe @catalinred will do some CSS magic again 😉

bazzadp:
Fixed the image. Would still be better to make this text if possible though IMHO.

src/config/2020.json
```diff
@@ -0,0 +1,225 @@
+{
+  "settings": [
+    {
+      "is_live": false,
+      "supported_languages": ["en","es","fr","ja"],
+      "ebook_languages": []
+    }
+  ],
+  "outline": [
```
rviscomi, line 9: Let's omit this for now until we have an actual 2020 outline. I have a question later about navigating to years' home pages consistently which might make this unnecessary for now.

src/config/2020.json
```diff
@@ -0,0 +1,225 @@
...
+      "name": "Editors"
+    },
+    "reviewers": {
+      "name": "Reviewers"
+    },
+    "translators": {
+      "name": "Translators"
+    }
+  },
+  "contributors": {
```
rviscomi, line 181: We could leave `contributors` empty and add to it when we have a Contributors page available.

src/main.py
```diff
@@ -59,11 +59,14 @@ def render_template(template, *args, **kwargs):
     language = get_language(lang)
     langcode_length = len(lang) + 1 # Probably always 2-character language codes but who knows!
 
-    # If the template does not exist, then redirect to English version
-    if (lang != '' and lang != 'en' and not(os.path.isfile('templates/%s' % template))):
-        return redirect('/en%s' % (request.full_path[langcode_length:]), code=302)
-
-    # Although a langauge may be enabled, all templates may not have been translated yet
+    # If the template does not exist, then redirect to English version if it exists, else home
+    if (lang != '' and not(os.path.isfile('templates/%s' % template))):
```
rviscomi, line 63: Unrelated thought. I wonder if it's necessary to call `isfile` each time as opposed to memoizing results, since we wouldn't expect the result to change.

src/static/css/2019.css
```diff
@@ -495,6 +495,13 @@ header.alt-bg a:focus, footer.alt-bg a:focus {
   align-items: center;
 }
 
+.year-2020 .btn,
+.year-2020 nav ul .toc,
+.year-2020 nav ul .contributors,
+.year-2020 nav ul .methodology {
+  display:none;
```
rviscomi, line 502: We should have a 2020 base template that simply omits these links, rather than hiding with CSS.

src/static/css/index.css
```diff
@@ -116,7 +116,8 @@ p {
   width: 86vw;
 }
 
-.intro-image-2019-wrapper {
+.intro-image-2019-wrapper,
+.intro-image-2020-wrapper {
```
rviscomi, line 120: WDYT about renaming this class to `.intro-image-wrapper` for year-agnostic styles?

src/templates/base/2019/accessibility_statement.html
```diff
@@ -36,6 +36,11 @@
 {% block styles %}
 {{ super() }}
 <link rel="stylesheet" href="/static/css/page.css?v=20200603172138">
+<style>
+.year-switcher {
+  display: none;
```
rviscomi, line 41: Similar to the 2020 styles, I think this is a job for the template.

src/templates/base/2019/base.html
```diff
@@ -274,6 +268,18 @@
             }
           });
         }
+        // Year switching
+        var yearSwitchers = document.querySelectorAll('.year-switcher select');
+        for (var i = 0; i < yearSwitchers.length; i++) {
```
rviscomi, line 273: The bodies of these for loops (lang and year) are functionally the same. Could we:

- write a `querySelectorAll` to include both lang and year switchers
- rewrite the loop variables to refer to generic switchers as opposed to year/lang-specific

src/templates/base/2019/base.html
```diff
@@ -341,18 +347,37 @@
 {% endmacro %}
 
 {% macro year_switcher(switcher_name) %}
+{# Only show if we currently support multi-year #}
+{% if year_live('2020') %}
```
rviscomi, line 351: How does this hold up when we have 2021?

I thought the `for y in supported_years` already handled this.

src/templates/base/2019/base.html
```diff
@@ -341,18 +347,37 @@
...
       {% for y in supported_years %}
         <option
           {% if y == year %}selected="selected"{% endif %}
-          value="{{ url_for('table_of_contents', **get_view_args(year=y)) }}">
-          {{ edition(y) }}
+          value="{{ url_for(request.endpoint, **get_view_args(lang=lang,year=y)) }}">
+          {{ y }}
         </option>
       {% endfor %}
+      {# If the current page doesn't support a year, then show a home link for that year #}
```
rviscomi, line 362: After demoing this, I'm torn on this behavior. It can create different experiences on different chapters. I might prefer the year switchers to always navigate to the year's home page. WDYT?

src/static/css/index.css
```diff
@@ -131,6 +132,15 @@ p {
   width: 40vw;
 }
 
+.intro-image-2020 {
```
rviscomi, line 135: WDYT about renaming this element to be year-agnostic and using the `.year-2020` class to specify how 2020 styles should be applied? For example:

```css
.year-2020 .intro-image { ... }
```

This affects 2019.

src/static/css/index.css
```diff
@@ -392,7 +402,8 @@ p {
     margin-top: 80px;
     margin-right: -46px;
   }
-  .intro-image-2019-wrapper {
+  .intro-image-2019-wrapper,
+  .intro-image-2020-wrapper {
```
rviscomi, line 406: Similarly, `.intro-image-wrapper`

src/templates/base.html
```diff
@@ -36,7 +36,7 @@
     {% endif %}
   </head>
 
-  <body>
+  <body class="year-{{ year }}">
```
rviscomi, line 39: Good idea!

src/templates/base/2019/base.html
```diff
@@ -127,15 +125,13 @@
             <span class="menu-btn__bar"></span>
           </button>
           <ul class="menu">
-            <li><a href="{{ url_for('table_of_contents', year=year, lang=lang) }}">{{ self.table_of_contents_title() }}</a></li>
-            <li><a href="{{ url_for('contributors', year=year, lang=lang) }}">{{ self.contributors_title() }}</a></li>
-            <li><a href="{{ url_for('methodology', year=year, lang=lang) }}">{{ self.methodology_title() }}</a></li>
+            <li><a href="{{ url_for('table_of_contents', year=year, lang=lang) }}" class="toc">{{ self.table_of_contents_title() }}</a></li>
```
rviscomi, line 128: Like I mentioned in another comment, I don't think we need classes on these links to hide them in CSS. I think the thing to do is overwrite this block in `2020/base.html`.

src/main.py
bazzadp: Remember that we cache all pages for 3 hours per CDN so, in effect, they are memorised for a bit for subsequent users in that location. Plus it's still pretty fast.

We could generate it as part of `npm run generate` of similar, and this goes back to you Static Site suggestion in #634. And similar to what I said back then, I don't think the performance hit is actually that noticeable, and the migration would be a large effort. So for this PR I say leave it as is.

src/static/css/2019.css
bazzadp: Until we decide how different 2019 and 2020 will be I say no. These are temporarily hidden - I fully expect to unhide them when launching 2020. So I don't think it's worth cloning the 2019 to 2020 templates just to remove these. If there are other reasons to create the 2020 templates (like it is a different design) then we can revisit this but for not I'd rather not have the extra maintenance when they will be pretty much direct clones.

src/static/css/index.css
bazzadp: Yes, but I would wait and do this part of #894 because at that point it is one wrapper.

src/templates/base/2019/accessibility_statement.html
bazzadp: You've more of a case with this one alright! 😀

I did try to override the macro's (defined in `base.html`) in this `accessibility_statement.html` template but it wouldn't work. The accessibility statement is a special case - it's the only template which is language but not year specific. We've still decide how to handle this if 2019 and 2020 differ (ref: https://github.com/HTTPArchive/almanac.httparchive.org/pull/665#discussion_r379591735). So for now I'm happy with this but let's revisit once we have more decisions on 2019 and 2020.

And btw any thoughts on Methodology? Will that differ per year? As could be a lot of repetition on that too.

src/templates/base/2019/base.html
bazzadp: Fair point. Done.

src/templates/base/2019/base.html
bazzadp: I disagree. If I'm into SEO (for example) I might want to flick between the different years and read chapters on each. Flicking back to home page creates a needless extra step.

At the end of the day, these are supposed to be short term workarounds. We should aim to launch the English version with most (if not all) the chapters at which point there is no different experience.

In that way it's similar with the translations where we flick back to English versions. I know you had similar reservations about that though. And yes the translations may take longer (and we've seen with 2019 that there is a risk that some will be in an uncompleted state for the long term), but that is unfortunate and I'd rather not hold back the languages that are fully available because some are not.

src/static/css/index.css
bazzadp: As per above this should be part of #894

src/static/css/index.css
bazzadp: As per above this should be part of #894

src/templates/base/2019/base.html
bazzadp: And, as per other comment, I'm against creating a `2020/base.html` right now as see no reason (other than this - which is a short term, temporary fix). We can remove these later if we want once we either have a reason for `2020/base.html` or launch 2020 properly with these pages using 2019 templates.

src/config/2020.json
bazzadp: I still think it's useful for have the ability to distinguish between a chapter that doesn't exist yet but should (302) and one that shouldn't and has been typed wrong (404). Even if we do change the navigation. Of course we can keep that logic and just not have chapters configured yet, but makes testing that painful (reminder to self - we should write test cases fo this!).

Also reminds us of the structure to save us having to copy it later.

So it is useful, and I don't see any reason or harm to have this in here now. So my preference would be to leave it.

src/templates/base/2019/accessibility_statement.html
rviscomi: Yes, Methodology pages will differ by year, although not by much. I'd like to see the methodologies as specific as possible for that year as a kind of "audit trail" in case anyone needed to see how we arrived at certain results. I fully expect there to be mostly minor differences year to year, but sometimes we may make more dramatic changes. In either case, we should make the doc as fully customized as possible. (I'll take ownership of that)

src/config/2020.json
bazzadp: Similar to above, I see no harm in adding it now. I also think it would make it easier for contributors to add themselves as we go along. Some people may do some development, then bore and drop out - like happened last year. It was a bit of an effort last time to get people added to this after the fact and we even forgot one or two.

src/templates/base/2019/accessibility_statement.html
bazzadp: OK. May need to be careful with Google seeing it as duplicate content.

Fixed the accessibility statement issue. Figured out [how to override macros in Jinja2](https://stackoverflow.com/questions/26582731/redefining-imported-jinja-macros).

src/templates/base/2019/base.html
bazzadp: `supported_years` is years supported by THIS template. I added an `all_supported_years` variable and generalised this code. Was being a bit lazy so fair enough for calling this out.

Resolved.

src/static/css/2019.css
bazzadp: I'd be OK to have the following in the `base/2019/base.html` template if you think this is better than CSS hiding - would also avoid us having to add the class:

```
        <ul>
          {% if year != '2020' %}
          <li><a href="{{ url_for('table_of_contents', year=year, lang=lang) }}" class="toc">{{ self.table_of_contents_title() }}</a></li>
          <li><a href="{{ url_for('contributors', year=year, lang=lang) }}" class="contributors">{{ self.contributors_title() }}</a></li>
          <li><a href="{{ url_for('methodology', year=year, lang=lang) }}" class="methodology">{{ self.methodology_title() }}</a></li>
          {% endif %}
          <li>
            {{ year_switcher('footer') }}
          </li>
          <li>
            {{ language_switcher('footer') }}
          </li>
        </ul>
```

bazzadp: OK figured out a better way to do this in https://github.com/HTTPArchive/almanac.httparchive.org/pull/893/commits/d5b2e678a96e63644b8697f213b4489e29805576

src/templates/base/2019/base.html
bazzadp: As per other comment, I figured out better way to do this in https://github.com/HTTPArchive/almanac.httparchive.org/commit/d5b2e678a96e63644b8697f213b4489e29805576 (and follow up in https://github.com/HTTPArchive/almanac.httparchive.org/pull/893/commits/a9baa2b564c14fa2b50c08f5ca5104956bac8f73)

src/static/css/index.css
bazzadp: Fixed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/893/commits/b5351ed6b55a2e02432f50f99765910e1c794ae9

src/static/css/index.css
bazzadp: Fixed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/893/commits/b5351ed6b55a2e02432f50f99765910e1c794ae9

src/static/css/index.css
bazzadp: Fixed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/893/commits/b5351ed6b55a2e02432f50f99765910e1c794ae9

PR #930, opened 2020-06-28
CSS logo

### LINKED ISSUE #567: i18n - Reimplement Almanac logo using text and CSS instead of SVG
The [PNG Logo](https://almanac.httparchive.org/static/images/logo.png) contains English wording. It looks like it's created from [the SVG logo](https://github.com/HTTPArchive/almanac.httparchive.org/blob/master/src/static/images/logo.svg) unfortunately, the SVG logo has been optimized for Performance and all original text has been replaced by paths.

Is there a way to get an editable version to localize the "By" wording (in French, "par").

catalinred:
Fixes #567
The logo is now made with CSS

- The logo degrades nicely when switching to an `rtl` text direction.
- Adress the Japan edge case when the proper logo translation is a bit different than just the preposition text. 
  Looking forward to @MSakamaki feedback here.
- The logo is perfectly scalable by adjusting `.navigation-logo`'s font-size size.
- I also fixed a horizontal scroll issue that I've noticed recently on the index page.

Let me know your thoughts!

### DIFF
--- a/src/templates/base/2019/base.html
+++ b/src/templates/base/2019/base.html
@@ -104,7 +104,19 @@
   <header class="alt-bg">
     <div class="container">
       <div class="top-header">
-        <a class="navigation-logo" href="{{ url_for('home', year=year, lang=lang) }}"><img src="/static/images/logo.svg" alt="{{ self.organization() }}" height="39" width="202"></a>
+        <a class="navigation-logo" href="{{ url_for('home', year=year, lang=lang) }}">
+          {% if language.lang_code != 'ja' %}
+            <span class="wa">{{ self.web_almanac() }}</span>
+            <span class="line-group">  
+              <span class="pre">{{ self.preposition() }}</span>
+              <span class="ha">{{ self.http_archive() }}</span>
+            </span>
+          {% else %}
+            <span class="ha line-group">{{ self.http_archive() }}</span>            
+            <span class="pre">{{ self.preposition() }}</span>
+            <span class="wa">{{ self.web_almanac() }}</span>                        
+          {% endif %}
+        </a>
         <nav aria-label="{{ self.page_navigation() }}">
           <ul>
             <li><a href="{{ url_for('table_of_contents', year=year, lang=lang) }}">{{ self.table_of_contents_title() }}</a></li>
@@ -142,7 +154,7 @@
             <li class="misc">
               <ul class="misc">
                 <li>
-                  <a href="https://httparchive.org/" class="navigation-logo" aria-labelledby="ha-logo-mobile">
+                  <a href="https://httparchive.org/" aria-labelledby="ha-logo-mobile">
                     <svg width="70" height="35" role="img">
                       <title id="ha-logo-mobile">{{ self.http_archive_link() }}</title>
                       <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#ha"></use>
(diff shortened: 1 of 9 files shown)


borisschapira:
Ok for the French modifications 👍

bazzadp:
This sounds awesome @catalinred. Will check it out later.

src/templates/base/2019/base.html
```diff
@@ -104,7 +107,19 @@
...
+            <span class="wa">{{ self.web_almanac() }}</span>
+            <span class="line-group">  
+              <span class="pre">{{ self.preposition() }}</span>
+              <span class="ha">{{ self.http_archive() }}</span>
+            </span>
+          {% else %}
+            <span class="ha line-group">{{ self.http_archive() }}</span>            
+            <span class="pre">{{ self.preposition() }}</span>
+            <span class="wa">{{ self.web_almanac() }}</span>                        
+          {% endif %}
```
bazzadp, line 121: This is very clever, but not loving the hard coding of the language.

What would you think about moving this to the templates? So replacing this with just:

```suggestion
          {{ self.web_almanac_logo() }}
```

And then having the following in `en/2019/base.html`:

```
{% block web_almanac_logo %}
<span class="wa">Web Almanac</span>
<span class="line-group">  
   <span class="pre">By</span>
   <span class="ha">HTTP Archive</span>
</span>
{% endblock %}
```

Yes it does put a bit of code/presentation logic into the language templates but we have a few other examples of this, gives us much greater flexibility for other languages in the future, and avoid repeating this code three times.

src/templates/base/2019/base_ebook.html
```diff
@@ -28,7 +28,7 @@
   @page {
     @top-left { content: ''; }
     @top-right { content: string(chapter-subtitle) string(chapter-title); }
-    @bottom-left { content: "{{ year }} {{ self.organization() }}"; }
+    @bottom-left { content: "{{ year }} {{ self.web_almanac() }} {{ self.preposition() }} {{ self.http_archive() }}"; }
```
bazzadp, line 31: This doesn't work for Japanese. So think we should maybe keep `self.organization()` (and the associated block).

src/templates/base/2019/base_ebook.html
```diff
@@ -45,7 +45,7 @@
     @top-left { content: string(chapter-subtitle) string(chapter-title); }
     @top-right { content: ''; }
     @bottom-left { content: counter(page); }
-    @bottom-right { content: "{{ year }} {{ self.organization() }}"; }
+    @bottom-right { content: "{{ year }} {{ self.web_almanac() }} {{ self.preposition() }} {{ self.http_archive() }}"; }
```
bazzadp, line 48: Same here

src/templates/base/2019/base_ebook.html
```diff
@@ -54,7 +54,7 @@
   @page :right {
     @top-left { content: ''; }
     @top-right { content: string(chapter-subtitle) string(chapter-title); }
-    @bottom-left { content: "{{ year }} {{ self.organization() }}"; }
+    @bottom-left { content: "{{ year }} {{ self.web_almanac() }} {{ self.preposition() }} {{ self.http_archive() }}"; }
```
bazzadp, line 57: Same here

src/templates/base/2019/base.html
catalinred: I did update this as requested! ;)

src/templates/base/2019/base_ebook.html
catalinred: Done.

src/templates/base/2019/base_ebook.html
catalinred: Done.

src/templates/base/2019/base_ebook.html
catalinred: Done.

catalinred:
Awesome, thanks for merging.

PR #931, opened 2020-06-28
Fix index font zoom

bazzadp:
Just noticed while reviewing another PR that when font zooming to 32px base font, that the chapter table of contents look a little squished (especially on Firefox):

Bad font zoom with fixed line height

This PR adds `rem` sizing to the `line-height` to fix this.

Better font zoom with relative line height

### DIFF
--- a/src/content/en/2020/README.md
+++ b/src/content/en/2020/README.md
@@ -1,1 +1,0 @@
-This is the directory in which content teams add the markdown version of their chapters.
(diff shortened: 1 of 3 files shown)


bazzadp:
Yeah thought you’d say that! 😀

I mostly keep it there to remind me - easier to think in `rem`‘s than px`. It’s the convention I’ve gone with elsewhere:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/9494a97c05a8c5ccdd04c2d180720c7ad9839878/src/static/css/2019.css#L301-L302

Also means `rem`-less browsers still get support, but thankfully [that really is pretty unusual these days](https://caniuse.com/#feat=rem), so it mostly is for my own understanding.

PR #932, opened 2020-06-28
Allow the development server to be backgrounded with optional command line argument

bazzadp:
Allows the werkzeug development server to be started in background mode with the following syntax:

```
   python main.py background &
```

This can then be used to automate runs, for example to get our GitHub action to regenerate the ebooks on commit.

Without this the server fails is started in background mode, with the following:

```
Traceback (most recent call last):
  File "/Users/barry/almanac.httparchive.org/src/main.py", line 356, in <module>
    app.run(host='127.0.0.1', port=8080, debug=True)
  File "/Users/barry/almanac.httparchive.org/src/env/lib/python3.8/site-packages/flask/app.py", line 943, in run
    run_simple(host, port, self, **options)
  File "/Users/barry/almanac.httparchive.org/src/env/lib/python3.8/site-packages/werkzeug/serving.py", line 1008, in run_simple
    run_with_reloader(inner, extra_files, reloader_interval, reloader_type)
  File "/Users/barry/almanac.httparchive.org/src/env/lib/python3.8/site-packages/werkzeug/_reloader.py", line 333, in run_with_reloader
    ensure_echo_on()
  File "/Users/barry/almanac.httparchive.org/src/env/lib/python3.8/site-packages/werkzeug/_reloader.py", line 322, in ensure_echo_on
    termios.tcsetattr(sys.stdin, termios.TCSANOW, attributes)
termios.error: (4, 'Interrupted system call')
```

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -10,6 +10,7 @@ from werkzeug.http import HTTP_STATUS_CODES
 from validate import validate
 import os.path
 import re
+import sys
 
 # Set WOFF and WOFF2 caching to return 1 year as they should never change
 # Note this requires similar set up in app.yaml for Google App Engine
@@ -24,7 +25,7 @@ app = MyFlask(__name__)
 # Flask default if not set is 12 hours but we want to match app.yaml
 # which is used by Google App Engine as it serves static files directly
 app.config['SEND_FILE_MAX_AGE_DEFAULT'] = 10800
-Talisman(app,
+talisman = Talisman(app,
          content_security_policy=csp,
          content_security_policy_nonce_in=['script-src'])
 logging.basicConfig(level=logging.DEBUG)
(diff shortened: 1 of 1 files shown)


src/main.py
```diff
@@ -342,4 +343,14 @@ def handle_bad_gateway(e):
...
-    app.run(host='127.0.0.1', port=8080, debug=True)
+
+    # If the 'background' command line argument is given:
+    #    python main.py background &
+    # then run in non-debug mode, as debug mode can't be backgrounded
+    # but debug mode is useful in general (as auto reloads on change)
+    if (len(sys.argv) > 1 and sys.argv[1] == 'background'):
+        # Turn off HTTPS redirects (automatically turned off for debug)
+        talisman.force_https=False
+        app.run(host='127.0.0.1', port=8080)
```
ibnesayeed, line 354: Pinning `host` to `127.0.0.1` or `localhost` will make it difficult to run in a Docker container or VM when the intent is to access from the host machine. Should we use `0.0.0.0` instead?

bazzadp: If we want to run docker, then perhaps we should look at this - as we're discussing in #933 

For now, I'm going to resolve this, as `127.0.0.1` is not something introduced in this PR so think changing that is separate issue to what we're trying to address here.

PR #939, opened 2020-06-29
GitHub linter

### LINKED ISSUE #938: Implement linters for Python, CSS, JS, JSON and YML files
As discussed in https://github.com/HTTPArchive/almanac.httparchive.org/issues/933#issuecomment-651228508 GitHub has a [superlinter](https://github.blog/2020-06-18-introducing-github-super-linter-one-linter-to-rule-them-all/) which will help to find typos and ensure a consistent coding style.

This can run through a GitHub action when a pull request is submitted so the requester gets quick feedback and can correct their code.

Downsides:
- Linting can lead to a large clean up when applying to old code
- It can enforce overly strict rules that annoy more than help. Key is to tune the rules to those useful for us. And change them if we want.

bazzadp:
Fixes #938 and implements linters on all our languages except HTML/Jinja2 templates.

Note it only runs on the files in the PR, and it does not block opening the PR (nor even merging if we want to ignore the errors - but let's try to avoid that!).

@ibnesayeed could you review the config and let me know your feedback?

@rviscomi / @catalinred quite a few changes to the files (particularly CSS) correcting minor errors or styling fixes. Would be good if one (or both!) of you could review those in case I made a typo. Have run the site and all looks good to me.

A lot of commits, as that was way of testing it - but they'll all be collapsed into one if/when we merge.

We can fine tune the rules if we find them too strict. Have already done that but let the ones in that I thought would be useful and ensure consistency.


.github/workflows/linter.yml
```diff
@@ -0,0 +1,62 @@
...
+#  * YML - default rules
+# Exception config files are in the .github/linters directory
+
+#############################
+# Start the job on all push #
+#############################
+on:
+  push:
+    branches-ignore:
+      - 'main'
```
ibnesayeed, line 27: Are we relying on contributors to ensure their code was linted in their branches and not running in in PRs?

.github/linters/.markdown-lint.yml
```diff
@@ -0,0 +1,38 @@
...
+MD029: false                  # Ordered list item prefix
+MD033: false                  # Allow inline HTML
+MD034: false                  # Allow base URLs
+MD036: false                  # Emphasis used instead of a heading
+MD040: false                  # Don't demand language for all code blocks
+
+#################
+# Rules by tags #
+#################
+blank_lines: false  # Error on blank lines
```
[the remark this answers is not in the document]
bazzadp: Fair point. Fixed.

.github/linters/.python-lint
```diff
@@ -0,0 +1,551 @@
...
+
+# List of valid names for the first argument in a metaclass class method.
+valid-metaclass-classmethod-first-arg=mcs
+
+
+[EXCEPTIONS]
+
+# Exceptions that will emit a warning when being caught. Defaults to
+# "Exception"
+overgeneral-exceptions=Exception
```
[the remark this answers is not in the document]
bazzadp: Fixed

.github/linters/.stylelintrc.json
```diff
@@ -0,0 +1,45 @@
...
+            }
+        ],
+        "value-keyword-case": [
+            "lower",
+            {
+                "ignoreKeywords": ["Art","A4"]
+            }
+        ]
+    }
+}
```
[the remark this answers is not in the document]
bazzadp: Fixed

.github/workflows/linter.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: docker://github/super-linter:v2.2.0
+        env:
+          VALIDATE_ALL_CODEBASE: false
+          DEFAULT_BRANCH: main
+          VALIDATE_YAML: true
+          VALIDATE_JSON: true
+          VALIDATE_PYTHON: true
+          VALIDATE_JAVASCRIPT_ES: true
+          VALIDATE_CSS: true
+          VALIDATE_MD: true
```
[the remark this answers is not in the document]
bazzadp: Fixed.

.github/workflows/linter.yml
bazzadp: That's a bit sucky that the [example action config](https://github.com/github/super-linter#example-connecting-github-action-workflow) is for pushed. Changed it to run on pull requests instead.

.github/workflows/linter.yml
```diff
@@ -0,0 +1,59 @@
...
+######################################
+# Start the job on all pull requests #
+######################################
+on: pull_request
+
+###############
+# Set the Job #
+###############
+jobs:
+  build:
```
ibnesayeed, line 30: Linting only on PRs means people will not get their code linted as they commit into their forks before creating a PR.

```suggestion
#########################
# Lint each push and PR #
#########################
on:
  - push
  - pull_request

###############
# Set the Job #
###############
jobs:
  lint:
    ########################################
    # Skip duplicate job for local brances #
    ########################################
    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
```

bazzadp: That’s true. Though on other hand some might get annoyed at running a whole lint job on every commit if code is work in progress. But guess they can just ignore the feedback.

Can you explain that last line to me? So that means it will run on all pushes, but only in pull requests from forks?

If so I’ve two questions:
1. We probably don’t need to run this on pushes to `main` in this primary repo, after accepting a PR. But then do want to run on `main` for forks as some people do develop on that.
2. Does this mean it will not run on ably PRs from this repo? Some of us work in branches in the main repo rather than forks and I would like to know when the PR is submitted that the linters are all green.

ibnesayeed: It is very important to run lint and test tasks on each push to ensure no push is causing any new issues. These CI processes run asynchronously, so they should not bother anyone. On the other hand, one can leverage them to incrementally improve the work in progress. Note that these jobs will only run when one or more commits are actually pushed to GH infrastructure, not when we are committing changes incrementally on our local development machines.

It is also important to run these tasks on PRs to ensure that the maintainers decide what failures they are willing to tolerate, not the contributor who may have created a PR with failing tests/lints in her/his fork.

This means, when maintainers create a branch in this repo (not in a separate fork) and then create a PR to the main branch, these jobs will be triggered twice at the same time in the same repo, one for the pus action and the other for the PR. This is where the `if:` condition comes in handy. It will let all the pushes run the job, but in case of PRs, it will only run name of the head repo is not the same as this repo (i.e., skipping jobs on PRs from the local branches). I will note here that in case of skipped jobs, we will see an ugly looking skipped task in the PR, but that is better than simultaneous duplicate run. I hope GH comes up with a better solution for this scenario.

Finally, we can skip this job to run when changes are pushed (or PRs are merged) into teh main branch, but there is actually no harm if we let it run once again asynchronously. That's why I did not exclude the main branch. This also means, if someone with write permission accidentally pushes something to the main branch directly, at least those changes will be linted/tested.

bazzadp: I have a slightly different take:

It is very important for them to run on PRs, as that means the results can be checked before the PR is merged. The result is then also shown in the pull result itself allowing a reviewer to check that as part of their PR review.

It would be very handy to run them on every commit, so developers can correct as they go along, and won't have a load of corrections to do at the end, after they've opened the pull request. The downside to this, is that there is no GUI to note the fact it fails (like in pull requests). Yes the pusher should get an email, and it's available in the Actions tab but not quite as nice as seeing a nicely linked status in the pull request. Another downside is GitHub actions can take time, and you are limited to the amount you can run. So two or three commits in quick succession could create a queue (how often have we all hit "push" – or "send" on an email – and then immediately spotted a typo!). The fact it only runs on the changed files is good, but the action still takes a few mins, to launch, checkout that branch, download the action docker file...etc.

So I agreed it would be ideal to have both, but to me doing it on a PR is a _must-have_ for me to protect the code, and doing it on a push is a _nice-to-have_.

My concern with not doing it on PRs for branches with-in this repo is would the results show in the pull request? Let's say I opened a branch (with-in this repo), then committed bad code to that branch, ignored the lint failures, and then opened a pull request. I thought with your change would me the pull request reviewer would not see the results of the action in the pull request. But testing it, it looks like that is not the case - it _does_ show the action of the last action in the pull request. Excellent - didn't realise it did that.

So have made your change and will merge if you are happy now?

bazzadp:
> In the future, we might fine-tune stylelint's `number-leading-zero` to `never` maybe.

Out of interest, is this a language/region specific thing? E.g. In French or Spanish is .5 more common than 0.5? Personally I think the latter is much clearer, but that could be just because I'm more used to it.

There's a few others we may want to consider turning off that I would be OK with (e.g. allow `0px` rather than insist on `0`), and some I'd more prefer to leave on but could be persuaded (e.g. allowing `#5C687D` as well as `#5c687d`, or allowing `#ffffff` as well as `#fff` - I'd prefer to standardise to make any search and/or replace easier).

I've tried to balance the rules with usefulness, versus painfulness. They are reasonably strict at the moment but didn't take a huge amount to get our code into shape, so hopefully won't be too burdensome. We can always tweak if that is the case and the linter configurations in this PR should give whoever wants to do that, a pointer on how to do that.

catalinred:
Definitely not a region or language specific thing but I usually prefer writing `.5` instead of `0.5` or `0` instead of `0px`. 

Not that size is a real problem in this case, but this results in less bytes and cleaner CSS declarations.

bazzadp:
> Definitely not a region or language specific thing but I usually prefer writing `.5` instead of `0.5` or `0` instead of `0px`.
> 
> Not that size is a real problem in this case, but this results in less bytes and cleaner CSS declarations.

OK, personally I think `0.5` is clearer and less likely to miss the small `.`, and that's the way it's currently set up. Then again, I also think `0px` is also clearer but swallowed my preference there and did go with the recommended `0` for that. Let's leave it as is for now and see if lots of others prefer differently and can tweak.

@ibnesayeed will wait for your feedback here and then merge. Think this will be a great addition to the development process so thanks for suggesting!

.github/workflows/linter.yml
```diff
@@ -0,0 +1,65 @@
...
+#  * Python - Only change from default is to set ignored-modules=flask, flask-talisman...etc.
+#  * CSS - Tweaked to ignore some Prince/Normalise exceptions and some whitespace exceptions
+#  * JS - default rules
+#  * JSON - default rules
+#  * YML - default rules
+# Exception config files are in the .github/linters directory
+
+######################################
+# Start the job on all pull requests #
+######################################
```
ibnesayeed, line 23: This comment does not reflect the actual situation.

```suggestion
##########%%%###############
# Lint on each push and PR #
############################
```

.github/workflows/linter.yml
```diff
@@ -0,0 +1,65 @@
...
+on:
+  - push
+  - pull_request
+
+###############
+# Set the Job #
+###############
+jobs:
+  build:
```
ibnesayeed, line 32: It is a not a build job (not that it will affect the process).

```suggestion
  lint:
```

.github/workflows/linter.yml
ibnesayeed: > My concern with not doing it on PRs for branches with-in this repo is would the results show in the pull request?

Yes, two tasks will show up in the PR that will be triggered due to two two different events (push and PR). However, they will essentially be doing the same thing simultaneously, so it is better to skip one of them. That's why I suggested the scary-looking `if:` condition.

.github/workflows/linter.yml
bazzadp: Fixed. In fact I've rewritten all the comment lines more in keeping with our other actions.

.github/workflows/linter.yml
bazzadp: Fixed.

.github/workflows/linter.yml
```diff
@@ -0,0 +1,48 @@
...
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+    # Set the agent to run on
+    runs-on: ubuntu-latest
+    steps:
+      # Checkout the code base
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      # Run Linter against code base #
+      - name: Lint Code Base
```
ibnesayeed, line 38: Thanks for removing unnecessary comment bloat. We can assume that those looking at the code can understand the obvious. I have removed some additional inline comments. Even if you chose not to remove the ones I suggest we should, please delete the trailing ` #` from two comments and correct misspelled `brances`.

```suggestion
name: Lint Code Base
on:
  - push
  - pull_request

jobs:
  lint:
    name: Lint Code Base
    # Skip duplicate job for local branches
    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
    runs-on: ubuntu-latest
    steps:
      - name: Checkout Code
        uses: actions/checkout@v2

      - name: Lint Code Base
```

ibnesayeed:
> OK, personally I think `0.5` is clearer and less likely to miss the small `.`, and that's the way it's currently set up. Then again, I also think `0px` is also clearer but swallowed my preference there and did go with the recommended `0` for that. Let's leave it as is for now and see if lots of others prefer differently and can tweak.

I too am in the same camp who used to write `0.5`, `0px`, and `#ffffff`, but if I remember correctly, Chrome DevTools (or some related article which I can't recall now) suggested me against these and the rationale was performance.

.github/workflows/linter.yml
bazzadp: Left the comments in for the steps but removed the rest

ibnesayeed: When the names of steps are clear enough, we do not need to say the same things in the comments. For example, if the name says, `Checkout Code`, then the comment `# Checkout the code base` does not add any value to it. I think, I will stop here, because I am clearly nitpicking now. ;-)

bazzadp: Nah you're right. Removed.

Issue #947, opened 2020-07-03
Make header/footer padding consistent


rviscomi:
On the 2020 landing page, the spacing inconsistency is more obvious because there is so little main body content:

![image](https://user-images.githubusercontent.com/1120896/86488988-dfdb6a00-bd30-11ea-84b1-07889a32955e.png)

The header, body, and footer should all have the same padding. Currently the footer has less padding.

Also ensure that the spacing is consistent on mobile.

bazzadp:
Good spot. While we're looking at this, I'm still not convinced the home page should go as wide as it does, as discussed in #574 - at least for the top part. Leaves a big old empty space in the middle on ultra-wide monitors:

![Home page on ultra-wide monitor with big grey section in the middle](https://user-images.githubusercontent.com/10931297/86489493-65c0da80-bd5c-11ea-8a29-74e62e1e1dea.png)

dsadhanala:
@bazzadp what you think about, treating home page hero section changes as separate issue?

Incase if we want to include that change also part of this issue, couple of questions.
* can you help me understand the expected behavior?
* do we want to scale font and layout spacings to adapt wide screen sizes?

dsadhanala:
@rviscomi QQ, I noticed, header and footer visual treatment for internal pages(other than home page) is different, example header content set to max size of 1280, where as footer resizes based on viewport no max set. Should we keep both header and footer content resize behavior consistent?

dsadhanala:
Footer content alignment looks little odd, especially logo aligned to left and leaving whitespace on the right, then navigation links aligned center. what you think about center aligned logo to be consistent with nav links alignment? below is the screenshot for your reference.

<img src="https://user-images.githubusercontent.com/5795823/86502498-ad456780-bd58-11ea-8f29-6ed55780406c.png" width="600px" />

rviscomi:
> Should we keep both header and footer content resize behavior consistent?

Good catch, yes that sounds like a good idea.

> what you think about center aligned logo to be consistent with nav links alignment?

Yeah that looks good. Could you also center the copyright and accessibility statement text?

dsadhanala:
thanks, all above mentioned changes are pushed into this PR. https://github.com/HTTPArchive/almanac.httparchive.org/pull/951

bazzadp:
> @bazzadp what you think about, treating home page hero section changes as separate issue?
> 
> Incase if we want to include that change also part of this issue, couple of questions.
> 
[...]

PR #951, opened 2020-07-04
Make header/footer padding consistent

dsadhanala:
Make header/footer padding consistent, fix for #947
- update footer styles to use 60px same as header content for non-mobile viewports

### DIFF
--- a/src/config/2020.json
+++ b/src/config/2020.json
@@ -220,6 +220,16 @@
       "avatar_url": "https:\/\/avatars0.githubusercontent.com\/u\/1120896?v=4&s=200",
       "github": "rviscomi",
       "twitter": "rick_viscomi"
+    },
+    "dsadhanala": {
+      "name": "Durga Prasad Sadhanala",
+      "teams": [
+        "developers"
+      ],
+      "avatar_url": "https:\/\/avatars3.githubusercontent.com\/u\/5795823?s=200&v=4",
+      "website": "https:\/\/web-ui.dev",
+      "github": "dsadhanala",
+      "twitter": "dsadhanala"
     }
   }
 }
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -752,7 +752,6 @@ footer {
   padding-right: 20px;
 }
 
-
 /* padding match with header content for non-mobile viewports */
 @media (min-width: 601px) {
   footer {
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -740,24 +740,13 @@ blockquote::before {
 }
 
 /* Footer */
-footer {
+footer .container {
   display: grid;
   grid-template-areas:
     'home-logo nav-items nav-items nav-items'
     'hr hr hr hr'
     'copyright copyright ha-logo social-media';
   grid-template-columns: 280px 1fr 1fr 120px 0 0;
-  /* 20px same as .container padding */
-  padding-left: 20px;
-  padding-right: 20px;
-}
-
-/* padding match with header content for non-mobile viewports */
-@media (min-width: 601px) {
-  footer {
-    padding-left: 60px;
-    padding-right: 60px;
-  }
 }
 
 footer .home-logo {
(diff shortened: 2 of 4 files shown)


src/static/css/2019.css
```diff
@@ -752,6 +752,14 @@ footer {
   padding-right: 20px;
 }
 
+/* padding match with header content for non-mobile viewports */
+@media (min-width: 601px) {
```
rviscomi, line 756: Not sure that the media query is necessary. Could you overwrite the padding left/right in the original `footer` style?

dsadhanala: I can but we do 20px for header content on mobile devices, so I just wanted to keep it consistent 60px for large screens and keep 20px for mobile. let me know that's not the expected behavior.

rviscomi: Oh ok. The convention we use is to make the desktop styles the default and use media queries for <= 600px to handle mobile styles.

dsadhanala: got it, thanks!

Since we want to keep header and footer for both internal and index page same, we no longer need this style update.
I'll publish new changes as discussed in the #947 to address few additional issues.

dsadhanala:
@rviscomi here are the screenshots for both home and internal pages.
as discussed in #947 made changes accordingly to keep both header and footer content flow consistently. can you please review the changes?

### home page
image

### internal pages
image

src/templates/base/2019/base.html
```diff
@@ -184,79 +184,81 @@
...
-            {{ year_switcher('footer') }}
-          </li>
-          <li>
-            {{ language_switcher('footer') }}
-          </li>
-        </ul>
-      </nav>
-      <div class="mobile-ha-social-media">
-        <a class="ha-logo" href="https://httparchive.org/" aria-labelledby="httparchive-logo-footer-mobile">
+      <div class="container">
```
dsadhanala, line 187: note, only change I made here is adding `<div class="container">` and wrap footer content in it.

bazzadp:
Any reason to not be consistent in header and footer between internal and home page?

bazzadp:
@dsadhanala, can you also add yourself to the `src/config/2020.json` file with the developers team?

dsadhanala:
> Any reason to not be consistent in header and footer between internal and home page?
@bazzadp It's based on current implementation in prod, because content on homepage is flexible to take over full viewport width, where as internal pages defined with max width 1280px. we can update home page also to limit max width to 1280px if that is intended experience.

Since I don't have historical knowledge about the current implementation in prod and I didn't want to introduce new behavior keep it as is, and only fixed spacing inconsistencies. Let me know the intended experience., I'll make changes accordingly.

src/config/2020.json
```diff
@@ -220,6 +220,16 @@
...
+    },
+    "dsadhanala": {
+      "name": "Durga Prasad Sadhanala",
+      "teams": [
+        "developers"
+      ],
+      "avatar_url": "https:\/\/avatars3.githubusercontent.com\/u\/5795823?s=200&v=4",
+      "website": "https:\/\/web-ui.dev",
+      "github": "dsadhanala",
+      "twitter": "dsadhanala"
```
bazzadp, line 232: Any chance you could move this up above Ricks? I try to keep it in first name alphabetical order to make it easier to find, avoid any thoughts of ordering and satisfy my OCD (mostly the last one to be honest!)

src/templates/base/2019/base.html
bazzadp: Thanks for the explanation - would really have confused without that!

src/config/2020.json
dsadhanala: no problem, I'll move this. thanks for the review!

dsadhanala:
@bazzadp @rviscomi all the feedback has been addressed, please take a look at this PR when you get a chance!

bazzadp:
Merged. Thanks! Great job.

PR #952, opened 2020-07-04
Web vitals

### LINKED ISSUE #944: Add Web Vitals to analytics
The web-vitals.js library can [augment](https://github.com/GoogleChrome/web-vitals#send-the-results-to-google-analytics) Google Analytics with the largest contentful paint (LCP), first input delay (FID), and cumulative layout shift (CLS) metrics. It'd be great to include these in our existing monitoring tools.

max-ostapenko:
Fixes #944


rviscomi:
Curious to hear @bazzadp's thoughts on the new browserify and babel dependencies.

src/tools/generate/web-vitals.js
```diff
@@ -0,0 +1,15 @@
+import {getCLS, getFID, getLCP, getFCP, getTTFB} from 'web-vitals';
```
rviscomi, line 1: TTFB is unused. Could we log it too?

bazzadp:
> Curious to hear @bazzadp's thoughts on the new browserify and babel dependencies.

Not loving it to be honest. We’ve a very light JS implementation at the moment and this seems a lot to add just for this. Do we really need them? [This](https://github.com/GoogleChrome/web-vitals#load-web-vitals-from-a-cdn) suggests not.

bazzadp:
Appreciate this is only needed to compile the script and actually added to the site so not massively concerned if we’re prefer to keep them in. But seems to could just download the pre-compiled version and remove the dependency.

src/tools/generate/web-vitals.js
max-ostapenko: Added.

src/templates/base.html
```diff
@@ -21,6 +21,7 @@
           gtag('js', new Date());
           gtag('config', 'UA-22381566-3');
       </script>
+      <script defer src="/static/js/bundle.js" nonce="{{ csp_nonce() }}"></script>
```
bazzadp, line 24: Don’t need the nonce here - should only need it for inlined JS.

Can we have a better name than bundle.js? web-vitals.js?

I do wonder if we should move the GTM and GA config into one script with this (common.js?). Though the fact this is `defer` (deliberately I’m guessing to get the full load?) probably means that’s not a good idea.

max-ostapenko:
@bazzadp I was concerned regarding adding unpkg.com to CSP, is it fine?
Alternatively, we could install web-vitals npm package and add filesystem copying command to generation scripts.

rviscomi:
I think I prefer the npm and static/js approach over unpkg, and both of those options over saving a local copy that will inevitably go stale.

bazzadp:
Oh definitely want it local rather than loaded from third-party.

It’s a far comment about it going stale. But it still requires a command to generate the new version - so how will we know when to run that? And what’s the difference between running that periodically and just downloading it from unpkg to our own static dir periodically?

I would have thought compile option was more for if you wanted to compile it into other code you were writing rather than the straight compile we’re doing here?

But look, at end of the day it’s a couple more dependencies in build, not in the website, so could be persuaded to just go with it as is. But so let me know thoughts on above questions I asked.

max-ostapenko:
Maybe this would be minimal effort solution:
```
fs
  .copy(
    './node_modules/web-vitals/dist/web-vitals.es5.umd.min.js',
    './static/js/web-vitals.js',
  )
  .catch(err => console.error(err));
```
triggered on `npm run generate`?

bazzadp:
I like that idea!

src/tools/generate/web-vitals.js
```diff
@@ -0,0 +1,30 @@
...
+function sendToGoogleAnalytics(metric) {
+  let event_value = 0;
+  switch (metric.name) {
+    case 'CLS':
+      // For CLS event value is first multiplied by 1000 for greater precision.
+      event_value = metric.delta * 1000;
+      break;
+    case 'TTFB':
+      // For TTFB event value excludes DNS lookup, connection negotiation, network latency, and unloading the previous document.
+      event_value = metric.value - metric.entries[0].requestStart;
```
rviscomi, line 12: For TTFB we do want network setup times (DNS, TCP, etc) so this `case` could be removed. (`value` and `delta` are equivalent for TTFB)

src/tools/generate/web-vitals.js
```diff
@@ -0,0 +1,30 @@
+import { getCLS, getFID, getLCP, getFCP, getTTFB } from 'web-vitals';
+
+function sendToGoogleAnalytics(metric) {
+  let event_value = 0;
+  switch (metric.name) {
+    case 'CLS':
+      // For CLS event value is first multiplied by 1000 for greater precision.
+      event_value = metric.delta * 1000;
```
rviscomi, line 8:
```suggestion
      event_value = Math.round(metric.delta * 1000);
```

bazzadp:
Also means it will be run by the GitHub Action which will do a fresh npm install each time and only commit it when it changes. So will always be up to date.

bazzadp:
Mac, can you also add yourself to the `src/config/2020.json` file in the developers teams?

src/templates/base.html
max-ostapenko: What I know about `defer` is that it reliably loads before `DOMContentLoaded`.
`async` may be loaded after.
I've added a listener on load and keeping `async`.

src/config/2020.json
```diff
@@ -220,6 +220,16 @@
...
+    },
+    "max-ostapenko": {
+      "name": "Max Ostapenko",
+      "teams": [
+        "analysts",
+        "developers"
+      ],
+      "avatar_url": "https:\/\/avatars1.githubusercontent.com\/u\/1611259?v=4&s=200",
+      "github": "max-ostapenko",
+      "twitter": "themax_o"
```
bazzadp, line 232: Any chance you could move this up above Ricks? I try to keep it in first name alphabetical order to make it easier to find, avoid any thoughts of precedency and satisfy my OCD (mostly the last one to be honest!) 😀

src/package.json
```diff
@@ -16,14 +16,16 @@
   "scripts": {
     "generate": "node ./tools/generate",
     "ebooks": "node ./tools/generate/generate_ebook_pdfs",
-    "deploy": "echo \"Y\" | gcloud app deploy --project webalmanac --stop-previous-version"
+    "deploy": "echo \"Y\" | gcloud app deploy --project webalmanac --stop-previous-version",
+    "browserify": "node ./tools/generate/generate_js_bundle"
```
bazzadp, line 20: Do we still need this now?

src/templates/base.html
```diff
@@ -13,13 +13,43 @@
...
       <script nonce="{{ csp_nonce() }}">
           window.dataLayer = window.dataLayer || [];
           function gtag() {
               dataLayer.push(arguments);
           }
           gtag('js', new Date());
           gtag('config', 'UA-22381566-3');
+
+          function sendWebVitalsGAEvents(metric) {
+            let event_value = 0;
```
bazzadp, line 26: Actually one more comment. Can we swap `let` for `var`? We try to avoid ES6 syntax to gracefully degrade rather than throw an error on any old or non-standard browsers. Obviously they won't support web-vitals but this could still cause an error.

```suggestion
            var event_value = 0;
```

src/templates/base.html
```diff
@@ -13,13 +13,43 @@
...
+            switch (metric.name) {
+              case 'CLS':
+                // For CLS event value is first multiplied by 1000 for greater precision.
+                event_value = Math.round(metric.delta * 1000);
+                break;
+              default:
+                event_value = metric.delta;
```
bazzadp, line 33: Actually shouldn't the non-CLS metrics be rounded too?
```suggestion
                eventValue = Math.round(metric.name === 'CLS' ? metric.delta * 1000 : metric.delta);
```
Should we just use [syntax like they suggest](https://github.com/GoogleChrome/web-vitals#using-gtagjs) which would replace a lot of this code and potentially make it easier to include any future changes by comparing to their suggested code?

src/templates/base.html
max-ostapenko: Sure, forgot to change it after moving away from nodejs code.

max-ostapenko:
I get Google advertising errors on public website:
`Refused to load the image 'https://www.google.pt/ads/ga-audiences?v=1&aip=1&t=sr&_r=4&tid=UA-22381566-3&cid=722833102.1592150741&jid=307643168&_v=j83&z=1389390864&slf_rd=1&random=3377659529' because it violates the following Content Security Policy directive: "img-src 'self' www.google-analytics.com www.googletagmanager.com stats.g.doubleclick.net *.githubusercontent.com *.google.com ssl.gstatic.com data:".
`

Should we fix it or switch it off?

src/templates/base.html
max-ostapenko: Yeah, you're right. Event value takes integer: https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#ev

bazzadp:
Yeah I get the same as in Ireland so use google.ie. Unfortunately there is [no easy way to whitelist every Google domain in a CSP](https://stackoverflow.com/questions/47197609/google-adwords-csp-script-src).

Not sure how to turn this off to be honest. Is it a setting in GA or GTM?

max-ostapenko:
`gtag('set', 'allow_google_signals', false );`
https://developers.google.com/analytics/devguides/collection/gtagjs/display-features

bazzadp:
> `gtag('set', 'allow_google_signals', false );`
> https://developers.google.com/analytics/devguides/collection/gtagjs/display-features

I'd be a fan of that, as better for privacy.

Guessing that's the same as removing from Google Analytics admin? Unfortunately I only have read-only access so we'll need @rviscomi to confirm if it's on there and whether to disable in there instead. If it's disabled in there already, then I say we add above. If it's not enabled and we're still getting this error, then can try above. Did you test it and it removes the error?

src/tools/generate/generate_js.js
```diff
@@ -0,0 +1,12 @@
+const fs = require('fs-extra');
+
+const generate_js = async () => {
+  fs.copy(
+    './node_modules/web-vitals/dist/web-vitals.es5.umd.min.js',
```
bazzadp, line 5: Actually I've had another look and the [web-vitals repo](https://github.com/GoogleChrome/web-vitals) does not include the `dist` folder by default so this only exists for you as you have manually built it, but won't work for everyone else wihtout adding the package.json dependencies and an extra step.

So our choices are:

1. Build manually periodically
2. Pull in from CDN like unpkg during `npm run generate`. This seems to get the latest version after a redirect: https://unpkg.com/web-vitals/dist/web-vitals.es5.min.js?module
3. Add to package.json and build as part of `npm run generate` before doing this copy.

To be honest I'm not a fan of 3 as we run `npm run generate` on every merge as part of a GitHub Action so adding dependencies, and builds to that just lengthens that. It's not liable to change THAT often to want to take that burden on in every commit.

@max-ostapenko / @rviscomi what are your thoughts on option 2?

max-ostapenko: If I understood the process correctly, it's being built upon `npm install` - I see it locally in /dist folder.
So the copy operation itself takes really no additional effort.

I prefer having it within npm dependencies to pulling from unpkg.com.

max-ostapenko: And a small note, I'm using UMD module for wider compatibility: `web-vitals.es5.umd.min.js`.
Comment if there is something in favour of ESM.

bazzadp: > If I understood the process correctly, it's being built upon npm install - I see it locally in /dist folder.

Isn’t that because you ran `npm install` **and** then `browserfy` to do the build? If you delete this folder and then do `npm install` again without `browserfy` does it create the dust folder?

My concern with adding browserfy and the build step, is the length of time it adds to ever commit. Not much I’m sure but it already takes a while to run.

max-ostapenko: After I removed browserify, I cleaned node_modules and reinstalled only required packages from scratch.
So web-vitals installed and fs.copy, nothing else in this branch.

src/templates/base.html
```diff
@@ -13,13 +13,32 @@
...
+              value: Math.round(metric.name === 'CLS' ? metric.delta * 1000 : metric.delta),
+              event_label: metric.id,
+              non_interaction: true,
+            });
+          }
+
+          window.addEventListener('load', function() {
+            webVitals.getFCP(sendWebVitalsGAEvents);
+            webVitals.getLCP(sendWebVitalsGAEvents, true);
+            webVitals.getCLS(sendWebVitalsGAEvents, true);
```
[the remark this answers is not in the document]
max-ostapenko: I thought that it would benefit in having overview over frequency:
_Total events per Event Label and Page_
and total value:
_Event Value per Event Label and Page_

Of course it comes with a cost of sending a sequence of same metric event requests (I've seen up to 3 events of same type locally).

If we set reportAllChanges=false, then the event is sent before navigating away, once per page:
_Event Value per Event Label and Page_
OR cumullative over all pageviews:
_Event Value per Page_

Do you think frequency will make sense?

P.S. Hope you'll share collected insights.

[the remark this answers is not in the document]
max-ostapenko: I also saw some LCP events that happen after rendering is finished (fires together with FID) may have 0 delta.
Don't know if it gives you some performance issue hint, but I'd like your feedback if it makes sense to collect such events.

src/tools/generate/generate_js.js
bazzadp: Yeah confirmed the same. Weird. Anyway that's a good thing for us! Resolving this.

src/templates/base.html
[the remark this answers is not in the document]
rviscomi: I'd prefer to have one metric reported per page load to keep the analytics simple. Reporting non-final values isn't recommended as @bazzadp pointed out and I think this means we will need to report the `value` rather than the `delta`.

cc @philipwalton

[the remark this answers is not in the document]
max-ostapenko: Got you.
But I'd still prefer to keep `metric.delta` as a event value. It's equal to `metric.value` in the first event, but in case there are several events sent - aggregated `metric.delta` will report correct value, but aggregated `metric.value` will inflate total event value in report.
And there will be no need to change it, if you decide to collect all events in future.

bazzadp:
Merged! Thanks @max-ostapenko !

src/templates/base.html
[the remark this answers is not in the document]
philipwalton: Final changes LGTM!

> I'd prefer to have one metric reported per page load to keep the analytics simple. Reporting non-final values isn't recommended as @bazzadp pointed out

@rviscomi FYI, CLS is the exception to this, see this explanation in the README:

> _**Important:** unlike other metrics, CLS continues to monitor changes for the entire lifespan of the page&mdash;including if the user returns to the page after it's been hidden/backgrounded. However, since browsers often [will not fire additional callbacks once the user has backgrounded a page](https://developers.google.com/web/updates/2018/07/page-lifecycle-api#advice-hidden), `onReport` is always called when the page's visibility state changes to hidden. As a result, the `onReport` function might be called multiple times during the same page load (see [Reporting only the delta of changes](#report-only-the-delta-of-changes) for how to manage this)._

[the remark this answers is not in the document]
rviscomi: Gotcha, so we'd have to aggregate event values by SUM when looking at session-level stats.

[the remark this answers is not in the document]
Tiggerito: Another thing to consider is that GA has a 500 hit per session limit. Firing a lot of events on a single page can quickly get you to that limit.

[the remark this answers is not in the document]
philipwalton: True, if you're using the `reportAllChanges` feature and sending a hit to GA with every callback. If you're not using `reportAllChanges` then the max number of hits you will send is three plus the number of times the tab is backgrounded in the same session.

I just did a quick check on my site, and over the last 28 days, the 99th percentile of "Web Vitals" events in a single session is 5 (the 99.9th percentile is 11), so probably nothing to worry about.

PR #954, opened 2020-07-04
Hero image with year as text

### LINKED ISSUE #894: Reimplement home page Hero Image with year as text
As mentioned in https://github.com/HTTPArchive/almanac.httparchive.org/pull/893#issuecomment-647521501

Currently the hero image on the home page comes in two flavours:
- On Desktop it's [an image with the year embedded in it](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/src/static/images/home-hero-2019.png).
- On Mobile it's an [image without the year](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/src/static/images/home-hero.png)

Can we make the desktop the same as mobile to allow ease of editing for 2020?

catalinred:
Fixes  #894

The year uses now an SVG `text` in order to simulate a "100%" font-size when on large displays and position above the hero image.

The only drawback of this approach is that we need to use magic numbers in order to set the SVG's viewBox.
e.g.:
`viewBox="0 0 47 16"` ~ for 2020 in `src/templates/base/2020/index.html `
`viewBox="0 0 42 16"` ~ for 2019  in the current `src/templates/base/2019/index.html`

I also improved a bit the placement of the year + hero image. 
In my opinion, it looks way better now considering the slanted header design. 

Current
almanac httparchive org_en_2019_

New
new_en_2019_

Let me know your thoughts!


catalinred:
Lint Code Base -> src/static/css/index.css seems to throw some errors, even if I already [committed](https://github.com/HTTPArchive/almanac.httparchive.org/pull/954/commits/2390eeb45c47ea9eb14ed1ccd7d8aeba95062369) a fix. 
I guess I'm missing something here?

bazzadp:
Weird. Even tried a rerun and it didn’t fix it. Wanna try another small change to see if that kicks it into life?

Btw away from computer for next day so will check this out tomorrow night if no one looks in the meantime.

catalinred:
> Wanna try another small change to see if that kicks it into life?

It worked. 🚀

> will check this out tomorrow night if no one looks in the meantime.
Sure, no problem.

bazzadp:
> > Wanna try another small change to see if that kicks it into life?
> 
> It worked. 🚀

Hmmm... will need to keep an eye on this. Hopefully just a one-off. and we know the work around now!

src/static/css/index.css
```diff
@@ -554,27 +520,6 @@ p {
...
-    font-size: 40vw;
-    line-height: 100%;
-    font-weight: bold;
-  }
-
-  .intro-image-wrapper {
-    display: block;
-    max-width: 100%;
-    margin-top: 60px;
-    margin-right: -46px;
```
bazzadp, line 571: By removing this -46px we get a change on mobile as this was there to offset the fact that the main body of the image is not in the centre of the image so looks weird against the above button:

Old:

Image is aligned under Start Exploring button

New:

Image is not aligned under Start Exploring button

It's a bit of a hack and a magic number to be honest and maybe the better solution it to change the image to make it centrally aligned?

src/templates/base/2019/index.html
```diff
@@ -36,11 +36,11 @@ <h2>{{ self.intro_sub_title() }}</h2>
...
     </div>
-    <div class="intro-image-year-wrapper">
-      <div class="intro-image-year"></div>
-    </div>
     <div class="intro-image-wrapper">
-      <div class="intro-image"></div>
+      <svg viewBox="0 0 42 16">
+        <text x="0" y="15">{{ year }}</text>
+      </svg>      
+      <img src="/static/images/home-hero.png" alt="">
```
bazzadp, line 43: Do we really need to use SVG for simple text?

I notice this breaks on IE11 as it doesn't seem to calculate the height of the SVG correctly. Not that we have much traffic from this but still that, along with the magic numbers required, are a downside to this approach.

If you changed to this:

```suggestion
      <div class="year">2019</div>     
      <img src="/static/images/home-hero.png" alt="">
```

Along with this CSS:

```css
.intro-image-wrapper .year {
  color: #f7f779;
  font-family: 'Poppins', sans-serif;
  font-weight: 700;
  font-size: 12rem;
  line-height: 0.58;
  margin-top: 0.2em;
}
```

We seem to get the same effect (though need slightly different numbers for mobile).

WDYT?

src/static/css/index.css
```diff
@@ -105,42 +105,29 @@ p {
...
-
-.intro-image-year {
-  background-image: url(/static/images/home-hero.png);
-  background-size: cover;
-  max-height: 470px;
-  max-width: 458px;
-  height: 41vw;
-  width: 40vw;
+  max-height: 485px;
+  max-width: 485px;
```
bazzadp, line 119: These magic numbers move it a bit too much on tablet mode (select iPad in Chrome Dev tools and compare it to production).

src/static/css/index.css
catalinred: I pushed another commit to revert this pseudo center alignment for this image.

src/templates/base/2019/index.html
catalinred: As you already know "mobile" breakpoint is hard to define and our current max-width:900px breakpoint might not be enough.

I am aware of the method you're showcasing above, but I avoided as you need to define different font sizes across different media queries breakpoints.

Made some research and I preferred using SVG because it preserves aspect ratio: https://css-tricks.com/fitting-text-to-a-container/#just-use-svg. I also add a CSS fix for the IE11 only. It degrades nicely now in my opinion.

src/static/css/index.css
catalinred: The previous magic numbers were:
max-height: 470px;
max-width: 458px;
height: 41vw;
width: 40vw;

And I reduced these to the above
 max-height: 485px;
 max-width: 485px;

I found them easier to understand and to update in the future, plus the hero image looks slightly better now as it overlaps a bit with the slanted background design.

bazzadp: Agree it’s great to reduce magic numbers. As I said I’m happy you’ve cleaned up a lot of CSS in this PR. And totally agree with your decision to move the logo down - you’re right it does look better.

However, did you look at iPad (landscape mode btw) in Chrome Dev tools? Seems a bit too far to the left compared to production. Noticed it first when testing real iPad but luckily also shows in Chrome in emulation mode. Though maybe I wouldn’t care if I wasn’t directly comparing it prod so maybe we should just accept this? Let me know your thoughts anyway.

src/static/css/index.css
bazzadp: Don’t think we should just widen the actually image file and remove this hack? This [caught me out elsewhere too btw](https://github.com/HTTPArchive/almanac.httparchive.org/pull/833#discussion_r428075979) and really think we should just fix the image if we want it “cantered” all the time.

src/static/css/index.css
catalinred: I understand what you mean. 
In this case, I think the missing bit is a margin between the main `grid` areas: intro and intro-image-wrapper.

`grid-gap: 2rem;` on `.intro-container` would solve this.

src/templates/base/2019/index.html
bazzadp: Ah OK. So your method allows the image to change size responsively and for the text to automatically resize as necessary (like happens between 900px and 1050px where image shrinks slightly slightly). OK happier now there is a reason. Resolving this.

And thanks for the IE11 hack. Not loving it as so IE specific, but will still take it. Hopefully soon enough I'll learn to get over supporting IE11 but since you've found the solution let's take it for now.

src/static/css/index.css
catalinred: Will have a look at it.

catalinred: I managed to push something on this matter. Let me know what you think @bazzadp

rviscomi:
I noticed that the image has moved down and to the right compared to the production version. Also the "2019" text slightly overlaps with the image.

### Before

almanac httparchive org_en_2019_

### After

127 0 0 1_8080_en_2019_

src/static/css/index.css
bazzadp: Thanks for that. Much better I think! Resolving.

src/static/css/index.css
```diff
@@ -106,41 +107,28 @@ p {
...
-.intro-image-year {
-  background-image: url(/static/images/home-hero.png);
-  background-size: cover;
-  max-height: 470px;
-  max-width: 458px;
-  height: 41vw;
-  width: 40vw;
+.intro-image-wrapper svg {
+  position: relative;
+  top: 28px;
```
bazzadp, line 118: Changing this to `25px` seems to solve the issue @rviscomi raised.

```suggestion
  top: 25px;
```

Also adding this:
```css
.intro-image-wrapper img {
  position: relative;
}
```
puts the image in front so even the slight overlap is less apparent. Not sure if there's a better way?

catalinred: Agree that `25px` is the right positioning value now that I've updated the image canvas.

catalinred:
> I've one minor suggestion, and also a request - can you run the new home-hero.png through tinypng.com? It compresses the image from 132KB back down to 40KB. We have the calibre image action but it doesn't work on forks so won't work for this PR.

Done, I guess knowing about the GH image action made me forget about tinypng or jpegmini.
Thanks for the heads up.

bazzadp:
> > I've one minor suggestion, and also a request - can you run the new home-hero.png through tinypng.com? It compresses the image from 132KB back down to 40KB. We have the calibre image action but it doesn't work on forks so won't work for this PR.
> 
> Done, I guess knowing about the GH image action made me forget about tinypng or jpegmini.
> Thanks for the heads up.

Yeah I've an [open PR to handle this better](https://github.com/calibreapp/image-actions/pull/54). Must chase!

src/static/css/index.css
```diff
@@ -682,3 +628,22 @@ p {
     margin: 0;
   }
 }
+
+@media (max-width: 900px) {
+  .intro-year {
+    display: block;
+    color: #f7f779;
+    font-size: min(190px, 40vw);
```
bazzadp, line 636: Was wondering why you hadn't answered my nit as well - but just realised I hadn't actually sent it!

Should we add a fallback font size here since `min()` [isn't quite universal yet](https://caniuse.com/#feat=css-math-functions)?

```suggestion
    font-size: 190px;
    font-size: min(190px, 40vw);
```

catalinred: Sure, good find!

PR #956, opened 2020-07-04
Add Docker-based development environment

### LINKED ISSUE #937: Create Dockerfile to run dev env
As discussed in https://github.com/HTTPArchive/almanac.httparchive.org/issues/933#issuecomment-651280322 let's set up a Dockerfile to allow easier launching of dev env

ibnesayeed:
Closes #937 with the following changes:

* Add new files: `Dockerfile` and `.gcloudignore`
* Update application server to listen on `0.0.0.0` interface as loopback interface of the container will not be accessible from outside
* Document the development procedure in Docker
* Fix a RegEx format deprecation warning in newer versions of Python

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -1,4 +1,4 @@
-ARG     PYTAG=latest
+ARG     PYTAG=3
 FROM    python:${PYTAG}
 
 # Enable unbuffered STDOUT logging
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -1,0 +1,23 @@
+ARG     PYTAG=latest
+FROM    python:${PYTAG}
+
+# Enable unbuffered STDOUT logging
+ENV     PYTHONUNBUFFERED=1
+
+ARG     NODEVER=12.x
+RUN     curl -sL https://deb.nodesource.com/setup_${NODEVER} | bash - \
+        && echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" > /etc/apt/sources.list.d/google-cloud-sdk.list \
+        && curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add -
+
+RUN     apt update && apt install -y \
+          google-cloud-sdk \
+          nodejs \
+        && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+COPY    requirements.txt package*.json ./
+RUN     pip install -r requirements.txt
+RUN     npm install
+COPY    . ./
+
+CMD     ["python", "main.py"]
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -5,14 +5,17 @@ FROM    python:${PYVER}
 ENV     PYTHONUNBUFFERED=1
 
 ARG     NODEVER=12.x
-RUN     curl -sL https://deb.nodesource.com/setup_${NODEVER} | bash - \
-        && echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" > /etc/apt/sources.list.d/google-cloud-sdk.list \
-        && curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add -
+RUN     curl -sL https://deb.nodesource.com/setup_${NODEVER} | bash -
 
 RUN     apt update && apt install -y \
-          google-cloud-sdk \
           nodejs \
         && rm -rf /var/lib/apt/lists/*
+
+# Add Google Cloud SDK only if image is built with `--build-arg SKIPGC=false`
+ARG     SKIPGC=true
+RUN     $SKIPGC || (echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" > /etc/apt/sources.list.d/google-cloud-sdk.list \
+        && curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add - \
+        && apt update && apt install -y google-cloud-sdk && rm -rf /var/lib/apt/lists/*)
 
 WORKDIR /app
 COPY    requirements.txt package*.json ./
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -1,5 +1,5 @@
-ARG     PYTAG=3
-FROM    python:${PYTAG}
+ARG     PYVER=3
+FROM    python:${PYVER}
 
 # Enable unbuffered STDOUT logging
 ENV     PYTHONUNBUFFERED=1
--- a/src/README.md
+++ b/src/README.md
@@ -153,5 +153,7 @@ docker container run --rm -it -v "$PWD":/app webalmanac pytest
 docker container run --rm -it -v "$PWD":/app -v /app/node_modules -p 8080:8080 webalmanac bash
 root@[CID]:/app# pytest
 root@[CID]:/app# python main.py
+^C
 root@[CID]:/app# npm run generate
+root@[CID]:/app# exit
 ```
--- a/src/README.md
+++ b/src/README.md
@@ -139,7 +139,7 @@ docker image build -t webalmanac .
 docker container run --rm -it -v "$PWD":/app -p 8080:8080 webalmanac
 ```
 
-3. Open http://localhost:8080 in your web browser to access the site then kill the server (using `Ctrl+C`)
+3. Open http://localhost:8080 in your web browser to access the site. You can kill the server when it is no longer needed using `Ctrl+C`.
 
 4. Make changes in the code using any text editor and run tests (need to build the image again if any Python or Node dependencies are changed):
 
--- a/src/README.md
+++ b/src/README.md
@@ -1,6 +1,6 @@
 # Developing the Web Almanac
 
-The Web Almanac can be developed on MacOs, Windows or Linux. It required node v12 and python v3 to be installed.
+The Web Almanac can be developed on macOS, Windows or Linux. It requires Node v12 and Python v3 to be installed. Alternatively, use Docker to avoid manually configuring the development environment.
 
 ## Run Locally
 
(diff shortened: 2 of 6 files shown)


src/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# Enable unbuffered STDOUT logging
+ENV     PYTHONUNBUFFERED=1
+
+ARG     NODEVER=12.x
+RUN     curl -sL https://deb.nodesource.com/setup_${NODEVER} | bash - \
+        && echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" > /etc/apt/sources.list.d/google-cloud-sdk.list \
+        && curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add -
+
+RUN     apt update && apt install -y \
+          google-cloud-sdk \
```
bazzadp, line 13: Do we need all these Google Cloud references? As I understand it they are only needed for release so only core maintainers need them so should we stick them under a comment or the like so every developer is not lumped with them?

src/README.md
```diff
@@ -122,3 +122,38 @@ npm run deploy
...
+docker image build -t webalmanac .
+```
+
+2. Run the application server (which is the default command of the Docker image, so no need to explicitly supply it as an argument):
+
+```
+docker container run --rm -it -v "$PWD":/app -p 8080:8080 webalmanac
+```
+
+3. Open http://localhost:8080 in your web browser to access the site then kill the server (using `Ctrl+C`)
```
bazzadp, line 142:
```suggestion
3. Open http://localhost:8080 in your web browser to access the site. You can kill the server when it is no longer needed using `Ctrl+C`.
```

src/Dockerfile
```diff
@@ -0,0 +1,23 @@
+ARG     PYVER=3
+FROM    python:${PYVER}
+
+# Enable unbuffered STDOUT logging
+ENV     PYTHONUNBUFFERED=1
```
bazzadp, line 5: Why do we need this out of interest? And is it something we should be doing even outside of Docker?

bazzadp:
Also can you add yourself to the 2020.json config in the developer team?

src/Dockerfile
ibnesayeed: > This looks great! Thanks very much for working on this. Very nicely explained (even for a complete Docker-novice like myself!).

Thanks! I did skip many details for the sake of brevity. For example, I did not go into the details of mounted volumes and mapped ports because those familiar with Docker would know these things.

> I've a few comments and also a question: is it possible to have a pre-built Docker image and save it in the repo? I'm thinking for the GitHub Actions there might be a good time saving for it having a pre-built image. Obviously it would need to be regenerated occasionally to dependencies up to date but still could be useful.

I thought about it, but realized that this is not a general purpose application that people would want to run independently as a service, so publishing a Docker image has little utility. That's why I did not optimize the image for running it as a service, instead, I kept it developer-friendly with all the necessary tools and packages included at the cost of longer build time and heavier image size on disk. Also, I did not add any metadata labels in it, that I would have, if I were to prepare it for distribution. That said, if a compelling utility emerges beyond developers pulling the image instead of building it locally, I am open to add a GH Workflow file to automatically build the image every time a commit or tag is pushed to the main branch and push the image to DockerHub.

ibnesayeed:
> This looks great! Thanks very much for working on this. Very nicely explained (even for a complete Docker-novice like myself!).

Thanks! I did skip many details for the sake of brevity. For example, I did not go into the details of mounted volumes and mapped ports because those familiar with Docker would know these things.

> I've a few comments and also a question: is it possible to have a pre-built Docker image and save it in the repo? I'm thinking for the GitHub Actions there might be a good time saving for it having a pre-built image. Obviously it would need to be regenerated occasionally to dependencies up to date but still could be useful.

I thought about it, but realized that this is not a general purpose application that people would want to run independently as a service, so publishing a Docker image has little utility. That's why I did not optimize the image for running it as a service, instead, I kept it developer-friendly with all the necessary tools and packages included at the cost of longer build time and heavier image size on disk. Also, I did not add any metadata labels in it, that I would have, if I were to prepare it for distribution. That said, if a compelling utility emerges beyond developers pulling the image instead of building it locally, I am open to add a GH Workflow file to automatically build the image every time a commit or tag is pushed to the main branch and push the image to DockerHub.

bazzadp:
Yeah our own Github Action's using the pre-built docker image to make them faster, was the main thing I was thinking off. But by reducing the requirements down a bit with the clean up |I'm doing that's maybe not such a big issue anyway.

src/Dockerfile
ibnesayeed: I added it so that those who need it and have the ability to use it can use it without manually installing the tool. Others, who do not need it can leave it alone in the image, it will not bother them. The point of this developer image is to make it as self-contained and friction-less as possible. I could have provided means to selectively install it using build args or multi-stage builds, but then the Dockerfile will have more build branches and not very easy to understand or maintain going forward. Additionally, we will have to document those variations in the build process, which will add more text about Docker than the main agenda of the repo.

That said, I have intentionally placed this instruction early in the Dockerfile to leverage layer caching, which means, successive builds of the image after code changes will not invalidate this layer and it will only be downloaded and installed once, unless something changes upstream.

bazzadp: Hmmmm I still think we should rip it out. I don't think we'll be able to deploy via this Docker Image which means deployments will always need to run outside of Docker so it's a waste of time and space installing this inside of Docker.

src/Dockerfile
ibnesayeed: I/O buffering is done at multiple levels for better performance such as operating systems and run time environments. Usually, the behavior of programs depends on where the STDOUT is pointed to. If we write to a file, data is usually written in blocks of fixed size and the OS/runtime waits for at least one block of data to be ready before it writes out, unless it is flushed explicitly. When the output is connected to a TTY, it yields data on each newline. This makes debugging easier, but when running in production, it is advised to log in files instead of a terminal for better performance.

When running processes in Docker, OS-level buffering is handled properly, depending on whether or not the output is connected to a TTY, but due to an increased level of indirection, Python's runtime does not get to know it, so it continues to buffer output internally. As result, when trying to print a debug message in a process running in a container attached to a virtual TTY, we might not see anything being logged on terminal for a while until there is enough data to yield. That's why, I have added this environment variable to tell Python not to enable I/O buffering. I will not do the same in a production environment. #LessonsLearnedTheHardWayOverTime ;-)

ibnesayeed:
> Also can you add yourself to the 2020.json config in the developer team?

Sure, I will do that in a separate PR.

src/Dockerfile
bazzadp: OK so a Docker-specific thing. Cool. Resolving.

ibnesayeed:
> Yeah our own Github Action's using the pre-built docker image to make them faster, was the main thing I was thinking off. But by reducing the requirements down a bit with the clean up |I'm doing that's maybe not such a big issue anyway.

If you see a utility there, we can certainly make the image available publicly, but I will note here that we will be using a docker image built on **at least one commit older** code base. This should not be an issue most of the time, but there might be places where we would prefer the most recent state of the code be used for the image in use.

bazzadp:
Let's leave it for now and open a separate PR for that in future if we see the use for it.

src/Dockerfile
ibnesayeed: I have made some changes here, it will now only be installed when asked explicitly at image build time.

bazzadp: Thanks! Resolving.

bazzadp:
@rviscomi you want to review this too or happy enough for me to merge now?

src/Dockerfile
ibnesayeed: To test this change, run the following:

```
$ docker image build -t webalmanac:nogc .
$ docker container run --rm -it webalmanac:nogc gcloud --version
docker: Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"gcloud\": executable file not found in $PATH": unknown.
$ docker image build -t webalmanac:gc --build-arg SKIPGC=false .
$ docker container run --rm -it webalmanac:gc gcloud --version
Google Cloud SDK 299.0.0
alpha 2020.06.26
beta 2020.06.26
bq 2.0.58
core 2020.06.26
gsutil 4.51
kubectl 2020.06.26
```

ibnesayeed:
> Let's leave it for now and open a separate PR for that in future if we see the use for it.

I am in the process of adding a workflow to build and push the image, but for that I will need access to settings of this repo to add secrets that will enable pushing images to DockerHub.

src/Dockerfile
bazzadp: I'll be perfectly honest with you and did not run this in Docker as don't feel I know enough about it and I intend to just develop locally for reasons given earlier. But did a manual review of the changes and am happy enough to merge if this makes life easier for others (yourself included).

bazzadp:
Is DockerHub the only way to host a Docker Image? It isn't something we can store locally and just use after a checkout?

To be honest I'd hold off creating another workflow for now, and let's see if we have a need for the image. I think the instructions you created as pretty easy to follow. The main reason for an image is to make the GitHub Actions themselves faster but, as I say, will see if they are fast enough after doing some of the spring cleaning so that isn't even needed for now.

ibnesayeed:
> Is DockerHub the only way to host a Docker Image?

No, there are many Docker repositories [GitHub Packages](https://github.com/features/packages) in one example, [Google Cloud Container Registry](https://cloud.google.com/container-registry) is another, and there are many others. DockerHub is the most popular one which is pre-configured in Docker CLI by default, so referencing images there does not require explicit host name.

> It isn't something we can store locally and just use after a checkout?

We always have a copy in the local cache, each time we build or pull an image and that's the one that is used when we run a container. If a new version is available in a remote registry, local container run command will not know about it and will use the local stale version. Not just the tagged image, but every layer is locally cached and is not garbage collected automatically, so those layers are reused if the content hash matches.

> To be honest I'd hold off creating another workflow for now, and let's see if we have a need for the image. I think the instructions you created as pretty easy to follow. The main reason for an image is to make the GitHub Actions themselves faster but, as I say, will see if they are fast enough after doing some of the spring cleaning so that isn't even needed for now.

Okay, I will push my changes in a separate branch of my fork for future reference.

ibnesayeed:
I have [placed a tentative workflow file](https://github.com/ibnesayeed/almanac.httparchive.org/blob/docker-img-workflow/.github/workflows/docker-image.yml) to build and push the image to DockerHub. Also, I have created a [placeholder image repo](https://hub.docker.com/r/httparchive/webalmanac) (and organization) in DockerHub.

ibnesayeed:
This PR is good to go from my end, let me know if you find any issues.

src/Dockerfile
ibnesayeed: I knew you did not test it, that's why I provided the outcome of the variations controlled by a build arg. By the way, there is an undocumented feature that would allow us to build images with different version combinations of Python and Node for testing without any changes in the Dockerfile, simply by passing appropriate build args.

PR #958, opened 2020-07-04
Update Flask version to fix a deprecation warning

### LINKED ISSUE #957: Update Flask dependency to remove a deprecation warning
As noted in https://github.com/HTTPArchive/almanac.httparchive.org/issues/937#issuecomment-653806450 the version of Flask we use (`1.0.2`) is causing a deprecation warning.

```
====================================================== test session starts =======================================================
platform linux -- Python 3.8.2, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
rootdir: /app
collected 12 items                                                                                                               

validate_test.py ............                                                                                              [100%]

======================================================== warnings summary ========================================================
/usr/local/lib/python3.8/site-packages/flask/sessions.py:14
  /usr/local/lib/python3.8/site-packages/flask/sessions.py:14: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    from collections import MutableMapping

-- Docs: https://docs.pytest.org/en/latest/warnings.html
================================================= 12 passed, 1 warning in 1.49s ==================================================
```

ibnesayeed:
Closes #957 by updating the Flask version. While the latest version of Flask is `1.1.2` and all the tests are passing with that, I only updated it to the latest patch version of of the `1.0.x` series to avoid any untested failures.

### DIFF
--- a/src/requirements.txt
+++ b/src/requirements.txt
@@ -1,4 +1,4 @@
-Flask==1.0.4
+Flask
 flask-talisman
 gunicorn==19.9.0
 matplotlib
--- a/src/requirements.txt
+++ b/src/requirements.txt
@@ -1,6 +1,6 @@
 Flask
 flask-talisman
-gunicorn==19.9.0
+gunicorn
 matplotlib
 mistune
 pandas
--- a/src/requirements.txt
+++ b/src/requirements.txt
@@ -1,4 +1,4 @@
-Flask==1.0.2
+Flask==1.0.4
 flask-talisman
 gunicorn==19.9.0
 matplotlib


bazzadp:
I’m a big fan of keeping things up to date so I say we go to latest version. More so I say remove the version number so we stay up to date. I’ll do some testing on this later but you can go ahead and update the PR in the meantime.

ibnesayeed:
> More so I say remove the version number so we stay up to date.

There were only two dependencies that were versions pinned, so I removed version numbers of both of them.

PR #959, opened 2020-07-05
Clean-up old code, and update version numbers

### LINKED ISSUE #953: Clean up unused code and dependencies
I don't think we use these files (added in #114 ) any more since we changed the way we generate chapters in  #222 
- generate_chapters.py
- generate_visualisations.py
- src/data

Also means we can get rid of the following from requirements file I think:
- matplotlib
- mistune
- pandas
- scour
- weasyprint (not part of above but removed by #833 )

We may want to add some of these back in depending what we do with #896 but easy enough to re-add if we want to.

Possibly could be done at same time as #949 ?

bazzadp:
Fixes #953 by removing unused dependencies from requirements.txt and also unused python files.
I also removed the `ar` language templates since we do not seem to be progressing that language at this time.

I also updated all our npm modules to latest versions. I like to keep up to date 😀 This does create some changes to the templates as due to prettier changes but they look OK.

I've left python on 3.7 instead of updating to 3.8 as, even though that's what I use locally, [GCP still lists 3.8 as beta](https://cloud.google.com/appengine/docs/standard/python3/runtime) and no pressing need to upgrade.


ibnesayeed:
> This does create some changes to the templates as v2 of prettier adds trailing semi-colons to inline style attributes but I'm OK with that.

Can't we move those inline styles to a stylesheet file and replace them with classes instead?

bazzadp:
> Can't we move those inline styles to a stylesheet file and replace them with classes instead?

In general we allow Authors to use inline-styles where Markdown is not sufficient and they want to write raw HTML. However looking at this we see there are only 3 different in-line styles so yeah let's move to CSS classes since w're changing them anyway.

Does make this PR even bigger now though since now changing `.md` files and `.html` files, but I've never been afraid of big PRs.

src/content/fr/2019/caching.md
```diff
@@ -412,15 +412,15 @@ Si l'on compare les ressources du domaine principal et celles des tierces partie
     </tr>
     <tr>
      <td>Bureau</td>
-     <td><p style="text-align: right">70.7&nbsp;%</p></td>
-     <td><p style="text-align: right">47.9&nbsp;%</p></td>
-     <td><p style="text-align: right">59.2&nbsp;%</p></td>
+     <td><p class="numeric">70.7&nbsp;%</p></td>
+     <td><p class="numeric">47.9&nbsp;%</p></td>
+     <td><p class="numeric">59.2&nbsp;%</p></td>
```
ibnesayeed, line 417:
```suggestion
     <td><p class="numeric">70.7%</p></td>
     <td><p class="numeric">47.9%</p></td>
     <td><p class="numeric">59.2%</p></td>
```

src/content/fr/2019/caching.md
```diff
@@ -412,15 +412,15 @@ Si l'on compare les ressources du domaine principal et celles des tierces partie
...
     </tr>
     <tr>
      <td>Mobile</td>
-     <td><p style="text-align: right">71.4&nbsp;%</p></td>
-     <td><p style="text-align: right">46.8&nbsp;%</p></td>
-     <td><p style="text-align: right">59.6&nbsp;%</p></td>
+     <td><p class="numeric">71.4&nbsp;%</p></td>
+     <td><p class="numeric">46.8&nbsp;%</p></td>
+     <td><p class="numeric">59.6&nbsp;%</p></td>
```
ibnesayeed, line 423:
```suggestion
     <td><p class="numeric">71.4%</p></td>
     <td><p class="numeric">46.8%</p></td>
     <td><p class="numeric">59.6%</p></td>
```

src/templates/fr/2019/chapters/caching.html
```diff
@@ -433,15 +433,15 @@ <h2 id="comment-les-ttl-de-cache-se-comparent-ils-à-lâge-des-ressourcesnbsp"><
           </tr>
           <tr>
             <td>Bureau</td>
-            <td><p style="text-align: right;">70.7&nbsp;%</p></td>
-            <td><p style="text-align: right;">47.9&nbsp;%</p></td>
-            <td><p style="text-align: right;">59.2&nbsp;%</p></td>
+            <td><p class="numeric">70.7&nbsp;%</p></td>
+            <td><p class="numeric">47.9&nbsp;%</p></td>
+            <td><p class="numeric">59.2&nbsp;%</p></td>
```
ibnesayeed, line 438:
```suggestion
            <td><p class="numeric">70.7%</p></td>
            <td><p class="numeric">47.9%</p></td>
            <td><p class="numeric">59.2%</p></td>
```

src/templates/fr/2019/chapters/caching.html
```diff
@@ -433,15 +433,15 @@ <h2 id="comment-les-ttl-de-cache-se-comparent-ils-à-lâge-des-ressourcesnbsp"><
...
           </tr>
           <tr>
             <td>Mobile</td>
-            <td><p style="text-align: right;">71.4&nbsp;%</p></td>
-            <td><p style="text-align: right;">46.8&nbsp;%</p></td>
-            <td><p style="text-align: right;">59.6&nbsp;%</p></td>
+            <td><p class="numeric">71.4&nbsp;%</p></td>
+            <td><p class="numeric">46.8&nbsp;%</p></td>
+            <td><p class="numeric">59.6&nbsp;%</p></td>
```
ibnesayeed, line 444:
```suggestion
            <td><p class="numeric">71.4%</p></td>
            <td><p class="numeric">46.8%</p></td>
            <td><p class="numeric">59.6%</p></td>
```

src/static/css/page.css
```diff
@@ -423,10 +427,20 @@ tbody tr:nth-child(even) {
...
   text-align: center;
+  font: bold 128px/1em Poppins, sans-serif;
   font: bold 8rem/1em Poppins, sans-serif;
   color: #62718b;
 }
 
+figure .really-big-number {
+  text-align: center;
+  font: bold 24px/1em Poppins, sans-serif;
+  font: bold 1.5rem/1em Poppins, sans-serif;
```
ibnesayeed, line 438: Are we duplicating `font` rule it to ensure proper rendition in browsers that do not support `rem` unit? If so, it looks like the support for `rem` is [pretty good](https://caniuse.com/#feat=rem) now.

src/content/fr/2019/caching.md
bazzadp: This is intentional as noted in the [translators guide](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Translators'-Guide). In french they write "70 %" (with a space) while in English we use "70%" (without a space). We use a non-breaking space to prevent orphaning a number from the percent sign when they fall on a line break.

src/content/fr/2019/caching.md
bazzadp: This is intentional as noted in the [translators guide](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Translators'-Guide). In french they write "70 %" (with a space) while in English we use "70%" (without a space). We use a non-breaking space to prevent orphaning a number from the percent sign when they fall on a line break.

src/templates/fr/2019/chapters/caching.html
bazzadp: This is intentional as noted in the [translators guide](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Translators'-Guide). In french they write "70 %" (with a space) while in English we use "70%" (without a space). We use a non-breaking space to prevent orphaning a number from the percent sign when they fall on a line break.

src/templates/fr/2019/chapters/caching.html
bazzadp: This is intentional as noted in the [translators guide](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Translators'-Guide). In french they write "70 %" (with a space) while in English we use "70%" (without a space). We use a non-breaking space to prevent orphaning a number from the percent sign when they fall on a line break.

src/static/css/page.css
bazzadp: Yes you are correct that it's rare to find a browser that doesn't support this. However I find it easier to think in `px` values so having it on the line above is helpful to me, with the added bonus of support if needed.

bazzadp:
> * I noticed some inconsistency in how percentages are reported, in some places there were non-breaking spaces before the `%` sign

This is intentional as noted in the [translators guide](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Translators'-Guide). In french they write "70 %" (with a space) while in English we use "70%" (without a space). We use a non-breaking space to prevent orphaning a number from the percent sign when they fall on a line break.

> * Should we also add `direction: ltr` for `numeric` class? Because if there were spaces or some tricky symbols, rendition may not be as expected in Right-to-Left language layouts.

Let's wait until we have some RTL languages and tackle this then as there's a few `text-align` and `float` stylings that may require reviewing.

> * Can't we change `<td><p class="numeric">0</p></td>` to something simpler like `<td class="numeric">0</td>`? (I am sure there is reason why it is the way it is currently)

Yeah I looked at that and it changes the styling as non longer have the `margin` in the table cells. However on further investigation all the other chapters use `class="numeric"` in tables rows directly without the `<p>` element so we should be consistent. So I've made the change here as suggested.

bazzadp:
Pulled the main inline=styles changes to a separate PR (#965) so as to to give a better commit reason and make this PR easier to review.

src/static/css/page.css
ibnesayeed: Which means you overwrite the rules immediately defined. I think, those performance related tools like lighthouse will not be happy about it.

ibnesayeed:
LGTM!

> Let's wait until we have some RTL languages and tackle this then as there's a few `text-align` and `float` stylings that may require reviewing.

If and when that happens I can review those changes as I have good experience with TRL languages.

PR #963, opened 2020-07-05
Add ibnesayeed in 2020.json

ibnesayeed:
As per https://github.com/HTTPArchive/almanac.httparchive.org/pull/956#issuecomment-653891658 I have added my details in the 2020.json file.

I was not sure about the order in which the records are inserted, so I kept it sorted by the record key. I can move it around if necessary.

### DIFF
--- a/src/config/2020.json
+++ b/src/config/2020.json
@@ -193,17 +193,6 @@
       "github": "bazzadp",
       "twitter": "tunetheweb"
     },
-    "ibnesayeed": {
-      "name": "Sawood Alam",
-      "teams": [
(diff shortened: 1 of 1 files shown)


bazzadp:
I sort alphabetically by first name. Makes it easier to find someone, avoids any suggestion of order, and satisfies my OCD 😀

Could have picked record key, but went with the firstname, surname.

Could you move it so it's consistent with 2019?

ibnesayeed:
I was confused, if the full name was used for sorting of the object key, because before my entry, it was consistent either way. ;-)

PR #967, opened 2020-07-06
Only modify changed or added files - not deleted ones

bazzadp:
The last generate chapters GitHub Action [failed](https://github.com/HTTPArchive/almanac.httparchive.org/runs/842478107?check_suite_focus=true) due to trying to update timestamps of deleted files.

We don't delete files often which was why this was only seen now.

This PR changes the diffs to only look at added (A) and modified (M) files.


ibnesayeed:
@bazzadp did you check it with file names renamed? I am not too sure about all the possibilities that we might encounter, but we may also chose to go by exclusion filter instead of inclusion. For example, `--diff-filter=d` will filter out deleted files (note the lower case letter).

PR #970, opened 2020-07-06
Fix spacing issues in header on tablet

### LINKED ISSUE #815: Create a 2020 landing page

bazzadp:
Closes #815 by fixing https://github.com/HTTPArchive/almanac.httparchive.org/issues/815#issuecomment-654444252

Ipad:

Ipad header

Desktop:

Desktop header

Table of Contents can still go to two lines on widths of between 900px and 1000px (iPad is 1024px so just misses this):

Tablet header with two line Table of Contents

However with the various languages, we'll need a very wide tablet breakpoint to avoid this completely so think this is acceptable.

Talking of which, this also looks better for French on iPad width (though ToC and Contributors still on two lines, while Methodology is not):

French iPad header

Compared to production:

French Ipad header on production

### DIFF
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -488,6 +488,7 @@ h2.header {
   text-decoration: none;
   font-size: 16px;
   font-size: 1rem;
+  margin-right: 40px;
 }
 
 .navigation-logo:hover,
@@ -586,6 +587,7 @@ header nav {
 
 header nav > ul > li > * {
   margin-right: 50px;
+  margin-right: 2vw;
 }
 
 header nav > ul > li:last-child > * {
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -586,7 +586,6 @@ header nav {
 }
 
 header nav > ul > li > * {
-  margin-right: 50px;
   margin-right: 2vw;
 }
 


src/static/css/2019.css
```diff
@@ -586,6 +587,7 @@ header nav {
 
 header nav > ul > li > * {
   margin-right: 50px;
+  margin-right: 2vw;
```
rviscomi, line 590: What's this?

bazzadp: It gives a slightly smaller or larger margin for larger viewports without having to introduce a new breakpoint.

Between 900px and 1200px it really needs to be 20px to avoid the two like ToC happening very early (<1120px) but that's a little squashed when we have more space on desktop so used `vw` units instead so there's a little more room to breath when we have the space.

I left the old 50px as a fallback but probably can remove that as `vw` [support looks to be pretty comprehensive](https://caniuse.com/#search=vw)

PR #973, opened 2020-07-07
Fix font size for logo year

bazzadp:
As discussed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/954/files#r450474029 we meant to change this to 25px but looks like we never did! So now it overlaps too much:

Year overlapping logo

Changing this to `25.5px;` (Firefox seems to need the extra .5), as discussed in the PR review fixes this:

Less overlap with 25.5px

I also add the `position: relative;` to the `img` so it floats above the year, which makes it look more like it sits on top:

Text appears to sit on top when image is relative

Without `positive: relative;`:

### DIFF
--- a/src/static/css/index.css
+++ b/src/static/css/index.css
@@ -116,7 +116,7 @@ p {
 
 .intro-image-wrapper svg {
   position: relative;
-  top: 25px;
+  top: 25.5px;
   fill: #f7f779;
 }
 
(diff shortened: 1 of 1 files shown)


bazzadp:
So did I! 😀 Should have double checked.

That's live now - thanks for hard work on this. And for putting up with the many rounds of feedback. Worth it though and good bit of CSS clean up done as part of this!

PR #976, opened 2020-07-07
Remove offset for image which is now no longer required

### LINKED ISSUE #815: Create a 2020 landing page

bazzadp:
Fixes bug noticed by @rviscomi https://github.com/HTTPArchive/almanac.httparchive.org/issues/815#issuecomment-654877828

Closes #815 

See - I knew there was a reason I didn't want all that CSS in the 2020 page! 😉

### DIFF
--- a/src/templates/base/2020/index.html
+++ b/src/templates/base/2020/index.html
@@ -45,9 +45,7 @@
   .intro h2 b {
     display: block;
   }
-  .intro-image-wrapper {
-    margin-right: -100px;
-  }
+  
   .methodology-container {
     margin-top: 200px;
     padding: 0;


bazzadp:
Np. Just released it there now.

PR #979, opened 2020-07-08
REM units for page.css for eBook

bazzadp:
Been doing some more work on the ebook as discussed @rviscomi . One problem we have got is with margins and paddings looking bad when changing page size and font-size.

This PR adds some more `rem` relative-based sizes to address this and putting this ahead of the ebook changes themselves (PR to follow!) to make them easier to review. Also benefits those that font zoom in browser.

At present I've followed the current standard we use of adding `rem` on top of the old `px` measurements. Mostly cause I still find it difficult to thing in `rem`'s! But also makes it easier to review this PR. At some point in future we should probably remove them `px` fallbacks as `rem` support is [pretty much everywhere nowadays](https://caniuse.com/#feat=rem).

### DIFF
--- a/src/static/css/page.css
+++ b/src/static/css/page.css
@@ -634,13 +634,13 @@ figure .fig-desktop {
   padding: 1rem;
 }
 
-.code-block .divider {	
-  position: relative;	
-  border: 1px solid #1a2b490a;	
-  width: calc(100% + 32px);	
-  width: calc(100% + 2rem);	
-  left: -16px;	
-  left: -1rem;	
+.code-block .divider {
+  position: relative;
+  border: 1px solid #1a2b490a;
+  width: calc(100% + 32px);
+  width: calc(100% + 2rem);
+  left: -16px;
+  left: -1rem;
 }
 
 .code-block pre {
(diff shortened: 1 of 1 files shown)


rviscomi:
Will this result in some text being slightly different sizes? Or is this generally an invisible change?

bazzadp:
It's invisible for default font-size. As you can see in the change, the `rem` values assume a base-font size of 16px (which is what all browser use).

So you notice I've changed things like this:

```css
  padding: 16px;
  margin: 32px;
```

to this:
```css

  padding: 16px;
  padding: 1rem;
  margin: 32px;
  margin: 2rem;
```

So they are the exact same sizes whether you use the `px` line or the `rem` line for default font-size of 16px.

However when you use browser zoom (or shrink to print the ebook) then you get wider/smaller margins and padding in line with increased/decreased default font-size. Which is exactly what you want - it's wrong to use the small padding as font's get smaller and padding then looks massive in comparison. And similarly in return.

In #731 I basically did the same thing for font-sizes. This just completes the job and also does it for `padding` and `margin`.

But to most people, unless you've editing your browser default font-size settings, you won't even notice this change as all sizes are the same.

Issue #989, opened 2020-07-10
2020 Content Progress


ibnesayeed:
Chapters | <span title="📄 (Draft), 🔍 (SQL), 📊 (Results)">Links</span> | <span title="Authors+Reviewers+Analysts≥All">Team</span> | <span title="**Jul 6th**: Project owners have selected an author to be the content team lead">1<span> | <span title="**Jul 13th**: The content team has at least one author, reviewer, and analyst (minimally viable team formed)">2<span> | <span title="**Jul 20th:** The content team has completed the chapter outline in the draft doc">3<span> | <span title="**Jul 27th**: Analysts have triaged the feasibility of all proposed metrics">4<span> | <span title="**Jul 27th**: Analysts have prepared all feasible metrics for testing and added queries to the sql directory">5<span> | <span title="**Sep 7th**: Analysts have queried all metrics and saved the output to the results sheet">6<span> | <span title="**Sep 14th**: The content team has reviewed the results sheet">7<span> | <span title="**Oct 12th**: Authors have completed the first draft in the doc">8<span> | <span title="**Oct 26th**: The content team has prototyped all data visualizations">9<span> | <span title="**Oct 26th**: The content team has reviewed the final draft, converted to markdown, and filed a PR to add it to the 2020 content directory">10<span>
:--------|:------|:----:|:-:|:-:|:-:|:-:|:-:|:-:|:-:|:-:|:-:|:-:
[...]

Issue #921, opened 2020-06-27
HTTP/2 2020


obto:
# Part IV Chapter 22: HTTP/2

## Content team
| Authors | Reviewers | Analysts | Draft | Queries | Results |
| ------- | --------- | -------- | ----- | ------- | ------- |
| @dotjs | @MikeBishop @LPardue @rmarx @ibnesayeed @pmeenan @Nooshu @gregorywolf @bazzadp | @gregorywolf | [Doc][~draft-doc] | [*.sql][~sql-dir] | [Sheet][~results-sheet] |

**Content team lead:** @dotjs

Welcome chapter contributors! You'll be using this issue throughout the [chapter lifecycle](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle) to coordinate on the content planning, analysis, and writing stages.

The **content team** is made up of the following contributors:

- [Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)
- [Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)
- [Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)

**New contributors:** If you're interested in joining the content team for this chapter, just leave a comment below and the content team lead will loop you in.

_Note: To ensure that you get notifications when tagged, you must be "[watching](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#notifications)" this repository._

### [**0. Form the content team**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#0-create-content-team)
+ [x] **Jul 6th**: Project owners have selected an author to be the content team lead
+ [x] **Jul 13th**: The content team has at least one author, reviewer, and analyst (minimally viable team formed)

### [**1. Plan content**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
+ [x] **Jul 20th:** The content team has completed the chapter outline in the [draft doc][~draft-doc]
+ [x] **Jul 27th**: Analysts have triaged the feasibility of all proposed metrics

### [**2. Gather data**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
[...]

Issue #897, opened 2020-06-27
JavaScript 2020


obto:
# Part I Chapter 2: JavaScript

## Content team
| Authors | Reviewers | Analysts | Draft | Queries | Results |
| ------- | --------- | -------- | ----- | ------- | ------- |
| @tkadlec | @ibnesayeed @jadjoubran @ahmadawais @denar90  | @rviscomi @paulcalvano  | [Doc][~draft-doc] | [*.sql][~sql-dir] | [Sheet][~results-sheet] |

**Content team lead:** @tkadlec

Welcome chapter contributors! You'll be using this issue throughout the [chapter lifecycle](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle) to coordinate on the content planning, analysis, and writing stages.

The **content team** is made up of the following contributors:

- [Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)
- [Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)
- [Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)

**New contributors:** If you're interested in joining the content team for this chapter, just leave a comment below and the content team lead will loop you in.

_Note: To ensure that you get notifications when tagged, you must be "[watching](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#notifications)" this repository._

### [**0. Form the content team**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#0-create-content-team)
+ [x] **Jul 6th**: Project owners have selected an author to be the content team lead
+ [x] **Jul 13th**: The content team has at least one author, reviewer, and analyst (minimally viable team formed)

### [**1. Plan content**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
+ [x] **Jul 20th:** The content team has completed the chapter outline in the [draft doc][~draft-doc]
+ [x] **Jul 27th**: Analysts have triaged the feasibility of all proposed metrics

### [**2. Gather data**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
[...]

PR #990, opened 2020-07-10
Automate chapter progress monitoring

### LINKED ISSUE #943: Automate the monitoring of chapter progress
This is a stretch, but I'd love to hear if anyone is interested to work on a small project like this.

### Problem
GitHub has a few features that make project management easier, like [project boards](https://github.com/HTTPArchive/almanac.httparchive.org/projects/1) and support for checkboxes to track issue progress. One challenge I had last year managing the project, and will continue to be a challenge this year, is keeping on top of 20+ chapters' worth of progress updates.

### Proposal
I would love to be able to automate some of this chapter monitoring so I have a single view where I can see how far along each chapter is at a glance, in terms of hitting milestones. The [GitHub API](https://developer.github.com/v3/issues/comments/#get-an-issue-comment) has support for getting issue comments, so I can imagine a system that fetches the first comment of each [chapter issue](https://github.com/HTTPArchive/almanac.httparchive.org/issues?q=is%3Aopen+label%3A%222020+chapter%22+sort%3Acreated-asc), parses the status of each milestone's checkbox, and saves the results to a grid of chapters/milestones.

One approach could be to create a Google Sheet with each chapter and its corresponding issue number. Then an Apps Script function periodically queries each issue and writes the status of each chapter's milestone to columns in the sheet.

Another approach could be to have a central GitHub issue that is automatically updated based on GitHub actions or something that can listen for edits to the specific issue comments and modify the central issue.

Any other ideas? Would anyone be interested in building this? cc @obto @HTTPArchive/developers

ibnesayeed:
This closes #943 by implementing a basic consolidated chapter progress tracking. This can be extended to poll more data from external sources in the future such as word counts of each chapter. Current workflow is designed to function as described below:

* The workflow is triggered anytime an issue is edited (such as checking or unchecking a checkbox in an issue or manually editing it)
  * It will only run when the first post of the issue is edited, not when a comment is edited (to avoid unnecessary work)
  * It will run on any issue edited, not just the ones we are interested in, which is a false-positive, but is not harmful and is not going to be too frequent
  * There are chances of race condition when, say, multiple checkboxes are checked in one go and the workflow ran out of order, however, this may not happen often and there will be an eventual consistency
* The workflow can also be triggered manually from the **Actions* tab in case report is inconsistent
* Individual chapter progress issues where the progress statuses are polled from are identified using `FILTER_LABEL = '2020 chapter'` constant, which should be adjusted next year
* The meta issue where the progress is reported is configured using `TRACKER_ISSUE_ID = 989` constant
* Once the workflow is triggered, it fetches all the individual chapter issues, parses their body to identify the tasks that are checked or unchecked, and renders the summary in a tabular form in the target reporting issue identified by `TRACKER_ISSUE_ID`
  * For simplicity sake, it assumes that each chapter issue will have exactly the same number of tasks and will have exactly the same order
* The report target issue must contain a placeholder region identified by `<!-- REPORT START -->` and `<!-- REPORT END -->` comments to receive the updated report
  * This style of replacement preserves any additional notes in the body, which, if we were to replace the whole body, we will need to move any additional notes in a follow up comment or in the form of a template inside the workflow file (the latter would be ugly and difficult to maintain)

Testing of this workflow is performed in a limited manner because it would require live repository with plenty of issues in it to test properly, hence there might be some bugs. Once merged in the `main` branch, we should be able to have a better option to fix any regressions.


bazzadp:
Can you give an example of the output?

ibnesayeed:
> Can you give an example of the output?

Suppose we had only three chapters and four tasks, the generated report will be something like the following (which will only replace the content within the placeholder markers):

```
<!-- REPORT START -->
Milestones | 1 | 2 | 3 | 4
:----------|:-:|:-:|:-:|:-:
[JavaScript](https://github.com/HTTPArchive/almanac.httparchive.org/issues/897) | ☑ | ☐ | ☐ | ☐
[Markup](https://github.com/HTTPArchive/almanac.httparchive.org/issues/899) | ☑ | ☑ | ☐ | ☐
[HTTP/2](https://github.com/HTTPArchive/almanac.httparchive.org/issues/921) | ☑ | ☐ | ☐ | ☐
<!-- REPORT END -->
```

Which will render something like:


Milestones | 1 | 2 | 3 | 4
:----------|:-:|:-:|:-:|:-:
[JavaScript](https://github.com/HTTPArchive/almanac.httparchive.org/issues/897) | ☑ | ☐ | ☐ | ☐
[Markup](https://github.com/HTTPArchive/almanac.httparchive.org/issues/899) | ☑ | ☑ | ☐ | ☐
[HTTP/2](https://github.com/HTTPArchive/almanac.httparchive.org/issues/921) | ☑ | ☐ | ☐ | ☐

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,61 @@
...
+          script: |
+            // Update the following two constants each year
+            const FILTER_LABEL = '2020 chapter'
+            const TRACKER_ISSUE_ID = 989
+
+            const parseTasks = text => {
+              const tasks = []
+              text.split('\n').forEach(line => {
+                if (/^(\s{2,})?([*+-]|\d+\.)\s+\[ \]\s+\S+/.test(line)) {
+                  tasks.push('☐')
```
rviscomi, line 24: I think an empty table cell would appear a bit cleaner and still convey "incomplete".
```suggestion
                  tasks.push('')
```

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,61 @@
...
+              repo: context.repo.repo,
+              direction: 'asc',
+              labels: FILTER_LABEL
+            })
+
+            let report = ''
+            for (const issue of taskIssues) {
+              tasksStatus = parseTasks(issue.body)
+              if(!report) {
+                report = `Milestones | ${tasksStatus.map((v, i) => i + 1).join(' | ')}\n`
```
rviscomi, line 44: Could we also parse the draft doc URL? For some issues it's still `#`.

If the doc URL is defined, it should appear as a clickable link, like "[Doc]()". But if the  URL is `#` it should just be blank, with no "Doc" word.

bazzadp:
Very nice. Playing with it in my own fork. Example here: https://github.com/bazzadp/almanac.httparchive.org/issues/73

Will have some review comments soon.

.github/workflows/progress-tracker.yml
ibnesayeed: I have actually added code to parse other URLs as well, in case we need them in the future.

Rather than consuming yet another grid column and repeat `Doc` in each row, I added draft link with an icon next to the chapter title like:

[JavaScript](#) • [🗎](#)

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,80 @@
+name: Progress Tracker
+
+on:
+  workflow_dispatch:
+  issues:
+    types:
+      - edited
```
bazzadp, line 7: I'm really not a fan of running this on every issue edit. As someone who makes a lot of typos this would run it a lot. Especially as we hope this repo is about to get very active.

I think we should run on a schedule. E.g. to run at 6am and 6pm every day (UTC):

```suggestion
on:
  workflow_dispatch:
  schedule:
    - cron:  '0 6,18 * * *'
```

@rviscomi could also kick it off manual if he wants to.

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,80 @@
...
+
+jobs:
+  track-progress:
```
bazzadp, line 10: If we are going to run it on every edit (see above), then we should at least bomb out early:

```suggestion
  track-progress:
      if: contains(github.event.issue.labels.*.name, '2020 chapter')
```

Annoyingly can't seem to use environment variables in the `contains` function so have to hard code this. Should make a comment above to note if changing FILTER_LABEL then need to change here too.

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,80 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v2
+        with:
+          script: |
+            // Update the following two constants each year
+            const FILTER_LABEL = '2020 chapter'
+            const TRACKER_ISSUE_ID = 989
```
bazzadp, line 18: Re-use the env variables set above.

```suggestion
            const FILTER_LABEL = process.env.FILTER_LABEL
            const TRACKER_ISSUE_ID = process.env.TRACKER_ISSUE_ID
```

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,80 @@
...
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: TRACKER_ISSUE_ID
+            })
+
+            github.issues.update({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: TRACKER_ISSUE_ID,
+              body: progressIssue.data.body.replace(/<!-- REPORT START -->[\s\S]*<!-- REPORT END -->/, `<!-- REPORT START -->\n${report}<!-- REPORT END -->`)
```
bazzadp, line 79: Does this do an update even if nothing's changed? Should we check first or simpler just to do the update and let GitHub worry about it?

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,80 @@
...
+  workflow_dispatch:
+  issues:
+    types:
+      - edited
+
+jobs:
+  track-progress:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v2
```
bazzadp, line 13: If we are going to run it on every edit (see above), then we should at least bomb out early instead of within the script:

```suggestion
      if: contains(github.event.issue.labels.*.name, env.FILTER_LABEL)
      - uses: actions/github-script@v2
```

.github/workflows/progress-tracker.yml
bazzadp: Actually see it only does it on edits of the issue it self (not on comments below them, or even adding labels). Maybe not so bad then. But would add a few suggestions below to limit this further for edits to non-chapter issues.

bazzadp:
This breaks if a non-chapter is tagged with the '2020 chapter' label:

Broken summary

I know we could just try not to use that label except on these issues but anyway to say if it doesn't have 10 tick status's then ignore to try to limit to the chapter issues only?

.github/workflows/progress-tracker.yml
ibnesayeed: I have further updated it to list it in a column as I have also added a column for teams.

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,80 @@
+name: Progress Tracker
```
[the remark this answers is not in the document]
ibnesayeed: Moved configs to ENV.

We need some way to figure out which ones are related to chapter tracking. Labels seem a reasonable approach. With the added check of the number of checkboxes in an issue, I think we can avoid most of the false-positives for now. If more complicated situations emerge, we can improve it.

.github/workflows/progress-tracker.yml
ibnesayeed: Since we have added `workflow_dispatch` event which will allow us to run it manually, anytime we want, I think an automated scheduled update will avoid unnecessary race conditions, so let's keep it this way. If needed we can increase the frequency to every hour our something, but every 12 hours seems good to me for now.

.github/workflows/progress-tracker.yml
ibnesayeed: Since I have decided to not go with triggering on issue edit, I think we can safely pass this for now. Though, I have limited the workflow to this repo only.

.github/workflows/progress-tracker.yml
ibnesayeed: Done!

.github/workflows/progress-tracker.yml
ibnesayeed: I was going to check for changes before updating the issue, but my early tests suggested that GitHub ignores the update if the content is same, so it was much simpler to leave it for GH to handle.

.github/workflows/progress-tracker.yml
ibnesayeed: Let's not go for running it on each issue update.

ibnesayeed:
> This breaks if a non-chapter is tagged with the '2020 chapter' label:

I have put some checks to minimize this situation. Now, the first non-zero tasks length issue will determine the desired tasks size and will only include that match it.

ibnesayeed:
Thanks for your reviews @bazzadp, My local changes grew bigger before I could accept your suggestions online, so I manually added you as a co-author of one of the commits in the commit message so you get the credit.

.github/workflows/progress-tracker.yml
bazzadp: I've actually come round to doing it on edits. Was worried every comment or label or raising of an issue would trigger this but it doesn't appear to so less worried than I was.

What do you think @rviscomi ?

.github/workflows/progress-tracker.yml
bazzadp: Good idea on limiting to this repo.

Depending what way we go on editing versus cron it we may still want to add this restriction so leaving this unresolved for now.

bazzadp:
Ah not worried about credit - there's more than enough of my fingerprints all over this code 😀

I copied your latest action [to my own fork](https://github.com/bazzadp/almanac.httparchive.org/blob/main/.github/workflows/progress-tracker.yml) to test it but it doesn't seem to be working [as it doesn't produce the right output](https://github.com/bazzadp/almanac.httparchive.org/issues/73). Any ideas? And yes I did update the issues and repo check to my own 😀

ibnesayeed:
> I copied your latest action [to my own fork](https://github.com/bazzadp/almanac.httparchive.org/blob/main/.github/workflows/progress-tracker.yml) to test it but it doesn't seem to be working [as it doesn't produce the right output](https://github.com/bazzadp/almanac.httparchive.org/issues/73). Any ideas? And yes I did update the issues and repo check to my own

I will look into it in a few.

bazzadp:
Have you access there. This is my throw-away testing repo so as not to pollute this main one so feel free to change and test in that. I rebase every so often anyway.

rviscomi:
Why not run on every edit? If we can filter out non-chapter issues and it won't exceed any quotas, it would be helpful to ensure that the tracker is up to date.

If scheduling is needed, can we go as often as 15 minutes? I know I can manually run the action, but it'd be great to know the tracker is reasonably close to up to date so I don't have to manually run it.

bazzadp:
I’m happy to revert back to every undate now I’ve confirmed it’s more limited than I thought (it’s won’t trigger for comments, nor edits on comments, nor tagging labels... etc.)

But should look at the extra checks I suggested above.

Can always revert to Cron if it ends up running too often. In fact doing it in edits might make it run **less often** 😀

ibnesayeed:
I have made some changes:

* Added debug messages when an issue is skipped or number of tasks is set
* Changed `TRACKER_ISSUE_ID` to `TRACKER_ISSUE_NUMBER`
* Updates some control-flow logic to better skip false positives
* Reverted the trigger behavior to issue edit with additional step filter to minimize false positives

.github/workflows/progress-tracker.yml
rviscomi: +1 on edit with checks seems like a good compromise

ibnesayeed:
This seems ready from my end. Tested it in @bazzadp's repo both by checking a task and by triggering the event manually.

.github/workflows/progress-tracker.yml
ibnesayeed: I have used this with some added condition for the manual trigger where the `event.issue` context will not be present.

.github/workflows/progress-tracker.yml
ibnesayeed: It would have been better if we could skip it on the job level instead of the step level, but I think we will go with the less brittle approach here to not specify labels in two different places and skip the step instead. This will be a little more work for GH, but that's okay.

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,126 @@
...
+
+env:
+  # Update these environment variables every year
+  FILTER_LABEL: '2020 chapter'
+  TRACKER_ISSUE_NUMBER: 989
```
bazzadp, line 6: Not loving the picking up of number of issues from first issue. Means the whole thing breaks if first issue is broken (like it is on my fork at the mo!). Also it took me a while to figure out what was going on as missed your comment, so wonder if we'll remember how this works next year.

I think we should add as a param instead:

```suggestion
  FILTER_LABEL: '2020 chapter'
  TRACKER_ISSUE_NUMBER: 989
  TOTAL_TRACKED_TASKS: 10
```

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,126 @@
...
+              })
+              return team
+            }
+
+            const formatLinks = links => {
+              const icons = {
+                draft: '🗎',
+                sql: '⛁',
+                results: '⚝'
+              }
```
bazzadp, line 76: Let's use some more meaningful icons. As using the same and only showing when they exist makes it difficult to tell which ones you are showing. Plus the one you've chosen is very difficult to read on windows and just a blank square on macOS.

```suggestion
              const icons = {
                draft: '✍',
                sql: '🔍',
                results: '📊'
              }
```

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,126 @@
...
+    name: Track Progress
+    if: github.repository == 'HTTPArchive/almanac.httparchive.org'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v2
+        if: github.event_name == 'workflow_dispatch' || contains(github.event.issue.labels.*.name, env.FILTER_LABEL)
+        with:
+          script: |
+            const FILTER_LABEL = process.env.FILTER_LABEL
+            const TRACKER_ISSUE_NUMBER = process.env.TRACKER_ISSUE_NUMBER
```
bazzadp, line 25:
```suggestion
            const FILTER_LABEL = process.env.FILTER_LABEL
            const TRACKER_ISSUE_NUMBER = process.env.TRACKER_ISSUE_NUMBER
            const TOTAL_TRACKED_TASKS = process.env.TOTAL_TRACKED_TASKS
```

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,126 @@
...
+
+            const taskIssues = await github.paginate(github.issues.listForRepo, {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              direction: 'asc',
+              labels: FILTER_LABEL
+            })
+
+            let report = ''
+            let totalTrackedTasks = 0
```
bazzadp, line 92:
```suggestion
```

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,126 @@
...
+            for (const issue of taskIssues) {
+              const tasksStatus = parseTasks(issue.body)
+              const tasksCount = tasksStatus.length
+              if(!tasksCount) {
+                core.info(`Skipping issue #${issue.number} with zero tasks`)
+                continue
+              }
+              if(!report) {
+                totalTrackedTasks = tasksCount
```
bazzadp, line 101:
```suggestion
```

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,126 @@
...
+                core.info(`Setting total tracked tasks to ${totalTrackedTasks} via issue #${issue.number}`)
+                report = `Chapters | Links | Team | ${tasksStatus.map((v, i) => i + 1).join(' | ')}\n`
+                report += `:--------|:------|:----:|:${tasksStatus.map(() => '-').join(':|:')}:\n`
+              }
+              if(tasksCount != totalTrackedTasks) {
```
bazzadp, line 106:
```suggestion
              if(tasksCount != TOTAL_TRACKED_TASKS) {
```

.github/workflows/progress-tracker.yml
ibnesayeed: I chose all three icons different, but if it was not rendering on your machine then it perhaps is a font issue. Anyways, I am updating them.

bazzadp: Ah. That explains why I saw one of the icons in Windows but none in Mac or iOS! Didn’t check the source to see if they all rendered and were different.

Above ones seem to be supported in both Apple and Windows world.

.github/workflows/progress-tracker.yml
ibnesayeed: Parameterizing it would make some of the code logic simpler, but there will be more things to update. That's why I tried to make it auto recognize. It would have been a lot more flexible if we had parsable short badges for each task, then we could have used a hash instead of an array and filled up any empty sports with not applicable and they could be reordered as well. Something like:

* [x] *LEAD* - Jul 5 - A lead selected
* [ ] *TEAM* - Jul 15 - At least one member in each team
* [ ] *OUTLINE* - Aug 3 - Chapter outline ready

Then we could use these short names as task identifiers in the column rather than relying on sequence number.

#JustSharingMyThoughts :wink:

bazzadp: True, but that makes the issues less readable.

I think what you’ve done is pretty awesome and will be very useful. I’m only suggesting this extra change to make it easier to understand next year as took me a bit to understand. Plus I think it makes it more robust to that first issue being edited so it no longer follows the pattern. So I think it’s worth the extra config.

.github/workflows/progress-tracker.yml
bazzadp: I'm still seeing that first one as the default square on my machine:

Screenshot of commit

It's better now as at least they are different but the square is ugly. Can you see the one I suggested on your machine? If so, could we use that?

ibnesayeed:
* I have simplified some logic where we are initializing and building the report table to skip any ticket that does not have the exact number of tasks as configured , so no need to check for zero entries separately (as that can be a valid configuration as well)
* Also, made sure that the environment variable is converted to an integer when it is expected to be an integer
* Emojis for links are changes to those that seem to be more widely available cross-platform

.github/workflows/progress-tracker.yml
ibnesayeed: > Above ones seem to be supported in both Apple and Windows world.

I am using Linux and they all rendered well here, including those that you suggested.

ibnesayeed:
@bazzadp I have not tested these new changes by running them, so please make sure to verify it works before approving.

.github/workflows/progress-tracker.yml
bazzadp: OK but looks like you didn't use my suggested one for the first one as not rendering for me?

ibnesayeed: > Can you see the one I suggested on your machine?

Yes!

> If so, could we use that?

Done!

bazzadp: All good now - thanks!

ibnesayeed: > OK but looks like you didn't use my suggested one for the first one as not rendering for me?

That's right, I tried something else which was more meaningful (a document with image and text), but when you reported it was not rendering well for you then I updated it to what you suggested.

ibnesayeed:
Don't merge it yet!

ibnesayeed:
> Don't merge it yet!

It's good to go now. I forgot to remove a debug message which is not relevant after we changed how we set the total tasks count.

rviscomi:
Not sure if this PR includes this use case, but many chapters still have `#` given as the draft doc URL and all of them have it as their results sheet URL. For those chapters I'd like to make that clear so I know that we still need to update their URLs and they're not clickable yet. Is that something that the action currently handles, or should we build that in a follow-up PR?

ibnesayeed:
> Not sure if this PR includes this use case, but many chapters still have `#` (a placeholder) given as the draft doc URL and all of them have it as their results sheet URL. For those chapters I'd like to make that clear so I know that we still need to update their URLs and they're not clickable yet. Is that something that the action currently handles, or should we build that in a follow-up PR?

The action checks for Draft, SQL, and Results links, if they are found and are not `#` then they are linked with their corresponding icon, otherwise no icon or link will be rendered for that particular link type.

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,121 @@
...
+            })
+
+            let report = ''
+            for (const issue of taskIssues) {
+              const tasksStatus = parseTasks(issue.body)
+              const tasksCount = tasksStatus.length
+              if(tasksCount != totalTrackedTasks) {
+                core.info(`Skipping issue #${issue.number} with ${tasksCount} tasks`)
+                continue
+              }
```
bazzadp, line 100: I actually think it's worth keeping that debugging in:

```suggestion
              } else {
                core.info(`Adding issue #${issue.number} with ${tasksCount} tasks`)  
              }  
```

ibnesayeed:
You can check the logic in the `parseLinks` and `formatLinks` functions to see how it works and feel free to improve it or suggest any improvements.

.github/workflows/progress-tracker.yml
```diff
@@ -0,0 +1,121 @@
...
+              return team
+            }
+
+            const formatLinks = links => {
+              const icons = {
+                draft: '📄',
+                sql: '🔍',
+                results: '📊'
+              }
+              return Object.entries(links).filter(l => l[1] && l[1] != '#').map(l => `[${icons[l[0]] || l[0]}](${l[1]})`).join(' ')
```
bazzadp, line 79:
```suggestion
              return Object.entries(links).filter(l => l[1] && l[1] != '#' && l[1] != 'https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/sql/2020/').map(l => `[${icons[l[0]] || l[0]}](${l[1]})`).join(' ')
```

Requested by @rviscomi as SQL is linking to SQL dir and not using `#`

rviscomi:
Sounds great. @bazzadp when it LGTY let's ship it! 🚀

.github/workflows/progress-tracker.yml
ibnesayeed: Yes, I have added more debugging messages now, but the one I removed was logically irrelevant.

.github/workflows/progress-tracker.yml
ibnesayeed: I think I have taken care of it, thanks for pointing this out.

ibnesayeed:
> Just two more changes and we're done - I swear!

Please do not hesitate to provide more feedback to make it better, as it is always a work in progress.

ibnesayeed:
I think `['#', sqlBaseDir].includes(l[1])` would have been sufficient instead of `['', '#', sqlBaseDir].includes(l[1])` because the way we are parsing links, empty links will not match in the first place.

rviscomi:
😍 Just found an issue whose milestone was out of date, edited it, and the action works great! Thanks again @ibnesayeed for building this!

ibnesayeed:
@rviscomi I liked the changes you made in the report tracker, especially moving report placeholder above and the legends that you added there (which I have updated to include emojis as well).

ibnesayeed:
Should we lock that report tracker issue so people cannot comment there? If we do so, we ought to make sure that the action bot can still update the issue.

rviscomi:
Locking only seems to prevent people who aren't on the HTTP Archive teams from commenting, but almost everyone already is.

ibnesayeed:
#1041 adds minor tweaks to the team members reporting in #989.

Issue #933, opened 2020-06-28
[Discuss] How to improve testing


bazzadp:
As discussed in the [Goals section of the Developer's Guide](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Developers'-Guide#goals-for-2020-edition), I'd like to consider how we can increase our automatic testing.

The tech stack is starting to get complicated and with 2020 coming along, we should look at how we can ensure we don't break old functionality when adding new functionality.

Thoughts so far include:
- [ ] Build upon the [pytest](https://docs.pytest.org/en/latest/) tests we already have. #1229 
- [x] Add Python/JS/CSS/JSON/YML linters - #938 and #939 
- [x] Add HTML/Jinja2 linters. #998 
- [x] Add Dockerfile - #937 
- [ ] Add Lighthouse test on opening a PR [via a GitHub Action](https://twitter.com/patrickhulce/status/1277048093083648000?s=20) and flag if any scores below a threshold. - #1178
- [x] Test generate some (or all?) URLs and ensure 200 response. Should be easier thanks to #932 - #1124 
- [ ] Run Cypress to test front end.
- [x] Add dependenbot and freeze version numbers for stability and visibility of changes - #1004 #1008 

Any other thoughts @HTTPArchive/developers ?

ibnesayeed:
Adding a `Dockerfile` with necessary development and testing environment will help contributors test their changes locally more easily before pushing.

bazzadp:
I've not used Docker much to be honest. What would that give us? Just conscious of adding extra tech and complexity so would want to know the benefits and how we maintain the docker file going forward.

At the moment you need node v12, and Python 3 installed (which are standard enough development tools that I would expect most developers to have installed IMHO). Then we do have some more requirements to install though `npm install` and `pip` (as detailed in `src/README.md`). Those dependencies do take a few minutes to install admittedly, but they are one command to run, and only need to be done once. It's then possible to run the site in MacOS, Windows and Linux (not tried linux, but presume so) and developers can develop locally.

I'm guessing Docker could help with those dependency installs? But not sure if installing Docker would take just as long for those of us that don't use it, and would give us another thing to maintain?

What I think we need is a combination of:
- Test scripts - like `pytest` that we have, to run though some test cases. Ideally developers should do this before submitting PRs, and we should do again when releasing.
[...]

PR #998, opened 2020-07-11
Add HTML Linter

bazzadp:
Note: Depends on #997 so leaving this in draft for now.

Part of #933 

Adds [HTMLHint](https://htmlhint.com/) linter through [Github Super-linter](https://github.com/github/super-linter/).

Did have to turn off too many rules despite using Jinja2 and EJS templates rather than pure HTML, which is good. Only one I'd like to have had is [id-unique](https://htmlhint.com/docs/user-guide/rules/id-unique) but we reuse it in some of our templates so can't have it on without a fair bit of refactoring.

I've also upgraded Super-Linter to v3.3.0. Would like to just use v3 but seems to be a bit broken at the moment. Will retest it before merging.


ibnesayeed:
I was wondering, should we also trigger HTML linting on generated HTML outputs? It may be a tricky things to do, but is possible to create an intermediary static file generation Action then run the linter on the output directory. This is something we can do at the very end when the site is to be released finally.

bazzadp:
Ideally yes, however that would take a little longer to run. Also we'll have to use the same config if we're using htmllint through the Super Linter. I think this (and other checks!) would be a good test to run nightly on `main` branch rather than slow every PR with them.

However I think this still a good check to have on PRs to try to reduce the change of bad code getting in there in the first place.

.github/linters/.htmlhintrc
```diff
@@ -0,0 +1,25 @@
...
+  "alt-require": true,
+  "attr-lowercase": ["viewBox"],
+  "attr-value-double-quotes": true,
+  "attr-value-not-empty": false,
+  "attr-no-duplication": true,
+  "attr-unsafe-chars": true,
+  "doctype-first": false,
+  "doctype-html5": true,
+  "empyty-tag-self-close": true,
+  "id-unique": false,
```
ibnesayeed, line 11: Do we not want to enforce ID uniqueness?

bazzadp: Yes, ideally but can't due to reuse in some templates as per my PR description above.

For example:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/fdbdbe9feab00b302c73078d77abd1ded1659d4f/src/templates/base/2019/base_chapter.html#L184
and also
https://github.com/HTTPArchive/almanac.httparchive.org/blob/fdbdbe9feab00b302c73078d77abd1ded1659d4f/src/templates/base/2019/base_chapter.html#L194
Only one of these ends up in the template as it's surround in `{% if ... %}/{% else %}/{% endif %}` Jinja2 constructs.

And also loads of instances in ebook templated which is generated.

Would be a bit of refactoring to prevent this, and not 100% sure we should to be honest. Could make the Jinja2 code more complex to write and read.

Ideally it would be nice if HTMLHint allowed us to ignore certain ids but it currently doesn't.

I say we go with what we have at present, as still loads of benefits to the other checks. We could then look to enforce it later, or look at linting the actual HTML produced in the end as you suggested elsewhere.

ibnesayeed: Yes, you are right, in conditional templates it is very possible and only a smart template linter can detect the branch in place. We can enforce this if and when we decide to lint the generated HTML later, so I will approve this PR.

Ideally, we would have added a comment next to this rule, explaining why we chose what we did, but unfortunately JSON does not support comments (unless we are using [JSON5](https://json5.org/))

rviscomi: I've seen some linters support a comment disabling certain rules for specific lines of code. Does this linter have a feature like that so we can enable the unique ID rule but turn it off in cases where the IDs are conditionally added?

bazzadp: I’ve configured it where possible. Note I allow viewBox to be an exception to the rule that attributes must be all lowercase a few lines above.

However this is rule [doesn’t allow any configuration but on or off](https://htmlhint.com/docs/user-guide/rules/id-unique).

Still. I was pleasantly surprised how few rules I had to turn off so think there’s still a lot of value in this. We’ll look at linting output HTML as part of a nightly build as think we can run other stuff (e.g. Lighthouse, test that every page returns a 200... etc.). I’ve broken the sitemap on previous releases when changing python so think that would be a good test script to be able to run offline on demand (e.g. when doing a release) and nightly on `main`.

PR #1000, opened 2020-07-11
Misc cleanups for 2020 landing page

bazzadp:


src/templates/en/2020/index.html
```diff
@@ -5,19 +5,19 @@
...
 {% block twitter_image_alt %}The {{ year }} Web Almanac{% endblock %}
 
-{% block date_published %}2019-11-04T12:00:00.000Z{% endblock %}
-{% block date_modified %}2020-07-06T00:00:00.000Z{% endblock %}
+{% block date_published %}2020-07-06T00:00:00.000Z{% endblock %}
+{% block date_modified %}2020-07-11T00:00:00.000Z{% endblock %}
 
 {% block intro_title %}Coming soon!{% endblock %}
-{% block intro_sub_title %}The 2020 <b>state of the web</b> report{% endblock %}
+{% block intro_sub_title %}The {{ year }} <b>state of the web</b> report{% endblock %}
```
rviscomi, line 12: 🤔 What's the point in making this dynamic when it's customized for the 2020 landing page? When we launch the 2020 home page, will this template get overwritten, or do you plan to reuse it in 2021 somehow?

rviscomi:
🎉 issue/PR number 1000 🎉

src/templates/en/2020/index.html
bazzadp: Yes that's my thinking. This won't change - and probably should be used for 2019. So might move it to `src/templates/base.html` at some point. And easier if moving the exact same text, rather than having to think if it ends up being the same.

Alternatively we might want to look back on this in 2021 and dig up this old version and use it.

Plus the French team went with this, and thought better to be consistent. And I prefer this 😀

bazzadp:
Should have send it coming and picked a more exciting one!!

PR #1004, opened 2020-07-12
Add Dependabot Update config for automated dependency updates

ibnesayeed:
This is a basic Dependabot update config file that will enable automatic dependency updates beyond just security fixes as discussed in https://github.com/HTTPArchive/almanac.httparchive.org/issues/933#issuecomment-657133713. For now, I am keeping it simple, but in the future we can customize to set update policy, assignee, and labels for each package ecosystem as described in [the docs](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates). We may also need to version-pin all the `requirements.txt` dependencies for this bot to take effect, but it may work on them nonetheless.


rviscomi:
@ibnesayeed are you branching from `production`? This PR and your last one have had a long stream of historical commits. Just curious if that's intentional and why not branch from `main` instead.

bazzadp:
Nice! This has enabled the [new Dependency Graph tab](https://github.com/HTTPArchive/almanac.httparchive.org/network/updates), which was inactive before:

Dependency Graph

And also recognised some NPM updates and submitted PRs for them.

Very cool.

ibnesayeed:
> @ibnesayeed are you branching from `production`? This PR and your last one have had a long stream of historical commits. Just curious if that's intentional and why not branch from `main` instead.

I did notice that that first time and thought I will do something about it, but forgot the next time. I did create the feature branch off of the main after syncing it, but I forgot to rebase it so I got all these merge commits between.

PR #1011, opened 2020-07-12
fix compression

ksakae1216:
Part of #1001 

Plz review.


bazzadp:
Thanks looks good! BTW while testing I noticed one stat on CSS that wasn't translated if you wouldn't mind doing the honours?:

CSS Japanese Featured Chapter screenshot

Issue #1016, opened 2020-07-13
Callout headlines to increase interest


natedame:
Hi @rviscomi @obto can I make an editorial and messaging suggestion?

The material presented in the almanac is intriguing and of value, but is structured primarily as - ahem, well - an almanac. *No criticism here* - I've been taking in the scope of the project and watching how things come together and how contributors/reviewers are organized. This is certainly a massive lift executed by volunteers, and the end result is impressive and exhaustive. I'm looking forward to contributing for the first time this year. But with an additional layer I think the almanac could be more engaging and possibly even have a wider reach.

I'd like to suggest that when 2020 goes live, **it includes 3-5 (or a few more?) callout headlines.** A callout headline would be one data point or finding of particular interest - **something a first-time homepage viewer would find interesting, and bloggers or reporters may find worthwhile commenting on** - items of interest and/or surprise, items that illustrate macro trends of note, etc., to add intrigue and depth to the almanac as a whole. 

These "callout headlines" could be on the homepage, each with some text and perhaps an image (graph or otherwise), linking to the appropriate chapter section that covers it. The purpose of these is to give readers a hook to pull them into the material, to guide them to noteworthy items, and perhaps even highlight items worth sharing or covering in other publications to increase the reach of the almanac.

The process of determining these headlines could be rather straightforward too:
- Each chapter author nominates 2+ possible callout headlines
- BONUS: Ask all contributors to vote for their favorite callout headline (most surprising, interesting, etc. - a simple survey or otherwise) to help lead editors review and prioritize
- Lead editors select primary callout headlines to feature
- BONUS: Perhaps each chapter has its own 1-2 callout headlines for their own pages

I'm sure there are lots of other ideas the group could think of in the spirit of "callout headlines." I'd be happy to brainstorm with others etc., or contribute in this arena. My experience is primarily with words, messages, and people (I'm not a developer) so I'd be happy to help coordinate or project manage this idea if it's something that's of interest.

obto:
I quite like the idea, and how it might drive visitors to read other sections they'd otherwise have skipped over. Doesn't sound like it'd add too much burden, besides needing to add a couple extra design elements to the homepage and chapter templates.

What would be a good next step here @natedame? Do you know of any sites doing something similar, so we could make sure we're all on the same page?

rviscomi:
How is this much different than the featured chapter call-out on the home page? Or is this something added to the chapter itself?

natedame:
[...]

PR #1018, opened 2020-07-13
Add unique id for each section to enable enhanced link tracking

bazzadp:
Adds a unique id to each home page section and enables [enhanced link attribution](https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-link-attribution), which allows us to see what bits of the home page people are actually clicking through the [Page Analytics Chrome Extension](https://support.google.com/analytics/answer/6047802?visit_id=637302587569733211-4025833548&rd=1#enable) as discussed in https://github.com/HTTPArchive/almanac.httparchive.org/issues/1016#issuecomment-657687326. Note this extension is deprecated but still works.

@rviscomi can you enable this in Google Analytics Admin screen? It's under the Property->Property Settings about half way down.

Enabling ELA in GA

### DIFF
--- a/src/templates/base.html
+++ b/src/templates/base.html
@@ -20,7 +20,9 @@
               dataLayer.push(arguments);
           }
           gtag('js', new Date());
-          gtag('config', 'UA-22381566-3');
+          gtag('config', 'UA-22381566-3', {
+            'link_attribution': true
+          });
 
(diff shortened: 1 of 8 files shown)


rviscomi:
Sure. Are these IDs all within 3 levels of the links you want to track?

bazzadp:
Good point! they are now with the latest commit.

PR #1019, opened 2020-07-13
Compress base images again

bazzadp:
Ran the home page through Lighthouse and it's complaining our images still aren't compressed enough despite what we did in #853 :

Lighthouse alert

So ran them through tinypng.com again.

We're also tantalisingly close to a perfect score on the home page, so hoping this will push us there:

Lighthouse scores for home page


bazzadp:
Good spot! We don't use it and it's not one of the [Apple recommended size](https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/app-icon/) so just removed it.

PR #1024, opened 2020-07-14
2020 contributors

### LINKED ISSUE #1020: Publish the 2020 contributors page
I'd love to see the `/<lang>/2020/contributors` page published so we can highlight everyone who has already signed up. I think it would also be really interesting from the reader's perspective to get an early glimpse of this year's author lineup, and maybe encourage new contributors to join.

I'm open to ideas for what to do with the "join the team" link. In the 2019 version we can remove it since we're no longer collecting form submissions, redirect the link to come to this repo's README, or something else. For the 2020 page, it can link to the README.

The link to the page should be added to the header/footer and the page itself should reuse the same template and translations from 2019.

bazzadp:
Closes #1020 

92 contributors at the mo, compared with 92 in 2019.

@borisschapira / @ksakae1216 could you check my translation of the new "Leads" team name?

### DIFF
--- a/src/templates/fr/2019/base.html
+++ b/src/templates/fr/2019/base.html
@@ -163,7 +163,7 @@ Sauf indication contraire, les statistiques des 20 chapitres du Web Almanac prov
   "designers": "Design",
   "developers": "Développement",
   "editors": "Édition",
-  "leads": "Chef",
+  "leads": "Gestion de projet",
   "reviewers": "Relecture",
   "translators": "Traduction",
   }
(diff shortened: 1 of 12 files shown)


src/templates/fr/2019/base.html
```diff
@@ -163,6 +163,7 @@
   "designers": "Design",
   "developers": "Développement",
   "editors": "Édition",
+  "leads": "Chef",
```
borisschapira, line 166: In French, "Chef" means "Male lead".
I'd rather use "Gestion de projet" (aka. Project Management) to be more inclusive (no need to apply a gender to the job title).

borisschapira: FYI: this transformation of job titles into thematics is already done for the other jobs.

bazzadp: Fixed! Thanks.

Issue #1015, opened 2020-07-13
Swap Year and Language selectors?


bazzadp:
While looking at #986 , @sudheendrachari has suggested we swap the Year and Language selectors.

There are some advantages to this:
- Language is longer so can lead to concerns of going off screen, especially with the “Help Translate” option.
- More consistent with our URLs

On down side, it does remove the Year away from the other content links (ToC, Contributors and Methodology). I would expect Language Spector to be used the least.

If we do go for this change, I’d prefer to do it sooner rather than later before people get too used to where it is now.

Any thoughts? Will put voting buttons below so plus 1 them depending on your opinion.

bazzadp:
Swap them around so Language and then Year

bazzadp:
Leave them in the order they currently are (Year then Language).

rviscomi:
No strong opinion. If there's an advantage to one way or the other, let's go with it! :)

bazzadp:
Now I see this in #1026 I'm less convinced. I really think the year belongs with ToC/contributors/Methodology and Language should be on far right.

What do you think @rviscomi ?

@sudheendrachari how much of an issue do you think it will be to leave it as is?

sudheendrachari:
We can surely try to leave it as is, and still try to style the dropdowns.
So to address #986 I tried few permutations of designs to see which fits best
 - Since language will be last item, we need to think about how we align drop down for it, considering they can be of variable width, in different languages. Couple of options here:
1. Align left edge of drop down to left edge of language button, and keep the dropdown width as fixed (The width can be pre-calculated since the distance of right edge of language button from viewport edge is known in advance, will be same for all languages more or less) This means text in it might wrap to next line if needed
<img width="782" alt="Screenshot 2020-07-14 at 10 38 46 PM" src="https://user-images.githubusercontent.com/10291571/87455273-cedd0380-c622-11ea-867d-1769e887fc91.png">

2. Align left edge of drop down to left edge of language button, but set the width such that it occupies maximum space till the right edge of the screen (barring few px, maybe) Since the dropdown will be positioned absolutely, i don't see a way to set variable width without using js (which will be only invoked once)
_Will look similar to approach 1 screenshot_

3. Align right edge of dropdown to right edge of language button, and let dropdown occupy width as needed while expanding to the left

**With text left aligned**

[...]

Issue #986, opened 2020-07-08
Style dropdown lists


rviscomi:
Related to #985, existing dropdown lists like the year and language pickers use `select` and `option` elements to represent list items. These are styled natively by the UA:

![image](https://user-images.githubusercontent.com/1120896/86954544-2ad7f180-c124-11ea-940a-bd8c63fb3fae.png)

I'd like to see these better fit with the Almanac style by customizing the appearance of the dropdown lists.

@HTTPArchive/developers any ideas for how this could look or anyone interested to build it?

sudheendrachari:
I would like to give this a shot, let me come up with a prototype with styles similar to dropdown in HTTP Archive. We can take a call based on that

rviscomi:
Thanks @sudheendrachari great to have you working on this! A prototype sounds like a great first step.

sudheendrachari:
I am no designer 😁 but here is what i could imagine and style these two dropdowns into:

Major changes: 
 - Reduced the rounded corners to only 5px radius (Completely rounded corners doesn't seem to fit well into dropdown style layout)
 - Instead of chevron added relevant icons for both
 - Instead of border, used background color to differentiate from other items in menu


Version 1:  (Trying to fit into the yellow color from the theme of home page)

<img width="1680" alt="version-1" src="https://user-images.githubusercontent.com/10291571/87249396-7f5cd300-c47c-11ea-8802-8dbc656f0ee2.png">
Version 2: (trying to fit into gray background of the header and footer)

Note:
 - List is supposed to be visible on click/hover
 - _2019_ and _Español_ are in _**hover**_ state
<img width="1680" alt="version-2" src="https://user-images.githubusercontent.com/10291571/87249433-aadfbd80-c47c-11ea-9684-a5c18238b1a0.png">

@rviscomi What are your thoughts on these? Also please add anyone else into the thread, would love to get some design reviews on this attempt 😁

bazzadp:
Thanks for those mockups @sudheendrachari 

My feedback:
- Agree with removing the rounded corners for the reason you give. Also solves #892 
- Not sure I agree with replacing the chevron with icons. The lead to misunderstanding. The date icon makes me think it’s a date picker rather than a year picker and not sure what the language picker icon is? I think the chevron is simple and well understood.
- Not sure I like the yellow. To me these are secondary call to actions and think yellow should be for primary CTAs.
[...]

PR #1026, opened 2020-07-14
Swap language and year dropdown in nav

sudheendrachari:
This change is related to #1015 in effort to swap the positions of language and year
selector dropdowns in nav

This is sort of a prerequisite for #986 where we attempt to style these dropdowns

Verified the change in header, footer and mobile view

Screenshot 2020-07-14 at 7 42 52 PM

Screenshot 2020-07-14 at 7 55 11 PM

Screenshot 2020-07-14 at 7 54 49 PM

### DIFF
--- a/src/templates/base/2019/base.html
+++ b/src/templates/base/2019/base.html
@@ -117,10 +117,10 @@
           <ul>
             {{ self.non_chapter_nav_links() }}
             <li>
-              {{ year_switcher('header') }}
+              {{ language_switcher('header') }}
             </li>
             <li>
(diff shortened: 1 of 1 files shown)


bazzadp:
Thanks for this but as per https://github.com/HTTPArchive/almanac.httparchive.org/issues/1015#issuecomment-658243012 I'm no longer convinced we should do this now I see it in action. Let's discuss over on that issue and then either close this or merge it depending on the discussion there.

bazzadp:
Closing as per discussion in #1015

PR #1027, opened 2020-07-14
Generate non-chapter pages in Sitemap based on JSON config

### LINKED ISSUE #1021: 2020 index page not in sitemap
The Sitemap depends on at least one chapter page being available before it will add the non-chapter pages like the 2020 home page and the contributor page we are about to add in #1020 

We should look to fix this.

bazzadp:
Fixes #1021 to add 2020 pages to sitemap


bazzadp:
Only because they don't exist in this branch! Will be included as soon as this and #1024 are both merged and `npm run generate` automatically runs again as part of our usual GitHub actions.

bazzadp:
Just merged and regenerated to be sure and yup, they are in there now!

PR #1029, opened 2020-07-14
CSP for Sitemap

bazzadp:
In #1010 we removed `unsafe-inline` from our `style-src` in our Content Security Policy.

However, annoyingly [Chrome uses inline styles to display XMLs files](https://bugs.chromium.org/p/chromium/issues/detail?id=924962) so they look busted with this, which caused me a mild panic when about to release the site! Seems to affect Safari too but not Firefox.

It's only a display issue but still - it's annoying, so this PR overrides the default CSP (including turning off nonce otherwise `unsafe-inline` is ignored) to the minimal necessary CSP to allow sitemap to display properly.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -247,6 +247,11 @@ def sitemap():
     xml = render_template('sitemap.xml')
     resp = app.make_response(xml)
     resp.mimetype = "text/xml"
+    # Chrome and Safari use inline styles to display XMLs files.
+    # https://bugs.chromium.org/p/chromium/issues/detail?id=924962
+    # Override default CSP (including turning off nonce) to allow sitemap to display
+    talisman.content_security_policy_nonce_in=[]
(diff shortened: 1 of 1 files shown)


rviscomi:
Where do we use XML files? Or would you happen to have a screenshot of the effects?

bazzadp:
Just in sitemap.xml as per title.

Screenshot below:

Broken sitemap

And this wasn't the complete fix - need another fix on top of this.

rviscomi:
Does this affect users or search engines or is it mostly an inconvenience for us when we manually check the sitemap?

bazzadp:
Just inconvenience for us. But easy enough to override on a route specific basis. Which is what I have done now.

rviscomi:
Ok thanks for clarifying (and fixing!)

bazzadp:
> Ok thanks for clarifying (and fixing!)

You break it, you fix it.

PR #1039, opened 2020-07-15
Reorganise HEAD to load CSS as quickly as possible

bazzadp:
Happened to notice that the Web Vitals script we added in #944 is the first thing loaded in our page:

WebPageTest loading of home page

Now it's not a big script, but still - it's not used until after the page loads so definitely doesn't need to be first. It is also `async` when I think `defer` is sufficient [and is recommended](https://github.com/GoogleChrome/web-vitals#load-web-vitals-from-a-cdn).

This made me have a bigger look at our `<HEAD>` and seems all in the wrong order. So reorganised it a bit better.

I've left Google Analytics call in the head as per advice from Google, but not 100% convinced it needs to be there, and tempted to move it to the end of the page. Still, as it requires another connection, and that takes time, it doesn't really change page load.

### DIFF
--- a/src/templates/base.html
+++ b/src/templates/base.html
@@ -11,7 +11,7 @@
     {% endblock %}
     <link rel="shortcut icon" href="/static/images/favicon.ico">
     <link rel="apple-touch-icon" href="/static/images/apple-touch-icon.png">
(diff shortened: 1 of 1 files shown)


bazzadp:
Going to merge this as simple enough reorg, and it passes HTML validation (both linting and nu validator).

PR #1041, opened 2020-07-16
Show team members in the tooltip

ibnesayeed:
* Refactor to DRY the code
* Remove an unnecessary exclusion check
* Add team member usernames in the tooltip on each number

Note: Team tooltips should be visible in #989 once merged.

### DIFF
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -82,7 +82,7 @@ jobs:
                 sql: '🔍',
                 results: '📊'
               }
-              return Object.entries(links).filter(l => !['', '#', sqlBaseDir].includes(l[1])).map(l => `[${icons[l[0]] || l[0]}](${l[1]})`).join(' ')
+              return Object.entries(links).filter(l => !['#', sqlBaseDir].includes(l[1])).map(l => `[${icons[l[0]] || l[0]}](${l[1]})`).join(' ')
             }
 
             const formatTeam = team => {
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -59,14 +59,18 @@ jobs:
               return links
             }
 
+            const parseUserNames = members => {
+              return [...members.matchAll(/@(?<username>\w+)/g)].map(u => u.groups.username)
+            }
+
             const parseTeam = text => {
               const team = {}
               text.split('\n').forEach(line => {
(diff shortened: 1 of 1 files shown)


bazzadp:
Nice little tweak. Could we take it to the next level and tooltip the milestones too? Based on step names in the main content issue?

ibnesayeed:
> Could we take it to the next level and tooltip the milestones too? Based on step names in the main content issue?

That does not add any value to the report as the information is available in a more accessible, persistent, and visible manner beneath the table. Title-based tooltips are not accessible on mobile devices. If we try to parse all the milestones from the reporting issue itself, we are mandating it to be there in the first place, this means more assumptions that will break in the future.

In the future, if we were to revamp this system, we can perhaps store all this information in a YAML/JSON file to avoid brittle parsing and can generate summary output or bootstrap/update chapter issues as content changes in that file.

bazzadp:
> > Could we take it to the next level and tooltip the milestones too? Based on step names in the main content issue?
> 
> That does not add any value to the report as the information is available in a more accessible, persistent, and visible manner beneath the table.

Yeah would just save a bit of scrolling up and down that's all

> Title-based tooltips are not accessible on mobile devices.

OK, but that is an argument against this PR too. This is a progressive enhancement - as you say the information is there already (though not for team members).

> If we try to parse all the milestones from the reporting issue itself, we are mandating it to be there in the first place, this means more assumptions that will break in the future.

Which contradicts your argument that it adds no value!

> In the future, if we were to revamp this system, we can perhaps store all this information in a YAML/JSON file to avoid brittle parsing and can generate summary output or bootstrap/update chapter issues as content changes in that file.

I'd be OK hardcoding these in the action now. Tiny bit extra config to update next year but much less brittle, and much easier to code for you.

Anyway, not a big deal and just a thought.

ibnesayeed:
If we were to rely on the milestones listed in the progress reporting post itself then we need to add a little bit more structure. For example, begin and end comments for milestones listing to avoid any other list item be counted as a milestone. Also, this will become mandatory for the system to function. In that case, we can remove `TOTAL_TRACKED_TASKS` config and rely on counting tasks listed in the reporting issue itself as the ultimate truth.

I honestly do not feel we should add unnecessary complications for the little gain. This PR on the other hand surfaces information up that was buried in many different issues so it sounded a more worthy addition, though not very accessible, but we can count on maintainers using non-mobile devices.

PR #1042, opened 2020-07-16
Make Contributor buttons accessible

### LINKED ISSUE #1040: Improve accessibility of contributors selector
The contributors selector is not accessible to screen readers:

Contributors team selector screenshot

I see the following problems:
- The filter buttons only used colour to indicate state. They should use `aria-pressed` or similar.
- They do not announce any update to the contributors section using `aria-live` or similar.
- It may be obvious to sighted users as to the state of multiple filters but this is less obvious to non-sighted users. Should we have a summary of all filters available to screen-readers only? Or available to all?

The rest of the site has very good accessibility as far as we are aware and we have a strong [Accessibility Statement](https://almanac.httparchive.org/en/accessibility-statement) that we try our best to honour so we need to fix this.

### LINKED ISSUE #1045: Sort contributor teams by locale
I noticed the list of teams is sorted alphabetically in [English](https://almanac.httparchive.org/en/2020/contributors):
image

In other languages like [French](https://almanac.httparchive.org/fr/2020/contributors), the list is still sorted using the English spelling:
image

Should the list of teams be sorted according to the localized team names?

bazzadp:
Fixes #1040
Fixes #1045

This also makes the 2020 contributors template an extension of 2019 template to avoid duplication of code.

@nico3333fr could you check I've implemented what you mean? And also the French translation?
I've deployed a test version here to make that easier for you to test: https://20200716t131108-dot-webalmanac.uk.r.appspot.com


src/templates/base/2019/contributors.html
```diff
@@ -404,8 +421,22 @@ <h1 class="title title-lg">
...
                 contributors.classList.add(...filterTeams);
             }
+
+            for (index = 0; index < buttons.length; ++index) {
+              var button = buttons[index];
+              if (hash.includes(button.classList)) {
+                button.setAttribute('aria-pressed',true);
+              } else {
+                button.setAttribute('aria-pressed',false);
+              }
```
rviscomi, line 431: Set the attribute's value directly to the boolean. Also compare using `className` (string) rather than `classList` (DOMTokenList).
```suggestion
              button.setAttribute('aria-pressed', hash.includes(button.className));
```

bazzadp:
BTW one weird thing that I noticed about this - and update since I posted above demo link.

What should the initial state of the button be?

Originally I had "pressed" since they are all yellow. However then when you press it (to select just this team), it doesn't change state since it is still press - which is weird that pressing the button doesn't change state.

So I changed it to all be "unpressed" initially, which makes more semantic sense to me when you do press a button. However it does create a clash with the colouring.

Maybe initially colouring should all be white rather than all yellow? Thoughts?

bazzadp:
Planning a release so going to go ahead and merge this. I think it's good, and certainly an improvement at very least.

Would love to hear your feedback though @nico3333fr and @ericwbailey if there are any further improvements we can make here.

src/templates/base/2019/contributors.html
```diff
@@ -262,27 +271,37 @@ <h1 class="title title-lg">
   </div>
   <div class="contributors {{ ' '.join(config.teams.keys()) }}">
     <div class="filter-container">
-      <p>
+      <p aria-live="polite" aria-atomic="true">
```
ericwbailey, line 274: Clever. I like it.

src/templates/base/2019/contributors.html
```diff
@@ -262,27 +271,37 @@ <h1 class="title title-lg">
   </div>
   <div class="contributors {{ ' '.join(config.teams.keys()) }}">
```
ericwbailey, line 272: What do you think about using a `form` element here, or using `role="form"`? I am thinking this would expose the filtering functionality as a landmark and make it potentially easier to discover.

src/templates/base/2019/contributors.html
```diff
@@ -262,27 +271,37 @@ <h1 class="title title-lg">
...
+      <p aria-live="polite" aria-atomic="true">
         {{ self.filter_by_team() }}
       </p>
-        {% for id, team in config.teams.items() %}
-          <button class="{{ id }}" type="button">{{ localizedTeamNames[id] if localizedTeamNames[id]|length else team.name }}</button>
+      <ul>
+        {% for id, team in localizedTeamNames.items() | accentless_sort() %}
+        {% if id in config.teams %} <!-- note not all team translations may be in this years config -->
+        <li>
+          <button class="{{ id }}" type="button" aria-pressed="false" title="{{ self.filter_by() }} {{ localizedTeamNames[id] if localizedTeamNames[id]|length else team.name }}">
```
ericwbailey, line 281: Great use of `aria-pressed` 😄 

Out of scope of this PR, but one thing I really like doing is tying visual presentation to semantic state. `[aria-pressed]` makes a great styling hook.

src/templates/base/2019/contributors.html
```diff
@@ -262,27 +271,37 @@ <h1 class="title title-lg">
...
-            {{ self.join_the_team_text() }}
-          </text>
-        </svg>
-        <img src="/static/images/avatars/12.jpg" alt="" height="100" width="100" loading="lazy" />
-      </a>
+    <ul class="contributor-grid">
+      <li>
+        <a href="https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/README.md" class="contributor-join" aria-labelledby="jointheteam">
+          <svg xmlns="http://www.w3.org/2000/svg" role="img" width="174" height="100" viewBox="0 0 203 117">
+            <title id="jointheteam">{{ self.join_the_team_title() }}</title>
```
ericwbailey, line 294: I think this is a correct usage pattern, but we might want to check for a double-announce or mis-announce when embedded SVG is used to provide a label. I personally prefer the visually-hidden CSS technique, as it allows things like browser translation.

src/templates/base/2019/contributors.html
```diff
@@ -262,27 +271,37 @@ <h1 class="title title-lg">
...
+            <text x="50%" y="47%">
+              {{ self.join_the_team_text() }}
+            </text>
+          </svg>
+          <img src="/static/images/avatars/12.jpg" alt="" height="100" width="100" loading="lazy" />
+        </a>
+      </li>
     {% for id, contributor in config.contributors.items() %}
-      <div class="contributor {{ ' '.join(contributor.teams) }}" id="{{ id }}">
+      <li class="contributor {{ ' '.join(contributor.teams) }}" id="{{ id }}">
```
ericwbailey, line 304: Nice catch!

src/templates/base/2019/contributors.html
bazzadp: Will have a look in follow up PR next week.

src/templates/base/2019/contributors.html
bazzadp: All @nico3333fr’s idea so full credit to him!

src/templates/base/2019/contributors.html
bazzadp: Yeah I started refactoring the CSS/JS but was growing in scope so decided to keep it as is for this PR for now so could concentrate on the accessibility problem. Agree we should come back to this later though.

src/templates/base/2019/contributors.html
bazzadp: Yeah this is based on usage pattern 8 from [Deque’s Accessible SVG post](https://www.deque.com/blog/creating-accessible-svgs/), which seems the best, even if it does cause a double announcement in at least one case. Do browsers not translate SVG components like the `<title>` tag here?

src/templates/base/2019/contributors.html
bazzadp: Yeah except I caused a bug 😔
https://github.com/HTTPArchive/almanac.httparchive.org/issues/1058

bazzadp:
Thanks for the great review @ericwbailey! And yeah like the test deploy to help review implementation rather than code. Though in this case I ended up merging before you review so you could have tested on the live site!

Any thoughts on my [comment above](https://github.com/HTTPArchive/almanac.httparchive.org/pull/1042#issuecomment-659657277) on the default state being slightly weird (it’s the only time yellow is used to mean not pressed). When I ended up deploying to the live site was actually different to the test link.

src/templates/base/2019/contributors.html
ericwbailey: As I understand it, anything passed through `aria-*` has the chance of not being picked up by browser/3rd party translation (like, I18n-style), as it's an attribute. I usually err on the side of avoiding it because of that.

ericwbailey:
Oop, forgot to address that. I like your idea of all white to begin with.

src/templates/base/2019/contributors.html
bazzadp: But it’s `aria-labelledby` not `aria-label` so points to an element for that very reason, so bit confused what you are saying? It’s an SVG element rather than an HTML element but I *think* that’s still OK.

bazzadp: And it’s a `<title>` SVG element not a `title` attribute. Is that the cause of confusion because you didn’t notice that and thought it was the attribute?

ericwbailey: Nope, just involving the ARIA attribute in the chain. Again, I don't think it would be a blocking merging issue if the issue was still open, just a corner case usability consideration.

PR #1048, opened 2020-07-17
2020 Config Updates

bazzadp:
Did another round of updates. Will get people to mind this themselves in future, but with so much in flux, and unfortunately having to remove some people as chapters were pulled, thought it was worth doing one more big sweep.

### DIFF
--- a/src/config/2020.json
+++ b/src/config/2020.json
@@ -50,8 +50,8 @@
         {
           "part": "I",
           "chapter": "7",
-          "title": "WebAssembly",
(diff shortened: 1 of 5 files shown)


bazzadp:
98 contributors now:
- 33 authors
- 62 reviewers
- 16 analysts
- 8 developers (haven't add them all until we get a code commit)
- 4 editors
- 4 leads
- 2 translators (haven't add them all until we get a code commit)
- and a partridge in a pear tree!

PR #1049, opened 2020-07-17
Update 2020.json

nrllh:
added leads to my profile


bazzadp:
@nrllh this is the overall project leads (myself, @rviscomi , @obto and @paulcalvano ) not individual chapter leads. Probably not the clearest admittedly...

PR #1050, opened 2020-07-17
Fix bug where code block causes scrolling issues on mobile

bazzadp:
Noticed an overflow issues on mobile for code blocks:

Code snipper overflowing text on mobile

Looks to have been caused by #939 where I removed duplicate code that wasn't actually duplicate!

This code removed:

Code snippet with an all important comma

Is not same as code, that I thought was duplicate!

Same code without comma

https://github.com/bazzadp/almanac.httparchive.org/blob/c0eb5425d44b353bd7a01cbc427f9348a011f88a/src/static/css/page.css#L580-L585

### DIFF
--- a/src/static/css/page.css
+++ b/src/static/css/page.css
@@ -643,7 +643,7 @@ figure .fig-desktop {
   left: -1rem;
 }
 
-.code-block pre {
+.code-block, pre {
   overflow-y: hidden;
   overflow-x: auto;
   margin: 8px 0;


src/static/css/page.css
```diff
@@ -643,7 +643,7 @@ figure .fig-desktop {
   left: -1rem;
 }
 
-.code-block pre {
+.code-block, pre {
```
rviscomi, line 646: One pattern I usually do is to separate selectors by a newline. We could enforce this in the linter if needed.
```suggestion
.code-block,
pre {
```

bazzadp: Always thought this was overkill. This bug has convinced me otherwise!

Raised #1053

rviscomi: Especially for selectors that are very similar, it's helpful to scan through and see the changes line up vertically, for example:

```css
h1 + .subtitle,
h2 + .subtitle,
h3 + .subtitle {
```

Issue #946, opened 2020-07-03
Finalize chapter sequence/organization


rviscomi:
Some chapters may need to be split/merged/dropped/added depending on changes in content brainstorming. For now, here is the list of candidate chapters for the 2020 edition:

### I. Page Content

1. CSS
2. JavaScript
3. Markup
4. Fonts
5. Media
6. Third Parties

### II. User Experience

7. SEO
8. Accessibility
9. Performance
10. Privacy
11. Security
12. Mobile Web
13. PWA

### III. Content Publishing
14. CMS
15. Ecommerce
16. JAMstack

### IV. Content Distribution
17. Page Weight
18. Compression
19. Caching
20. Resource Hints
21. HTTP/2

rviscomi:
@obto how does this sequencing look to you? If good could you update the issues' metadata in our sheet and in their issues?

@ibnesayeed is it possible to match this sequence in #989?

bazzadp:
Presume we’d need to do one of below to allow ordering as currently think it’s ordered by issue number:

1. Rename this issue titles (e.g. from “JavaScript” to “1.1 JavaScript”)
Or
2. Add meta data to each issue (either as a hidden tag or not)
Or
3. Add config with the correct ordering to the GitHub Action

rviscomi:
I imagined being able to parse the chapter number from the H1 in each chapter's comment. For example: `/^#.*Chapter (\d+):/`. Some are still `Chapter [Num]` so we'd need to update them first.

bazzadp:
Completely missed that!

ibnesayeed:
> @ibnesayeed is it possible to match this sequence in #989?

Yes! However, the more I think about it, the more I am convinced that in the future we should move to some structured data file like YAML to store these properties and use issue tickets to communicate. Changes in the data file can be pushed to change the issue ticket, not the other way.

> ... as currently think it’s ordered by issue number:

That's correct. The default order of the API response is descending order of issue numbers, but I have explicitly asked it to return in ascending order.

> I imagined being able to parse the chapter number from the H1 in each chapter's comment. For example: `/^#.*Chapter (\d+):/`. Some are still `Chapter [Num]` so we'd need to update them first.

I was worried that chapter counter might reset under each new part, which would have been a little more work because roman numerals are not sort-friendly unless we convert them to integers.

> Completely missed that!

You are not alone! I too was thinking about updating the issue title. ;-)

rviscomi:
[...]

PR #1059, opened 2020-07-18
Refactoring, sorting, and enhancing progress report

ibnesayeed:
* Refactors the code
    * Moves code outside of functions that were constants, but unnecessarily getting reinitialized
    * Builds nested objects after parsing issues then serializes them instead of serializing as processing
* Reads tracked tasks/milestones from the reporting issue itself
    * No need to set `TOTAL_TRACKED_TASKS` config
    * Must include a list of tasks surrounded by `<!-- TASKS BEGIN -->` and `<!-- TASKS END -->` comments in the reporting post (#989 is updated already with this change)
    * Adds tooltip to each numbered column with milestone description
* Parses heading from the content to extract Part Number, Chapter Number, and Title
    * Uses extracted title, but falls back to issue title (earlier trailing year was removed, but not anymore, as that sounded a hacky approach)
    * Prefixes chapter number in front of the title in the table cell and adds a tooltip on it to show Part and Chapter numbers
* Sorts chapters according to their chapter numbers (as per #946)
    * Currently, unumbered chapters will bubble on top
* Adds tooltip to Links and Teams table column heads (eliminates the need of manual legend placement, as is the case in #989 currently)
    * It is better because things will reflect automatically if, say, icons change
* Adds chapter-wise statistical summary
    * Which includes Total, Min, Max, Mean, and Median number of contributors per chapter
* Lists all the team members in sorted order under each team
    * Adds number of chapters they are contributing to with corresponding role
    * Adds all the chapters in tooltip on each member's username

### DIFF
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -213,7 +213,7 @@ jobs:
               owner: context.repo.owner,
               repo: context.repo.repo,
               issue_number: trackerIssueNumber,
-              body: progressContent.replace(/<!-- REPORT START -->[\s\S]*<!-- REPORT END -->/, `<!-- REPORT START -->\n${report}\n\n${stats}\n\n<details><summary>Contributors</summary>\n\n${teams}\n</details>\n<!-- REPORT END -->`)
+              body: progressContent.replace(/<!-- REPORT START -->[\s\S]*<!-- REPORT END -->/, `<!-- REPORT START -->\n${report}\n\n${stats}\n\n<details><summary><b>Contributor Details</b></summary>\n\n${teams}\n</details>\n<!-- REPORT END -->`)
             })
 
             core.info(`Updated report in issue #${trackerIssueNumber} with ${chapters.length} chapters`)
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -28,6 +28,18 @@ jobs:
             const totalTrackedTasks = parseInt(process.env.TOTAL_TRACKED_TASKS, 10)
             const sqlBaseDir = process.env.SQL_BASE_DIR
 
+            const icons = {
+              draft: '📄',
+              sql: '🔍',
+              results: '📊'
+            }
+
+            const linkMap = {
+              draft: /^\[~draft-doc\]:\s*(?<link>\S+)/,
+              sql: /^\[~sql-dir\]:\s*(?<link>\S+)/,
+              results: /^\[~results-sheet\]:\s*(?<link>\S+)/
+            }
+
             const parseTasks = text => {
               const tasks = []
               text.split('\n').forEach(line => {
@@ -44,11 +56,6 @@ jobs:
             const parseLinks = text => {
               const links = {}
               text.split('\n').forEach(line => {
-                const linkMap = {
-                  draft: /^\[~draft-doc\]:\s*(?<link>\S+)/,
-                  sql: /^\[~sql-dir\]:\s*(?<link>\S+)/,
-                  results: /^\[~results-sheet\]:\s*(?<link>\S+)/
-                }
                 for (const [k, v] of Object.entries(linkMap)) {
                   const m = line.match(v)
                   if (m) {
@@ -57,6 +64,11 @@ jobs:
                 }
               })
               return links
+            }
+
+            const parseHeading = text => {
+              const m = text.match(/^#\s+Part\s+(?<part>\S+)\s+Chapter\s+(?<number>\S+)\s*:\s*(?<title>.+)/)
+              return m ? m.groups : {part: '', number: '', title: ''}
             }
 
             const parseUserNames = members => {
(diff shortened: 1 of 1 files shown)


ibnesayeed:
Updated first post with the description of recent changes regarding tracked tasks parsing.

rviscomi:
Is there a way to preview what the output of this would be before merging?

ibnesayeed:
> Is there a way to preview what the output of this would be before merging?

Please have a look at https://github.com/bazzadp/almanac.httparchive.org/issues/73 which does not have many different cases as the number if issues are fewer there, but you will get an idea. Do note that there are tooltips added in almost everything.

.github/workflows/progress-tracker.yml
```diff
@@ -77,55 +120,75 @@ jobs:
...
-              }
               return Object.entries(links).filter(l => !['#', sqlBaseDir].includes(l[1])).map(l => `[${icons[l[0]] || l[0]}](${l[1]})`).join(' ')
             }
 
             const formatTeam = team => {
               return Object.entries(team).map(([k, v]) => `<span title="${k}: ${v.length ? v.join(', ') : 'TBD'}">${v.length}</span>`).join('/')
             }
 
+            const formatRow = chapter => {
+              return `${chapter.number ? `<span title="Part: ${chapter.part}, Chapter: ${chapter.number}">${chapter.number}</span> - ` : ''}[${chapter.title}](${chapter.url}) | ${formatLinks(chapter.links)} | ${formatTeam(chapter.team)} | ${chapter.tasks.join(' | ')}`
```
rviscomi, line 131: Just a small tweak to the delimiter
```suggestion
              return `${chapter.number ? `<span title="Part: ${chapter.part}, Chapter: ${chapter.number}">${chapter.number}</span>. ` : ''}[${chapter.title}](${chapter.url}) | ${formatLinks(chapter.links)} | ${formatTeam(chapter.team)} | ${chapter.tasks.join(' | ')}`
```

.github/workflows/progress-tracker.yml
```diff
@@ -77,55 +120,75 @@ jobs:
...
+                })
+              })
             })
 
+            const report = [
+              `Chapters | <span title="${Object.entries(icons).map(([k, v]) => `${v} (${k})`).join(', ')}">Links</span> | <span title="${Object.keys(teamMembers).join('/')}">Team</span> | ${trackedTasks.map((v, i) => `<span title="${v}">${i + 1}<span>`).join(' | ')}`,
+              `:--------|:------|:----:${'|:-:'.repeat(totalTrackedTasks)}`
+            ].concat(chapters.map(formatRow)).join('\n')
+
+            const teams = Object.entries(teamMembers).map(([k, v]) => {
```
rviscomi, line 183: With ~100 contributors, I think this section would get too unwieldy, and it's also not necessary to "@" notify each contributor.

I still like the idea of showing some aggregate stats. Could we do something like this instead?

### Contributor stats

  | Authors | Reviewers | Analysts
-- | -- | -- | --
Total | 30 | 45 | 20
Min | 1 | 1 | 0
Max | 3 | 9 | 2
Median | 2 | 3 | 1
Average | 2.5 | 5 | 1.1

ibnesayeed: > With ~100 contributors, I think this section would get too unwieldy

I thought about it, but I decided to add it anyway because there is one piece of information that is not readily available anywhere, "in how many chapters a specific member is involved in and in what ways?", which I think is an important thing to know to properly distribute workload.

> and it's also not necessary to "@" notify each contributor.

That's a very good point, I will certainly remove auto-mention feature.

> I still like the idea of showing some aggregate stats. Could we do something like this instead?

Summarized statistical table is a great idea, but we can perhaps do it in a separate iteration (due to time constraints at my end).

ibnesayeed: @rviscomi I think I have taken care of the issue of getting too unwieldy by hiding the list under a <details> element, so it will be there, but won't bother.

bazzadp: Rather than making the contributors a `<details>` element could you not make each sub-team the `<details>` element? So can always see the team counts but can expand for names in a team.

ibnesayeed: > Rather than making the contributors a `<details>` element could you not make each sub-team the `<details>` element? So can always see the team counts but can expand for names in a team.

We can do it, but I think the current approach is more friendly in some ways (I am open to be convinced otherwise):

* It is a compromise between clicking vs. scrolling, and right now individual contributor usernames are not in a separate line as bullet points, so I do not thing there will be too much scrolling (perhaps we will have a better idea of how much space it takes once it is merged here and reports are generated).
* Currently, contributors report is composed as "h3" headings for each team name followed by a paragraph of comma separated usernames, which is independent of how it is finally placed in the report, i.e., only at the very end we decide to place it in a "details" container (which can be something else too), but if we decide it otherwise, "details" and "summary" elements will be engraved during the iteration over the data.

It may also depend on how the information is used often, a specific role at a time or a specific person (across rolls) at a time.

ibnesayeed: @rviscomi I have added stats table as well. Please have a look at https://github.com/bazzadp/almanac.httparchive.org/issues/73 for an example.

PR #1060, opened 2020-07-19
Support hyphens in usernames and sort contributors by most to least chapters

rviscomi:
I noticed the analyst "max" showing up, but I know the full username should be "max-ostapenko". The `parseUserNames` function should support usernames with hyphens.

I also wanted to change the sort order of the contributors per team from alphabetical to number of chapters, so it's easier to see who is contributing the most. Changing this part was difficult because of the single character variable names and long lines, so I also fixed those things to make it more maintainable.

I also made a fix and improvement to the median calculation. It should take the middlemost value when the length is divisible by two. I also removed the `toFixed(2)` so the values appear as either whole numbers, or end in .5 (the only two possibilities).

### DIFF
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -75,14 +75,14 @@ jobs:
               if (!nums.length) {
                 return [0, 0, 0, 0, 0]
               }
-              nums.sort((a, b) => a - b)
+              nums.sort()
               const half = Math.floor(nums.length / 2)
               return [
                 nums.reduce((a,b) => a + b, 0),
                 Math.min(...nums),
                 Math.max(...nums),
                 (nums.reduce((a,b) => a + b, 0) / nums.length).toFixed(2),
-                (nums.length % 2 ? nums[half] : (nums[half - 1] + nums[half]) / 2.0).toFixed(2)
+                (nums.length % 2 == 0 ? nums[half] : (nums[half - 1] + nums[half]) / 2.0)
               ]
             }
 
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -118,7 +118,7 @@ jobs:
             }
 
             const parseUserNames = members => {
-              return [...members.matchAll(/@(?<username>\w+)/g)].map(u => u.groups.username)
+              return [...members.matchAll(/@(?<username>[\w-]+)/g)].map(u => u.groups.username)
             }
 
             const parseTeam = text => {
@@ -205,8 +205,13 @@ jobs:
               ':-----------------|------:|----:|----:|-----:|------:'
             ].concat(Object.entries(teamStats).map(([k, v]) => `${k} | ${v.join(' | ')}`)).join('\n')
 
-            const teams = Object.entries(teamMembers).map(([k, v]) => {
-              return `### ${k} (${Object.keys(v).length})\n\n${Object.entries(v).sort().map(([m, c]) => `<span title="Chapters: ${c.join(', ')}">${m} (${c.length})</span>`).join(', ')}`
+            const teams = Object.entries(teamMembers).map(([team, contributors]) => {
+              const contributorList = ${Object.entries(contributors).sort(([x, a], [y, b]) => {
+                return b.length - a.length
+              }).map(([name, chapters]) => {
+                return `<span title="Chapters: ${chapters.join(', ')}">${name} (${chapters.length})</span>`).join(', ')
+              })
+              return `### ${team} (${Object.keys(contributors).length})\n\n${contributorList}`
             }).join('\n\n')
 
             github.issues.update({
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -205,13 +205,13 @@ jobs:
               ':-----------------|------:|----:|----:|-----:|------:'
             ].concat(Object.entries(teamStats).map(([k, v]) => `${k} | ${v.join(' | ')}`)).join('\n')
 
-            const teams = Object.entries(teamMembers).sort().map(([team, contributors]) => {
-              const contributorList = Object.entries(contributors).sort(([x, a], [y, b]) => {
+            const teams = Object.entries(teamMembers).sort().map(([team, members]) => {
+              const memberList = Object.entries(members).sort(([x, a], [y, b]) => {
                 return b.length - a.length
-              }).map(([name, chapters]) => {
-                return `<span title="Chapters: ${chapters.join(', ')}">${name} (${chapters.length})</span>`
+              }).map(([member, chapters]) => {
+                return `<span title="Chapters: ${chapters.join(', ')}">${member} (${chapters.length})</span>`
               }).join(', ')
-              return `### ${team} (${Object.keys(contributors).length})\n\n${contributorList}`
+              return `### ${team} (${Object.keys(members).length})\n\n${memberList}`
             }).join('\n\n')
 
             github.issues.update({
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -206,11 +206,11 @@ jobs:
             ].concat(Object.entries(teamStats).map(([k, v]) => `${k} | ${v.join(' | ')}`)).join('\n')
 
             const teams = Object.entries(teamMembers).sort().map(([team, contributors]) => {
-              const contributorList = ${Object.entries(contributors).sort(([x, a], [y, b]) => {
+              const contributorList = Object.entries(contributors).sort(([x, a], [y, b]) => {
                 return b.length - a.length
               }).map(([name, chapters]) => {
-                return `<span title="Chapters: ${chapters.join(', ')}">${name} (${chapters.length})</span>`).join(', ')
-              })
+                return `<span title="Chapters: ${chapters.join(', ')}">${name} (${chapters.length})</span>`
+              }).join(', ')
               return `### ${team} (${Object.keys(contributors).length})\n\n${contributorList}`
             }).join('\n\n')
 
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -75,14 +75,14 @@ jobs:
               if (!nums.length) {
                 return [0, 0, 0, 0, 0]
               }
-              nums.sort()
+              nums.sort((a, b) => a - b)
               const half = Math.floor(nums.length / 2)
               return [
                 nums.reduce((a,b) => a + b, 0),
                 Math.min(...nums),
                 Math.max(...nums),
                 (nums.reduce((a,b) => a + b, 0) / nums.length).toFixed(2),
-                (nums.length % 2 == 0 ? nums[half] : (nums[half - 1] + nums[half]) / 2.0)
+                nums[half]
               ]
             }
 
@@ -205,7 +205,7 @@ jobs:
               ':-----------------|------:|----:|----:|-----:|------:'
             ].concat(Object.entries(teamStats).map(([k, v]) => `${k} | ${v.join(' | ')}`)).join('\n')
 
-            const teams = Object.entries(teamMembers).map(([team, contributors]) => {
+            const teams = Object.entries(teamMembers).sort().map(([team, contributors]) => {
               const contributorList = ${Object.entries(contributors).sort(([x, a], [y, b]) => {
                 return b.length - a.length
               }).map(([name, chapters]) => {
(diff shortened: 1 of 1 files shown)


rviscomi:
This hasn't been tested. Do I need edit access to @bazzadp's fork to test it?

ibnesayeed:
> I noticed the analyst "max" showing up, but I know the full username should be "max-ostapenko". The `parseUserNames` function should support usernames with hyphens.

That's a good catch, I missed the fact that `-` is also a legitimate username character.

> I also wanted to change the sort order of the contributors per team from alphabetical to number of chapters, so it's easier to see who is contributing the most.

I thought about it, but I was not too sure if I would go with this because most of the people will have only a single chapter in a specific role. This means, apart from just a first few, rest of the list will look rather random, hence I decided not to sort by their contributions. However, now that you have done it, I would suggest you first sort the list by name as I was doing before, and then apply the frequency sort on top, so withing the group of the same number of contributions, usernames are sorted.

> Changing this part was difficult because of the single character variable names and long lines, so I also fixed those things to make it more maintainable.

I did realize that as the file grew organically, it became more difficult to follow and I tried to use single character names for variables that were in effect in a very limited scope, so thanks for making this change. However, if I were to decompose `teamMembers`, I would call it `[team, members]` and not `[team, contributors]` to keep the mental map of the two terms as I though about a contributor to be an umbrella term that encapsulate all contributors, but when talking about one team at a time, the association I would uses would be a member.

.github/workflows/progress-tracker.yml
```diff
@@ -75,14 +75,14 @@ jobs:
               if (!nums.length) {
                 return [0, 0, 0, 0, 0]
               }
-              nums.sort((a, b) => a - b)
+              nums.sort()
```
ibnesayeed, line 78: Please revert this change, because by default sorting is not numeric. If the number goes in two digits, you will see messy results like this:

```
[3, 11, 222].sort()
> [11, 222, 3]

File: .github/workflows/progress-tracker.yml [hunk 2]
```diff
@@ -75,14 +75,14 @@ jobs:
...
               const half = Math.floor(nums.length / 2)
               return [
                 nums.reduce((a,b) => a + b, 0),
                 Math.min(...nums),
                 Math.max(...nums),
                 (nums.reduce((a,b) => a + b, 0) / nums.length).toFixed(2),
-                (nums.length % 2 ? nums[half] : (nums[half - 1] + nums[half]) / 2.0).toFixed(2)
+                (nums.length % 2 == 0 ? nums[half] : (nums[half - 1] + nums[half]) / 2.0)
```
ibnesayeed, line 85: Removing `.toFixed(2)` from median calculation is wise, but by adding `== 0 ` in the conditional, you changed the boolean condition. Now, you are returning `nums[half]` when the length is even.

ibnesayeed:
> This hasn't been tested. Do I need edit access to @bazzadp's fork to test it?

Yes, you can copy and paste only the script portion of the code in @bazzadp's fork (leaving along top configs and step conditionals. Then trigger a run either manually or by making a change in one of the related issues.

.github/workflows/progress-tracker.yml
```diff
@@ -205,8 +205,13 @@ jobs:
               ':-----------------|------:|----:|----:|-----:|------:'
             ].concat(Object.entries(teamStats).map(([k, v]) => `${k} | ${v.join(' | ')}`)).join('\n')
 
-            const teams = Object.entries(teamMembers).map(([k, v]) => {
-              return `### ${k} (${Object.keys(v).length})\n\n${Object.entries(v).sort().map(([m, c]) => `<span title="Chapters: ${c.join(', ')}">${m} (${c.length})</span>`).join(', ')}`
+            const teams = Object.entries(teamMembers).map(([team, contributors]) => {
```
ibnesayeed, line 208: This will ensure usernames are sorted withing the group of same number of contributions.

```suggestion
            const teams = Object.entries(teamMembers).sort().map(([team, contributors]) => {
```

.github/workflows/progress-tracker.yml
rviscomi: Oh good catch. That's unexpected.

.github/workflows/progress-tracker.yml
rviscomi: Good catch! Thinking about this more, I'm ok with the median of an even number of items returning index `half` and simplifying this. It makes more intuitive sense to me to see a real value in the list than an average of the middle values, for the purpose of approximating the number of contributors, since there is no "half of a contributor". Averages can obviously be fractional.

.github/workflows/progress-tracker.yml
rviscomi: 👍

rviscomi: I think this sorts teams, not members. I moved this down before the sort by length and made it a case insensitive sort for convenience.

Also, I realized that I can test this action in this repo before merging by setting the branch in the "run workflow" manual run. You can see the results in #989.

bazzadp:
Gave you access to my repo there @rviscomi in case you wanna test without polluting this repo with lots of failed runs.

Getting a little obsessed with this? 😁

rviscomi:
I needed to test the sorting change in this repo because it has lots of contributors/chapters. I think it's ready to merge and I promise not to touch it for at least another few hours. 😇

.github/workflows/progress-tracker.yml
```diff
@@ -82,7 +82,7 @@ jobs:
                 Math.min(...nums),
                 Math.max(...nums),
                 (nums.reduce((a,b) => a + b, 0) / nums.length).toFixed(2),
-                (nums.length % 2 ? nums[half] : (nums[half - 1] + nums[half]) / 2.0).toFixed(2)
+                nums[half]
```
ibnesayeed, line 85: Why did you change the logic of median calculation? Was anything wrong in it? I had some testing locally before adding the code I did. Removing `.toFixed(2)` and surrounding parenthesis was all that could have been removed.

```suggestion
                nums.length % 2 ? nums[half] : (nums[half - 1] + nums[half]) / 2.0
```

.github/workflows/progress-tracker.yml
ibnesayeed: I think it is important to stick with the standard definition and do take mean of the two middle values when there are an even number of elements in the list. This has implications when the two middle values are far apart, for example, `[1, 2, 99, 100]`, if we were to take one of the middle values, the answer would be `2`, but actual median should be `50.5`. Fractions are fine in median as well, the same way they are in the mean.

.github/workflows/progress-tracker.yml
ibnesayeed: > Also, I realized that I can test this action in this repo before merging by setting the branch in the "run workflow" manual run.

This is a very good catch, I totally forgot about this option.

.github/workflows/progress-tracker.yml
rviscomi: That's an unrealistically extreme example :)

I'd be surprised if any contributors sign up for more than 5 chapters, and I would rather see the median value reflect an actual number of chapters rather than a fractional approximation.

.github/workflows/progress-tracker.yml
rviscomi: Commented in https://github.com/HTTPArchive/almanac.httparchive.org/pull/1060#discussion_r456934285

.github/workflows/progress-tracker.yml
ibnesayeed: I gave that extreme example intentionally to highlight the issue. If we are calling it median then there is no reason to deviate from the standard definition of median and cause confusion to those who do not know what we did. The whole Web Almanac is all about representing statistics and if we take such shortcuts internally then it would not reflect good on our external reporting. If you are too keen about keeping it an integer (though I do not think it is necessary) then simply use `Math.round()` on it after conditional calculation.

rviscomi: As a compromise, how about we rename the field from median to p50 or 50th percentile?

ibnesayeed: 50th percentile is the same as median, so changing the term will not make a difference here. I think we are arguing for non-issue. Whatever definition we choose is only going to make a difference in Reviewers numbers (and Authors and Analysis are going to have a median of 1 most of the time) and perhaps not by far, so I would stick to the standard definition and just use rounding if I am really after a whole number.

ibnesayeed: While we are here, since we have sorted the list of numbers, Min and Max can be calculated as `nums[0]` and `nums[-1]` without the need to traversing through the list again, but that would be an unnecessary optimization. :wink:

rviscomi: Hmm, I think by definition percentiles all map back to real values in the dataset, so in this case the 50th percentile of an even number of values would be one of those values, either floor/ceil of the middle, and that'd be what we're measuring here.

To your point about this design decision reflecting poorly on our project methodology, I think that's a bit of a stretch. Even [APPROX_QUANTILES](https://cloud.google.com/bigquery/docs/reference/standard-sql/approximate_aggregate_functions#approx_quantiles) that we use on datasets with millions of values behaves this way.

```sql
SELECT APPROX_QUANTILES(x, 100)[OFFSET(50)] AS percentile_50
FROM UNNEST([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]) AS x;
```

The value here is 5, which is a real value from the dataset, as opposed to 5.5 from your calculation. We could take `nums[half-1]` to be more consistent with this behavior for even-length datasets, but I'd say we're pretty far into the weeds about statistics and starting to lose focus on what this code is actually trying to do. The tracking issue that this code generates is a tool for the PMs to get a better sense of how the chapters are staffed. For me as PM, the code I've proposed here is more aligned with what I'd expect to see. Your points about technical correctness are well taken but I think in practice we can opt for the simpler route.

ibnesayeed: > I think by definition percentiles all map back to real values in the dataset

[Not really!](https://www.mathsisfun.com/data/quartiles.html)

The quantile calculations you showed from Google Cloud docs has the terms "approx" in its name. And who knows, it is choosing an item from the list or simply rounding the result, that is not obvious from the sample example. :wink:

Here is a simple, yet exaggerated example using R, one of the popular statistical tools.

```
$ R
> a <- c(1, 2, 8, 10)
> quantile(a)
   0%   25%   50%   75%  100% 
 1.00  1.75  5.00  8.50 10.00 
> summary(a)
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
   1.00    1.75    5.00    5.25    8.50   10.00
```

And here is the calculation on the current actual data we have:

```
> authors <- c(4, 1, 3, 2, 2, 1, 3, 1, 1, 1, 2, 1, 2, 1, 1, 1, 1, 1, 1, 1, 1)
> reviewers <- c(5, 3, 3, 5, 3, 2, 11, 3, 9, 1, 3, 3, 5, 2, 2, 3, 3, 1, 1, 5, 8)
> analysts <- c(0, 1, 0, 0, 1, 1, 3, 1, 2, 2, 2, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1)
> sort(authors)
 [1] 1 1 1 1 1 1 1 1 1 1 1 1 1 1 2 2 2 2 3 3 4
> sort(reviewers)
 [1] 1 1 1 2 2 2 3 3 3 3 3 3 3 3 5 5 5 5 8 9 11
> sort(analysts)
 [1] 0 0 0 0 1 1 1 1 1 1 1 1 1 1 1 1 1 2 2 2 3

> summary(authors)
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
  1.000   1.000   1.000   1.524   2.000   4.000 
> summary(reviewers)
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
  1.000   2.000   3.000   3.857   5.000  11.000 
> summary(analysts)
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
  0.000   1.000   1.000   1.048   1.000   3.000
```

In this case, even with the actual definition of these terms we are getting a median that points to an actual number, and in case it is not, the number will not be too far off.

> but I'd say we're pretty far into the weeds about statistics and starting to lose focus on what this code is actually trying to do

I agree! :100:

> The tracking issue that this code generates is a tool for the PMs to get a better sense of how the chapters are staffed. For me as PM, the code I've proposed here is more aligned with what I'd expect to see. Your points about technical correctness are well taken but I think in practice we can opt for the simpler route.

Calculating actual median is not any less simple, but if you are really after a real data in your dataset then go by the route you chose. Statistics is not a divine spell, make it what helps you handle the problem at hand. If you chose to do so, you may want to get rid of the temporary constant `half` an put the calculation inline when finding the middle value as it will not be reused.

rviscomi: You're right about percentiles, thanks for calling me out on that. I'm so used to BigQuery's "approx" quantiles that I absorbed that into my understanding of percentiles 😅 

Sounds like we're good to move forward, thanks for your review!

PR #1064, opened 2020-07-20
Fix translation “accessibility" in Japanese

masuP9:
There are two translation of "accessibility", 「アクセシビリティ」 and 「アクセ**ン**シビリティ」. 「アクセ**ン**シビリティ」is probably typo. So I unified to 「アクセシビリティ」.


bazzadp:
Thanks @masuP9 ! @ksakae1216 you happy with these changes?

By the way @masuP9 you don't need to translate the `src/templates/ja/2019/chapters/` files as they are automatically generated from the `.md` files so you can save a bit of effort here. Similarly for the `ebook.html` file. No harm to translate it so don't need to remove from this pull request, but a bit of wasted effort. You were right to translate `featured_chapters.html`.

masuP9:
@bazzadp Thanks for letting me know which files I need to translate. I understand 👍

ksakae1216:
Yes, Good translation.
Thanks @masuP9

PR #1071, opened 2020-07-20
Add collective team in the progress report

ibnesayeed:
* Adds `All` as a derived field in teams that reflects in the both the tables
* Changes `Authors/Reviewers/Analysts` format to `Authors+Reviewers+Analysts=All`
* Refactors stats calculation function without changing the logic

### DIFF
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -41,7 +41,8 @@ jobs:
             const teamMembers = {
               Authors: {},
               Reviewers: {},
-              Analysts: {}
+              Analysts: {},
+              All: {}
             }
 
             const parseListItems = text => {
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -72,17 +72,18 @@ jobs:
             core.info(`Tracking progress of ${totalTrackedTasks} tasks across chapters`)
 
             const calcStats = nums => {
-              if (!nums.length) {
+              const size = nums.length
+              if (!size) {
                 return [0, 0, 0, 0, 0]
               }
               nums.sort((a, b) => a - b)
-              const half = Math.floor(nums.length / 2)
+              const sum = nums.reduce((a,b) => a + b, 0)
               return [
-                nums.reduce((a,b) => a + b, 0),
-                Math.min(...nums),
-                Math.max(...nums),
-                (nums.reduce((a,b) => a + b, 0) / nums.length).toFixed(2),
-                nums[half]
+                sum,
+                nums[0],
+                nums[size - 1],
+                (sum / size).toFixed(2),
+                nums[Math.floor(size / 2)]
               ]
             }
 
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -129,6 +129,7 @@ jobs:
                   team.Authors = parseUserNames(authors)
                   team.Reviewers = parseUserNames(reviewers)
                   team.Analysts = parseUserNames(analysts)
+                  team.All = [...team.Authors, ...team.Reviewers, ...team.Analysts]
                 }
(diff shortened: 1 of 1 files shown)


ibnesayeed:
This PR does not make team members list per chapter unique, in case the same person is in more than one roles because that will mess the stats up and I am not sure which way is more desired.

.github/workflows/progress-tracker.yml
```diff
@@ -129,6 +131,7 @@ jobs:
                   team.Authors = parseUserNames(authors)
                   team.Reviewers = parseUserNames(reviewers)
                   team.Analysts = parseUserNames(analysts)
+                  team.All = [...team.Authors, ...team.Reviewers, ...team.Analysts]
```
rviscomi, line 134: I think unique count of total contributors would be more useful. Could do something like `Array.from(new Set([...]))`.

.github/workflows/progress-tracker.yml
```diff
@@ -139,7 +142,7 @@ jobs:
             }
 
             const formatTeam = team => {
-              return Object.entries(team).map(([k, v]) => `<span title="${k}: ${v.length ? v.join(', ') : 'TBD'}">${v.length}</span>`).join('/')
+              return Object.entries(team).map(([k, v]) => `<span title="${k}: ${v.length ? v.join(', ') : 'TBD'}">${v.length}</span>`).join('+').replace(/\+([^+]*)$/, '=$1')
```
rviscomi, line 145: How about keeping "A/B/C" format and dealing with the fact that they won't add up to the unique total by formatting it as "A/B/C (D total)"?

.github/workflows/progress-tracker.yml
ibnesayeed: I actually wrote the code for making it unique, but reverted it for a few reasons:

* I was not sure how often a single person can be in more than one roles in the same chapter
* The sum can look odd in situations like `1+2+2=3`, but this is a representational thing that can be done differently
* In the Contributor details section, numbers will represent only one entry per chapter

That said, if you think that would be a better representation then I can bring my change back that I put together to make it unique initially.

.github/workflows/progress-tracker.yml
ibnesayeed: I have changed it slightly differently to keep it logically correct while still being compact.

.github/workflows/progress-tracker.yml
```diff
@@ -72,17 +73,18 @@ jobs:
...
               nums.sort((a, b) => a - b)
-              const half = Math.floor(nums.length / 2)
+              const sum = nums.reduce((a,b) => a + b, 0)
               return [
-                nums.reduce((a,b) => a + b, 0),
-                Math.min(...nums),
-                Math.max(...nums),
-                (nums.reduce((a,b) => a + b, 0) / nums.length).toFixed(2),
-                nums[half]
+                sum,
```
rviscomi, line 83: Just realized this doesn't work for unique contributors.

For example, in https://github.com/bazzadp/almanac.httparchive.org/issues/73 there are only 2 people (you and Barry) but the All/Total cell says 4. I'd expect this value to also deduplicate repeated contributors.

ibnesayeed: You don't want to deduplicate there as that is a chapter-wise stats and it represents a different quantity. The summary table is not showing in the Total columns what you feel it should. It has always been different. For example, even before we introduced `All`, `Reviewers` were adding up to 3, though there were only two unique people involved.

ibnesayeed: In short, we are deduplication multiple roles of a person in a single chapter, but we are not deduplicating roles of individuals across chapters, and we should not. If you are looking for an absolute number of unique contributors and in how many chapters they each contributed in one or more roles, the place you should be looking at is under the Contributor Details under the the All section.

PR #1075, opened 2020-07-22
Adding myself to the json :)

alextait1:


src/config/2020.json
```diff
@@ -880,6 +880,16 @@
       "website": "http://www.opentagclosetag.com/",
       "github": "oluoluoxenfree"
     },
+    "alextait1": {
+      "name": "Alex Tait",
```
rviscomi, line 884: Looks great. Could you move this up to be in sorted order by first name?

alextait1: done :)

PR #1077, opened 2020-07-22
Update contributors

rviscomi:
Synced with #989


bazzadp:
Noticed the avatar URL added in #1075 didn’t have the dimension query params in it. Does that mean it downloads a larger than necessary avatar? If so can we correct that URL as part of this?

rviscomi:
Good catch. Fixed.

PR #1095, opened 2020-07-25
Build site on Package.JSON change

bazzadp:
Dependabot updated web-vitals in #1088 but that didn't trigger a build so let's add `package.json` to the list of files our build action watches.

Also fixed incorrect `tools` folder reference.

### DIFF
--- a/.github/workflows/generate_chapters.yml
+++ b/.github/workflows/generate_chapters.yml
@@ -18,9 +18,10 @@ on:
     paths:
       - src/content/**
       - src/templates/**
-      - tools/generate/**
+      - src/tools/generate/**
       - src/static/css/**
       - src/static/js/**
+      - src/package.json
 jobs:
   build:
     runs-on: ubuntu-latest


bazzadp:
Did you see this one btw @ibnesayeed ? You only reviewed the other one so not sure if you noticed there were actually two?

ibnesayeed:
Oh, I did not realize it was waiting for my review. I get email notifications for everything, so the noise is high and I might miss things. I will review it soon.

bazzadp:
Thanks. Pretty small and simple but always good to get a second pair of eyes if I can!

.github/workflows/generate_chapters.yml
```diff
@@ -18,9 +18,10 @@ on:
     paths:
       - src/content/**
       - src/templates/**
-      - tools/generate/**
+      - src/tools/generate/**
```
ibnesayeed, line 21: I am not too familiar with this section of the site. Do we really want to watch here as well? I guess, the answer is going to be "yes", most likely.

bazzadp: Yeah I figure if we change that generate scripts, then we should regenerate 😀

If nothing changes then won’t matter and just a bit of a wasted run. But these scripts don’t change often enough for that to be a problem.

PR #1096, opened 2020-07-25
Update create pull request to v3

bazzadp:
See details here: https://github.com/peter-evans/create-pull-request/releases
and here: https://github.com/peter-evans/create-pull-request/

Also cleaned up some of the indentation.

I [tested this on my own repo](https://github.com/bazzadp/almanac.httparchive.org/runs/909258660?check_suite_focus=true)


.github/workflows/generate_chapters.yml
```diff
@@ -75,19 +75,20 @@ jobs:
         echo "regenerating chapters"
         npm run generate
     - name: Create Pull Request
-      uses: peter-evans/create-pull-request@v2
+      id: cpr
+      uses: peter-evans/create-pull-request@v3
```
ibnesayeed, line 79: With the new version, they have some breaking changes and one of them is how the author of the PR is assigned. I don't think there is anything wrong in the new style, but if you want to be consistent with previous style then you might want to supply the author input manually as suggested in the [release note](https://github.com/peter-evans/create-pull-request/releases/tag/v3.0.0).

.github/workflows/generate_chapters.yml
```diff
@@ -75,19 +75,20 @@ jobs:
...
-            - Auto-generated by [create-pull-request][1] GitHub Action
+          Updated datestamps and regenerated the chapters
+          - Please merge today or we'll get new datestamps tomorrow and you'll have to merge again!
+          - Auto-generated by [create-pull-request][1] GitHub Action
 
-            [1]: https://github.com/peter-evans/create-pull-request
-        labels: generate chapters
+          [1]: https://github.com/peter-evans/create-pull-request
+        labels: |
+          generate chapters
```
ibnesayeed, line 91: I do not see the pressing need to make this change, but it does not make a difference as such. We usually use this style for multi-line values.

```suggestion
        labels: generate chapters
```

.github/workflows/generate_chapters.yml
bazzadp: Yeah I'm happy to use the real author so actually like this change.

bazzadp:
Thanks @ibnesayeed !

Issue #22, opened 2019-05-21
Finalize assignments: Chapter 20. HTTP/2


rviscomi:
Section | Chapter | Authors | Reviewers
-- | -- | -- | --
IV. Content Distribution | 20. HTTP/2 | @bazzadp | @bagder @rmarx @dotjs

**Due date:** To help us stay on schedule, please complete the action items in this issue by June 3.

**To do:**
- [x] Assign subject matter expert (author)
- [x] Assign peer reviewers
- [x] Finalize metrics

**Current list of metrics:**

- Adoption rate of HTTP/2 by site (home page only) and by requests (all request on page) over the years. Trend graph over all available years.
- Measure of all HTTP versions (0.9, 1.0, 1.1, 2, QUIC) for main page of all sites, and for HTTPS sites. Table for last crawl. For example:

| Version          | All sites          | HTTPS only sites |
| ------------- | ------------- | ----------------- |
| HTTP/0.9       | 0%                  | 0%                        |
| HTTP/1.0       | 2%                   | 0%                        |
| HTTP/1.1        | 48%                | 20%                      |
| HTTP/2          | 44%                | 70%                      |
| gQUIC            | 6%                  | 10%                      |

For gQUIC it will be sites that return `Alt-Svc` HTTP Header which starts with `quic`.
- Average percentage of resources loaded over HTTP/2 (or gQUIC) versus HTTP/1.1 per site. Trend graph over all available years.
- Number of HTTP (not HTTPS) sites which return `upgrade` HTTP header containing `h2`. Once off stat for last crawl.
- Number of HTTPS sites using HTTP/2 which return `upgrade` HTTP header containing `h2`. Once off stat for last crawl.
- Number of HTTPS sites not using HTTP/2 which return `upgrade` HTTP header containing `h2`. Once off stat for last crawl.
- % of sites affected by CDN prioritization issues (H2 and served by CDN) - https://github.com/andydavies/http2-prioritization-issues#cdns--cloud-hosting-services. Once off stat for last crawl.
- Count of HTTP/2 sites grouped by `server` HTTP header value but strip version numbers (e.g. Apache and Apache 2.4.28 and Apache 2.4.29 should all report as Apache, but Apache Tomcat should report as Tomcat. Probably need to massive the results to achieve this). Once off stat for last crawl.
- Count of non-HTTP/2 sites grouped by `server` HTTP header value but strip version numbers. Once off stat for last crawl.
[...]

PR #1098, opened 2020-07-25
HTTP/2 2020 queries

gregorywolf:
Progress on #921 

All queries have been updated to use the almanac data set
- [x] - adoption_of_http_2_by_site_and_requests.sql
- [x] - avg_number_of_h2_pushed_resources_and_avg_bytes.sql
- [x] - avg_number_of_h2_pushed_resources_and_avg_bytes_by_content_type.sql
- [x] - avg_percent_of_resources_loaded_over_h2_or_h1_1_per_site.sql
- [x] - count_of_h2_sites_grouped_by_server.sql
- [x] - count_of_h2_sites_using_h2_push.sql
- [x] - count_of_non_h2_sites_grouped_by_server.sql
- [x] - count_of_preload_http_headers_with_nopush_attribute_set.sql
- [x] - detailed_alt_svc_headers.sql
- [x] - detailed_upgrade_headers.sql
- [x] - measure_number_of_tcp_connections_per_site.sql
- [x] - measure_of_all_http_versions_for_main_page_of_all_sites.sql
- [x] - number_of_http_sites_returning_upgrade_http_header_containing_h2.sql
- [x] - number_of_https_sites_not_using_h2_returning_upgrade_http_upgrade_header_containing_h2.sql
- [x] - number_of_https_sites_using_h2_returning_upgrade_http_header_containing_h2.sql
- [x] - percent_of_sites_affected_by_cdn_prioritization_issues.sql
- [x] - tls_1_3_adoption_for_h2


rviscomi:
Thanks @gregorywolf! Could you edit the PR description to include a checklist of the metrics needed by each chapter and check off the ones implemented in this PR so far? This will help us see at a glance how much work is still left to do.

gregorywolf:
@rviscomi I have updated the PR description as requested. I have also extracted out some suggested metrics from the chapter discussion which I need to determine how to generate. Any assistance would be much appreciated :)

bazzadp:
Hey @gregorywolf I've edited the initial comment to convert your SUCCESS/FAILS to tick boxes. This then shows how complete they are (14 out of 20) when looking at the issue:

Issue status

> How many sites offer H3 by looking at the alt-svc headers

Take a look at [20_04 from last year](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2019/20_HTTP_2/20_04.sql)
 and [20_04a from last year](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2019/20_HTTP_2/20_04a_05a_06a.sql)

> TLS 1.3 adoption for H2

That should be easy enough to get. Look at HTTP/2 queries and Security queries from last year.

> How many HTTP Headers are sent
> Size of HTTP headers

Again [20_04 from last year](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2019/20_HTTP_2/20_04.sql) might give you a way to get started with this.

We also wanted a way to measure HPACK gains, but [I asked about this last year](https://github.com/HTTPArchive/almanac.httparchive.org/issues/22#issuecomment-495771051) and it wasn't possible then so doubt it's possible now.

Hope that helps!

rviscomi:
Thanks @gregorywolf!

gregorywolf:
> Tested against sample_data.requests

ies have been updated to run against

bazzadp:
@gregorywolf please see [this comment](https://github.com/HTTPArchive/almanac.httparchive.org/issues/1258#issuecomment-691222367) about the fact the prioritisation table has been renamed and dated now.

gregorywolf (commented):
I updated the name of the table to be used in the query

sql/2020/22_HTTP_2/adoption_of_http_2_by_site_and_requests.sql
```diff
@@ -0,0 +1,20 @@
+#standardSQL
+# 22_HTTP_2 - Adoption of HTTP/2 by site and requests
+SELECT 
+  client,
+  firstHtml, 
+  JSON_EXTRACT_SCALAR(payload, "$._protocol") AS http_version, 
+  COUNT(*) AS num_requests,
```
rviscomi, line 7: `COUNT(0)` for consistency with other queries
```suggestion
  COUNT(0) AS num_requests,
```

sql/2020/22_HTTP_2/adoption_of_http_2_by_site_and_requests.sql
```diff
@@ -0,0 +1,20 @@
...
+  ROUND(COUNT(0) * 100 / SUM(COUNT(0)) OVER (PARTITION BY client), 2) AS pct
```
rviscomi, line 8: One change from last year is to use decimals for percents rather than `* 100`. This makes it easier to format with the % suffix in Sheets for use in data viz.
```suggestion
  ROUND(COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client), 4) AS pct
```

sql/2020/22_HTTP_2/adoption_of_http_2_by_site_and_requests.sql
```diff
@@ -0,0 +1,20 @@
+#standardSQL
+# 22_HTTP_2 - Adoption of HTTP/2 by site and requests
```
rviscomi, line 2: Do you also want to distinguish between first and third party requests? The test for first party is if `NET.HOST(url) = NET.HOST(page)`.

rviscomi, line 2: FWIW here and throughout it's ok to drop the `22_HTTP_2` prefix.

sql/2020/22_HTTP_2/avg_number_of_h2_pushed_resources_and_avg_bytes.sql
```diff
@@ -0,0 +1,26 @@
+#standardSQL
+# 22_HTTP_2 - Average number of HTTP/2 Pushed Resources and Average Bytes
+SELECT 
+  client,
+  COUNT(DISTINCT page) AS num_pages,
+  ROUND(AVG(num_requests),2) AS avg_pushed_requests,
```
rviscomi, line 6: Would percentiles be more useful? Averages tend to be easily skewed. For example, the median calculation would look like this:
```suggestion
  APPROX_QUANTILES(num_requests, 1000)[OFFSET(500)] AS median_pushed_requests,
```

(The file would also need to be renamed accordingly)

Many chapters also generate a few significant percentiles at the same time using a pattern like this:

```sql
SELECT
  percentile,
  client,
  APPROX_QUANTILES(metric, 1000)[OFFSET(percentile * 10)] AS value
FROM
  #(table or query),
GROUP BY
  percentile,
  client
ORDER BY
  percentile,
  client
```

sql/2020/22_HTTP_2/avg_number_of_h2_pushed_resources_and_avg_bytes.sql
```diff
@@ -0,0 +1,26 @@
...
+  ROUND(AVG(kb_transfered),2) AS avg_kb_transfered
+FROM (
+
+SELECT
```
rviscomi, line 10: Nit: whitespace
```suggestion
  SELECT
```

sql/2020/22_HTTP_2/avg_number_of_h2_pushed_resources_and_avg_bytes.sql
```diff
@@ -0,0 +1,26 @@
...
+  COUNT(DISTINCT page) AS num_pages,
+  ROUND(AVG(num_requests),2) AS avg_pushed_requests,
+  ROUND(AVG(kb_transfered),2) AS avg_kb_transfered
+FROM (
+
+SELECT 
+    client,
+    page,
+    SUM(CAST(JSON_EXTRACT_SCALAR(payload, "$._bytesIn") AS INT64)/1024) AS kb_transfered,
+    COUNT(*) AS num_requests
```
rviscomi, line 14:
```suggestion
    COUNT(0) AS num_requests
```

rviscomi (changes requested):
Looks great! Some early feedback after looking at the first few queries. The comments apply to the other queries as well.

sql/2020/22_HTTP_2/adoption_of_http_2_by_site_and_requests.sql
gregorywolf: Sounds good. Agree with the change. Changing all count(*) to count(0) in all files

sql/2020/22_HTTP_2/adoption_of_http_2_by_site_and_requests.sql
gregorywolf: Sounds good. Dropping the `22_HTTP_2` prefix in all files

sql/2020/22_HTTP_2/adoption_of_http_2_by_site_and_requests.sql
gregorywolf: Sounds good. I made the change to all files that had similar constructs

sql/2020/22_HTTP_2/avg_number_of_h2_pushed_resources_and_avg_bytes.sql
gregorywolf: Understood. What I did was groom all files to remove whitespace between any SQL statements. However, for readability I decided to leave whitespace between FUNCTION and SQL code. I also left whitespace before and after UNION ALL statement, agin for easier readability. Let me know if you disagree. If so i can remove whitespace from ALL files regardless of code.

sql/2020/22_HTTP_2/avg_number_of_h2_pushed_resources_and_avg_bytes.sql
gregorywolf: Yup. Propagated this change through all of my files

sql/2020/22_HTTP_2/avg_number_of_h2_pushed_resources_and_avg_bytes.sql
gregorywolf: I agree that averages are not the best. Following your suggestions and guidance I have done the following:

1. Renamed `avg_number_of_h2_pushed_resources_and_avg_bytes_by_content_type` to `number_of_h2_pushed_resources_and_avg_bytes_by_content_type`

2.

sql/2020/22_HTTP_2/adoption_of_http_2_by_site_and_requests.sql
gregorywolf: Hi. I understand what you are saying and think it makes good sense. Are you suggesting that I modify this one query or use this query as the basis for an extra query. Some SQL guidance would be helpful. Thanks.

gregorywolf:
Hi @rviscomi I have addressed all but a single one of your comments, which I added my comment/question. In addition to your feedback I also made some additional required changes that I discovered during the process

sql/2020/22_HTTP_2/adoption_of_http_2_by_site_and_requests.sql
rviscomi: @dotjs what would be your preference?

bazzadp: My 2 cents: Personally I’m not so sure.

The firstHTML is good to measure separately (do the pages support HTTP/2?), and interesting to see that X% of requests support HTTP/2 overall. Do I care the 1st party/third party breakout of those requests? Not so much for this chapter IMHO since we have the firstHTML one so know whether the site supports it or not. Plus that will exclude sharded asset domains anyway.

Then again could just query it and see if anything interesting shows up.

gregorywolf: @rviscomi - @bazzadp and I discussed this and the collective opinion is that this is not extremely useful information

bazzadp:
@gregorywolf I can see values like `"_protocol": "h3-Q050"` in the HARs from WebPageTest so presume they will be in the `_protocol` field (currently being discussed in #1293). That will be your HTTP/3 versions but the quic version will change so you should search for `like 'h3%'` rather than that specific value.

sql/2020/22_HTTP_2/measure_number_of_tcp_connections_per_site.sql
```diff
@@ -0,0 +1,41 @@
...
+  "mobile" AS client,
+  JSON_EXTRACT_SCALAR(payload, "$._protocol") AS protocol,
+  COUNT(0) AS num_pages,
+  APPROX_QUANTILES(_connections, 100)[SAFE_ORDINAL(50)] AS median,
+  APPROX_QUANTILES(_connections, 100)[SAFE_ORDINAL(75)] AS p75,
+  APPROX_QUANTILES(_connections, 100)[SAFE_ORDINAL(95)] AS p95
+FROM 
+  `httparchive.requests.2020_08_01_mobile` AS requests
+INNER JOIN
+  `httparchive.summary_pages.2020_08_01_mobile` AS summary
```
bazzadp, line 13: Can we use the `httparchive.almanac.requests` table and run this on one go? Seems a bit inconsistent to use that in some places and these tables in others.

sql/2020/22_HTTP_2/measure_number_of_tcp_connections_per_site.sql
```diff
@@ -0,0 +1,41 @@
...
+ON
+  requests.url = summary.url
+WHERE
+  JSON_EXTRACT_SCALAR(payload, "$._is_base_page") = "true"
```
bazzadp, line 17: Should we use `firstHTML` from `httparchive.almanac.requests` table? More consistent with other queries.

sql/2020/22_HTTP_2/number_of_h2_pushed_resources_and_avg_bytes.sql
```diff
@@ -0,0 +1,28 @@
...
+  client,
+  COUNT(DISTINCT page) AS num_pages,
+  APPROX_QUANTILES(num_requests, 1000)[OFFSET(500)] AS median_pushed_requests,
+  APPROX_QUANTILES(kb_transfered, 1000)[OFFSET(500)] AS median_kb_transfered,
+FROM (
+ SELECT 
+    percentile,
+    client,
+    page,
+    SUM(CAST(JSON_EXTRACT_SCALAR(payload, "$._bytesIn") AS INT64)/1024) AS kb_transfered,
```
bazzadp, line 13: I wonder could we use `respSize` here? See https://httparchive.slack.com/archives/C016QLMT0BZ/p1600189006109100 though as many be unreliable for HTTP/2?

sql/2020/22_HTTP_2/percent_of_resources_loaded_over_h2_or_h1_1_per_site.sql
```diff
@@ -0,0 +1,28 @@
...
+FROM (
+  SELECT 
+    percentile,
+    client,
+    page,
+    COUNT(0) AS num_requests, 
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="http/0.9",1,0)) AS http_0_9, 
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="http/1.0",1,0)) AS http_1_0, 
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="http/1.1",1,0)) AS http_1_1,
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="HTTP/2",1,0)) AS http_2
```
bazzadp, line 16: As discussed we should add `h3` (e.g. `h3-Q050`) to this and also check if protocol is case sensitive or not.

sql/2020/22_HTTP_2/percent_of_sites_affected_by_cdn_prioritization_issues.sql
```diff
@@ -0,0 +1,30 @@
...
+  IF(pages.cdn = "", "Not using CDN", pages.cdn) AS CDN,
+  IF(prioritization_status IS NOT null, prioritization_status, "Unknown") prioritizes_correctly,
+  COUNT(0) AS num_pages,
+  ROUND(COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client), 4) AS pct
+FROM (
+    SELECT
+      date,
+      client,
+      url,
+      JSON_EXTRACT_SCALAR(payload, "$._cdn_provider") as cdn
```
bazzadp, line 14: cdn_provider is available as a column so don't need to extract this manually yourself.

bazzadp (changes requested):
It's worth looking at the `httparchive.almanac.requests` columns as many of the values you are pulling out of the payload (server, cdn_provider).

Also seeing some inconsistency in the tables and queries as noted below.

sql/2020/22_HTTP_2/measure_number_of_tcp_connections_per_site.sql
gregorywolf: Makes sense. Let me rewrite the query to make sure I get the same output

sql/2020/22_HTTP_2/measure_number_of_tcp_connections_per_site.sql
gregorywolf: Agreed that makes more sense. I'll update the query

sql/2020/22_HTTP_2/percent_of_resources_loaded_over_h2_or_h1_1_per_site.sql
gregorywolf: Hopefully the requests table will get updated and if so then i will be able to drop all of the manual extraction from payload.

In regards to H3 I am not seeing protocol = `h3-Q050` The only entires I am seeing are:

image

IMHO I think if we use:
`SUM(IF(protocol="http/0.9",1,0)) AS http_0_9,
    SUM(IF(protocol="http/1.0",1,0)) AS http_1_0,
    SUM(IF(protocol="http/1.1",1,0)) AS http_1_1,
    SUM(IF(protocol="HTTP/2",1,0)) AS http_2,
    SUM(IF(protocol IN ("QUIC","http/2+quic/46"),1,0)) AS quic`

we should be ok. Thoughts? Perhaps I am missing the intent of the comment...?

sql/2020/22_HTTP_2/percent_of_sites_affected_by_cdn_prioritization_issues.sql
gregorywolf: Makes sense! Will make the change

sql/2020/22_HTTP_2/detailed_upgrade_headers.sql
```diff
@@ -0,0 +1,34 @@
...
+  }
+""";
+
+SELECT 
+  client,
+  firstHtml,  
+  JSON_EXTRACT_SCALAR(payload, "$._protocol") AS protocol,
+  IF(url LIKE "https://%","https","http") AS http_or_https,
+  getUpgradeHeader(payload) AS upgrade,
+  COUNT(0) AS num_requests
```
bazzadp, line 24: Upgrade header (and all response headers actually) exists in the `respOtherHeaders` column so use that (similar for other queries):

```suggestion
# standardSQL
# Detailed upgrade headers for 20.04, 20.05 and 20.06
SELECT 
  client,
  firstHtml,  
  JSON_EXTRACT_SCALAR(payload, "$._protocol") AS protocol,
  IF(url LIKE "https://%","https","http") AS http_or_https,
  regexp_extract(regexp_extract(respOtherHeaders, r'(?is)Upgrade = (.*)'), r'(?im)^([^=]*?)(?:, [a-z-]+ = .*)') is not null AS upgrade,
  COUNT(0) AS num_requests
```

sql/2020/22_HTTP_2/detailed_upgrade_headers.sql
```diff
@@ -0,0 +1,34 @@
...
+FROM 
+  `httparchive.almanac.requests`
+WHERE
+  date='2020-08-01'
+GROUP BY
+  client,
+  firstHtml,
+  protocol,
+  http_or_https,
+  upgrade
```
rviscomi, line 34:
```suggestion
# standardSQL
# Detailed upgrade headers for 20.04, 20.05 and 20.06
CREATE TEMPORARY FUNCTION getUpgradeHeader(payload STRING)
RETURNS STRING LANGUAGE js AS """
  try {
    var $ = JSON.parse(payload);
    var headers = $.response.headers;
    return headers.find(h => h.name.toLowerCase() === 'upgrade').value.trim();
  } catch (e) {
    return '';
  }
""";

SELECT
  client,
  firstHtml,
  JSON_EXTRACT_SCALAR(payload, '$._protocol') AS protocol,
  IF(url LIKE 'https://%', 'https', 'http') AS http_or_https,
  NORMALIZE_AND_CASEFOLD(getUpgradeHeader(payload)) AS upgrade,
  COUNT(0) AS num_requests,
  SUM(COUNT(0)) OVER (PARTITION BY client) AS total,
  COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client) AS pct
FROM
  `httparchive.almanac.requests`
WHERE
  date = '2020-08-01'
GROUP BY
  client,
  firstHtml,
  protocol,
  http_or_https,
  upgrade
HAVING
  num_requests >= 100
ORDER BY
  num_requests DESC
```

sql/2020/22_HTTP_2/measure_number_of_tcp_connections_per_site.sql
```diff
@@ -0,0 +1,41 @@
...
+  `httparchive.requests.2020_08_01_desktop` AS requests
+INNER JOIN
+  `httparchive.summary_pages.2020_08_01_desktop` AS summary
+ON
+  requests.url = summary.url
+WHERE
+  JSON_EXTRACT_SCALAR(payload, "$._is_base_page") = "true"
+GROUP BY 
+  client, 
+  protocol
```
rviscomi, line 41:
```suggestion
# standardSQL
# Measure number of TCP Connections per site.
SELECT
  percentile,
  client,
  protocol,
  COUNT(0) AS num_pages,
  APPROX_QUANTILES(_connections, 1000)[OFFSET(percentile * 10)] AS connections
FROM (
  SELECT
    client,
    page,
    JSON_EXTRACT_SCALAR(payload, '$._protocol') AS protocol,
  FROM
    `httparchive.almanac.requests`
  WHERE
    date = '2020-08-01' AND
    firstHtml)
JOIN (
  SELECT
    _TABLE_SUFFIX AS client,
    url AS page,
    _connections,
  FROM
    `httparchive.summary_pages.2020_08_01_*`)
USING
  (client, page),
  UNNEST([10, 25, 50, 75, 90]) AS percentile
GROUP BY
  percentile,
  client,
  protocol
ORDER BY
  percentile,
  client,
  protocol
```

sql/2020/22_HTTP_2/measure_of_all_http_versions_for_main_page_of_all_sites.sql
```diff
@@ -0,0 +1,20 @@
...
+FROM 
+   `httparchive.almanac.requests`
+WHERE
+   date='2020-08-01' AND 
+   firstHtml
+GROUP BY
+  client,
+  protocol
+ORDER BY 
+  num_pages / total DESC
```
rviscomi, line 20:
```suggestion
# standardSQL
# Measure of all HTTP versions (0.9, 1.0, 1.1, 2, QUIC) for main page of all sites, and for HTTPS sites. Table for last crawl.
SELECT
  client,
  JSON_EXTRACT_SCALAR(payload, '$._protocol') AS protocol,
  COUNT(0) AS num_pages,
  SUM(COUNT(0)) OVER (PARTITION BY client) AS total,
  COUNTIF(url LIKE 'https://%') AS num_https_pages,
  COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client) AS pct_pages,
  COUNTIF(url LIKE 'https://%') / SUM(COUNT(0)) OVER (PARTITION BY client) AS pct_https
FROM
   `httparchive.almanac.requests`
WHERE
   date = '2020-08-01' AND
   firstHtml
GROUP BY
  client,
  protocol
ORDER BY
  num_pages / total DESC
```

sql/2020/22_HTTP_2/number_of_h2_pushed_resources_and_avg_bytes.sql
```diff
@@ -0,0 +1,28 @@
...
+    date='2020-08-01' AND
+    JSON_EXTRACT_SCALAR(payload, "$._protocol") = "HTTP/2" AND
+    JSON_EXTRACT_SCALAR(payload, "$._was_pushed") = "1"
+  GROUP BY 
+    percentile,
+    client,
+    page
+)
+GROUP BY
+  client
```
rviscomi, line 28:
```suggestion
# standardSQL
# Number of HTTP/2 Pushed Resources and distribution of Bytes
SELECT
  percentile,
  client,
  COUNT(DISTINCT page) AS num_pages,
  APPROX_QUANTILES(num_requests, 1000)[OFFSET(percentile * 10)] AS pushed_requests,
  APPROX_QUANTILES(kb_transfered, 1000)[OFFSET(percentile * 10)] AS kb_transfered,
FROM (
 SELECT
    client,
    page,
    SUM(CAST(JSON_EXTRACT_SCALAR(payload, '$._bytesIn') AS INT64) / 1024) AS kb_transfered,
    COUNT(0) AS num_requests
  FROM
    `httparchive.almanac.requests`
  WHERE
    date = '2020-08-01' AND
    JSON_EXTRACT_SCALAR(payload, '$._protocol') = 'HTTP/2' AND
    JSON_EXTRACT_SCALAR(payload, '$._was_pushed') = '1'
  GROUP BY
    client,
    page),
  UNNEST([10, 25, 50, 75, 90]) AS percentile
GROUP BY
  percentile,
  client
ORDER BY
  percentile,
  client
```

rviscomi (changes requested):
Rather than going through a PR in your fork, I've suggested my changes in this PR, which you can approve individually as needed. There's some overlap with @bazzadp's suggestions too.

gregorywolf (commented):
All changes look fine

sql/2020/22_HTTP_2/percent_of_sites_affected_by_cdn_prioritization_issues.sql
gregorywolf: Change made and committed

sql/2020/22_HTTP_2/percent_of_resources_loaded_over_h2_or_h1_1_per_site.sql
gregorywolf: I added `SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") IN ("QUIC","http/2+quic/46"),1,0)) AS quic` to the query in order to collect H3

sql/2020/22_HTTP_2/count_of_non_h2_sites_grouped_by_server.sql
```diff
@@ -0,0 +1,35 @@
...
+  } catch (e) {
+    return '';
+  }
+""";
+
+SELECT 
+  client,
+  getServerHeader(payload) AS server_header,
+  COUNT(0) AS num_pages,
+  ROUND(COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client), 4) AS pct
```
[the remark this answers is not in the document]
gregorywolf: @rviscomi is running all of the H2 queries so cost is now a moot point

sql/2020/22_HTTP_2/measure_number_of_tcp_connections_per_site.sql
gregorywolf: @rviscomi modified the query and i just committed the suggestion

sql/2020/22_HTTP_2/detailed_upgrade_headers.sql
gregorywolf: Changes merged into other changes made by @rviscomi

gregorywolf:
@rviscomi @bazzadp Hi. Between Rick and I I believe all of the required changes have been made. A number of the changes that Barry has requested were base don cost savings, which make sense. However, since Rick plans to run all of the H2 queries the efficiency changes are moot at this point.

gregorywolf (commented):
I have gone through all of the comments and have input the requested changes where it made sense. A number of modifications were also made by @rviscomi

sql/2020/22_HTTP_2/number_of_h2_pushed_resources_and_avg_bytes_by_content_type.sql
```diff
@@ -0,0 +1,32 @@
+# standardSQL
+# Number of HTTP/2 Pushed Resources and Average Bytes by Content type
+SELECT 
+  client,
+  content_type,
+  COUNT(DISTINCT page) AS num_pages,
+  APPROX_QUANTILES(num_requests, 1000)[OFFSET(500)] AS median_pushed_requests,
```
rviscomi, line 7: Use the `percentile` field. See `number_of_h2_pushed_resources_and_avg_bytes.sql` for example.

sql/2020/22_HTTP_2/number_of_http_sites_returning_upgrade_http_header_containing_h2.sql
```diff
@@ -0,0 +1,30 @@
...
+  COUNT(0) AS num_requests
+FROM 
+  `httparchive.almanac.requests`
+WHERE
+  date='2020-08-01' AND 
+  url LIKE "http://%" AND
+  getUpgradeHeader(payload) LIKE "%h2%"
+GROUP BY
+  client,
+  firstHtml
```
rviscomi, line 30: This should be in the `WHERE` clause so that you're only analyzing the response for the document.

sql/2020/22_HTTP_2/number_of_http_sites_returning_upgrade_http_header_containing_h2.sql
```diff
@@ -0,0 +1,30 @@
...
+SELECT 
+  client,
+  firstHtml,  
+  COUNT(0) AS num_requests
+FROM 
+  `httparchive.almanac.requests`
+WHERE
+  date='2020-08-01' AND 
+  url LIKE "http://%" AND
+  getUpgradeHeader(payload) LIKE "%h2%"
```
rviscomi, line 27: Consider moving this out of the `WHERE` clause, and changing `num_requests` to a conditional count:

```sql
COUNTIF(getUpgradeHeader(payload) LIKE "%h2%") AS num_requests
```

That way you can use `COUNT(0) AS total` and also calculate the percent of HTTP sites that can upgrade to H2.

sql/2020/22_HTTP_2/number_of_https_sites_not_using_h2_returning_upgrade_http_upgrade_header_containing_h2.sql
```diff
@@ -0,0 +1,31 @@
...
+SELECT 
+  client,
+  firstHtml,  
+  COUNT(0) AS num_requests
+FROM 
+  `httparchive.almanac.requests`
+WHERE
+  date='2020-08-01' AND 
+  url LIKE "https://%" AND
+  JSON_EXTRACT_SCALAR(payload, "$._protocol") != "HTTP/2" AND
```
rviscomi, line 27: Ditto from comments on previous query. Filter only for `firstHtml` requests and `COUNTIF`.

sql/2020/22_HTTP_2/number_of_https_sites_using_h2_returning_upgrade_http_header_containing_h2.sql
```diff
@@ -0,0 +1,31 @@
...
+SELECT 
+  client,
+  firstHtml,  
+  COUNT(0) AS num_requests
+FROM 
+  `httparchive.almanac.requests`
+WHERE
+  date='2020-08-01' AND 
+  url LIKE "https://%" AND
+  JSON_EXTRACT_SCALAR(payload, "$._protocol") = "HTTP/2" AND
```
rviscomi, line 27: Ditto

sql/2020/22_HTTP_2/percent_of_resources_loaded_over_h2_or_h1_1_per_site.sql
```diff
@@ -0,0 +1,29 @@
+# standardSQL
+# Percentage of resources loaded over HTTP/2 or HTTP/1.1 per site
+SELECT
+  client,
+  APPROX_QUANTILES((http_1_1/num_requests) * 100, 1000)[OFFSET(500)] AS median_avg_pct_http_1_1,
```
rviscomi, line 5: Ditto on percentiles.

sql/2020/22_HTTP_2/percent_of_sites_affected_by_cdn_prioritization_issues.sql
```diff
@@ -0,0 +1,30 @@
...
+      `httparchive.almanac.requests` 
+    WHERE 
+      firstHtml AND
+      JSON_EXTRACT_SCALAR(payload, "$._protocol") ="HTTP/2" AND
+      date='2020-08-01'
+  ) AS pages
+LEFT JOIN 
+  `httparchive.almanac.h2_prioritization_cdns` AS h2_pri
+ON pages.cdn = h2_pri.cdn
+AND pages.date = h2_pri.date
```
rviscomi, line 25: `USING` is helpful when the fields have the same name. This way you also don't need to alias the tables `h2_pri` and `pages`.
```suggestion
USING (cdn, date)
```

rviscomi (changes requested):

sql/2020/22_HTTP_2/number_of_https_sites_not_using_h2_returning_upgrade_http_upgrade_header_containing_h2.sql
bazzadp: Ditto for including (and differentiating HTTP/3)

sql/2020/22_HTTP_2/number_of_https_sites_using_h2_returning_upgrade_http_header_containing_h2.sql
bazzadp: And ditto for me too.

bazzadp (changes requested):
Lots of comments about HTTP/3 and last query looks wrong.

sql/2020/22_HTTP_2/number_of_h2_pushed_resources_and_avg_bytes_by_content_type.sql
gregorywolf: Updated the query by adding percentile field. Used `number_of_h2_pushed_resources_and_avg_bytes.sql` as example. Checked the results against `sample_data.requests` table and looks correct.

sql/2020/22_HTTP_2/number_of_http_sites_returning_upgrade_http_header_containing_h2.sql
gregorywolf: @rviscomi I am not following. You are stating that THIS should be in the `Where` clause....Please define THIS

rviscomi: `firstHtml`

gregorywolf: Made the changes and tested successfully

bazzadp: @rviscomi is suggesting only to measure this on the initial page and not all requests. In which case you should remove from `SELECT` and `GROUP BY` clause and add to `WHERE` clause (no point in being in `SELECT` or `GROUP BY` if always true).

I'm not sure I agree to be honest. It may be interesting to see if ANY request does this and not just document itself. And with the group by we get to choose whatever stat we want. So think I'd leave it as is to be honest. I came up with this stat last year so I'm using my "I know more than Rick on this one (for once!)" card! 😀

At end of the day it was only on [108 sites last year](https://almanac.httparchive.org/en/2019/http2#http2-issues) so probably not worth including this year unless that's got a lot worse. But let's run as is to have a look.

We should however rename this file from "number of sites..." to "number of requests..." which is probably why Rick suggested this change.

gregorywolf: Leaving firstHtml in Select and Group By and renamed the filename to reflect requests instead of sites

sql/2020/22_HTTP_2/percent_of_resources_loaded_over_h2_or_h1_1_per_site.sql
```diff
@@ -0,0 +1,29 @@
...
+  SELECT 
+    percentile,
+    client,
+    page,
+    COUNT(0) AS num_requests, 
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="http/0.9",1,0)) AS http_0_9, 
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="http/1.0",1,0)) AS http_1_0, 
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="http/1.1",1,0)) AS http_1_1,
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="HTTP/2",1,0)) AS http_2,
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") IN ("QUIC","http/2+quic/46"),1,0)) AS quic
```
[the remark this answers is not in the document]
gregorywolf: I agree that the existing code is "fragile." I modified using suggested code. In addition I added each HTTP version to Select statement. I also think (IMHO) that removing percentile and unnest logic makes sense. Thoughts? I commented out the lines I think should be dropped so it is really obvious

sql/2020/22_HTTP_2/percent_of_sites_affected_by_cdn_prioritization_issues.sql
gregorywolf: Change made and syntax check passed

sql/2020/22_HTTP_2/percent_of_resources_loaded_over_h2_or_h1_1_per_site.sql
gregorywolf: @rviscomi  I actually removed the percentiles from the query logic. Thoughts?

sql/2020/22_HTTP_2/number_of_http_sites_returning_upgrade_http_header_containing_h2.sql
gregorywolf: @rviscomi @bazzadp I made the following changes:

Added the `COUNTIF(getUpgradeHeader(payload) LIKE "%h2%") AS num_requests` and `Count(0) as total` to the `SELECT`

Added `JSON_EXTRACT_SCALAR(payload, '$._protocol') as http_version` to `SELECT` and `GROUP BY`

Added `JSON_EXTRACT_SCALAR(payload, '$._protocol') != ''` to `WHERE` statement to filter out any null values for `http_version` I think it makes sense to filter out nulls....let me know if you disagree

sql/2020/22_HTTP_2/number_of_https_sites_not_using_h2_returning_upgrade_http_upgrade_header_containing_h2.sql
gregorywolf: @rviscomi @bazzadp I made the following changes:

Added the `COUNTIF(getUpgradeHeader(payload) LIKE "%h2%") AS num_requests` and `Count(0) as total` to the `SELECT`

Added `JSON_EXTRACT_SCALAR(payload, '$._protocol') as http_version` to `SELECT` and `GROUP BY`

Added `JSON_EXTRACT_SCALAR(payload, '$._protocol') != ''` to `WHERE` statement to filter out any null values for `http_version` I think it makes sense to filter out nulls....let me know if you disagree

sql/2020/22_HTTP_2/number_of_https_sites_using_h2_returning_upgrade_http_header_containing_h2.sql
gregorywolf: @rviscomi @bazzadp I made the following changes:

Added the `COUNTIF(getUpgradeHeader(payload) LIKE "%h2%") AS num_requests` and `Count(0) as total` to the `SELECT`

Added `JSON_EXTRACT_SCALAR(payload, '$._protocol') as http_version` to `SELECT` and `GROUP BY`

Added `JSON_EXTRACT_SCALAR(payload, '$._protocol') != ''` to `WHERE` statement to filter out any null values for `http_version` I think it makes sense to filter out nulls....let me know if you disagree

sql/2020/22_HTTP_2/number_of_http_sites_returning_upgrade_http_header_containing_h2.sql
gregorywolf: @bazzadp Need some clarification on your comment `Look at all upgrade except “Webprotocol” to include any HTTP/3 type upgrades (which shouldn’t really be sent as should use Alt-Svc but still interested to see).`

bazzadp: I don't see the protocol changes you mentioned @gregorywolf . You can ignore my comments. Let's leave looking at `h2` only. But yes lets add the protocol.

sql/2020/22_HTTP_2/tls_1_3_adoption_for_h2.sql
```diff
@@ -0,0 +1,21 @@
+# standardSQL
+# Distribution of TLS versions
```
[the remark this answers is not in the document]
gregorywolf: Agreed! Now collecting HTTP version - `JSON_EXTRACT_SCALAR(payload, '$._protocol') as http_version`

Also renamed the query - `tls_adoption_by_http_version.sql`

sql/2020/22_HTTP_2/count_of_h2_sites_grouped_by_server.sql
```diff
@@ -0,0 +1,22 @@
...
+  NORMALIZE_AND_CASEFOLD(REGEXP_EXTRACT(resp_server, r'\s*([^/]*)\s*')) AS server_header,
+  COUNT(0) AS num_pages,
+  SUM(COUNT(0)) OVER (PARTITION BY client) AS total,
+  ROUND(COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client), 4) AS pct
+FROM
+  `httparchive.almanac.requests`
+WHERE
+  date = '2020-08-01' AND
+  firstHtml AND
+  JSON_EXTRACT_SCALAR(payload, '$._protocol') = 'HTTP/2'
```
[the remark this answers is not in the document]
gregorywolf: Change made

sql/2020/22_HTTP_2/count_of_h2_sites_using_h2_push.sql
```diff
@@ -0,0 +1,19 @@
...
+FROM (
+  SELECT
+    client,
+    page,
+    JSON_EXTRACT_SCALAR(payload, '$._was_pushed') = '1' AS was_pushed
+  FROM
+    `httparchive.almanac.requests`
+  WHERE
+    date = '2020-08-01' AND
+    JSON_EXTRACT_SCALAR(payload, '$._protocol') = 'HTTP/2')
```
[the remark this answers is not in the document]
gregorywolf: Changes made. Also changed the comment inside the query and the name of the query to reflect both H2 and H3

sql/2020/22_HTTP_2/count_of_non_h2_sites_grouped_by_server.sql
```diff
@@ -0,0 +1,22 @@
...
+  NORMALIZE_AND_CASEFOLD(REGEXP_EXTRACT(resp_server, r'\s*([^/]*)\s*')) AS server_header,
+  COUNT(0) AS num_pages,
+  SUM(COUNT(0)) OVER (PARTITION BY client) AS total,
+  ROUND(COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client), 4) AS pct
+FROM
+  `httparchive.almanac.requests`
+WHERE
+  date = '2020-08-01' AND
+  firstHtml AND
+  JSON_EXTRACT_SCALAR(payload, '$._protocol') != 'HTTP/2'
```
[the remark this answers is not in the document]
gregorywolf: Changes made. Comment on line 2 updated as well as the query name to reflect H2 and H3

sql/2020/22_HTTP_2/number_of_h2_pushed_resources_and_avg_bytes.sql
```diff
@@ -0,0 +1,30 @@
...
+ SELECT
+    client,
+    page,
+    SUM(CAST(JSON_EXTRACT_SCALAR(payload, '$._bytesIn') AS INT64) / 1024) AS kb_transfered,
+    COUNT(0) AS num_requests
+  FROM
+    `httparchive.almanac.requests`
+  WHERE
+    date = '2020-08-01' AND
+    JSON_EXTRACT_SCALAR(payload, '$._protocol') = 'HTTP/2' AND
```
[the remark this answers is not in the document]
gregorywolf: Changes made

sql/2020/22_HTTP_2/number_of_h2_pushed_resources_and_avg_bytes_by_content_type.sql
```diff
@@ -0,0 +1,32 @@
...
+    page,
+    JSON_EXTRACT_SCALAR(payload, "$._contentType") as content_type,
+    SUM(CAST(JSON_EXTRACT_SCALAR(payload, "$._bytesIn") AS INT64)/1024) AS kb_transfered,
+    COUNT(0) AS num_requests
+  FROM 
+    `httparchive.almanac.requests`,
+    UNNEST([10, 25, 50, 75, 90]) AS percentile
+  WHERE 
+    date='2020-08-01' AND
+    JSON_EXTRACT_SCALAR(payload, "$._protocol") = "HTTP/2" AND
```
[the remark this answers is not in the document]
gregorywolf: Changes made

sql/2020/22_HTTP_2/percent_of_sites_affected_by_cdn_prioritization_issues.sql
```diff
@@ -0,0 +1,30 @@
+# standardSQL
+# % of sites affected by CDN prioritization issues (H2 and served by CDN)
```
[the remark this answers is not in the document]
gregorywolf: Changes made

gregorywolf (commented):
All changes made

sql/2020/22_HTTP_2/number_of_h2_and_h3_pushed_resources_and_bytes_by_content_type.sql
```diff
@@ -0,0 +1,41 @@
+# standardSQL
+# Number of H2 and H3 Pushed Resources and Bytes by Content type
+SELECT 
+  percentile,
+  client,
+  http_version,
+  content_type,
+  COUNT(DISTINCT page) AS num_pages,
+  APPROX_QUANTILES(num_requests, 1000)[OFFSET(500)] AS pushed_requests,
+  APPROX_QUANTILES(kb_transfered, 1000)[OFFSET(500)] AS kb_transfered,
```
rviscomi, line 10: Use the percentiles :)
```suggestion
  APPROX_QUANTILES(num_requests, 1000)[OFFSET(percentile * 10)] AS pushed_requests,
  APPROX_QUANTILES(kb_transfered, 1000)[OFFSET(percentile * 10)] AS kb_transfered,
```

sql/2020/22_HTTP_2/number_of_http_requests_returning_upgrade_http_header_containing_h2.sql
```diff
@@ -0,0 +1,32 @@
...
+  firstHtml,
+  JSON_EXTRACT_SCALAR(payload, '$._protocol') as http_version,
+  COUNTIF(getUpgradeHeader(payload) LIKE "%h2%") AS num_requests,
+  COUNT(0) AS total
+FROM 
+  `httparchive.almanac.requests`
+WHERE
+  date='2020-08-01' AND 
+  url LIKE "http://%" AND
+  getUpgradeHeader(payload) LIKE "%h2%"
```
rviscomi, line 28: This is handled in the `COUNTIF` above
```suggestion
  url LIKE "http://%"
```

sql/2020/22_HTTP_2/number_of_https_requests_not_using_h2_or_h3_returning_upgrade_http_upgrade_header_containing_h2.sql
```diff
@@ -0,0 +1,37 @@
...
+  firstHtml,
+  JSON_EXTRACT_SCALAR(payload, '$._protocol') as http_version,
+  COUNTIF(getUpgradeHeader(payload) LIKE "%h2%") AS num_requests,
+  COUNT(0) AS total
+FROM 
+  `httparchive.almanac.requests`
+WHERE
+  date = '2020-08-01' AND 
+  url LIKE "https://%" AND
+  getUpgradeHeader(payload) LIKE "%h2%" AND
```
rviscomi, line 29:
```suggestion
```

sql/2020/22_HTTP_2/number_of_https_requests_using_h2_returning_upgrade_http_header_containing_h2.sql
```diff
@@ -0,0 +1,34 @@
...
+  JSON_EXTRACT_SCALAR(payload, '$._protocol') as http_version,
+  COUNTIF(getUpgradeHeader(payload) LIKE "%h2%") AS num_requests,
+  COUNT(0) AS total
+FROM 
+  `httparchive.almanac.requests`
+WHERE
+  date='2020-08-01' AND 
+  url LIKE "https://%" AND
+  JSON_EXTRACT_SCALAR(payload, "$._protocol") = "HTTP/2" AND
+  getUpgradeHeader(payload) LIKE "%h2%"
```
rviscomi, line 30:
```suggestion
  JSON_EXTRACT_SCALAR(payload, "$._protocol") = "HTTP/2"
```

sql/2020/22_HTTP_2/percentage_of_resources_loaded_over_HTTP_by_version_per_site.sql
```diff
@@ -0,0 +1,36 @@
+# standardSQL
+# Percentage of resources loaded over HTTP by version per site
+SELECT
+  client,
+  APPROX_QUANTILES((http_0_9/num_requests) * 100, 1000)[OFFSET(500)] AS pct_http_0_9,
```
rviscomi, line 5: `percentile` is unused

rviscomi (changes requested):

sql/2020/22_HTTP_2/percentage_of_resources_loaded_over_HTTP_by_version_per_site.sql
gregorywolf: `500` was changed to `percentile * 10` on lines 5 - 9

sql/2020/22_HTTP_2/number_of_h2_and_h3_pushed_resources_and_bytes_by_content_type.sql
```diff
@@ -0,0 +1,41 @@
...
+    content_type),
+  UNNEST([10, 25, 50, 75, 90]) AS percentile
+GROUP BY
+  percentile,
+  client,
+  http_version,
+  content_type
+ORDER BY
+  percentile,
+  client
```
rviscomi, line 41: Suggestion to use the preprocessed `type` field rather than `$._contentType`. This groups similar types together since you don't need to differentiate between exact MIME types.

```suggestion
# standardSQL
# Number of H2 and H3 Pushed Resources and Bytes by Content type
SELECT 
  percentile,
  client,
  http_version,
  type,
  COUNT(DISTINCT page) AS num_pages,
  APPROX_QUANTILES(num_requests, 1000)[OFFSET(percentile * 10)] AS pushed_requests,
  APPROX_QUANTILES(kb_transfered, 1000)[OFFSET(percentile * 10)] AS kb_transfered,
FROM (
  SELECT 
    client,
    page,
    JSON_EXTRACT_SCALAR(payload, '$._protocol') as http_version,
    type,
    SUM(CAST(JSON_EXTRACT_SCALAR(payload, "$._bytesIn") AS INT64)/1024) AS kb_transfered,
    COUNT(0) AS num_requests
  FROM 
    `httparchive.almanac.requests`  
  WHERE 
    date='2020-08-01' AND
    JSON_EXTRACT_SCALAR(payload, "$._was_pushed") = "1" AND
    (LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "http/2" OR
     LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "%quic%" OR
     LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "h3%" OR
     LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "http/3%")  
  GROUP BY 
    client,
    http_version,
    page,
    type),
  UNNEST([10, 25, 50, 75, 90]) AS percentile
GROUP BY
  percentile,
  client,
  http_version,
  type
ORDER BY
  percentile,
  client
```

rviscomi, line 41: Suggestion to use the preprocessed `type` field rather than `$._contentType`. This groups similar types together since you don't need to differentiate between exact MIME types.

```suggestion
# standardSQL
# Number of H2 and H3 Pushed Resources and Bytes by Content type
SELECT 
  percentile,
  client,
  http_version,
  type,
  COUNT(DISTINCT page) AS num_pages,
  APPROX_QUANTILES(num_requests, 1000)[OFFSET(percentile * 10)] AS pushed_requests,
  APPROX_QUANTILES(kb_transfered, 1000)[OFFSET(percentile * 10)] AS kb_transfered,
FROM (
  SELECT 
    client,
    page,
    JSON_EXTRACT_SCALAR(payload, '$._protocol') as http_version,
    type,
    SUM(CAST(JSON_EXTRACT_SCALAR(payload, "$._bytesIn") AS INT64)/1024) AS kb_transfered,
    COUNT(0) AS num_requests
  FROM 
    `httparchive.almanac.requests`  
  WHERE 
    date='2020-08-01' AND
    JSON_EXTRACT_SCALAR(payload, "$._was_pushed") = "1" AND
    (LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "http/2" OR
     LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "%quic%" OR
     LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "h3%" OR
     LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "http/3%")  
  GROUP BY 
    client,
    http_version,
    page,
    type),
  UNNEST([10, 25, 50, 75, 90]) AS percentile
GROUP BY
  percentile,
  client,
  http_version,
  type
ORDER BY
  percentile,
  client
```

rviscomi, line 41: Suggestion to use the preprocessed `type` field rather than `$._contentType`. This groups similar types together since you don't need to differentiate between exact MIME types.

```suggestion
# standardSQL
# Number of H2 and H3 Pushed Resources and Bytes by Content type
SELECT 
  percentile,
  client,
  http_version,
  type,
  COUNT(DISTINCT page) AS num_pages,
  APPROX_QUANTILES(num_requests, 1000)[OFFSET(percentile * 10)] AS pushed_requests,
  APPROX_QUANTILES(kb_transfered, 1000)[OFFSET(percentile * 10)] AS kb_transfered,
FROM (
  SELECT 
    client,
    page,
    JSON_EXTRACT_SCALAR(payload, '$._protocol') as http_version,
    type,
    SUM(CAST(JSON_EXTRACT_SCALAR(payload, "$._bytesIn") AS INT64)/1024) AS kb_transfered,
    COUNT(0) AS num_requests
  FROM 
    `httparchive.almanac.requests`  
  WHERE 
    date='2020-08-01' AND
    JSON_EXTRACT_SCALAR(payload, "$._was_pushed") = "1" AND
    (LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "http/2" OR
     LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "%quic%" OR
     LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "h3%" OR
     LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "http/3%")  
  GROUP BY 
    client,
    http_version,
    page,
    type),
  UNNEST([10, 25, 50, 75, 90]) AS percentile
GROUP BY
  percentile,
  client,
  http_version,
  type
ORDER BY
  percentile,
  client
```

sql/2020/22_HTTP_2/percentage_of_h2_and_h3_sites_affected_by_cdn_prioritization.sql
```diff
@@ -0,0 +1,36 @@
...
+  IF(prioritization_status IS NOT NULL, prioritization_status, "Unknown") prioritizes_correctly,
+  COUNT(0) AS num_pages,
+  ROUND(COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client), 4) AS pct
+FROM (
+    SELECT
+      date,
+      client,
+      JSON_EXTRACT_SCALAR(payload, '$._protocol') as http_version,
+      url,
+      _cdn_provider as cdn
```
rviscomi, line 16: One thing about this field is that there may be a chain of CDNs, for example `Cloudflare, Google`. Do you need to parse one of those out? Otherwise it won't be able to be joined with the per-CDN values in `h2_prioritization_cdns`. cc @bazzadp 

Here are the [saved results](https://docs.google.com/spreadsheets/d/1M1tijxf04wSN3KU0ZUunjPYCrVsaJfxzuRCXUeRQ-YU/edit#gid=1152953475) for this query for reference.

sql/2020/22_HTTP_2/percentage_of_resources_loaded_over_HTTP_by_version_per_site.sql
```diff
@@ -0,0 +1,36 @@
...
+    UNNEST([10, 25, 50, 75, 90]) AS percentile
+  WHERE
+    date='2020-08-01'
+  GROUP BY
+    percentile,
+    client,
+    page
+)
+GROUP BY 
+  client
```
rviscomi, line 36: Fixing up use of `percentile`, plus some minor improvements

```suggestion
# standardSQL
# Percentage of resources loaded over HTTP by version per site
SELECT
  percentile,
  client,
  APPROX_QUANTILES(http_0_9, 1000)[OFFSET(percentile * 10)] AS pct_http_0_9,
  APPROX_QUANTILES(http_1_0, 1000)[OFFSET(percentile * 10)] AS pct_http_1_0,
  APPROX_QUANTILES(http_1_1, 1000)[OFFSET(percentile * 10)] AS pct_http_1_1,
  APPROX_QUANTILES(http_2, 1000)[OFFSET(percentile * 10)] AS pct_http_2,
  APPROX_QUANTILES(quic, 1000)[OFFSET(percentile * 10)] AS pct_quic
FROM (
  SELECT
    client,
    COUNTIF(JSON_EXTRACT_SCALAR(payload, "$._protocol") = "http/0.9") / COUNT(0) AS http_0_9, 
    COUNTIF(JSON_EXTRACT_SCALAR(payload, "$._protocol") = "http/1.0") / COUNT(0) AS http_1_0, 
    COUNTIF(JSON_EXTRACT_SCALAR(payload, "$._protocol") = "http/1.1") / COUNT(0) AS http_1_1,
    COUNTIF(JSON_EXTRACT_SCALAR(payload, "$._protocol") = "HTTP/2") / COUNT(0) AS http_2,
    COUNTIF(REGEXP_CONTAINS(JSON_EXTRACT_SCALAR(payload, "$._protocol"), r"(?i)(quic|^h3|^http/3)")) AS quic
  FROM 
    `httparchive.almanac.requests`
  WHERE
    date = '2020-08-01'
  GROUP BY
    client,
    page),           
    UNNEST([10, 25, 50, 75, 90]) AS percentile
GROUP BY
  percentile,
  client
ORDER BY
  percentile,
  client
```

rviscomi (changes requested):
Results for all queries, including suggestions, saved to the [chapter sheet](https://docs.google.com/spreadsheets/d/1M1tijxf04wSN3KU0ZUunjPYCrVsaJfxzuRCXUeRQ-YU/edit?usp=sharing).

One outstanding question about CDN chains that may require rerunning the query.

sql/2020/22_HTTP_2/percentage_of_h2_and_h3_sites_affected_by_cdn_prioritization.sql
gregorywolf: @rviscomi I understand your point and the short answer is I do not know the answer right now. I will have to speak with either @bazzadp or @paulcalvano to get their thoughts

bazzadp: Yeah I didn't fully get what those values were to be honest. Chained CDNs makes sense. The main one will be the one that the user hits - is that the first one in the last or the last one? @gregorywolf could you investigate a few examples?

Also are there many of these? If not maybe just ignore as noise. We never noticed this last year (and btw also missed out on WordPress last year - worth noting in the results @gregorywolf as imagine that alone could account for a measurable uptick).

bazzadp: Looks like 0.478% of sites are using chained CDNs. I don't think that's significant enough to worry about to be honest. And chaining will introduce all sorts of other prioritisation weirdness anyway. So I'd ignore.

rviscomi: @andydavies or @pmeenan would know more about the CDN value chain and what the positions mean.

See the [results sheet](https://docs.google.com/spreadsheets/d/1M1tijxf04wSN3KU0ZUunjPYCrVsaJfxzuRCXUeRQ-YU/edit#gid=1152953475) for a sense of how common this is. It's in the ~thousands of pages.

pmeenan: The position doesn't mean anything other than the order they were detected when looking at the CNAME and headers.

bazzadp: OK let's ignore but add a note that chained CDNs will not include correct results in this chapter.

To be honest, both would need to support this to get optimal usage (and even then it's questionable!) and only the "Cloudflare, Fastly" one will from a quick glance meaning 0.13% of sites are affected which is even smaller.

So I say ignore but make a comment on the sheet (maybe linking to this conversation?)

rviscomi (approved):

rviscomi:
With the CDN discussion resolved, all of the queries have been reviewed and their results saved to the sheet, so this PR should be ready to merge and the HTTP/2 chapter should now be ready for the content team to review the results and start writing! 🎉 📝

bazzadp (approved):
A few minor comments from me but happy to merge.

sql/2020/22_HTTP_2/number_of_h2_and h3_pushed_resources_and_bytes_transferred.sql
```diff
@@ -0,0 +1,37 @@
...
+    JSON_EXTRACT_SCALAR(payload, '$._was_pushed') = '1' AND
+    (LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "http/2" OR
+     LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "%quic%" OR
+     LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "h3%" OR
+     LOWER(JSON_EXTRACT_SCALAR(payload, "$._protocol")) LIKE "http/3%")   
+  GROUP BY
+    client,
+    http_version,
+    page),
+  UNNEST([10, 25, 50, 75, 90]) AS percentile
```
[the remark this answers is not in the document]
gregorywolf: Discussed with @rviscomi and decided to leave in place. The author can filter the pivot table if necessary

sql/2020/22_HTTP_2/number_of_h2_and_h3_pushed_resources_and_bytes_by_content_type.sql
```diff
@@ -0,0 +1,41 @@
...
+    type),
+  UNNEST([10, 25, 50, 75, 90]) AS percentile
+GROUP BY
+  percentile,
+  client,
+  http_version,
+  type
+ORDER BY
+  percentile,
+  client
```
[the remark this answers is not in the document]
gregorywolf: Discussed with @rviscomi and decided to leave in place. The author can filter the pivot table if necessary

Issue #1083, opened 2020-07-23
Add contributor section to 2020 landing page


rviscomi:
Now that we've published the 2020 [Contributors](https://almanac.httparchive.org/en/2020/contributors) page, we could add the related section to the [landing](https://almanac.httparchive.org/en/2020/) page to highlight the number of contributors.

Not having a "featured chapter" section makes the design a bit awkward because there's not a lot of room below the angled header background graphic. I've mocked up a version of the 2020 landing page without the `alt-bg` inverted styles on the contributor section:

![0 0 0 0_8080_en_2020_ (1)](https://user-images.githubusercontent.com/1120896/88313981-46292c00-cce2-11ea-9d10-a25226589610.png)

Open to other design ideas!

Related issue: #1082

bazzadp:
To save redesigning this, what about something else in the Featured Quote section instead?

What about quotes/praise about the 2019 edition? Didn't you have a list somewhere?

rviscomi:
That'd require some redesigning of the feature chapter section too, but that also works for me. Let me know which approach you'd prefer.

bazzadp:
Completed in #1383

Issue #48, opened 2019-06-15
Upgrade App Engine environment to Python 3


anoblet:
Does it make sense to upgrade to Python 3 as early as possible?

rviscomi:
It wouldn't be a bad idea before we write too much 2.7 code. Let me think about it.

rviscomi:
Yeah I think we should commit to 3 now. Especially given that this can become a multi-year project and 2 is being deprecated next year, we don't want to fall behind.

anoblet:
Is updating app.yaml the only thing left to do? I don't have much experience with Python/Flask and don't want to break anything.

rviscomi:
Everyone must also upgrade their local development environment to Python 3 to simulate the production environment..

1. Go to https://www.python.org/downloads/
2. Download and install 3.7.x
3. Run `python3 --version` to verify that it works

@anoblet Our scripts should also switch to using the explicit `python3` command rather than `python`.

anoblet:
For me, scripts defaulted to 3 (I guess because it was the first version I installed) I needed to specify -p in virtualenv to use python 2. I'm not sure I know the difference between a standard/flexible environment and might need a little time to understand the differences.

anoblet:
The only difference I can find between the current (flex) codebase and the standard codebase are the
```
env: flex
- url: /.*
  script: auto
```

rviscomi:
We change this:

```yaml
runtime: python
env: flex
runtime_config:
  python_version: 3
```

to this:

```yaml
runtime: python37
```

PR #1103, opened 2020-07-26
Move featured chapter quotes from Jinja2 template to JSON file

bazzadp:
This PR moves the featured chapter quotes to JSON which has several advantages:
- Allows dynamically changing the quote via JavaScript - currently the quote is randomised server side, which means it is cached for 3 hours and refreshing the page, does not show a different quote. That is fixed in this PR. Have left server-side randomisation in place too for those who don't have JavaScript.
- Reduces the amount of repeated template code in each language and consolidates it into the core `base/2019/index.html` template
- Allows easier generation of the quotes from the chapters as discussed in https://github.com/HTTPArchive/almanac.httparchive.org/issues/1016#issuecomment-663485421
- Removes need of an additional base_featured_quotes.html EJS template to do that generation.
- Allows easier possibility of other quotes in a similar fashion as discussed in https://github.com/HTTPArchive/almanac.httparchive.org/issues/1083#issuecomment-663123933


src/templates/base/2019/index.html
```diff
@@ -43,7 +43,100 @@ <h2>{{ self.intro_sub_title() }}</h2>
...
+      if (featured_quote) {
+        featured_quote.innerHTML = featured_chapters[randomChapter].quote;
+      }
+
+      var featured_titles = document.querySelectorAll('.featured-chapter-name');
+      for (var i = 0; i < featured_titles.length; i++) {
+        featured_titles[i].innerHTML = chapter_names[randomChapter];
+      }
+
+      for (let i=1; i <= 3; i++) {
```
SaptakS, line 115: Any specific reason we are using `let` only in this one instance whereas we use `var` almost everywhere else?

bazzadp: Good spot! Should be `var` to support older browsers. About to write a doc on this for all our new developers!!

src/main.py
```diff
@@ -190,6 +191,15 @@ def accentless_sort(value):
   return sorted(value, key=lambda i: strip_accents(i[1]).lower())
 
 
+def get_featured_quotes(lang, year):
+    featured_quotes = {}
+    featured_quote_filename = 'templates/%s/%s/featured_chapters.json' % (lang, year)
```
SaptakS, line 196: Any particular reason why we use `%s` against `.format()` ?

bazzadp: To be perfectly honest I wasn't even aware [this was a thing!](https://realpython.com/python-string-formatting/) (I'm not a Python developer by day!). It looks like this repo [originally started on Python 2](https://github.com/HTTPArchive/almanac.httparchive.org/issues/48) so maybe that's the reason?

Anyway, that's the convention currently used in this code base so let's not make this one line out of sync. If we want to change this, then we should do it in a separate PR across the code base - though not seeing a pressing need to do this.

src/templates/base/2019/index.html
bazzadp: Fixed. Though IE allows `let` in a for loop, which is why I didn't spot this earlier when testing in there! Still better to be consistent.

bazzadp:
> I notice that the `.json` files don't have a EOF line

Fixed!

src/main.py
SaptakS: > It looks like this repo originally started on Python 2 so maybe that's the reason?

Makes sense

> Anyway, that's the convention currently used in this code base so let's not make this one line out of sync. If we want to change this, then we should do it in a separate PR across the code base - though not seeing a pressing need to do this.

Agreed

src/templates/en/2019/featured_chapters.json
```diff
@@ -0,0 +1,183 @@
...
+    "quote": "HTTP/2 was the first major update to the main transport protocol of the web in nearly 20 years. It arrived with a wealth of expectations: it promised a free performance boost with no downsides. More than that, we could stop doing all the hacks and work arounds that HTTP/1.1 forced us into, due to its inefficiencies. Bundling, spriting, inlining, and even sharding domains would all become anti-patterns in an HTTP/2 world, as improved performance would be provided by default. This chapter examines how this relatively new technology has fared in the real world.",
+    "stat_1": "95%",
+    "stat_label_1": "Users who can use HTTP/2",
+    "stat_2": "27.83%",
+    "stat_label_2": "Requests with HTTP/2 prioritisation issues",
+    "stat_3": "8.38%",
+    "stat_label_3": "Sites supporting QUIC"
+   }
+  }
+
```
SaptakS, line 183:
```suggestion

```

I think due to whitespace in the last line, the EOF isn't detected.

bazzadp: Fixed thanks!

bazzadp:
Thanks for the review @SaptakS ! I’m pretty confident in regression issues.

@rviscomi wouldn’t mind you casting your eye over this if you have the time as a bit of a change in direction. And I feel the JavaScript is a bit verbose and you always have better ways to write this to look neater! For context this is setup for #1104

rviscomi:
Will do. I plan to get to this in a few days if there's no hurry.

src/templates/base/2019/index.html
```diff
@@ -43,7 +43,100 @@ <h2>{{ self.intro_sub_title() }}</h2>
...
+  {%- set random_chapter = featured_quotes.keys()|list()|random %}
+  {%- set featured_chapter_quotes = featured_quotes.get(random_chapter) %}
+  {%- if not featured_chapter_quotes.get('stat_1') %}{% set stat1_hidden = "hidden" %}{% endif %}
+  {%- if not featured_chapter_quotes.get('stat_2') %}{% set stat2_hidden = "hidden" %}{% endif %}
+  {%- if not featured_chapter_quotes.get('stat_3') %}{% set stat3_hidden = "hidden" %}{% endif %}
+  {% if featured_chapter_quotes %}
+  <script nonce="{{ csp_nonce() }}">
+    // Get a random quote using JS for those reloading cached CDN page.
+    // We do as much as possible before HTML so we can replace the server
+    // generated quote quickly without a flash of the server side quote
```
rviscomi, line 57: I'm not sold on this. The contributors page was one thing because I can see the value of giving each contributor equal prominence. But a chapter can be featured for a few hours at a time and I don't think there's any expectation from readers or authors that this should randomly update on refresh.

The overhead in both the HTML payload and the JS code just don't seem worth it for a feature that would be used infrequently and adds little to the UX.

If you're still convinced that this is the right change, could we continue this discussion in another issue and decouple it from the JSON change in this PR?

src/templates/en/2020/index.html
```diff
@@ -22,3 +22,6 @@
 
 <a href="https://github.com/HTTPArchive/almanac.httparchive.org#contributing" class="btn">Join the team</a>
 {% endblock %}
+
+{% block featured_chapter%}Featured Chapter{% endblock %}
```
rviscomi, line 26: The 2020 home page doesn't include a featured chapter section yet. Is this needed?

In a separate issue, could we also explore DRY solutions to reduce the duplication of language-specific text blocks across year templates?

src/templates/base/2019/index.html
bazzadp: Yeah you’re probably right. Having a chapter “featured” for a few hours is more expected. Was so preoccupied with whether or not I could that they didn't stop to think if I should 😁

Given that, should we even move to JSON or just stick with the Jinja2 template instead? We could still dynamically generate that with an EJS template like we do for the ebook. It would be an extra EJS template but as long as we use the .ejs.html naming convention we use for ebook that should be fine. Or do you think we will ever use these quotes elsewhere where having them in JSON format would be helpful?

bazzadp:
I'm going to close as decided to stick with templates.

PR #1110, opened 2020-07-27
Allow language and country code languages when choosing best site for visitor

bazzadp:
To support the different Chinese languages (`zh-CN` and `zh-TW`) first check language and country code rather than just language.

Fall back to language only (`zh`) if needs be but most browsers are configured to have both country specific, and then generic so this fallback is unlikely to be used. My `Accept-Language` header for example includes British English, American English and then just English:

```
Accept-Language: en-GB,en-US;q=0.9,en;q=0.8
```

This is only used when visiting the top level site (https://almanac.httparchive.org/) so extra processing time is not a concern.

### DIFF
--- a/src/validate.py
+++ b/src/validate.py
@@ -98,12 +98,21 @@ def parse_accept_language(header, supported_langs):
 
     if header is not None:
 
-        accepted_languages = re.findall(r'(?:^|\s|,)(\w+)', header)
-
+        # Looks for languages of the format en or en-US first
+        accepted_languages = re.findall(r'(?:^|\s|,)(\w+-?\w*)', header)
         logging.debug('Accepted languages: %s' % accepted_languages)
 
         # The header could contain multiple languages, in order of precedence
         # Limit the number of accepted languages tested to 10.
+        for lang in accepted_languages[:10]:
+            if lang in supported_langs:
+                # Return the first found supported language.
+                logging.debug('Using "%s" as the highest precedent language.' % lang)
+                return lang
+
+        # If that didn't find anything, then strip off the country code and try just the language
+        accepted_languages = re.findall(r'(?:^|\s|,)(\w+)', header)
+        logging.debug('Accepted languages (no country): %s' % accepted_languages)
         for lang in accepted_languages[:10]:
             if lang in supported_langs:
                 # Return the first found supported language.


src/validate.py
```diff
@@ -98,8 +98,8 @@ def parse_accept_language(header, supported_langs):
 
     if header is not None:
 
-        accepted_languages = re.findall(r'(?:^|\s|,)(\w+)', header)
-
+        # Looks for languages of the format en or en-US first
+        accepted_languages = re.findall(r'(?:^|\s|,)(\w+-?\w*)', header)
```
rviscomi, line 102: Is this a thing? Could we capture both lang and lang-locale at the same time?
```suggestion
        accepted_languages = re.findall(r'(?:^|\s|,)((\w+)-?\w*)', header)
```

bazzadp: Fraid not:

```
DEBUG:root:Trying to extract Accept-Language header.
DEBUG:root:Accepted languages: [('en-GB', 'en'), ('en-US', 'en'), ('en', 'en')]
DEBUG:root:Accepted languages (no country): ['en', 'en', 'en']
DEBUG:root:Using "en" as the highest precedent language.
```
I did think about functionalising the for loop but it's really quite small so figured more readable to leave as is.

The other option is something like this:

```python
        # The header could contain multiple languages, in order of precedence
        # Limit the number of accepted languages tested to 10.
        for lang in accepted_languages[:10]:
            if lang in supported_langs:
                # Return the first found supported language.
                logging.debug('Using "%s" as the highest precedent language and country.' % lang)
                return lang
            if '-'  in lang:
                # If language-country then test language only
                lang_only = lang[:lang.find('-')]
                if lang_only != [] and lang_only in supported_langs:
                    # Return the first found supported language.
                    logging.debug('Using "%s" as the highest precedent language.' % lang)
                    return lang_only
```

But more unreadable, and also means that:

```
Accept-Language: zh-TW,en;q=0.9,zh;q=0.8
```

Would accept `zh` over `en`. Probably are getting into edge cases there, but still. Think current code is most reable.

Plus the second regex will rarely (if ever) be called as most people set up browsers with the country-less language anyway as I said.

rviscomi: This still feels like an easy regex fix. Maybe something like this? (untested)

```suggestion
        accepted_languages = re.findall(r'(?:^|\s|,)(\w+-?\w*|\w+)', header)
```

bazzadp: I am finding the country-less languages already.

My Regex is:
```python
accepted_languages = re.findall(r'(?:^|\s|,)(\w+-?\w*)', header)
```
Note the second `?` and the `*` meaning the entire dash part is optional:

This means, this matches the following:
```
Accept-Language: en-GB,en-US;q=0.9,en;q=0.8
```
to this:
```
Accepted languages: ['en-GB', 'en-US', 'en']
```

The thing we're trying to solve with the second for loop is if you have this:
```
Accept-Language: en-GB,en-US;q=0.9
```
That is, where you don't have *just* a language set without a country code. That works currently but will break with my new change without the second for loop.

Alternatively we could do the following:

```python
        accepted_languages_country = re.findall(r'(?:^|\s|,)(\w+-?\w*)', header)
        accepted_languages_no_country = re.findall(r'(?:^|\s|,)(\w+)', header)
        accepted_languages = accepted_languages_country + accepted_languages_no_country
        logging.debug('Accepted languages: %s' % accepted_languages)
```
and then just have one loop. Less code, but a (very slight!) penalty to those that don't need the second regex.

rviscomi: Ok thanks for explaining

PR #1111, opened 2020-07-27
Adds nowrap whitespace for japanese language in contributor-team style

### LINKED ISSUE #1074: Team name word wrapping in Japanese Contributors page
For example, some [contributors](https://almanac.httparchive.org/ja/2019/contributors) are members of multiple teams and the team names span multiple lines:

image

On Katie and Una's tiles the team `ブレインストーマー` is broken up onto multiple lines. Would it make more sense in Japanese to set the `white-space: nowrap;` style on the `.contributor-team` rule, so that the characters in the team name are all rendered to the same line, like this:

image

This could also be useful for other languages with longer translations.

SaptakS:
Fixes #1074 

Adds nowrap whitespacing for japanese language in contributor-team styling. I have used a `in` logic instead of `==` logic so that if we need to add same style for more languages at a later stage we can just add that in the list.

### DIFF
--- a/src/templates/base/2019/contributors.html
+++ b/src/templates/base/2019/contributors.html
@@ -147,6 +147,7 @@
   font-size: 14px;
   font-size: 0.875rem;
   color: #757575;
+  white-space: nowrap;
 }
 .contributor-team::after {
   content: ', ';
@@ -154,11 +155,6 @@
 .contributor-team:last-child::after {
   content: '';
 }
-{% if language.lang_attribute in ['ja'] %}
-.contributor-team {
-  white-space: nowrap;
-}
-{% endif %}
 
 {% for id in config.teams.keys() %}
 .contributors.{{ id }} .contributor.{{ id }} {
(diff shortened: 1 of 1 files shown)


src/templates/base/2019/contributors.html
```diff
@@ -154,6 +154,11 @@
 .contributor-team:last-child::after {
   content: '';
 }
+{% if language.lang_attribute in ['ja'] %}
```
bazzadp, line 157: Is there any harm to doing this for every language?

And can we add to existing definition on line 146?

SaptakS: We can actually. Given that we never do want it to wrap to the next line. And in that case definitely can be moved to line 146

SaptakS: Done

catalinred: Something like: 
[lang="ja"] .contributor-team  {
  white-space: nowrap;
}
would be just enough in my opinion.

rviscomi: It may not be needed for non-JA languages but I agree with the approach taken here to apply the styles evenly.

PR #1112, opened 2020-07-27
Removes copyright year from footer

### LINKED ISSUE #948: Remove the year from the footer copyright
@bazzadp what do you think is the best way to do this? The 2020 template extends 2019, which sets ©️ 2019 in each lang's localized base template.

I think it's ok if the copyright year was defined in one central place and all pages/langs/editions inherit the latest year. Bonus points if we don't need to manually increment it each year.

SaptakS:
Fixes #948

### DIFF
--- a/src/config/2020.json
+++ b/src/config/2020.json
@@ -1154,6 +1154,16 @@
       "avatar_url": "https://avatars1.githubusercontent.com/u/16757512?v=4&s=200",
       "github": "sboukortt"
     },
+    "saptaks": {
+      "name": "Saptak Sengupta",
+      "teams": [
+        "developers"
+      ],
+      "avatar_url": "https://avatars2.githubusercontent.com/u/9530293?v=4&s=200",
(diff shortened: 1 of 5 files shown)


bazzadp:
Good stuff! Can you submit a PR to add yourself to `src/config/2020.json` in `development` team? They are in alphabetical order by first name btw.

You can do this either as part of this PR or as a separate one.

rviscomi:
_Aside: @bazzadp can we have a linter that checks that the contributor was added in the right sort order? 😅_

bazzadp:
> _Aside: @bazzadp can we have a linter that checks that the contributor was added in the right sort order? 😅_

My OCD suffices for that for now 😁

SaptakS:
> Good stuff! Can you submit a PR to add yourself to `src/config/2020.json` in `development` team? They are in alphabetical order by first name btw.
> 
> You can do this either as part of this PR or as a separate one.

I will add it as part of this PR itself.

PR #1114, opened 2020-07-27
Fix contributor numbers when more than 100 of them

### LINKED ISSUE #1082: Optimize home page contributor section for 3 digits
We have a good problem this year of having over 100 contributors. The 2019 edition had "only" 92 contributors, so the UI only had to handle 2 digits. By expanding to 3 digits in 2020, some of the [contributor section](https://almanac.httparchive.org/en/2019/#contributors) styles may need to be optimized. As we continue to add translations of the 2019 edition, it may also end up exceeding 100 contributors as some point.

image

We have four breakpoints for various viewport widths, and the styling of this section changes in each one. For some in-between widths like 1300px shown above, the button style breaks. We should set `white-space: nowrap;` on the `.btn` styles to avoid this.

More generally, it seems like the `letter-spacing` of the `people-number` class could be improved as well as some of the relative positioning.

It may be a good idea to dynamically add something like a `three-digit` class to better differentiate between large and small numbers in the styles.

bazzadp:
Fixes #1082

### DIFF
--- a/src/static/css/index.css
+++ b/src/static/css/index.css
@@ -192,6 +192,10 @@ p {
   padding-left: 180px;
 }
 
+.contributors .btn {
+  white-space: nowrap;
+}
+
 .people {
   position: relative;
   height: 400px;
@@ -405,7 +409,7 @@ p {
   }
 
   .people-number {
-    font-size: 300px;
+    font-size: 200px;
     overflow: visible;
     line-height: 400px;
     top: 50px;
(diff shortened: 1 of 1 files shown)


rviscomi:
Could you share a few screenshots of how it appears with different breakpoints?

bazzadp:
Sure.

With this change:

image
image
image
image
image

image

Production:

image


image

image

image

image

image

bazzadp:
I changed breakpoint of second one to make the big numbers smaller quicker. So this now looks like this:

image

rviscomi:
Looks great! Yeah your change makes sense to give the text more room.

PR #1121, opened 2020-07-29
Add Python Test GitHub Action

bazzadp:
Run pytest on any Pull Request that impacts our Python code to avoid issues like #1120

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -1,0 +1,28 @@
+name: Run Python Tests
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - src/config/**
+      - src/**.py
+      - src/requirements.txt
+jobs:
+  pytest:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+      - name: Set up Python 3.x
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+      - name: Install Requirements
+        run: |
+          python -m pip install --upgrade pip
+          cd src
+          pip install -r requirements.txt
+      - name: Run pytest
+        run: |
+          cd src
+          pytest
+  
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -12,10 +12,10 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
-      - name: Set up Python 3.x
+      - name: Set up Python 3.7
         uses: actions/setup-python@v2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/python-tests.yml
```diff
@@ -0,0 +1,28 @@
+name: Run Python Tests
+on:
```
ibnesayeed, line 2: This will not perform tests on each push, which means contributors will not know about any failing tests until they create a PR or trigger the action manually. I do not know how desired this behavior is, but I wanted to note my observations. For example, in code linting workflow we covered these situations.

.github/workflows/python-tests.yml
```diff
@@ -0,0 +1,28 @@
...
+        with:
+          python-version: '3.7'
+      - name: Install Requirements
+        run: |
+          python -m pip install --upgrade pip
+          cd src
+          pip install -r requirements.txt
+      - name: Run pytest
+        run: |
+          cd src
```
ibnesayeed, line 27: I am not too sure if we need to change the working directory again as it was done in the previous step. If the directory context state is not preserved across steps then this will be needed, but I am don't know how the system behaves. You might want to test it somewhere else in a small sample workflow to confirm it.

.github/workflows/python-tests.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.7'
+      - name: Install Requirements
+        run: |
+          python -m pip install --upgrade pip
```
ibnesayeed, line 22: Do we really need to upgrade pip here? If the system can work with what `setup-python` ships with, then we can avoid yet another process that involves network requests.

.github/workflows/python-tests.yml
bazzadp: Good point. Had that originally and then decided to simplify but you're right - it's a good change so will put back.

.github/workflows/python-tests.yml
bazzadp: I tested it and we do need to do this. Alternative is to merge into same step but quite like them being separate so can see exactly which step fails more easily.

.github/workflows/python-tests.yml
bazzadp: I was taking it from their documentation: https://docs.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions
It's fairly quick so happy to leave in.

.github/workflows/python-tests.yml
bazzadp: Fixed!

PR #1123, opened 2020-07-29
Add action to auto generate Ebooks

bazzadp:
This auto generates the ebooks when the GitHub action is manually trigger.
This removes the need to install PrinceXML locally.
This should be run before deploying the site to production.


bazzadp:
Have had a change of heart on this. Generating the ebooks on *every* commit to main is overkill as this only needs to be done before a release. So changing to a manually Trigger Github action instead. I've updated the initial comment to reflect this.

Still intend to run the site and run tests in the many Generate Chapters GitHub action but don't think we need the performance hit of generating the ebooks in that.

.github/workflows/generate_ebooks.yml
```diff
@@ -0,0 +1,77 @@
...
+      with:
+        version:  12.x
+    # Install Python 3.7 as used by Google Cloud Platform
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.7'
+    - name: Install Python requirements
+      run: |
+        echo "Installing Python requirements"
```
ibnesayeed, line 35: Name of the step makes it obvious, no need to echo it explicitly.

```suggestion
```

.github/workflows/generate_ebooks.yml
```diff
@@ -0,0 +1,77 @@
...
+        python -m pip install --upgrade pip
+        cd src
+        pip install -r requirements.txt
+    - name: Install node modules
+      run: |
+        echo "Running npm install"
```
ibnesayeed, line 41: Name of the step makes it obvious, no need to echo it explicitly.

```suggestion
```

.github/workflows/generate_ebooks.yml
```diff
@@ -0,0 +1,77 @@
...
+        cd src
+        npm install
+    - name: Install PrinceXML
+      run: |
+        echo "Installing PrinceXML"
```
ibnesayeed, line 46: Name of the step makes it obvious, no need to echo it explicitly.

```suggestion
```

.github/workflows/generate_ebooks.yml
```diff
@@ -0,0 +1,77 @@
...
+        cd /tmp
+        sudo apt-get install -y aptitude
+        sudo aptitude install -y gdebi
```
ibnesayeed, line 49: Is it necessary to install `aptitude` first and then use it? Why can't we just use `apt`?

.github/workflows/generate_ebooks.yml
```diff
@@ -0,0 +1,77 @@
...
+        wget https://www.princexml.com/download/${{ env.PRINCE_PACKAGE }}
+        sudo gdebi --non-interactive ${{ env.PRINCE_PACKAGE }}
+    - name: Run website
+      run: |
+        echo "Running website"
```
ibnesayeed, line 54: Name of the step makes it obvious, no need to echo it explicitly.

```suggestion
```

.github/workflows/generate_ebooks.yml
```diff
@@ -0,0 +1,77 @@
...
+        cd src
+        python main.py background &
```
ibnesayeed, line 56: Did you test it that processes running in background persist across steps?

.github/workflows/generate_ebooks.yml
```diff
@@ -0,0 +1,77 @@
...
+    - name: Generating Ebooks
+      run: |
+        echo "Generating ebooks"
```
ibnesayeed, line 59: Name of the step makes it obvious, no need to echo it explicitly.

```suggestion
```

.github/workflows/generate_ebooks.yml
```diff
@@ -0,0 +1,77 @@
...
+        commit-message: Generate Ebooks
+        body: |
+          Regenerated Ebooks through GitHub action
+          - Auto-generated by [create-pull-request][1] GitHub Action
+
+          [1]: https://github.com/peter-evans/create-pull-request
+        labels: generate chapters
+    - name: Check outputs
+      run: |
+        echo "Pull Request Number - ${{ steps.cpr.outputs.pull-request-number }}"
```
ibnesayeed, line 77: This can go in a single line.

```suggestion
      run: echo "Pull Request Number - ${{ steps.cpr.outputs.pull-request-number }}"
```

.github/workflows/generate_ebooks.yml
bazzadp: GitHub actions have this annoying thing where they list all the run commands under the first command instead of under the name:

image

When expanded you can see the real commands:

image

The Echo was an attempt to at least make that grouping name more meaingful:

image

image

Any ideas for a better way to solve this?

.github/workflows/generate_ebooks.yml
bazzadp: That's the recommended way to install PrinceXML: https://www.princexml.com/doc/installing/#installing-prince-on-linux-freebsd

.github/workflows/generate_ebooks.yml
bazzadp: Yes. This action has been well tested. Took me ages to get it working!! 😀

.github/workflows/generate_ebooks.yml
ibnesayeed: I usually set a secret named `ACTIONS_STEP_DEBUG` and set it to `true` so that I can see more detailed output when I need it. Otherwise, in a normal code run knowing that all the steps passed is more than enough.

bazzadp: OK removed all the `echo`s.

.github/workflows/generate_ebooks.yml
ibnesayeed: Honestly speaking, I would not go through this lengthy route. I would dome something like the following that will take care of any dependencies as well:

```
# Set the package file name as ENV in the beginning
PRINCE_PACKAGE=prince_13.5-1_ubuntu18.04_amd64.deb

cd /tmp
wget https://www.princexml.com/download/prince_13.5-1_ubuntu18.04_amd64.deb
DEBIAN_FRONTEND=noninteractive sudo apt install -y /tmp/${{ env.PRINCE_PACKAGE }}
```

bazzadp: This seems to work too - so switched to that as a lot quicker. Can switch back if it proves problematic. Thanks!

PR #1124, opened 2020-07-30
Automate website status tests

bazzadp:
Makes progress on #933 

This adds a script `npm run test` to test the local website with every URL in the sitemap, and a few others (the sitemap itself, some URLs that should redirect, and a 404 page).

It also make this script run with the Generate GitHub Action so the tests are run on every merge to main.

It also allows the Generate action to be run manually incase you want to test the site. I'm leaving the name the same [until GitHub allow you to delete old ones!](https://github.community/t/delete-old-workflow-results/16152/16)

### DIFF
--- a/src/tools/test/test_status_codes.js
+++ b/src/tools/test/test_status_codes.js
@@ -7,7 +7,7 @@ const base_url = "http://127.0.0.1:8080";
 let failures = 0;
 let passes = 0;
 
-const test_status_code = async (page,status,location) => {
+const test_status_code = async (page, status, location) => {
 
   if (location == undefined) {
     location = null;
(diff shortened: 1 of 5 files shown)


src/tools/test/test_status_codes.js
```diff
@@ -0,0 +1,73 @@
...
+    failures++;
+  }
+};
+
+const test_sitemap_pages = async (page,code) => {
+  const xml = await fs.readFile(`templates/sitemap.xml`, 'utf-8');
+  const sitemap = JSON.parse(convert.xml2json(xml,{compact: true}));
+  const urls = sitemap['urlset']['url'];
+  for ( var url in urls ) {
+    page = urls[url]['loc']['_text'];
```
ibnesayeed, line 47: Should we trim any whitespaces, if present? (XML text node might be in a different line, which may not be the case if we know how we are generating the sitemap, but I guess there is not harm to harden this):

```suggestion
    page = urls[url]['loc']['_text'].trim();
```

src/tools/test/test_status_codes.js
```diff
@@ -0,0 +1,73 @@
...
+    page = page.replace('https://almanac.httparchive.org','');
```
ibnesayeed, line 48: Is it okay to hardcode the domain name for replacement here or should we attempt to add a regex matcher, in case the domain name changes in the future?

src/tools/test/test_status_codes.js
bazzadp: Good call. We generate the sitemap, but no harm to add this in case we change the format.

src/tools/test/test_status_codes.js
bazzadp: I think that's a bit overkill. Changing the domain will be a big deal, that will require a lot of changes so happy to leave this as is.

PR #1128, opened 2020-07-30
Fix Chinese language selector

bazzadp:
We currently have this:

Drop down arrow covers Chinese language

When we should have this:

Drop down arrow to right of Chinese language

Wish there was a way of coding this based on length of current option selected, but `select` elements has very little styling support. Hopefully be fixes properly by #986 but this'll do for now.

### DIFF
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -669,6 +669,10 @@ header .cta {
   min-width: 100px;
 }
 
+.language-switcher select {
+  min-width: 120px;
+}
+
 .language-switcher select::-ms-expand,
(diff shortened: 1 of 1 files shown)


rviscomi:
Weird! What is it about these characters that make the `<select>` shrink?

bazzadp:
It's not so much the `<select>` shrinking - it's the fact we have the drop down arrow as an `::after` element and use absolute position to bring it into the `<select>` circle. So it is a bit of a magic number to give it that space but not too much space.

Plus `<selects>` really are the unloved children of the CSS world!

Issue #902, opened 2020-06-27
Fonts 2020


obto:
# Part I Chapter 4: Fonts

## Content team
| Authors | Reviewers | Analysts | Draft | Queries | Results |
| ------- | --------- | -------- | ----- | ------- | ------- |
| @raphlinus  | @malchata @RoelN @notwaldorf @mandymichael @svgeesus @rsheeter @jpamental @davelab6 | @AbbyTsai | [Doc][~draft-doc] | [*.sql][~sql-dir] | [Sheet][~results-sheet] |

**Content team lead:** @jpamental

Welcome chapter contributors! You'll be using this issue throughout the [chapter lifecycle](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle) to coordinate on the content planning, analysis, and writing stages.

The **content team** is made up of the following contributors:

- [Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)
- [Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)
- [Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)

**New contributors:** If you're interested in joining the content team for this chapter, just leave a comment below and the content team lead will loop you in.

_Note: To ensure that you get notifications when tagged, you must be "[watching](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#notifications)" this repository._

### [**0. Form the content team**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#0-create-content-team)
+ [x] **Jul 6th**: Project owners have selected an author to be the content team lead
+ [x] **Jul 13th**: The content team has at least one author, reviewer, and analyst (minimally viable team formed)

### [**1. Plan content**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
+ [x] **Jul 20th:** The content team has completed the chapter outline in the [draft doc][~draft-doc]
+ [x] **Jul 27th**: Analysts have triaged the feasibility of all proposed metrics

### [**2. Gather data**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
[...]

Issue #991, opened 2020-07-10
Lint SQL


rviscomi:
The BigQuery SQL files (eg [01_01a.sql](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2019/01_JavaScript/01_01a.sql)) for each metric are written by many analysts. It would be great to automatically enforce an SQL style guide with the linter, so that discrepancies can be caught by analysts in the PRs. Not sure how much control/customization we have at our disposal.

@paulcalvano can be in charge of defining what the SQL style guide should be.

Related: #938

bazzadp:
It’s not in the Github Superlinter yet which would make this easier. Issue to follow: https://github.com/github/super-linter/issues/128

Still may be worth looking at the linters proposed and considering the rules we want.

rviscomi:
@fhoffa this is a hail mary, but does BigQuery have an open source linter/formatter?

rviscomi:
Spoke with Felipe offline and BigQuery doesn't have anything open source to help with this.

This would be nice to have for the 2020 queries, but not required. At a minimum we should have some kind of style guide or reference doc in the wiki to help analysts understand what we're looking for. @paulcalvano is that doable?

bazzadp:
Good news - The [GitHub Superlinter just merged in sql-lint](https://github.com/github/super-linter/pull/528).

Not so good news, is that it is not [currently configurable](https://github.com/github/super-linter/blob/master/docs/disabling-linters.md#sql) so we cant turn off rules we don't like - which might be an issue for some of the weird SQL we use in Big Query. However [sql-lint just added that facility](https://github.com/joereynolds/sql-lint/issues/155) but not tagged in a release yet.

I tried to run it locally and it's complaining about functions, which looks to have been raised as [an issue with sql-lint](https://github.com/joereynolds/sql-lint/issues/154) so will monitor that issue and see if it fixes for us when resolved.

So progress, but still a bit to go. Would be lovely to get this in, in the next month but not sure how realistic that it.

joereynolds:
Hey @bazzadp,

I'm working on fixing these issues (I'm the dev on `sql-lint`). Bit of a busy time for me but it's on the way, hope that gives some reassurance!

bazzadp:
Thanks for responding @joereynolds ! Great to have that reassurance. I’m watching those so should be informed of updates whenever you get a chance!

rviscomi:
[...]

PR #1133, opened 2020-07-31
Fonts 2020 queries 

AbbyTsai:
Queries for Fonts 2020 chapter(#902)
Adoption / font usage
- [x] 04.01 web fonts usage 
- [x] 04.02 web fonts usage by country
- [x] 04.03 popular font host by country

Source / font host  /  FCP, LCP
- [x] 04.04 Local vs Host 
- [x] 04.05 Web fonts usage breakdown 

Implementation  /  FCP, LCP
- [x] 04.06 Font-display 
- [x] 04.07 resource-hints 
- [x] 04.08 unicode-range
- [x] 04.09 subset   

Feature
- [x] 04.10 font formats
- [x]  04.10a font formats font face
- [x] 04.11 popular typeface by country
- [x] 04.11a popular typeface

font-face
  - [x] 04.12 font-weight / font-style / font-stretch

Variable fonts
- [x] 04.14 variable font usage 
- [x]  04.14a variable font comparison with FCP
- [x] 04.17 axes value in variable font
- [x] 04.18 variable fonts with animation 

Fonts
- [x] 04.19 Color font format
- [x] 04.20 Icon font
- [x] 04.21 OpenType axis


bazzadp (changes requested):
I'll leave someone who knows the HTTP Archive better than me to comment on the queries themselves but could you give all the filenames .sql extensions so GitHub (and other editors) does some syntax highlighting which will make reviewing easier? And also if we ever adding SQL linting as described in #991 

Also could you remove spaces from the filenames as might reference them from the web?

@paulcalvano I've added these two points to the [analysts wiki](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide) but let us know if you disagree.

AbbyTsai:
After a happily completed test running of 21 queries, I found 7 of those have been delivered by 0 result, in which some might be due to lack of font data of august, so expected them have to be run in September, however some might have problems, including another 2 with null. Then I would love to have your review to help me see the different side. Thanks.

bazzadp:
> After a happily completed test running of 21 queries, I found 7 of those have been delivered by 0 result, in which some might be due to lack of font data of August, so expected them have to be run in September, however some might have problems, including another 2 not working well. Then I would love to have your review to help me see the different side. Thanks.
> 0 result : 7, 9, 14, 15, 18, 19, 21 (vf related, resource-hints, subsetting, color font, openType)
> Not working : 3,12 (popular font by country, font weight)

Probably a good idea to edit the first comment and make a comment against each of the issues you'd like help with. That way you can update and remove the comment as those are sorted.

sql/2020/04_Fonts/04_07.font_resource_hints_with_fcp.sql
```diff
@@ -0,0 +1,88 @@
+#standardSQL
```
paulcalvano, line 1: This returns 0 results.

paulcalvano (commented):
The following queries are reviewed:

  - 04.02
  - 04.03
  - 04.04
  - 04.05
  - 04.10

04.06, 04.07, 04.08, 04.09, 04.11, 04.12, 04.13 
 - These look correct to me, but the JavaScript that is being used to parse the custom metrics is a bit over my head.  @rviscomi is going to help me finish reviewing these tomorrow.
 - These queries run successfully, but 04.07 does not return any results.
 
Will review the rest tomorrow as well.  Sorry this is taking so long.

sql/2020/04_Fonts/04_06.font_display_with_fcp.sql
```diff
@@ -0,0 +1,71 @@
...
+    return [];
+}
+''';
+
+SELECT
+  client,
+  font_display,
+  COUNT(DISTINCT page) AS freq_display,
+  total_page,
+  ROUND(COUNT(DISTINCT page) * 100 / total_page, 2) AS pct_displa
```
paulcalvano, line 35:
```suggestion
  ROUND(COUNT(DISTINCT page) * 100 / total_page, 2) AS pct_display,
```

sql/2020/04_Fonts/04_02.web_fonts_usage_by_country.sql
```diff
@@ -0,0 +1,27 @@
...
+ON 
+  CONCAT(origin, '/') = url
+WHERE 
+  bytesFont IS NOT NULL
+GROUP BY
+ client,
+ country
+ORDER BY
+ client,
+ country
```
rviscomi, line 27: When joining data from the Chrome UX Report and HTTP Archive datasets, also be sure to join by the device/client.

This suggestion includes that change plus some minor whitespace fixes and the`ROUND`/kbytes feedback from the previous query.

```suggestion
#standardSQL
#web_fonts_usage_by_country
SELECT
  _TABLE_SUFFIX AS client,
  country,
  COUNT(0) AS freq_url,
  APPROX_QUANTILES(bytesFont, 1000)[OFFSET(500)] / 1024 AS median_font_kbytes
FROM (
  SELECT
    DISTINCT origin,
    device,
    `chrome-ux-report`.experimental.GET_COUNTRY(country_code) AS country
  FROM 
    `chrome-ux-report.materialized.country_summary`
  WHERE 
    yyyymm = 202008)
JOIN 
  `httparchive.summary_pages.2020_08_01_*`
ON 
  CONCAT(origin, '/') = url AND
  IF(device = 'desktop', 'desktop', 'mobile') = _TABLE_SUFFIX
WHERE 
  bytesFont IS NOT NULL
GROUP BY
  client,
  country
ORDER BY
  client,
  country
```

sql/2020/04_Fonts/04_03.popular_font_hosts_by_country.sql
```diff
@@ -0,0 +1,42 @@
...
+  JOIN (
+    SELECT
+      DISTINCT origin,
+      `chrome-ux-report`.experimental.GET_COUNTRY(country_code) AS country
+    FROM
+      `chrome-ux-report.materialized.country_summary`
+    WHERE
+      yyyymm=202008)
+  ON
+    CONCAT(origin, '/')=page
```
rviscomi, line 32: Also join on device/client

(This feedback applies to all CrUX queries in the PR)

sql/2020/04_Fonts/04_03.popular_font_hosts_by_country.sql
```diff
@@ -0,0 +1,42 @@
...
+FROM 
+(
+  SELECT
+    client,
+    country,
+    NET.HOST(url) AS url_font,
+    COUNT(0) AS freq_font,
+    SUM(COUNT(0)) OVER (PARTITION BY country) AS country_total,
+    ROUND(COUNT(0)/SUM(COUNT(0)) OVER (PARTITION BY country),2) AS pct_country,
+    ROW_NUMBER() OVER (PARTITION BY client, country ORDER BY COUNT(0) DESC) AS rank
```
rviscomi, line 20: This appears to be unused

sql/2020/04_Fonts/04_04.local_vs_host_with_fcp.sql
```diff
@@ -0,0 +1,36 @@
...
+    WHEN NET.HOST(page) != NET.HOST(url) THEN "host"     
+    ELSE "local"
+  END as host_local,
+  COUNT(0) AS freq,
+  SUM(COUNT(0)) OVER (PARTITION BY client) AS total,
+  ROUND(COUNT(0)*100/SUM(COUNT(0)) OVER (PARTITION BY client),2) AS pct,
+  ROUND(COUNTIF(fast_fcp>=0.75)*100/COUNT(0),0) AS pct_fast_fcp,
+  ROUND(COUNTIF(NOT(slow_fcp >=0.25)
+      AND NOT(fast_fcp>=0.75))*100/COUNT(0),0) AS pct_moderate_fcp,
+  ROUND(COUNTIF(slow_fcp>=0.25)*100/COUNT(0),0) AS pct_slow_fcp,
```
rviscomi, line 15: Nit: use the [Core Web Vitals](web.dev/vitals) terminology of good, ni (needs improvement), and poor

Also whitespace and `ROUND` suggestions.
```suggestion
  COUNTIF(fast_fcp >= 0.75) / COUNT(0) AS pct_good_fcp,
  COUNTIF(NOT(slow_fcp >= 0.25)
      AND NOT(fast_fcp >= 0.75)) / COUNT(0) AS pct_ni_fcp,
  COUNTIF(slow_fcp >= 0.25) / COUNT(0) AS pct_poor_fcp,      
```

sql/2020/04_Fonts/04_05.web_font_usage_breakdown_with_fcp.sql
```diff
@@ -0,0 +1,34 @@
+#standardSQL
+#web_font_usage_breakdown_with_fcp
+SELECT
+  client,
+  NET.HOST(url) AS host,
+  COUNT(0) AS freq_host,
+  SUM(COUNT(0)) OVER(PARTITION BY client) AS TOTAL,
+  ROUND(COUNT(0)*100/SUM(COUNT(0)) OVER(PARTITION BY client), 2) AS pct_host,
+  ROUND(COUNTIF(fast_fcp>=0.75)*100/COUNT(0),0) AS pct_fast_fcp,
```
rviscomi, line 9: Same feedback as previous query

sql/2020/04_Fonts/04_06.font_display_with_fcp.sql
```diff
@@ -0,0 +1,71 @@
...
+  COUNT(DISTINCT page) AS freq_display,
+  total_page,
+  ROUND(COUNT(DISTINCT page) * 100 / total_page, 2) AS pct_display,
+  ROUND(COUNTIF(fast_fcp>=0.75)*100/COUNT(0),0) AS pct_fast_fcp_display,
+  ROUND(COUNTIF(NOT(slow_fcp >=0.25)
+    AND NOT(fast_fcp>=0.75))*100/COUNT(0),0) AS pct_mode_fcp_display,
+  ROUND(COUNTIF(slow_fcp>=0.25)*100/COUNT(0),0) AS pct_slow_fcp_display,
+FROM
+  `httparchive.almanac.parsed_css`,
+  UNNEST(getFontDisplay(css)) AS font_display
```
rviscomi, line 42: This will create one row for every `font-display` on a page, which will count the same page's FCP multiple times when grouping by the `font_display` value.

I think the way to address this would be to deduplicate values per page before aggregating.
```suggestion
FROM (
  SELECT
    DISTINCT client,
    page,
    font_display
  FROM
    `httparchive.almanac.parsed_css`,
    UNNEST(getFontDisplay(css)) AS font_display
  WHERE
    date = '2020-08-01')
```

sql/2020/04_Fonts/04_06.font_display_with_fcp.sql
```diff
@@ -0,0 +1,71 @@
...
+    fast_fcp,
+    slow_fcp,
+  FROM
+    `chrome-ux-report.materialized.metrics_summary`
+  WHERE
+    yyyymm=202008)
+ON
+  CONCAT(origin, '/')=page
+WHERE
+  date='2020-08-01'
```
rviscomi, line 65: This is handled by the suggestion above
```suggestion
```

sql/2020/04_Fonts/04_07.font_resource_hints_with_fcp.sql
```diff
@@ -0,0 +1,88 @@
...
+#font_resource_hints_with_fcp(??NoResult)
+CREATE TEMPORARY FUNCTION getResourceHints(payload STRING)
+RETURNS ARRAY < STRUCT < name STRING, href STRING >>
+    LANGUAGE js AS '''
+var hints = new Set(['preload', 'prefetch', 'preconnect', 'prerender', 'dns-prefetch']);
+try {
+    var $ = JSON.parse(payload);
+    var almanac = JSON.parse($._almanac);
+    return almanac['link-nodes'].reduce((results, link) => {
```
rviscomi, line 10: The query is returning 0 results because this line is calling `reduce` on something that isn't an array, so the `catch` handler below returns an empty array.

```suggestion
    return almanac['link-nodes'].nodes.reduce((results, link) => {
```

sql/2020/04_Fonts/04_07.font_resource_hints_with_fcp.sql
```diff
@@ -0,0 +1,88 @@
...
+    _TABLE_SUFFIX)
+USING
+  (client)
+WHERE
+  NET.HOST(hint_url)=NET.HOST(url)
+  AND SUBSTR(hint_url,0,5)=SUBSTR(url,0,5)
+GROUP BY
+  client,
+  name,
+  url,
```
rviscomi, line 83:
```suggestion
```

rviscomi (changes requested):
Hi @AbbyTsai great work so far! I've only reviewed the first 7 queries but I wanted to pass it back to you now for early feedback. Some comments apply to all queries, like removing `ROUND` and some Chrome UX Report best practices.

sql/2020/04_Fonts/04_01.web_fonts_usage.sql
```diff
@@ -0,0 +1,28 @@
+#standardSQL
+#web_fonts_usage
+SELECT
+IF
+  (ENDS_WITH(_TABLE_SUFFIX,'desktop'),
+    'desktop',
+    'mobile') AS client,
+  CAST(SUBSTR(_table_suffix, 1, 4) AS int64) year,
+  CAST(SUBSTR(_table_suffix, 6, 2) AS int64) month,
```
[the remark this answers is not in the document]
AbbyTsai: Great tip to learn. thanks for sharing.

sql/2020/04_Fonts/04_01.web_fonts_usage.sql
```diff
@@ -0,0 +1,28 @@
...
+  COUNTIF(reqFont>0) AS freq_fonts,
+  COUNT(0) AS total,
+  ROUND(COUNTIF(reqFont>0) * 100 / COUNT(0),2) AS pct_fonts,
+  ROUND(APPROX_QUANTILES(bytesFont, 1000)[
+  OFFSET
+    (500)]/1024,2) AS median_byteFont,
```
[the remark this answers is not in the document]
AbbyTsai: Thanks to the warm explanation.

sql/2020/04_Fonts/04_02.web_fonts_usage_by_country.sql
AbbyTsai: Thank you to point out the procedure of combining CrUX report and HTTP Archive dateset.

sql/2020/04_Fonts/04_04.local_vs_host_with_fcp.sql
AbbyTsai: Thanks for the clarification.

sql/2020/04_Fonts/04_06.font_display_with_fcp.sql
AbbyTsai: Thank you to give the warm explanation of `UNNEST`.
I'm thinking that shall we use `distinct page` as well to deduplicate pages. Does it make sense?
Moreover, this metric may probably aim to compare which use or not use font-dispaly.
however, I couldn't find a way on how to query those not using.
May I have your thoughts? Thank you.

rviscomi: > I'm thinking that shall we use distinct page as well to deduplicate pages. Does it make sense?

The `SELECT DISTINCT` will ensure that client, page, and font_display are all unique. It would probably be clearer if I wrote it like that rather than `DISTINCT client`.

> Moreover, this metric may probably aim to compare which use or not use font-dispaly. however, I couldn't find a way on how to query those not using.

In `getFontDisplay` rather than returning an empty array, we could return `[null]`. We'd have to do this after the `reduce` function and in the `catch` block. That should give us a `NULL` value to group by in the aggregation, representing pages without `font-display`.

rviscomi: I'm not sure if it would work, but another idea is to change the type of JOIN between parsed_css and the UNNEST. Right now they're joined using the comma operator, which is an implicit CROSS JOIN. Maybe using an explicit LEFT JOIN would preserve pages with no `font_display` value and they'd be `NULL` by default.

Let me know if that doesn't make sense and I can try to make a proof-of-concept query.

sql/2020/04_Fonts/04_03.popular_font_hosts_by_country.sql
AbbyTsai: I think `row_number()` is used to sort country by which `rank <=10` (line 42)  
Would like to hear your further thought if it is not the one you mean.

rviscomi: Oh, I assumed the `WHERE rank <= 10` corresponded to the `rank` field defined on line 10, but I think you're right. It might be worth renaming one of these to be clearer.

sql/2020/04_Fonts/04_06.font_display_with_fcp.sql
AbbyTsai: No more words could be said than thank you to the warm help. 
The cool way you mentioned actually did a wonderful job to query null value using `LEFT JOIN UNNEST`. 
I'll try to apply this to 04_7/8/9 which all have the same goal to compare using or not using property.

sql/2020/04_Fonts/04_03.popular_font_hosts_by_country.sql
AbbyTsai: I like the suggestion which make them easy known.

sql/2020/04_Fonts/04_01.web_fonts_usage.sql
```diff
@@ -0,0 +1,28 @@
...
+  reqFont IS NOT NULL
+  AND bytesFont IS NOT NULL
+GROUP BY
+  client,
+  year,
+  month
+ORDER BY
+  client,
+  year,
+  month DESC
```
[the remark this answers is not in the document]
AbbyTsai: done.

sql/2020/04_Fonts/04_05.web_font_usage_breakdown_with_fcp.sql
AbbyTsai: done.

sql/2020/04_Fonts/04_07.font_resource_hints_with_fcp.sql
AbbyTsai: Thanks for the suggestion which is working well,
however, the things seem to go a wrong way again being return 0 when I try to join CrUX report.
May I have your help in direction?

rviscomi: @AbbyTsai I ran the latest version of the query and it produces [results](https://docs.google.com/spreadsheets/d/1OsAogFyPFOkYF185pHHV42mQ87XFCmU5LE3PUnDHlFQ/edit?usp=sharing). There are a few things to clean up or optimize, but it seems to be working.

sql/2020/04_Fonts/04_07.font_resource_hints_with_fcp.sql
```diff
@@ -0,0 +1,84 @@
...
+    hint.name,
+    hint.href AS hint_url
+  FROM
+    `httparchive.pages.2020_08_01_*`
+    LEFT JOIN UNNEST(getResourceHints(payload)) AS hint)
+LEFT JOIN (
+  SELECT
+    _TABLE_SUFFIX AS client,
+    url AS page,
+    type
```
rviscomi, line 48: This is unused

sql/2020/04_Fonts/04_07.font_resource_hints_with_fcp.sql
```diff
@@ -0,0 +1,84 @@
...
+  (client, page)
+JOIN (
+  SELECT DISTINCT
+    origin, device,
+    fast_fcp,
+    slow_fcp,
+  FROM
+    `chrome-ux-report.materialized.device_summary`
+  WHERE
+    yyyymm=202008)
```
rviscomi, line 63: The `device_summary` table is inconsistent with the `country_summary` table in that it's partitioning field is actually `date` rather than `yyyymm`. Sorry for the confusion!

So this query can be optimized:
```suggestion
    date='2020-08-01')
```

sql/2020/04_Fonts/04_07.font_resource_hints_with_fcp.sql
```diff
@@ -0,0 +1,84 @@
...
+  COUNTIF(fast_fcp>=0.75)*100/COUNT(0) AS pct_good_fcp_hints,
+  COUNTIF(NOT(slow_fcp>=0.25)
+    AND NOT(fast_fcp>=0.75))*100/COUNT(0) AS pct_ni_fcp_hints,
+  COUNTIF(slow_fcp>=0.25)*100/COUNT(0) AS pct_poor_fcp_hints,
+FROM (
+  SELECT
+    _TABLE_SUFFIX AS client,
+    url AS page,
+    hint.name,
+    hint.href AS hint_url
```
rviscomi, line 40: This is unused

rviscomi (changes requested):

sql/2020/04_Fonts/04_07.font_resource_hints_with_fcp.sql
AbbyTsai: sorry to request the time on review again, as I couldn't find a road without 0 result after running revised query. Thank you.

rviscomi: It would be weird if the same query was producing results for me but not for you. Are you sure you're running [the latest version](https://github.com/HTTPArchive/almanac.httparchive.org/blob/Fonts-2020-queries/sql/2020/04_Fonts/04_07.font_resource_hints_with_fcp.sql)? Is it possible you've run out of quota? Let me know if there are any error messages related to quota or anything else.

rviscomi:
Hi @AbbyTsai, what are you currently blocked on? Let us know how we can help get this over the finish line.

sql/2020/04_Fonts/04_04.local_vs_host_with_fcp.sql
```diff
@@ -0,0 +1,37 @@
+#standardSQL
```
rviscomi, line 1: Question from @AbbyTsai:

> guidance on when to be measured by the % pages and when by % requests in almanac.request and parsed_css.

sql/2020/04_Fonts/04_05.web_font_usage_breakdown_with_fcp.sql
```diff
@@ -0,0 +1,36 @@
+#standardSQL
```
rviscomi, line 1: Question from @AbbyTsai:

> guidance on when to be measured by the % pages and when by % requests in almanac.request and parsed_css.

sql/2020/04_Fonts/04_10.font_formats.sql
```diff
@@ -0,0 +1,17 @@
+#standardSQL
```
rviscomi, line 1: Question from @AbbyTsai:

> guidance on when to be measured by the % pages and when by % requests in almanac.request and parsed_css.

sql/2020/04_Fonts/4_10a.font_formats_font_face.sql
```diff
@@ -0,0 +1,52 @@
+#standardSQL
```
rviscomi, line 1: Question from @AbbyTsai:

> guidance on scanning the format inside url @font-face, as discussed with Barry I may lose that part. i.e. url(xxx.woff2)==> woff2

sql/2020/04_Fonts/04_21.openType_axis.sql
```diff
@@ -0,0 +1,54 @@
+#standardSQL
```
rviscomi, line 1: Question from @AbbyTsai:

> guidance on extract the values of table_sizes.

rviscomi (commented):
Flagging outstanding questions by @AbbyTsai so the other reviewers and I can focus on resolving those.

AbbyTsai:
@HTTPArchive/analysts    
Thanks to all reviewers. I'd like to sum up my questions here and seek your help on them.
04_04,05,10:curiosity about the impact of metric between %pages and %requests push me to test the result of both. for example, as [test run](https://docs.google.com/spreadsheets/d/1jjvZqYay5KmTle4atzFWqtbkz9ohw25KFmNtCS-7n3s/edit#gid=838326315) of 04_04, we might roughly see 75% web requests of 3rd-party fonts, in other way, 59% web pages adapt 3rd-party fonts. I'm confused on which ratio would be an appropriate one to describe the usage of web font hosts.
04_10a: I'm thinking is it feasible/worth to extra query the format within url(.woff2). the query might already include most of the formats, as it's a general sense to declare a format within format("woff2"). 
04_14:not sure what caused the count of pct_good+pct_ni+pct_poor not equal 1. 
04_21:how to extract the values of table_sizes for opentype as mine return no result.

rviscomi:
@csxgg I'd like to get your thoughts on the approach taken in several of this chapter's queries to join font usage data from HA with origin-level performance data in CrUX. I know in some of your performance-focused correlational analyses you've opted not to join with CrUX and use lab data in HA instead. Do you think the results would be more significant if we used lab-based FCP data? For example, see [04_04.local_vs_host_with_fcp.sql](https://github.com/HTTPArchive/almanac.httparchive.org/blob/Fonts-2020-queries/sql/2020/04_Fonts/04_04.local_vs_host_with_fcp.sql).

csxgg:
> @csxgg I'd like to get your thoughts on the approach taken in several of this chapter's queries to join font usage data from HA with origin-level performance data in CrUX. I know in some of your performance-focused correlational analyses you've opted not to join with CrUX and use lab data in HA instead. Do you think the results would be more significant if we used lab-based FCP data?

Rick, you summarized my concern with CrUX pretty well. Since CrUX is user generated data, there are lots of sources of variability that we are not measuring, such as the device, network speed, types of content the users likes to look, etc. With the lab generated data, even though the performance data will be less representative of real usage, it is much "cleaner" since we no longer need to worry about these sources of variability, although there may still be some additional sources of variability that are intrinsic to the website itself that we may not be measuring.
 
> For example, see [04_04.local_vs_host_with_fcp.sql](https://github.com/HTTPArchive/almanac.httparchive.org/blob/Fonts-2020-queries/sql/2020/04_Fonts/04_04.local_vs_host_with_fcp.sql). The overarching question it's trying to answer is "is there a performance cost to serving web fonts from a third party host?". The query uses CrUX data at the origin level to get the performance data, but I'm not sure if this is valid. I think it also double counts pages that make multiple font requests. Do you have any advice to improve these queries from a statistical perspective?

I think we could consider drawing the performance data from the "pages" table as you had showed me.

rviscomi (changes requested):
Implemented the suggested change from the PR discussion to use the lab-based performance data in 04_04 rather than joining with origin-based field data in CrUX.

Queries that use CrUX for segmenting URLs by country are OK, but I think we should use this approach for queries that use it for performance data.

sql/2020/04_Fonts/04_04.local_vs_host_with_fcp.sql
```diff
@@ -0,0 +1,43 @@
...
+  CONCAT(origin, '/')=page AND
+  IF(device = 'desktop', 'desktop', 'mobile')=client        
+WHERE
+  type = 'font'
+  AND date='2020-08-01'
+GROUP BY
+  client,
+  host_local
+ORDER BY
+  client, host_local, freq_req DESC
```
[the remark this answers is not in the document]
rviscomi: @jpamental @davelab6 the way I'd interpret these results would be that pages that load fonts from external hosts tend to have faster FCP and LCP values than those that exclusively host fonts locally (read as same domain, not from the client's local machine) or a mix of both hosting strategies.

The big caveat with this kind of analysis is that this doesn't prove external font hosting is necessarily faster than local font hosting, it's just showing the correlation.

Let us know if this is the kind of results you're looking for.

AbbyTsai:
@rviscomi @csxgg Thank you. Great to hear these awesome solutions.

Here's update of query trying to kick lab-based performance data as review mentioned, however I'm not sure how to fix why the usage (3,5) of the same host(url) isn't accumulated together.  
These questions (3,5,10,21) and query would love to have your eye on help and review. Thank you.

AbbyTsai:
Hello, HTTPArchive/analysts, 
The queries here have be updated, except the test of 04_21 still returns no result. May we call for review. Thanks.

sql/2020/04_Fonts/04_08.font_unicode_range_with_fcp.sql
```diff
@@ -0,0 +1,75 @@
...
+  WHERE
+    date = '2020-08-01'
+  GROUP BY
+    client,
+    page,
+    unicode)
+JOIN (
+  SELECT
+    _TABLE_SUFFIX AS client,
+    COUNT(0) AS total_page,
```
rviscomi, line 58: I think you should select the `page` field, remove the `GROUP BY` for this subquery, and join with `USING (client, page)`.

rviscomi:
Hi @AbbyTsai. I know there have been a few questions about some of the queries, could you give us an update on which queries are still in need of additional work? I'll try to prioritize those for you so we can get this merged.

AbbyTsai:
so far the queries may seem getting golden brown color on with big hands of reviewers, especially thanks Rick to help me fixing many of those. thus, I'd like to request a review of merge to meet the timeline, and saying that we may still work on a possible change afterwards.     .

rviscomi (approved):
Thank you @AbbyTsai for getting these queries in shape, I know it took a lot of time and effort and we appreciate that.

Let's merge these in and open up a new PR with any necessary changes.

PR #1062, opened 2020-07-19
SEO 2020 queries

Tiggerito:
SEO 2020 Queries (#908)

### Crawlability & Indexability
* [ ] HTTP Status Code
Not possible as all home pages are 200 OK
* [x] Robots.txt - Status Code
pages_robots_txt_by_device_and_status.sql - status_code
* [x] Robots.txt - Validity
lighthouse.sql - robots_txt_valid
* [ ] Robots.txt - Syntax Compliance
Not sure how this is different to Validity?
* [x] Meta robots tags inclusion
pages_wpt_bodies_by_device.sql = pct_has_meta_robots
* [x] X-robots tags inclusion
pages_wpt_bodies_by_device.sql = pct_robots_has_x_robots_tag
* [x] Robots directives (generic vs UA-specific)
pages_robots_txt_by_device_and_useragent.sql
* [x] Robots values
pages_robots_txt_by_device_and_useragent.sql
* [x] % of pages with a canonical tag
pages_wpt_bodies_by_device.sql - has_canonical
* [x] % of pages with a self canonical tag
pages_wpt_bodies_by_device.sql - pct_has_self_canonical
* [x] % of pages with a canonicalized tag
pages_wpt_bodies_by_device.sql - pct_is_canonicalized
* [x] % of pages specifying canonical via http header
pages_wpt_bodies_by_device.sql - pct_http_canonical
* [x] % of pages with different canonical values in HTTP header and html tag
pages_wpt_bodies_by_device.sql - pct_has_conflict_http_header_changed_canonical
* [ ] Usage of Rendering strategies
* [x] Where the meta robots is included
pages_wpt_bodies_by_device.sql - pct_has_meta_robots
pages_wpt_bodies_by_device.sql - pct_has_x_robots_tag
pages_wpt_bodies_by_device.sql - pct_has_meta_robots_and_x_robots_tag
* [x] Conflict between raw and rendered meta robots
pages_wpt_bodies_by_device.sql - pct_rendering_changed_robots_meta_tag
* [x] Where the meta canonical is included
pages_wpt_bodies_by_device.sql - pct_http_canonical
pages_wpt_bodies_by_device.sql - pct_has_raw_canonical
pages_wpt_bodies_by_device.sql - pct_has_rendered_canonical
pages_wpt_bodies_by_device.sql - pct_has_rendered_but_not_raw_canonical
* [x] Conflict between raw and rendered canonical
pages_wpt_bodies_by_device.sql - pct_has_conflict_rendering_changed_canonical
[...]

PR #1137, opened 2020-08-01
Markup 2020 queries

Tiggerito:
## Introduction

Metrics are in sections. 

All metrics segment their data and there are common segments like device (mobile, desktop). Each metric will indicate how the data is segmented. In some cases I will be able to merge metrics into queries that use the same segments to reduce costs.

Some metrics are segemented by percentile. This means totals are split into the 10, 25, 50, 75 and 90th percentiles.

I've use some tags to help identify common segments:

[SB-D] = segmented by device
[SB-DP] = segmented by device and percentile
[SB-CUST] = custom segmenting

I'm also adding IDs to each metric so they can be referenced from the [queries ](https://github.com/HTTPArchive/almanac.httparchive.org/tree/markup-sql-2020/sql/2020/03_Markup)and the [metrics sheet](https://docs.google.com/spreadsheets/d/1MqMNHuXzZP54qP-OmYbk5cevVf6FQpQza4GbKQMyq9A/edit#gid=0) etc.

### 1. General

- [x]  M101 - **doctype values, percent, segmented by device and doctype** [SB-CUST]
summary_pages.doctype
summary_pages_by_device_and_doctype.sql
[summary_pages_by_device_and_doctype_sample_10k](https://docs.google.com/spreadsheets/d/1WDoslcp_1Mm8vEtdsBXrm0ax-Np64ZWpmTLciS4Mcc8/edit#gid=562096630)

- [x]  M102 - **doctype is present, percent, segmented by device** [SB-D]
summary_pages.doctype
summary_pages_by_device.sql
[summary_pages_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1-o35swPBXlBdKN1KZhLExXi5UfxCfnVX463iL0_Tf4Q/edit#gid=875349140)

- [x]  M103 - **number of comments per page, segmented by device and percentile** [SB-DP]
pages.payload._wpt_bodies->raw_html.comment_count
pages_wpt_bodies_by_device_and_percentile.sql
[pages_wpt_bodies_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1ysKUTtP3igVh_1N_89tV2AqGymLK7oGVB9647voVJsQ/edit#gid=1306523771)

- [x]  M104 - **comments are present, percent, segmented by device** [SB-D]
pages.payload._wpt_bodies->raw_html.comment_count
pages_wpt_bodies_by_device.sql
[pages_wpt_bodies_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1BxKdRh1v1f5KQGwHlEO3RnL8zFqt4fr1CDFM5jHF1H4/edit#gid=1873734015)

- [x]  M105 - **number of conditional comments per page, segmented by device and percentile** [SB-DP]
pages.payload._wpt_bodies->raw_html.conditional_comment_count
pages_wpt_bodies_by_device_and_percentile.sql
[pages_wpt_bodies_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1ysKUTtP3igVh_1N_89tV2AqGymLK7oGVB9647voVJsQ/edit#gid=1306523771)

- [x]  M106 - **conditional comments are present, percent, segmented by device** [SB-D]
pages.payload._wpt_bodies->raw_html.conditional_comment_count
pages_wpt_bodies_by_device.sql
[pages_wpt_bodies_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1BxKdRh1v1f5KQGwHlEO3RnL8zFqt4fr1CDFM5jHF1H4/edit#gid=1873734015)

- [x]  M107 - **document size, segmented by device and percentile** [SB-DP]
summary_pages.bytesHtml
summary_pages_by_device_and_percentile.sql
[summary_pages_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1IocKSzXes2f1tuZwaOf4gvYfiZD8w-cmClCXxHMidtw/edit#gid=1135890889)

- [x]  M108 - **max document size, segmented by device** [SB-D]
summary_pages.bytesHtml
summary_pages_by_device.sql
[summary_pages_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1-o35swPBXlBdKN1KZhLExXi5UfxCfnVX463iL0_Tf4Q/edit#gid=875349140)

- [x]  M109 - **min document size, segmented by device** [SB-D]
summary_pages.bytesHtml
summary_pages_by_device.sql
[summary_pages_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1-o35swPBXlBdKN1KZhLExXi5UfxCfnVX463iL0_Tf4Q/edit#gid=875349140)

- [x]  M110 - **average document size, segmented by device** [SB-D]
summary_pages.bytesHtml
summary_pages_by_device.sql
[summary_pages_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1-o35swPBXlBdKN1KZhLExXi5UfxCfnVX463iL0_Tf4Q/edit#gid=875349140)

### 2. Elements

- [x]  M201 - **top 10,000 element types, full count over all pages, segmented by device and element type** [SB-CUST]
pages.payload._element_count
[fig 1 - frequency](https://almanac.httparchive.org/en/2019/markup#fig1)
pages_element_count_by_device_and_element_type_frequency.sql
[pages_element_count_by_device_and_element_type_frequency_sample_10k](https://docs.google.com/spreadsheets/d/1NKGWBluSVJ5Ujxt4SHzY2BZAd-Nai9d-Tle7xIvxmn8/edit#gid=209340073)

- [x]  M202 - **top 10,000 element types, percent, segmented by device and element type** [SB-CUST]
I think this relates to last year and is number of pages that include an element at least once 
pages.payload._element_count
[fig 1 - frequency](https://almanac.httparchive.org/en/2019/markup#fig1)
pages_element_count_by_device_and_element_type_frequency.sql
[pages_element_count_by_device_and_element_type_frequency_sample_10k](https://docs.google.com/spreadsheets/d/1NKGWBluSVJ5Ujxt4SHzY2BZAd-Nai9d-Tle7xIvxmn8/edit#gid=209340073)

- [x]  M203 - **how many different element types are there on a page, segmented by device** [SB-CUST]
from last year. Not requested this year.
pages.payload._element_count
pages_element_count_by_device_and_element_types_used_per_page.sql
[pages_element_count_by_device_and_element_types_used_per_page_sample_10k](https://docs.google.com/spreadsheets/d/1d1LBsu59vgj1BXpjOzuZuAQ_vQ4ZbiHBSnyxUbZL5tE/edit#gid=1911463327)

- [x]  M204 - **pages with a script tag, percent, segmented by device** [SB-D]
excluding application/ld+json
pages.payload._almanac->scripts
pages_almanac_by_device.sql
[pages_almanac_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1oHORBXqiCBODF7EPNXN2zGjPOa5oWDj5aAsMHtn7pZI/edit#gid=2036515744)

- [x]  M205 - **script tags per page, segmented by device and percentile** [SB-DP]
excluding application/ld+json
pages.payload._almanac->scripts
pages_almanac_by_device_and_percentile.sql
[pages_almanac_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1gsGUjm2aIf4obhcjwCfF23nlziaI-PV729Fzu6TZBfI/edit#gid=1205767109)

- [x]  M206 - **pages with an inline script tag, percent, segmented by device** [SB-D]
excluding application/ld+json
pages.payload._almanac->scripts
pages_almanac_by_device.sql
[pages_almanac_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1oHORBXqiCBODF7EPNXN2zGjPOa5oWDj5aAsMHtn7pZI/edit#gid=2036515744)

- [x]  M207 - **inline script tags per page, segmented by device and percentile** [SB-DP]
excluding application/ld+json
pages.payload._almanac->scripts
pages_almanac_by_device_and_percentile.sql
[pages_almanac_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1gsGUjm2aIf4obhcjwCfF23nlziaI-PV729Fzu6TZBfI/edit#gid=1205767109)

- [x]  M208 - **pages with a src script tag, percent, segmented by device** [SB-D]
pages.payload._almanac->scripts
pages_almanac_by_device.sql
[pages_almanac_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1oHORBXqiCBODF7EPNXN2zGjPOa5oWDj5aAsMHtn7pZI/edit#gid=2036515744)

- [x]  M209 - **src script tags per page, segmented by device and percentile** [SB-DP]
excluding application/ld+json
pages.payload._almanac->scripts
pages_almanac_by_device_and_percentile.sql
[pages_almanac_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1gsGUjm2aIf4obhcjwCfF23nlziaI-PV729Fzu6TZBfI/edit#gid=1205767109)

- [x]  M210 - **pages without a script tag, percent, segmented by device** [SB-D]
pages.payload._almanac->scripts
pages_almanac_by_device.sql
[pages_almanac_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1oHORBXqiCBODF7EPNXN2zGjPOa5oWDj5aAsMHtn7pZI/edit#gid=2036515744)

- [x]  M211 - **pages with a noscript tag, percent, segmented by device** [SB-D]
pages.payload._markup->noscripts
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M212 - **noscript tags per page, segmented by device and percentile** [SB-DP]
pages.payload._markup->noscripts
pages_markup_by_device_and_percentile.sql
[pages_markup_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1ApK7pPuvum9H8mz0Mp1DyvcB6cM8RA0gzoCfyOTPyRI/edit#gid=429356452)

- [x]  M213 - **pages with a noscript GTM tag, percent, segmented by device** [SB-D]
pages.payload._markup->noscripts
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M214 - **pages with details/summary tags, percent, segmented by device** [SB-D]
pages_payload._element_count
pages_element_count_by_device.sql
[pages_element_count_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1ayEDcvt_k2FvNTnLFdTK9mUWuztJWiTjP8Q7DIMLdAw/edit#gid=119962298)

- [ ]  M215 - **How often do elements appear per document? - ???**
pages_payload._element_count
Is this really M202?
Suggestion on how [to do this in slack from Rick](https://httparchive.slack.com/archives/C016QLMT0BZ/p1596881174369700?thread_ts=1596880054.369600&cid=C016QLMT0BZ). Do percentile and group by element and limit to top elements
This seems like a good use case for percentiles, specifically the median. You could count the frequency of each element type per page, aggregate the median frequency across all pages, and limit the results to the top 10 element types by frequency.

- [x]  M216 - **pages using obsolete elements, percent, segmented by device and element** [SB-CUST]
pages.payload._markup->obsolete_elements
pages.payload._element_count
pages_element_count_by_device_and_obsolete_elements.sql
[pages_element_count_by_device_and_obsolete_elements_sample_10k](https://docs.google.com/spreadsheets/d/1aboKPZxFGWsTQIia6bCYKUVfMoFZnhNNBtEALPiTnhw/edit#gid=1116531659)

- [x]  M217 - **image types used for favicons, percent, segmented by device and type** [SB-CUST]
This is based on file extensions.
pages.payload._almanac.link-nodes
pages_almanac_by_device_and_favicon_image_type.sql
[pages_almanac_by_device_and_favicon_image_type_sample_10k](https://docs.google.com/spreadsheets/d/1KRz3QjLp0-h2l8mZCjQ97OFxZjeQoYnNo-ldWQlJRZU/edit#gid=362462411)

- [x]  M218 - **pages with a favicon, percent, segmented by device** [SB-D]
pages.payload._markup->favicon
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M219 - **How do people use the `viewport` meta element**
summary_pages.meta_viewport
summary_pages_by_device_and_viewport.sql
[summary_pages_by_device_and_viewport_sample_10k](https://docs.google.com/spreadsheets/d/1wmH9E4570LiOLImnEVD6eJ59VHM9REinrxuWmpEjVdY/edit#gid=873089975)

- [x]  M220 - **pages without an h1, percent, segmented by device** [SB-D]
pages.payload._wpt_bodies->headings
pages_wpt_bodies_by_device.sql
[pages_wpt_bodies_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1BxKdRh1v1f5KQGwHlEO3RnL8zFqt4fr1CDFM5jHF1H4/edit#gid=1873734015)

- [x]  M221 - **number of h1s to h8s on a page, segmented by device, percentile and heading level** [SB-CUST]
pages.payload._wpt_bodies->headings
pages_wpt_bodies_by_device_and_percentile_and_heading_level.sql
[pages_wpt_bodies_by_device_and_percentile_and_heading_level_sample_10k](https://docs.google.com/spreadsheets/d/1qoaxvXV62vOHR1Oty0ixp9dCnbYmHrwxTU_f-OrEiEg/edit#gid=1004123736)

- [x]  M222 - **Does the heading logical sequence make any sense** [SB-CUST]
Need more details. safe to be same or one lower or go up any amount
pages.payload._almanac->headings_order
pages_almanac_by_device.sql
[pages_almanac_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1oHORBXqiCBODF7EPNXN2zGjPOa5oWDj5aAsMHtn7pZI/edit#gid=2036515744)

- [x]  M223 - **pages with an svg element, percent, segmented by device** [SB-D]
pages.payload._markup->svgs
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M224 - **number of svg element on a page, segmented by device and percentile** [SB-DP]
pages.payload._markup->svgs
pages_markup_by_device_and_percentile.sql
[pages_markup_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1ApK7pPuvum9H8mz0Mp1DyvcB6cM8RA0gzoCfyOTPyRI/edit#gid=429356452)

- [x]  M225 - **pages with an svg image tag, percent, segmented by device** [SB-D]
pages.payload._markup->svgs
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M226 - **number of svg image tag on a page, segmented by device and percentile** [SB-DP]
pages.payload._markup->svgs
pages_markup_by_device_and_percentile.sql
[pages_markup_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1ApK7pPuvum9H8mz0Mp1DyvcB6cM8RA0gzoCfyOTPyRI/edit#gid=429356452)

- [x]  M227 - **pages with an svg object tag, percent, segmented by device** [SB-D]
pages.payload._markup->svgs
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M228 - **number of svg object tag on a page, segmented by device and percentile** [SB-DP]
pages.payload._markup->svgs
pages_markup_by_device_and_percentile.sql
[pages_markup_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1ApK7pPuvum9H8mz0Mp1DyvcB6cM8RA0gzoCfyOTPyRI/edit#gid=429356452)

- [x]  M229 - **pages with an svg embed tag, percent, segmented by device** [SB-D]
pages.payload._markup->svgs
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M230 - **number of svg embed tag on a page, segmented by device and percentile** [SB-DP]
pages.payload._markup->svgs
pages_markup_by_device_and_percentile.sql
[pages_markup_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1ApK7pPuvum9H8mz0Mp1DyvcB6cM8RA0gzoCfyOTPyRI/edit#gid=429356452)

- [x]  M231 - **pages with an svg iframe tag, percent, segmented by device** [SB-D]
pages.payload._markup->svgs
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M232 - **number of svg iframe tag on a page, segmented by device and percentile** [SB-DP]
pages.payload._markup->svgs
pages_markup_by_device_and_percentile.sql
[pages_markup_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1ApK7pPuvum9H8mz0Mp1DyvcB6cM8RA0gzoCfyOTPyRI/edit#gid=429356452)

- [x]  M233 - **pages with a svg in them, percent, segmented by device** [SB-D]
i.e. any identified method (svg_total)
pages.payload._markup->svgs
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M234 - **size of the head section in characters, segmented by device and percentile** [SB-DP]
pages.payload._wpt_bodies->raw_html.head_size
pages_wpt_bodies_by_device.sql
[pages_wpt_bodies_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1BxKdRh1v1f5KQGwHlEO3RnL8zFqt4fr1CDFM5jHF1H4/edit#gid=1873734015)

- [x]  M235 - **how many pages contain a link using a specific protocol, percent, segmented by device, protocol** [SB-CUST]
pages.payload._wpt_bodies->...protocols
pages_wpt_bodies_by_device_and_protocol.sql
[pages_wpt_bodies_by_device_and_protocol_sample_10k](https://docs.google.com/spreadsheets/d/1seyPiOcNZzQUifkfkL-8EdiiW1fugm_9b7LI9Yj3dBU/edit#gid=1062628530)

- [ ]  M236 - **What values are used for `src` and `srcset` on `img` and `source` elements? - ???**
Need more details
pages.payload._almanac->images

- [x]  M241 - **how many pages contain an element by element and device** [SB-CUST]
from last year. Not requested this year.
[fig 1 - per site](https://almanac.httparchive.org/en/2019/markup#fig1)
pages.payload._element_count
pages_element_count_by_device_and_element_type_present.sql
[pages_element_count_by_device_and_element_type_present_sample_10k](https://docs.google.com/spreadsheets/d/1d3CKi0OiZuw9WzAkUZQJ0kb0vDq4MmvJuCMYKzTuaww/edit#gid=912512400)

- [x]  M242 - **number of pages containing element types including a dash segmented by element and device** [SB-CUST]
from last year. Not requested this year.
pages.payload._element_count
pages_element_count_by_device_and_custom_dash_elements.sql
[pages_element_count_by_device_and_custom_dash_elements_sample_10k](https://docs.google.com/spreadsheets/d/12fS5Crf5DaNXz5N6XmqiRR8SsKVP_gsUX3S2f3t3YwA/edit#gid=1167686108)

- [x]  M243 - **top element types including a dash segmented by element and device** [SB-CUST]
from last year. Not requested this year.
pages.payload._element_count
pages_element_count_by_device_and_custom_dash_elements.sql
[pages_element_count_by_device_and_custom_dash_elements_sample_10k](https://docs.google.com/spreadsheets/d/12fS5Crf5DaNXz5N6XmqiRR8SsKVP_gsUX3S2f3t3YwA/edit#gid=1167686108)

### 3. Element/Attributes

- [x]  M301 - **how many button elements on a page, segmented by device and percentile** [SB-DP]
pages.payload._markup->buttons
pages_markup_by_device_and_percentile.sql
[pages_markup_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1ApK7pPuvum9H8mz0Mp1DyvcB6cM8RA0gzoCfyOTPyRI/edit#gid=429356452)

- [x]  M302 - **pages with button elements, percent, segmented by device** [SB-D]
pages.payload._markup->buttons
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M303 - **pages with button elements with no type, percent, segmented by device** [SB-D]
pages.payload._markup->buttons
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M304 - **top button element types, percent, segmented by device and type** [SB-CUST]
pages.payload._markup->buttons
pages_markup_by_device_and_button_types.sql
[pages_markup_by_device_and_button_types_sample_10k](https://docs.google.com/spreadsheets/d/1_qXtgDqi1oGSfdYfdZYJZFfxYYrNvXmvWKuQIqpEmtQ/edit#gid=1201328401)

- [x]  M305 - **how many input elements with type image on a page, segmented by device and percentile** [SB-DP]
pages.payload._markup->inputs
pages_markup_by_device_and_percentile.sql
[pages_markup_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1ApK7pPuvum9H8mz0Mp1DyvcB6cM8RA0gzoCfyOTPyRI/edit#gid=429356452)

- [x]  M306 - **how many input elements with type button on a page, segmented by device and percentile** [SB-DP]
pages.payload._markup->inputs
pages_markup_by_device_and_percentile.sql
[pages_markup_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1ApK7pPuvum9H8mz0Mp1DyvcB6cM8RA0gzoCfyOTPyRI/edit#gid=429356452)

- [x]  M307 - **how many input elements with type submit on a page, segmented by device and percentile** [SB-DP]
pages.payload._markup->inputs
pages_markup_by_device_and_percentile.sql
[pages_markup_by_device_and_percentile_sample_10k](https://docs.google.com/spreadsheets/d/1ApK7pPuvum9H8mz0Mp1DyvcB6cM8RA0gzoCfyOTPyRI/edit#gid=429356452)

- [x]  M310 - **pages with autoplaying video elements, percent, segmented by device** [SB-D]
autoplay equates to true
pages.payload._almanac->videos
pages_almanac_by_device.sql
[pages_almanac_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1oHORBXqiCBODF7EPNXN2zGjPOa5oWDj5aAsMHtn7pZI/edit#gid=2036515744)

- [x]  M311 - **pages with non autoplaying video elements, percent, segmented by device** [SB-D]
autoplay equates to false
pages.payload._almanac->videos
pages_almanac_by_device.sql
[pages_almanac_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1oHORBXqiCBODF7EPNXN2zGjPOa5oWDj5aAsMHtn7pZI/edit#gid=2036515744)

- [x]  M312 - **pages with autoplaying audio elements, percent, segmented by device** [SB-D]
autoplay equates to true
pages.payload._markup->audio
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M313 - **pages with non autoplaying audio elements, percent, segmented by device** [SB-D]
autoplay equates to false
pages.payload._markup->audio
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [ ]  M320 - **pages that are WordPress, percent, segmented by device - ???** [SB-D]
or maybe use technologies or leave this up to CMS chapter
pages.payload._almanac->meta_nodes

### 4. Attributes

- [x]  M400 - **most frequently use attributes, percent, segmented by device and attribute** [SB-CUST]
pages.payload._almanac->attributes_used_on_elements
pages_almanac_by_device_and_attribute_name_frequency.sql
[pages_almanac_by_device_and_attribute_name_frequency_sample_10k](https://docs.google.com/spreadsheets/d/18XKM7lYGhX0Jr3Om2G8fGiFFz7m09VikMBDIZF_-pDY/edit#gid=1485659129)

- [x]  M401 - **How often do what attributes appear per document?** [SB-CUST]
Not sure how to report this. Maybe just the top attributes and a percentile for them?
pages.payload._almanac->attributes_used_on_elements
pages_almanac_by_device_and_attribute_name_present.sql
[pages_almanac_by_device_and_attribute_name_present_sample_10k](https://docs.google.com/spreadsheets/d/1u-swMK9LWUU7giP9sbBHcaWnY0oi2epdk7fETUawvT4/edit#gid=443103081)

- [x]  M402 - **most frequently use data- attributes, percent, segmented by device and attribute** [SB-CUST]
pages.payload._almanac->attributes_used_on_elements
pages_almanac_by_device_and_data_attribute_name_frequency.sql
[pages_almanac_by_device_and_data_attribute_name_frequency_sample_10k](https://docs.google.com/spreadsheets/d/15j7Clwo5TyoDhHlC8Sc_IqqjEGBYroQCNwunYPv-yHE/edit#gid=848934321)

- [x]  M403 - **pages identified as an app, percent, segmented by device** [SB-D]
pages.payload._markup->app
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M404 - **pages with no html lang attribute, percent, segmented by device** [SB-D]
pages.payload._almanac->html_node.lang
pages_almanac_by_device.sql
[pages_almanac_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1oHORBXqiCBODF7EPNXN2zGjPOa5oWDj5aAsMHtn7pZI/edit#gid=2036515744)

- [x]  M405 - **most frequently use html lang attribute, percent, segmented by device and lang** [SB-CUST]
pages.payload._almanac->html_node.lang
pages_almanac_by_device_and_html_lang.sql
[pages_almanac_by_device_and_html_lang_sample_10k](https://docs.google.com/spreadsheets/d/1oJthuUYkv5IG2uzcz3ErVQ9zN76yoKruvwqF2kqrPO4/edit#gid=1780862283)

- [x]  M410 - **pages with html dir set, percent, segmented by device** [SB-D]
pages.payload._markup->dirs
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M411 - **pages with html dir set to ltr, percent, segmented by device** [SB-D]
pages.payload._markup->dirs
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M412 - **pages with html dir set to rtl, percent, segmented by device** [SB-D]
pages.payload._markup->dirs
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M413 - **pages with html dir set to auto, percent, segmented by device** [SB-D]
pages.payload._markup->dirs
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M414 - **pages with dir on other elements, percent, segmented by device** [SB-D]
pages.payload._markup->dirs
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)

- [x]  M420 - **pages with all target _banks including rel="noopener noreferrer", percent, segmented by device** [SB-D]
pages.payload._wpt_bodies->anchors.target_blank
pages_wpt_bodies_by_device.sql
[pages_wpt_bodies_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1BxKdRh1v1f5KQGwHlEO3RnL8zFqt4fr1CDFM5jHF1H4/edit#gid=1873734015)

- [x]  M421 - **pages with some target _banks not using rel="noopener noreferrer", percent, segmented by device** [SB-D]
pages.payload._wpt_bodies->anchors.target_blank
pages_wpt_bodies_by_device.sql
[pages_wpt_bodies_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1BxKdRh1v1f5KQGwHlEO3RnL8zFqt4fr1CDFM5jHF1H4/edit#gid=1873734015)

- [x]  M430 - **pages with a link rel="amphtml", percent, segmented by device** [SB-D]
pages.payload._markup->amp
pages_markup_by_device.sql
[pages_markup_by_device_sample_10k](https://docs.google.com/spreadsheets/d/1C2x2R6MSdJOl8zIWzQgHVPoodPHqx43xyp0D3KlOS4U/edit#gid=158219390)


Tiggerito:
@rviscomi @bazzadp 

I've been testing out a new way to construct these custom metrics payload queries. [Here's my experiment](https://github.com/HTTPArchive/almanac.httparchive.org/blob/markup-sql-2020/sql/2020/03_Markup/pages_by_device_experiment.sql).

It's building on the idea of using a single query to gather as much data as possible.

I was worried that all the repeated parsing of json would make things very slow. I eventually came up with creating one js function per custom metric that returns a STRUCT with all the data needed. So only one parse of the json per row.

This should mean that the query should scale up well.

At one point I tried to parse the whole payload in a js function, and it was sloooow. query went from 15 sec to 50 sec. Using JSON_EXTRACT_SCALAR is far faster but does not have the power to process the json like we often need. Parsing a single custom metric in js is fine. So I get to each custom metric via JSON_EXTRACT_SCALAR and pass that to the functions.

I've only got it dealing with a few fields at the moment. Adding new ones seem to have no impact on speed.

As expected, it was about twice the speed of an older query that used two separate functions to get each value.

The current experiment query uses the function method to gather data from two custom metrics. If it does start to struggle then doing a query per custom metric would work. 

I did one query on real data. It took 39.1 seconds and cost $2.90.

rviscomi:
+cc @paulcalvano 

@Tiggerito for readers who are browsing the finished chapters and want to see how a particular metric was calculated, would this mean that they would be linked to an SQL file that also calculates many other metrics? I think we should opt for many smaller queries rather than a few larger ones. The more narrowly focused the queries are, the easier they are for subsequent analysts (be it the 2021++ team or external readers/researchers) to grok and reuse.

To your implicit goal of reducing computational overhead, I think we should explore ways to make the custom metrics more easily accessible. For example, we could preprocess all custom metrics so that each metric has its own column in a special purpose table. Something like `SELECT COUNTIF(markup.images.loading.lazy) FROM httparchive.almanac.custom_metrics` where the `_markup` field from the `pages` dataset is converted into a BigQuery STRUCT object with properties corresponding to the [`markup.js`](https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/markup.js) schema.

Tiggerito:
> +cc @paulcalvano
> 
> @Tiggerito for readers who are browsing the finished chapters and want to see how a particular metric was calculated, would this mean that they would be linked to an SQL file that also calculates many other metrics? I think we should opt for many smaller queries rather than a few larger ones. The more narrowly focused the queries are, the easier they are for subsequent analysts (be it the 2021++ team or external readers/researchers) to grok and reuse.
> 
> To your implicit goal of reducing computational overhead, I think we should explore ways to make the custom metrics more easily accessible. For example, we could preprocess all custom metrics so that each metric has its own column in a special purpose table. Something like `SELECT COUNTIF(markup.images.loading.lazy) FROM httparchive.almanac.custom_metrics` where the `_markup` field from the `pages` dataset is converted into a BigQuery STRUCT object with properties corresponding to the [`markup.js`](https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/markup.js) schema.

I will test more about using JSON_EXTRACT_SCALAR to get the metric out of the payload. That alone seemed to save a lot of cpu time. 

For mine I may take a middle ground. Often a set of data makes sense together, so combining them in one function using a STRUCT output would make sense. I learnt it from a simple example last year that returned a few values in one go.

I'm concerned with having many smaller queries because of the maintenance involved (we have a lot to create, rough guess 100 for this chapter). Partly why the shared AS_PERCENT would be of help. I also understand that merging into one causes other issues. Maybe a middle ground again with appropriate/logical grouping?

Pre converting the metrics to STRUCT would be great. I'm kind of post doing that for just the data I need. Just having each in its own column would probably help a lot. 

This grouping only works with the common query types, by device and by device and percentile. It can create a lot of columns but a few rows. I'm trying to find a way to flip rows and columns as that would create far more readable/processable results.

Tiggerito:
Definitely using JSON_EXTRACT_SCALAR is faster for the payload. 5.4 seconds v 43 seconds for this simple query. I think it's worth at least telling people to do this change:

```
# payload parsed in the function to get _element_count
# 43.0 seconds for sample 10k
CREATE TEMPORARY FUNCTION parse_element_count(payload STRING)
RETURNS BOOL LANGUAGE js AS '''
    var $ = JSON.parse(payload); // payload in the function
    var elements = JSON.parse($._element_count);
    return true;
''';

SELECT
  _TABLE_SUFFIX AS client,
  COUNTIF(parse_element_count(payload)) AS parsed,
  COUNT(0) AS freq
FROM
  `httparchive.sample_data.pages_*`
GROUP BY
  client
```
```
# payload parsed in JSON_EXTRACT_SCALAR to get _element_count
# 5.4 seconds for sample 10k
CREATE TEMPORARY FUNCTION parse_element_count(element_count STRING)
RETURNS BOOL LANGUAGE js AS '''
    var elements = JSON.parse(element_count);
    return true;
''';

SELECT
  _TABLE_SUFFIX AS client,
  COUNTIF(parse_element_count(JSON_EXTRACT_SCALAR(payload, '$._element_count'))) AS parsed, # payload in JSON_EXTRACT_SCALAR
  COUNT(0) AS freq
FROM
  `httparchive.sample_data.pages_*`
GROUP BY
  client
```

rviscomi:
> I'm concerned with having many smaller queries because of the maintenance involved (we have a lot to create, rough guess 100 for this chapter). Partly why the shared AS_PERCENT would be of help. I also understand that merging into one causes other issues. Maybe a middle ground again with appropriate/logical grouping?

Yeah if there's a cluster of related metrics then a single query is ok. Grouping unrelated metrics would be overkill though, even if more efficient. If you run into quota issues, you can pass me the finished query which I can run using my HTTP Archive billing account and save to your chapter's results sheet.

Tiggerito:
> > Definitely using JSON_EXTRACT_SCALAR is faster for the payload.
> 
> Interesting findings! This is a good tip to share with the rest of the @HTTPArchive/analysts.
> 
> To summarize: if we let BigQuery extract the custom metric using native JSON functions before passing it into the JS UDF, that seems to speed up the query significantly. We still need UDFs for complex processing, but the basic extraction can be done more efficiently in the SQL code.

Yep. I guess the JS UDF struggles with the large payload. but can cope fine with single custom metrics.

I also suspect JSON_EXTRACT_SCALAR may cache well, meaning multiple calls on the same json string would also be fast. Not sure how to test that.

Tiggerito:
I've covered almost all of the requested data now. Only the one query [pages_markup_by_device.sql](https://github.com/HTTPArchive/almanac.httparchive.org/blob/markup-sql-2020/sql/2020/03_Markup/pages_markup_by_device.sql) ended up quite big at about 20 fields.  There are some obvious ways I can split out some field groups if required.

I've also enhanced the [README](https://github.com/HTTPArchive/almanac.httparchive.org/blob/markup-sql-2020/sql/2020/03_Markup/README.md) with some guidelines on the structure I used.

Tiggerito:
@rviscomi @paulcalvano @bazzadp 

All the queries are now pointing to live tables and ready for review.

I tested a few metrics based queries, which most of mine are. They come in at $4 each.

And I used them to practice setting up the spreadsheet:

https://docs.google.com/spreadsheets/d/1Ta7amoUeaL4pILhWzH-SCzMX9PsZeb1x_mwrX2C4eY8/edit?ts=5f4e700c#gid=150962402

Note, I had to copy the Data Viz Gallery so that I could copy the charts.

rviscomi (changes requested):
Similar to https://github.com/HTTPArchive/almanac.httparchive.org/pull/1062#discussion_r486488217, I'd recommend removing test code from the queries.

Tiggerito:
> Similar to [#1062 (comment)](https://github.com/HTTPArchive/almanac.httparchive.org/pull/1062#discussion_r486488217), I'd recommend removing test code from the queries.

Yep, I'll make the same changes here.

Tiggerito:
> Similar to [#1062 (comment)](https://github.com/HTTPArchive/almanac.httparchive.org/pull/1062#discussion_r486488217), I'd recommend removing test code from the queries.

All test code removed.

sql/2020/03_Markup/README.md
```diff
@@ -0,0 +1,164 @@
+# Guide 
+
+## file naming
+file names start with the table being queried. Next is the custom metric if appropriate. Then "_by_" followed by how the query is grouped. Finally anything more specific needed.
+
+Queries contain one or more metrics that match the queries grouping structure.
```
rviscomi, line 6: I'm torn on this. I know it saves on quota, like a query carpool, but it also makes the queries more overwhelming to understand how a single metric is evaluated.

In #1138 I laid out some ideas for directly linking figures to corresponding SQL and results. Carpooling would map many figures to the same SQL file, and it won't be clear where the stats are coming from.

If the obstacle isn't writing each individual query, but rather running them and exceeding quota, then we have some ways to mitigate that. Either with more coupon codes or someone with access to the HTTP Archive billing account can run them for you. Let me know if that could work for you.

Another benefit to fragmentation is that the SQL files become more browsable, if each file is named according to its metric.

sql/2020/03_Markup/pages_almanac_by_device.sql
```diff
@@ -0,0 +1,111 @@
...
+} catch (e) {}
+return result;
+''';
+
+SELECT
+  client,
+  COUNT(0) AS total,
+
+  # has scripts excluding jsonld ones
+  AS_PERCENT(COUNTIF(almanac_info.none_jsonld_scripts_total > 0), COUNT(0)) AS pct_contains_none_jsonld_scripts_m204,
```
rviscomi, line 78: What's the rationale for the `_m204`-like suffixes?

sql/2020/03_Markup/pages_almanac_by_device.sql
```diff
@@ -0,0 +1,111 @@
...
+# real run estimated at $4.08 and took 48 seconds
+
+# to speed things up there is only one js function per custom metric property. It returns a STRUCT with all the data needed
+# current test gathers 3 bits of incormation from the custom petric properties
+# I tried to do a single js function processing the whole payload but it was very slow (50 sec) because of parsing the full payload in js
+# this uses JSON_EXTRACT_SCALAR to first get the custom metrics json string, and only passes those into the js functions
+# Estimate about twice the speed of the original code. But should scale up far better as the custom metrics are only parsed once.
+
+# helper to create percent fields
+CREATE TEMP FUNCTION AS_PERCENT (freq FLOAT64, total FLOAT64) RETURNS FLOAT64 AS (
```
rviscomi, line 13: Idea: if these functions will be useful to many queries, we can define a reusable version that lives in the `almanac` dataset.

sql/2020/03_Markup/pages_almanac_by_device.sql
```diff
@@ -0,0 +1,111 @@
...
+  AS_PERCENT(COUNTIF(almanac_info.contains_videos_with_autoplay), COUNT(0)) AS pct_contains_videos_with_autoplay_m310,
+
+  # pages without autoplaying video elements M311
+  AS_PERCENT(COUNTIF(almanac_info.contains_videos_without_autoplay), COUNT(0)) AS pct_contains_videos_without_autoplay_m311,
+
+  # pages with no html lang attribute M404
+  AS_PERCENT(COUNTIF(almanac_info.html_node_lang IS NULL OR LENGTH(almanac_info.html_node_lang) = 0), COUNT(0)) AS pct_no_html_lang_m404,
+
+FROM
+    (
```
rviscomi, line 102: Nit: inconsistent indentation. Prefer 2 spaces for tabs like above.

sql/2020/03_Markup/pages_almanac_by_device.sql
```diff
@@ -0,0 +1,111 @@
...
+  src_scripts_total INT64,
+  inline_scripts_total INT64,
+  good_heading_sequence BOOL,
+  contains_videos_with_autoplay BOOL,
+  contains_videos_without_autoplay BOOL,
+  html_node_lang STRING
+> LANGUAGE js AS '''
+var result = {};
+try {
+    var almanac = JSON.parse(almanac_string);
```
rviscomi, line 31: Nit: Inconsistent indentation. 4 spaces here but 2 below. Use 2 for consistency.

sql/2020/03_Markup/pages_almanac_by_device.sql
rviscomi, line 78: Nit: Could this name be clearer? The condition is `> 0` but the name says `contains_none`.

sql/2020/03_Markup/pages_almanac_by_device_and_attribute_name_frequency.sql
```diff
@@ -0,0 +1,40 @@
...
+  `httparchive.pages.2020_08_01_*`,
+  UNNEST(get_almanac_attribute_info(JSON_EXTRACT_SCALAR(payload, '$._almanac'))) AS almanac_attribute_info
+GROUP BY
+  client,
+  almanac_attribute_info.name
+ORDER BY
+  client,
+  pct_m400 DESC
+LIMIT
+  10000
```
rviscomi, line 40: Consider a smaller limit. Having many rows in the spreadsheet slows it down and there may not be many interesting results down here. Alternatively you could filter using something like `HAVING pct > 0.001`.

sql/2020/03_Markup/pages_almanac_by_device_and_attribute_name_frequency.sql
```diff
@@ -0,0 +1,40 @@
...
+''';
+
+SELECT
+  _TABLE_SUFFIX AS client,
+  almanac_attribute_info.name,
+  SUM(almanac_attribute_info.freq) AS freq, # total count from all pages
+  AS_PERCENT(SUM(almanac_attribute_info.freq), SUM(SUM(almanac_attribute_info.freq)) OVER (PARTITION BY _TABLE_SUFFIX)) AS pct_m400 
+FROM
+  `httparchive.pages.2020_08_01_*`,
+  UNNEST(get_almanac_attribute_info(JSON_EXTRACT_SCALAR(payload, '$._almanac'))) AS almanac_attribute_info
```
rviscomi, line 32: Nit: using `almanac` in the variable/function names is redundant. (Throughout)

sql/2020/03_Markup/pages_almanac_by_device_and_html_lang.sql
```diff
@@ -0,0 +1,43 @@
...
+      return almanac.html_node.lang.trim().toLowerCase();
+    }
+
+} catch (e) {}
+return '';
+''';
+
+SELECT
+  client,
+  COUNT(0) AS total,
```
rviscomi, line 27: Should this be `freq`?

sql/2020/03_Markup/pages_almanac_by_device_and_html_lang.sql
```diff
@@ -0,0 +1,43 @@
...
+  almanac_html_lang as html_lang,
+
+  AS_PERCENT(COUNT(0), SUM(COUNT(0)) OVER (PARTITION BY client)) AS pct_m405
```
rviscomi, line 30: Should the denominator be the number of pages? Or is the sum always equal to that?

sql/2020/03_Markup/pages_element_count_by_device_and_custom_dash_elements.sql
```diff
@@ -0,0 +1,54 @@
...
+USING (_TABLE_SUFFIX),
+  UNNEST(get_element_types_with_a_dash(JSON_EXTRACT_SCALAR(payload, '$._element_count'))) AS element_type_with_a_dash # so we end up with pages + element_type_with_a_dash rows
+GROUP BY
+  client,
+  total,
+  element_type_with_a_dash
+ORDER BY
+  pct DESC,
+  client
+LIMIT 10000
```
rviscomi, line 54: Ditto about smaller limits.

rviscomi (changes requested):
Early feedback. Looks great so far, just some organizational questions and a few suggestions.

sql/2020/03_Markup/README.md
Tiggerito: I really don't want to spend even more hours (And don't have them) splitting this all out, having to re-test, deal with the massively multiplied maintenance issue and then have to go through the data gathering process many more times than I need to. In some cases my job will be multiplied maybe 30xs and so will the bill. And I have to do this over two chapters where their 20 odd queries each will bloat to over 100.

I'd also argue about it making them more overwhelming. Dealing with 100 queries that vary by small amounts would be more overwhelming than seeing the same pattern used in the one place. And making one fix or improvement would take a lot longer and would most likely cause code to get out of sync and broken. 

I personally don't see how it makes the sheet more complex. I'd personally would prefer working with groups of results in one tab, instead of 100s of separate tabs. See my first example for markup :

https://docs.google.com/spreadsheets/d/1Ta7amoUeaL4pILhWzH-SCzMX9PsZeb1x_mwrX2C4eY8/edit?ts=5f4e700c#gid=150962402

If one tab gets too congested the data could be split into a new tab with smaller groups of info. I've been pushing the team to look at what I have done so far, to try and find out if they are OK with the structure. No feedback yet.

For both chapters I've put a lot of effort in directing me and the the others to where the data is. Noting down the script and field name in their documentation and mine. In this case I even created numbered labels for the fields . It would be nice if we could create links to a specific tab or cell on a sheet 👍

sql/2020/03_Markup/pages_almanac_by_device.sql
Tiggerito: This helped me map requirements to fields, to make sure I did not miss any, and make it easier for everyone to find and be sure they are looking at the right values. 

They are referenced in the pull requests list which includes the query name (which relates to the data sheet tab):

https://github.com/HTTPArchive/almanac.httparchive.org/pull/1137

And the requirements sheet they made:

https://docs.google.com/spreadsheets/d/1MqMNHuXzZP54qP-OmYbk5cevVf6FQpQza4GbKQMyq9A/edit#gid=0

Which does not seem to have the query/tab names in there. I will have to add it as I pull in the data.

sql/2020/03_Markup/pages_almanac_by_device.sql
Tiggerito: That was mentioned when I came up with the function. I just went ahead and used it, knowing I could easily switch if a shared function was created. Maybe next year.

sql/2020/03_Markup/pages_almanac_by_device.sql
Tiggerito: There's probably a few things like that. I keep tidying up when I spot things. A lot of the variation of formatting comes from where I stole the code from!

I personally let my development environment tidy up my own code. I've not worked out how to do that in Visual Studio Code for BigQuery yet. I also don't know all the formatting rules. Indentation is one thing that seemed inconsistent.

sql/2020/03_Markup/pages_almanac_by_device.sql
Tiggerito: Another code theft issue. I've noticed that Studio Code does deal with some indention improvements for JS, but it bases its rule on the current code. So sometimes it auto indents in different ways. I basically could not be bothered going through all those scripts trying to get them all to follow the same indentation pattern. Imagine if I had 100s of script to tidy up!

sql/2020/03_Markup/pages_almanac_by_device_and_attribute_name_frequency.sql
Tiggerito: That came from last year. I've changed it to 1,000. I could trim it manually later if needed.

sql/2020/03_Markup/pages_almanac_by_device_and_attribute_name_frequency.sql
Tiggerito: Major refactor. I specifically named them at first because I was testing getting multiple metrics in the one query. I still like to name them as each accepts a different parameter and it's a reminder of that. Don't pass _markup into an almanac function. If JavaScript was strongly typed and allowed method overloading I would not name them ;-)

I did think about using common names for all queries at one point. I think I backed away as it would be hard to reverse if required. I would be comfortable in going that direction now I know how it all worked out. But I don't want to do a major refactor like that now. The author is nagging me because we've missed a deadline.

Another confusion I saw was that I used _info for the SQL field names, but not in the JavaScript. I picked up quite a few errors from mixing them up because of that.

sql/2020/03_Markup/pages_almanac_by_device_and_html_lang.sql
Tiggerito: Probably. I found a few cases of mixed use and I'm never sure which to use when. Changed.

sql/2020/03_Markup/pages_almanac_by_device_and_html_lang.sql
Tiggerito: I'm not great on how the totalling works. I think this is fine because there is always one value per page?

Now I have to re-add my test code. And then I forgot to change the comments, just did a live run.

Test run shows total of freq for a client equals the number of pages (10k) indicating it should be fine.  For the live run freq for both clients totalled 11,941,282 which also sounds about right.

Added an ORDER BY for next time but was easy to do that in the sheet.

Tiggerito: Seems Japanese beats Spanish

sql/2020/03_Markup/pages_element_count_by_device_and_custom_dash_elements.sql
Tiggerito: Done

sql/2020/03_Markup/pages_almanac_by_device.sql
Tiggerito: A specific request from them.  I've tried to clarify the comment:

# has scripts that are not jsonld ones. i.e. has a none jsonld script.

sql/2020/03_Markup/pages_almanac_by_device.sql
```diff
@@ -0,0 +1,111 @@
...
+            good = false;
+          }
+          previousLevel = level;
+      }); 
+      result.good_heading_sequence = good;
+    }
+
+    if (almanac.videos) {
+      var autoplay_count = almanac.videos.nodes.filter(n => n.autoplay == "" || n.autoplay).length; // valid values are blank or autoplay. Im just checking it exists...
+      //var autoplay_count = almanac.videos.nodes.filter(n => n.autoplay == "" || n.autoplay == "autoplay").length;
```
rviscomi, line 59: Remove unused code

sql/2020/03_Markup/pages_almanac_by_device_and_attribute_name_frequency.sql
rviscomi: Still showing as 10,000? The limits are kind of arbitrary but you could also use the `HAVING` suggestion.

sql/2020/03_Markup/pages_almanac_by_device_and_attribute_name_present.sql
```diff
@@ -0,0 +1,46 @@
...
+    
+} catch (e) {
+    
+}
+return []; 
+''';
+
+SELECT
+  _TABLE_SUFFIX AS client,
+  attribute_name,
```
rviscomi, line 27: Limit the results, eg `LIMIT 1000` or `HAVING pages >= 100`. I think there would be a lot of unusable data in the tail < 100 pages.

sql/2020/03_Markup/pages_almanac_by_device_and_favicon_image_type.sql
```diff
@@ -0,0 +1,73 @@
...
+      result.image_type_extension = "NO_DATA";
+    }
+
+} catch (e) {result.image_type_extension = "ERROR: "+e.message;}
+return result;
+''';
+
+SELECT
+  client,
+  almanac_info.image_type_extension AS image_type_extension,
```
rviscomi, line 56: Also considering limiting the results since this could be a high cardinality field to group by.

sql/2020/03_Markup/pages_almanac_by_device_and_html_lang.sql
```diff
@@ -0,0 +1,44 @@
...
+# returns all the data we need from _almanac
+CREATE TEMPORARY FUNCTION get_almanac_html_lang(almanac_string STRING)
+RETURNS STRING LANGUAGE js AS '''
+try {
+    var almanac = JSON.parse(almanac_string); 
+
+    if (Array.isArray(almanac) || typeof almanac != 'object') return '';
+
+    if (almanac.html_node && almanac.html_node.lang) {
+      return almanac.html_node.lang.trim().toLowerCase();
```
rviscomi, line 18: Does this include region codes, eg `en-US`, or just language codes? Do you need to parse out the language codes?

sql/2020/03_Markup/pages_element_count_by_device.sql
```diff
@@ -0,0 +1,69 @@
...
+    result.count = Object.values(element_count).reduce((total, freq) => total + (parseInt(freq, 10) || 0), 0);
+    
+    result.contains_custom_element = Object.keys(element_count).filter(e => e.includes('-')).length > 0;
+    result.contains_details_element = Object.keys(element_count).filter(e => e ==='details').length > 0;
+    result.contains_summary_element = Object.keys(element_count).filter(e => e ==='summary').length > 0;
+
+    var obsoleteElements = new Set(["applet","acronym","bgsound","dir","frame","frameset","noframes","isindex","keygen","listing","menuitem","nextid","noembed","plaintext","rb","rtc","strike","xmp","basefont","big","blink","center","font","marquee","multicol","nobr","spacer","tt"]);
+
+    result.contains_obsolete_element = !!Object.keys(element_count).find(e => {
+        return obsoleteElements.has(e);
```
rviscomi, line 35: Do you need to normalize the case?
```suggestion
        return obsoleteElements.has(e.toLowerCase());
```

sql/2020/03_Markup/pages_element_count_by_device.sql
```diff
@@ -0,0 +1,69 @@
...
+return result;
+''';
+
+SELECT
+  client,
+  COUNT(0) AS total,
+
+  CAST(ROUND(AVG(element_count_info.count)) AS INT64) AS elements_avg, # not sure what value this is? average elements per page? same for all device rows
+  MIN(element_count_info.count) AS elements_min, # y-axis min, same for all device rows
+  MAX(element_count_info.count) AS elements_max, # y-axis max, same for all device rows
```
rviscomi, line 48: Should these be percentiles?

sql/2020/03_Markup/pages_element_count_by_device_and_custom_dash_elements.sql
```diff
@@ -0,0 +1,54 @@
...
+# COUNT(COUNT(0)) OVER (PARTITION BY _TABLE_SUFFIX) AS total,
+
+CREATE TEMP FUNCTION AS_PERCENT (freq FLOAT64, total FLOAT64) RETURNS FLOAT64 AS (
+  ROUND(SAFE_DIVIDE(freq, total), 4)
+);
+
+CREATE TEMPORARY FUNCTION get_element_types_with_a_dash(element_count_string STRING)
+RETURNS ARRAY<STRING> LANGUAGE js AS ''' 
+try {
+    if (!element_count_string) return ["ERROR: payload._element_count is missing"]; // 2019 had a few cases
```
rviscomi, line 16: This is useful diagnostic info but not sure if we need it in the final results. Ideally the results are cleaned up for public consumption.

sql/2020/03_Markup/pages_element_count_by_device_and_element_type_frequency.sql
```diff
@@ -0,0 +1,40 @@
...
+} catch (e) {
+    return []; 
+}
+''';
+
+SELECT
+  _TABLE_SUFFIX AS client,
+  element_type_info.name,
+  SUM(element_type_info.freq) AS freq_m201, # total count from all pages
+  AS_PERCENT(SUM(element_type_info.freq), SUM(SUM(element_type_info.freq)) OVER (PARTITION BY _TABLE_SUFFIX)) AS pct_m202
```
rviscomi, line 29: This will be biased towards elements that are used a lot, which may be what you're looking for. Instead, consider only counting elements once per page, so these stats would be in terms of the number/percent of pages having the element.

sql/2020/03_Markup/pages_element_count_by_device_and_element_type_frequency.sql
```diff
@@ -0,0 +1,40 @@
...
+  `httparchive.pages.2020_08_01_*`,
+  UNNEST(get_element_types_info(JSON_EXTRACT_SCALAR(payload, '$._element_count'))) AS element_type_info
+GROUP BY
+  client,
+  element_type_info.name
+ORDER BY
+  client,
+  pct_m202 DESC
+LIMIT
+  10000
```
rviscomi, line 40: Consider a smaller limit or `HAVING` clause.

sql/2020/03_Markup/pages_element_count_by_device_and_element_type_present.sql
```diff
@@ -0,0 +1,46 @@
...
+# See related: sql/2019/03_Markup/03_02a.sql
+
+CREATE TEMP FUNCTION AS_PERCENT (freq FLOAT64, total FLOAT64) RETURNS FLOAT64 AS (
+  ROUND(SAFE_DIVIDE(freq, total), 4)
+);
+
+CREATE TEMPORARY FUNCTION get_element_types(element_count_string STRING)
+RETURNS ARRAY<STRING> LANGUAGE js AS '''
+try {
+    if (!element_count_string) return ["ERROR: payload._element_count is missing"]; // 2019 had a few cases
```
rviscomi, line 12: Ditto on diagnostic info

sql/2020/03_Markup/pages_element_count_by_device_and_element_type_present.sql
```diff
@@ -0,0 +1,46 @@
...
+
+    return Object.keys(element_count); 
+} catch (e) {
+    return ["ERROR: "+e.message]; 
+}
+''';
+
+ SELECT
+  _TABLE_SUFFIX AS client,
+  element_type,
```
rviscomi, line 27: Are all of the element types the same case, or should this be `LOWER(element_type)`?

rviscomi, line 27: Include a limit/having clause

sql/2020/03_Markup/pages_element_count_by_device_and_element_types_used_per_page.sql
```diff
@@ -0,0 +1,35 @@
...
+} catch (e) { 
+  return null;
+}
+
+''';
+
+SELECT
+  _TABLE_SUFFIX AS client,
+  get_element_types_used(JSON_EXTRACT_SCALAR(payload, '$._element_count')) AS element_types_used,
+  COUNT(0) AS freq
```
rviscomi, line 25: I think this is a job for percentiles

sql/2020/03_Markup/pages_element_count_by_device_and_obsolete_elements.sql
```diff
@@ -0,0 +1,45 @@
...
+# See related: sql/2019/03_Markup/03_01b.sql
+
+CREATE TEMP FUNCTION AS_PERCENT (freq FLOAT64, total FLOAT64) RETURNS FLOAT64 AS (
+  ROUND(SAFE_DIVIDE(freq, total), 4)
+);
+
+CREATE TEMPORARY FUNCTION get_element_types(element_count_string STRING)
+RETURNS ARRAY<STRING> LANGUAGE js AS '''
+try {
+    if (!element_count_string) return ["ERROR: payload._element_count is missing"]; // 2019 had a few cases
```
rviscomi, line 12: Ditto on diagnostic info

sql/2020/03_Markup/pages_element_count_by_device_and_obsolete_elements.sql
```diff
@@ -0,0 +1,45 @@
...
+    if (typeof element_count != 'object') return ["ERROR: payload._element_count is a " + typeof(element_count)];
+
+    return Object.keys(element_count); 
+} catch (e) {
+    return ["ERROR: "+e.message]; 
+}
+''';
+
+CREATE TEMPORARY FUNCTION is_obsolete(element STRING) AS (
+  element IN ("applet","acronym","bgsound","dir","frame","frameset","noframes","isindex","keygen","listing","menuitem","nextid","noembed","plaintext","rb","rtc","strike","xmp","basefont","big","blink","center","font","marquee","multicol","nobr","spacer","tt")
```
rviscomi, line 26:
```suggestion
  LOWER(element) IN ("applet","acronym","bgsound","dir","frame","frameset","noframes","isindex","keygen","listing","menuitem","nextid","noembed","plaintext","rb","rtc","strike","xmp","basefont","big","blink","center","font","marquee","multicol","nobr","spacer","tt")
```

sql/2020/03_Markup/pages_element_count_by_device_and_obsolete_elements.sql
```diff
@@ -0,0 +1,45 @@
...
+);
+
+SELECT
+  _TABLE_SUFFIX AS client,
+  element_type AS obsolete_element_type,
+  COUNT(0) AS freq,
+  SUM(COUNT(0)) OVER (PARTITION BY _TABLE_SUFFIX) AS total_pages_with_obsolete_elements, # not accurate unless you only care about pages with obsolete element
```
rviscomi, line 33: Is this the total pages or the total number of obsolete elements?

sql/2020/03_Markup/pages_element_count_by_device_and_obsolete_elements.sql
```diff
@@ -0,0 +1,45 @@
...
+}
+''';
+
+CREATE TEMPORARY FUNCTION is_obsolete(element STRING) AS (
+  element IN ("applet","acronym","bgsound","dir","frame","frameset","noframes","isindex","keygen","listing","menuitem","nextid","noembed","plaintext","rb","rtc","strike","xmp","basefont","big","blink","center","font","marquee","multicol","nobr","spacer","tt")
+);
+
+SELECT
+  _TABLE_SUFFIX AS client,
+  element_type AS obsolete_element_type,
```
rviscomi, line 31: FWIW I'm not suggesting using limit/having clauses here since the obsolete elements list is relatively small.

sql/2020/03_Markup/pages_markup_by_device_and_button_types.sql
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+} catch (e) {}
+return result;
+''';
+
+SELECT
+  _TABLE_SUFFIX AS client,
+  button_type_info.name AS button_type,
+  SUM(button_type_info.freq) AS freq,
```
rviscomi, line 32: Do you intend to make the frequency the number of occurrences of the button type, or the number of pages having that button type? I think the latter may be more meaningful to readers since it doesn't vary by pages that may use a button type many many times.

sql/2020/03_Markup/pages_markup_by_device_and_button_types.sql
```diff
@@ -0,0 +1,44 @@
...
+      return Object.entries(markup.buttons.types).map(([name, freq]) => ({name, freq})); 
+    }
+
+} catch (e) {}
+return result;
+''';
+
+SELECT
+  _TABLE_SUFFIX AS client,
+  button_type_info.name AS button_type,
```
rviscomi, line 31: If this value can be high cardinality, again I'd suggest LIMIT/HAVING.

sql/2020/03_Markup/pages_wpt_bodies_by_device_and_protocol.sql
```diff
@@ -0,0 +1,46 @@
...
+
+} catch (e) {}
+return result;
+''';
+
+SELECT
+  _TABLE_SUFFIX AS client,
+  COUNT(DISTINCT url) AS pages,
+  total,
+  protocol,
```
rviscomi, line 30: Casefold. Also LIMIT/HAVING if this is high cardinality. Probably not.
```suggestion
  LOWER(protocol) AS protocol,
```

sql/2020/03_Markup/summary_pages_by_device.sql
```diff
@@ -0,0 +1,25 @@
...
+# live is very cheap
+
+CREATE TEMP FUNCTION AS_PERCENT (freq FLOAT64, total FLOAT64) RETURNS FLOAT64 AS (
+  ROUND(SAFE_DIVIDE(freq, total), 4)
+);
+
+SELECT
+  _TABLE_SUFFIX AS client,
+  COUNTIF(TRIM(doctype) <> '') AS freq_doctype,
+  SUM(COUNT(0)) OVER (PARTITION BY _TABLE_SUFFIX) AS total,
```
rviscomi, line 13: Can this be simplified? `client` is the only field being grouped, so I'm not sure that you need to SUM/OVER. Here and in `pct` fields.
```suggestion
  COUNT(0) AS total,
```

sql/2020/03_Markup/summary_pages_by_device.sql
```diff
@@ -0,0 +1,25 @@
...
+  AS_PERCENT(COUNTIF(TRIM(doctype) <> ''), SUM(COUNT(0)) OVER (PARTITION BY _TABLE_SUFFIX)) AS pct_doctype_m102, 
+  MIN(bytesHtml) AS min_bytes_html_m109, 
+  MAX(bytesHtml) AS max_bytes_html_m108, 
+  ROUND(AVG(bytesHtml),0) AS avg_bytes_html_m110,
```
rviscomi, line 17: Should these be percentiles?

rviscomi: Actually, I see this is it's own standalone query. Can these be removed?

rviscomi (changes requested):

sql/2020/03_Markup/pages_almanac_by_device_and_attribute_name_frequency.sql
Tiggerito: Looks like I missed a few. Did a search and replaced all to 1,000 or less.

sql/2020/03_Markup/pages_almanac_by_device.sql
Tiggerito: Done

sql/2020/03_Markup/pages_almanac_by_device_and_attribute_name_present.sql
Tiggerito: Changed to 1,000 and added the LIMIT to the second one.

sql/2020/03_Markup/pages_almanac_by_device_and_favicon_image_type.sql
Tiggerito: Done

sql/2020/03_Markup/pages_almanac_by_device_and_html_lang.sql
Tiggerito: It's just the raw values trimmed and in lowercase. I've already done a live query on this. Let's see if it's enough for the team. If not I can invent something new. It includes frequency data so it should be processable in the sheet.

sql/2020/03_Markup/pages_element_count_by_device.sql
Tiggerito: Already done in the custom metric.

sql/2020/03_Markup/pages_element_count_by_device.sql
Tiggerito: They're not used. I think pulled in from last year. I'll drop them. A percentile of just element count would be of use and would cover all three. I'll add that.

sql/2020/03_Markup/pages_element_count_by_device_and_custom_dash_elements.sql
Tiggerito: Changed to return an empty array on any error or if nothing is detected.

sql/2020/03_Markup/pages_element_count_by_device_and_element_type_frequency.sql
Tiggerito: pages_element_count_by_device_and_element_type_present.sql does that. They wanted it both ways.

sql/2020/03_Markup/pages_element_count_by_device_and_element_type_frequency.sql
Tiggerito: done

sql/2020/03_Markup/pages_element_count_by_device_and_element_type_present.sql
Tiggerito: the custom metric returns them all in lowercase.

`let tag = el.tagName.toLowerCase()`

Tiggerito: Done

sql/2020/03_Markup/pages_element_count_by_device_and_element_type_present.sql
Tiggerito: Same done

sql/2020/03_Markup/pages_element_count_by_device_and_element_types_used_per_page.sql
Tiggerito: Added to pages_element_count_by_device_and_percentile.sql and removed this.

sql/2020/03_Markup/pages_element_count_by_device_and_obsolete_elements.sql
Tiggerito: ditto on returning an empty array.

sql/2020/03_Markup/pages_element_count_by_device_and_obsolete_elements.sql
Tiggerito: Already lowercase. Should I be defensive and add it?

sql/2020/03_Markup/pages_markup_by_device_and_button_types.sql
Tiggerito: done

sql/2020/03_Markup/pages_wpt_bodies_by_device_and_protocol.sql
Tiggerito: Already lowercase in the custom metric. Added a LIMIT.

sql/2020/03_Markup/summary_pages_by_device.sql
Tiggerito: These are to cover specific requests made by the team. I added bytesHtml as a percentile already as I thought that may be of more use.

sql/2020/03_Markup/pages_markup_by_device_and_button_types.sql
Tiggerito: I think I've worked this one out. I struggle with these totals and frequencies:

```
  _TABLE_SUFFIX AS client,
  #total AS pages_total,
  button_type_info.name AS button_type,
  COUNTIF(button_type_info.freq > 0) AS freq_page_with_button, 
  AS_PERCENT(COUNTIF(button_type_info.freq > 0), total) AS pct_page_with_button,
  SUM(button_type_info.freq) AS freq_button, 
  #SUM(SUM(button_type_info.freq)) OVER (PARTITION BY _TABLE_SUFFIX) as buttons_total,
  AS_PERCENT(SUM(button_type_info.freq), SUM(SUM(button_type_info.freq)) OVER (PARTITION BY _TABLE_SUFFIX)) AS pct_button
```

pct_page_with_button should be based on pages that contain the button at least once. What you suggested.
pct_button is using the raw count of buttons, so is more like a ratio for the types of buttons used.

Would that make sense?

sql/2020/03_Markup/pages_element_count_by_device_and_obsolete_elements.sql
Tiggerito: I've re written it to mainly be based on total pages. I've kept that field but renamed to to something it more accurately represents. The ratio of pages using specific obsolete elements.

sql/2020/03_Markup/pages_markup_by_device_and_button_types.sql
Tiggerito: I've submitted the change (minus commented out lines)

Tiggerito:
@rviscomi If your happy with my comments, I think I've covered everything.

rviscomi (approved):

rviscomi:
Massive effort, thanks @Tiggerito!

PR #1142, opened 2020-08-02
Ebook social links

### LINKED ISSUE #1140: [Ebook] Author alignment and social links
For example, see my author bio in the [2019 ebook](https://almanac.httparchive.org/static/pdfs/web_almanac_2019_en.pdf):

image

I'd suggest two changes:

1. Change the alignment to more closely match the [web](https://almanac.httparchive.org/en/2019/performance#authors) version in which the photo is to the left of the bio.

2. Add footnote citations to the social links so that their URLs are readable in the footer, similar to the links in the bio.

image

On a related note, the Appendix B section for contributors has the social icons but no footnotes. Maybe we should drop the social icons from this section of the ebook? Otherwise the footnotes alone would take up pages.

image

@bazzadp what are your thoughts on these changes?

bazzadp:
Fixes #1140 

Profile picture is now at the side and author links now link the social media profiles:

Author bio with links

For contributors I decided to leave the Twitter handle in where there is one, to allow some way of readers to follow contributors (well at least most of them). Figured GitHub user id is less useful for this and websites will have long URLs so Twitter is best:

Remove social media icons and replace with Twitter username


src/static/css/ebook.css
```diff
@@ -439,6 +439,20 @@ section.chapter {
     width: 4.29rem;
   }
 
+  /* Prince does not yet support Grid https://www.princexml.com/roadmap/ */
+  @supports not (display: grid) {
+    .content .authors li {
+      display: flex;
+      margin-bottom: 10px;
+      margin-bottom: 1rem;
```
SaptakS, line 447: Why the repeated `margin-bottom` attribute here?

bazzadp: It's partially a force of habit from the old days when browser support of `rem` was spotty, but truth be told, it's more about how my brain seems incapable of thinking in `rem`s - so I develop in `px`, then convert to `rem` and leave the old `px` value in place to remind me of what the `px` value was.

It's especially useful for places like here where the default `rem` value is set to `10px` instead of the usual `16px` that people come to expect.

Not that is matter much what the actual `px` value is - how it looks is more important and it could be any unit but still - this old dog seems incapable of learning new tricks sometimes 😀

This is explained in our main `2019.css` file:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/ba5b2dbab8758a70d1adf56f6ac78ad15b929475/src/static/css/2019.css#L313-L341

src/static/css/ebook.css
```diff
@@ -465,6 +479,10 @@ section.chapter {
     line-height: none;
   }
 
+  .authors .social .fn::footnote-call {
+    vertical-align: sub;
```
rviscomi, line 483: Text links use superscript. Why give social icons subscript?

src/templates/base/2019/base_ebook.html
```diff
@@ -345,30 +345,7 @@ <h1 class="title title-lg">
...
-          {% endif %}
-          {% if contributor.website %}
-          <a href="{{ contributor.website }}" aria-labelledby="contributors-{{ id }}-website">
-            <svg role="img">
-              <title id="contributors-{{ id }}-website">{{ website(contributor.name) }}</title>
-              <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#globe"></use>
-            </svg>
-          </a>
-          {% endif %}
+            {% if contributor.twitter %} (@{{ contributor.twitter }}){% endif %}
```
bazzadp, line 348: @rviscomi this is the contributors social section!

src/static/css/ebook.css
bazzadp: It’s because the images are vertically centred according to the text and, as the images were larger than the line-height, the superscript actually looked like it was the middle of the images. Subscript looked cleaner. Did look at changing all links to subscript but didn’t really like it. Could look at reverting this to superscript and then moving it up a bit more so it really is superscript to the images if you want?

bazzadp: This is fixed (and screenshot in first comment updated) but still required a manual vertical-height to be set. But at least it's super-script again now so consistent.

src/templates/base/2019/base_ebook.html
rviscomi: Thanks!

Small design suggestion. What would you think of more distinction between name/Twitter by having them on separate lines, different font size, and/or different shade text?

Similar to how names/handles appear on Twitter itself:

image

bazzadp: What do you reckon - with Twitter icon or without?

With Twitter icon

Without Twitter icon

rviscomi: Both look great! If I had to pick, the version without.

bazzadp: Think the full works is too much?

All social media with each contributor

bazzadp: And with LinkedIn:

Contributor Social Icons

bazzadp: I actually like it, even if it is a little busy. It's an appendix anyway.

I also implemented a similar way in the Authors section and removed links when the URL is used:

Better Author Bios

I like this so much I'm tempted to change the website to this format as well - WDYT?

rviscomi: LGTM. Only nit is to ensure the icons and the text are vertically aligned.

bazzadp: They are vertically aligned and look perfect in Chrome, but Prince seems to be adding an offset for some reason. Did try various things to remove it but never got there. Will have another look with a fresh pair of eyes.

bazzadp: Fixed it with `vertical-align: text-top;`:

Author bio

PR #1156, opened 2020-08-04
Fixes deeplinks for figures with ":" instead of "."

### LINKED ISSUE #1154: No clickable deeplink for CDN figure 2
See https://almanac.httparchive.org/en/2019/cdn#fig-2

image

I'd expect the text "Figure 2" to link to the URL above, but there is no link. Is this because it is suffixed with `:` rather than `.`? Should be an easy fix.

SaptakS:
Fixes #1154


SaptakS:
I was kind of in a dilemma between whether to push or not to given I saw that it has a GitHub action `.yml`. But thanks for clearing it.

bazzadp:
No harm top do it - just a bit of wasted effort as it will be over written. See #1157 for the auto generated one for this one.

And the reason we do it on merge, rather than as part of pull request is to avoid merge conflicts when the timestamp is updated (also done as part of this) and also to make pull requests easier as less to review! 😀

SaptakS:
Makes sense

Issue #907, opened 2020-06-27
Accessibility 2020


obto:
# Part II Chapter 8: Accessibility

## Content team
| Authors | Reviewers | Analysts | Draft | Queries | Results |
| ------- | --------- | -------- | ----- | ------- | ------- |
| @oluoluoxenfree  @alextait1 |  @aardrian @ericwbailey @obto | @obto | [Doc][~draft-doc] | [*.sql][~sql-dir] | [Sheet][~results-sheet] |

**Content team lead:** @oluoluoxenfree

Welcome chapter contributors! You'll be using this issue throughout the [chapter lifecycle](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle) to coordinate on the content planning, analysis, and writing stages.

The **content team** is made up of the following contributors:

- [Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)
- [Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)
- [Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)

**New contributors:** If you're interested in joining the content team for this chapter, just leave a comment below and the content team lead will loop you in.

_Note: To ensure that you get notifications when tagged, you must be "[watching](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#notifications)" this repository._

### [**0. Form the content team**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#0-create-content-team)
+ [x] **Jul 6th**: Project owners have selected an author to be the content team lead
+ [x] **Jul 13th**: The content team has at least one author, reviewer, and analyst (minimally viable team formed)

### [**1. Plan content**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
+ [x] **Jul 20th:** The content team has completed the chapter outline in the [draft doc][~draft-doc]
+ [x] **Jul 27th**: Analysts have triaged the feasibility of all proposed metrics

### [**2. Gather data**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
[...]

Issue #1138, opened 2020-08-01
Streamline figures in markdown


rviscomi:
I'd like to propose a change in the way figures are described by authors in the markdown and rendered internally.

![image](https://user-images.githubusercontent.com/1120896/89106384-65693d00-d3f7-11ea-9c89-1552d2f05885.png)

In [css.md](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/src/content/en/2019/css.md), the corresponding markdown (using embedded HTML) for the figure above is:

```html
<figure>
  <a href="/static/images/2019/css/fig12.png">
    <img src="/static/images/2019/css/fig12.png" alt="Figure 12. Adoption of flexbox." aria-labelledby="fig12-caption" aria-describedby="fig12-description" width="600" height="371" data-width="600" data-height="371" data-seamless data-frameborder="0" data-scrolling="no" data-iframe="https://docs.google.com/spreadsheets/d/e/2PACX-1vQO5CabwLwQ5Lj1_9bbEFnFM1qEqCorymaBHrcaNiMSJ7sYDKHUI5iish5VAS-SxN447UTW-1-5-OjE/pubchart?oid=2021161093&amp;format=interactive">
  </a>
  <div id="fig12-description" class="visually-hidden">Bar chart showing 49% of desktop pages and 52% of mobile pages using flexbox.</div>
  <figcaption id="fig12-caption">Figure 12. Adoption of flexbox.</figcaption>
</figure>
```

One thing I'd like to improve is the figure numbering. The figure ID is automatically generated, but we're still manually writing `Figure 12` in the `<figcaption>` and managing the IDs needed for the accessible description. Can all of this be automated? I'd also like to make all figure numbers unique and prefixed by chapter number. So rather than `Figure 12` this would be `Figure 2.12`.

The `fig12.png` image shouldn't be named according to its figure number. These names aren't descriptive and must be kept in sync with the figure ordering in the chapter. Instead, we should name figure images descriptively, like `flexbox-adoption.png`. We should use this name as the `<figure id>` and permalink so that reordering figures wouldn't make old links obsolete.

[...]

PR #1159, opened 2020-08-04
A11Y queries

obto:
All queries for #907. Currently in progress

- [x] 1. % mobile sites with sufficient text color contrast with its background
- [x] 2. Disabled zooming and scaling via the viewport tag
- [x] 3. % of mobile sites with a valid html lang attribute
- [x] 4. Most common html lang attributes
- [x] 5. Alt text median, mean, and max (ignoring empty and missing alts)
- [x] 6. Most common lengths of alt text (-1 for none. 2000+ grouped together)
- [x] 7. Alt text ending in an image extension
- [x] 8. How often the title attribute is used with an alt attribute, and how often they are the same values
- [x] 9. How often pages contain an element with a given attribute
- [x] 10. % of sites using each type of aria role
- [x] 11. % sites with links to to the same page, # or javascript:void
- [x] 12. % of pages having skip links
- [x] 13. Most popular accesskey or aria-keyshortcuts keys
- [x] 14. Table stats. Total all, captioned and presentational
- [x] 15. th and td stats. Scope and headers usage
- [x] 16. Divs or spans with role of button or link
- [x] 17. Sites containing elements with role="tab", aria-selected and aria-controls attributes
- [x] 18. Form controls with placeholder but no label
- [x] 19. Sites that have associated labels for their form elements
- [x] 20. Sites using sr-only or visually-hidden classes
- [x] 21. Page title stats (usage, descriptive, changed on render)
- [x] 22. Duplicate ID occurrences
- [x] 23. Positive tabindex value occurrences
- [x] 24. Anchors with role="button"
- [x] 25. Video elements attribute usage
- [x] 26. Video elements track usage


obto:
This PR is ready for review. Right now the queries use the sample table data, but that can quickly be swapped out afterwards.

Should make reviewing much less expensive as well.

rviscomi (changes requested):
Could you rename the SQL files from `08_01` format to more descriptive names, eg `zoom_scale.sql`? We're breaking from the convention used last year because the queries weren't easily discoverable and this will also fit into some of the work we're doing for #1138.

sql/2020/08_Accessibility/08_10.sql
```diff
@@ -0,0 +1,35 @@
...
+  FROM
+    `httparchive.pages.2020_08_01_*`
+  GROUP BY _TABLE_SUFFIX
+)
+USING (_TABLE_SUFFIX)
+GROUP BY
+  client,
+  role,
+  total_sites
+ORDER BY
```
rviscomi, line 34: HAVING >= 100?

rviscomi (changes requested):
Did a thorough review of the first half, sending it back to you to iterate while I find time to do the rest.

Really good stuff!

sql/2020/08_Accessibility/08_05.sql
```diff
@@ -0,0 +1,21 @@
+#standardSQL
+# 08_05: Alt text median, mean, and max (ignoring empty and missing alts)
+SELECT
+  client,
+  COUNT(0) AS total_non_zero_alts,
+  APPROX_QUANTILES(alt_length, 100)[SAFE_ORDINAL(50)] AS median,
```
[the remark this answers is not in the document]
obto: Good call. I'll add 100 as a percentile here as i think it'd be humorous to see the longest alt tag

sql/2020/08_Accessibility/08_05.sql
```diff
@@ -0,0 +1,21 @@
+#standardSQL
+# 08_05: Alt text median, mean, and max (ignoring empty and missing alts)
+SELECT
+  client,
+  COUNT(0) AS total_non_zero_alts,
```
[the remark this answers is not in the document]
obto: Good call. Going to remove it

sql/2020/08_Accessibility/08_07.sql
```diff
@@ -0,0 +1,72 @@
...
+RETURNS ARRAY<STRUCT<extension STRING, total INT64>> LANGUAGE js AS '''
+try {
+  const a11y = JSON.parse(payload);
+
+  return Object.keys(a11y.file_extension_alts.file_extensions).map(key => {
+    return {
+      extension: key,
+      total: a11y.file_extension_alts.file_extensions[key],
+    };
+  });
```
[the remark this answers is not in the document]
obto: Very much so

obto:
Updated files based on feedback. Would you like me to rename each file name in this PR? or in a separate one?

rviscomi:
Descriptive file names in this PR would be great 🙏

obto:
@rviscomi Descriptive filenames have been added. Going to start adding query results to the spreadsheet

bazzadp:
@obto I posted this in the slack, but I’ve a couple of Lighthouse queries that you might want to use for this chapter:
- [List of all Lighthouse audits for a category](https://github.com/HTTPArchive/almanac.httparchive.org/blob/pwa-sql-2020/sql/2020/14_PWA/lighthouse_pwa_audits.sql) I use `pwa` in my query but easily changes to `accessibility`. Results [here](https://docs.google.com/spreadsheets/u/0/d/1i0_3Z8dDewjCgD0JVR5chXSd7OQVZXemlZuTYlRESgg/)
- [Lighthouse score for a category by percentile](https://github.com/HTTPArchive/almanac.httparchive.org/blob/pwa-sql-2020/sql/2020/14_PWA/lighthouse_pwa_score.sql).

You probably have a lot more detail in your many queries but still, these two queries are easily run since I’ve written them, so might be interesting to provide to the Authors to see what other nuggets these stats yield.

obto:
@bazzadp Great idea. Just added the LH score by percentile.

The audit by category will need to be optimized since it timed out after running out of memory. So I'm going to refrain from futzing with it and adding any more queries till we get this PR merged and the authors check out the data :)

obto:
All queries ran successfully and can be seen in the spreadsheet: https://docs.google.com/spreadsheets/d/1UjEBhq0TfYxUpdpq5IuxjeHB4yqhJq4NOKEd6Dwwrdk/edit#gid=1067352317

bazzadp:
> The audit by category will need to be optimized since it timed out after running out of memory. So I'm going to refrain from futzing with it and adding any more queries till we get this PR merged and the authors check out the data :)

I think I've improved it enough there by excluding really, really large sites (16 sites). Checkout [latest version](https://github.com/HTTPArchive/almanac.httparchive.org/blob/pwa-sql-2020/sql/2020/14_PWA/lighthouse_pwa_audits.sql).

sql/2020/08_Accessibility/alt_text_length.sql
```diff
@@ -0,0 +1,28 @@
+#standardSQL
+# Alt text length
+SELECT
+  client,
+  MAX(alt_length) as max_alt_length,
+
+  percentile,
+  APPROX_QUANTILES(alt_length, 1000)[SAFE_ORDINAL(percentile * 10)] AS alt_length
```
rviscomi, line 8: This should be compatible with the 100th percentile instead of `MAX` 
```suggestion
  APPROX_QUANTILES(alt_length, 1000)[OFFSET(percentile * 10)] AS alt_length
```

sql/2020/08_Accessibility/alt_text_length.sql
```diff
@@ -0,0 +1,28 @@
...
+      SAFE_CAST(alt_length_string AS INT64) AS alt_length
+    FROM
+      `httparchive.pages.2020_08_01_*`,
+      UNNEST(
+        JSON_EXTRACT_ARRAY(JSON_EXTRACT_SCALAR(payload, '$._almanac'), '$.images.alt_lengths')
+      ) AS alt_length_string
+  ),
+  UNNEST([10, 25, 50, 75, 90]) AS percentile
+WHERE
+  alt_length > 0 AND alt_length IS NOT NULL
```
rviscomi, line 22: I think the `IS NOT NULL` is redundant with the `> 0`
```suggestion
  alt_length > 0
```

sql/2020/08_Accessibility/common_alt_text_length.sql
```diff
@@ -0,0 +1,33 @@
...
+  client,
+  SUM(COUNT(0)) OVER (PARTITION BY client) AS total_alt_tags,
+
+  alt_length_clipped AS alt_length,
+  COUNT(0) AS occurrences,
+  COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client) AS pct_all_occurrences
+FROM (
+  SELECT
+    client,
+    IF(alt_length >= 2000, 2000, alt_length) AS alt_length_clipped
```
rviscomi, line 13: Totally optional, but I think `LEAST` would be the semantic way to do this in BigQuery
```suggestion
    LEAST(2000, alt_length) AS alt_length_clipped
```

sql/2020/08_Accessibility/common_element_attributes.sql
```diff
@@ -0,0 +1,37 @@
...
+    `httparchive.pages.2020_08_01_*`
+  GROUP BY _TABLE_SUFFIX
+)
+USING (_TABLE_SUFFIX)
+GROUP BY
+  client,
+  attribute,
+  total_sites
+HAVING
+  total_sites_using > 1000
```
rviscomi, line 35: I don't think this threshold is high enough - there are 16,000 rows in the sheet. There could be a lot of unusable data in there, making it harder to sift through and it slows down the sheet's performance.

\>= 1% still leaves 1,000 results

Also consider filtering the attributes to a known set of a11y attributes.
```suggestion
  pct_sites_using >= 0.01
```

sql/2020/08_Accessibility/common_video_attributes.sql
```diff
@@ -0,0 +1,51 @@
...
+}
+''';
+SELECT
+  client,
+  total_sites,
+
+  total_sites_with_video,
+  pct_sites_with_video,
+
+  attribute,
```
rviscomi, line 19: Do you care about all attributes or only the ones related to a11y?

sql/2020/08_Accessibility/duplicate_id.sql
```diff
@@ -0,0 +1,25 @@
...
+# Duplicate ID occurrences
+SELECT
+  client,
+  SUM(COUNT(0)) OVER (PARTITION BY client) AS total_sites,
+  SUM(COUNTIF(total_duplicate_ids > 0)) OVER (PARTITION BY client) AS total_with_duplicate_ids,
+  SUM(COUNTIF(total_duplicate_ids > 0)) OVER (PARTITION BY client) / SUM(COUNT(0)) OVER (PARTITION BY client) AS pct_with_duplicate_ids,
+  MAX(total_duplicate_ids) AS max_duplicate_ids,
+
+  percentile,
+  APPROX_QUANTILES(total_duplicate_ids, 1000)[SAFE_ORDINAL(percentile * 10)] AS total_duplicate_ids
```
rviscomi, line 11: Ditto re p100

sql/2020/08_Accessibility/skip_links.sql
```diff
@@ -0,0 +1,34 @@
...
+    `httparchive.pages.2020_08_01_*`
+  GROUP BY
+    _TABLE_SUFFIX
+)
+USING (_TABLE_SUFFIX)
+WHERE
+  getEarlyHash(JSON_EXTRACT_SCALAR(payload, '$._almanac')) > 0
+GROUP BY
+  client,
+  total
```
rviscomi, line 34: Same results, just a bit cleaner
```suggestion
SELECT
  _TABLE_SUFFIX AS client,
  COUNTIF(getEarlyHash(JSON_EXTRACT_SCALAR(payload, '$._almanac')) > 0) AS pages,
  COUNT(0) AS total,
  COUNTIF(getEarlyHash(JSON_EXTRACT_SCALAR(payload, '$._almanac')) > 0) / COUNT(0) AS pct
FROM
  `httparchive.pages.2020_08_01_*`
GROUP BY
  client
```

sql/2020/08_Accessibility/title_used_with_alt.sql
```diff
@@ -0,0 +1,23 @@
+#standardSQL
+# How often the title attribute is used with an alt attribute, and how often they are the same values
+SELECT
+  client,
+  SUM(total_alt) AS total_alts,
```
rviscomi, line 5: Sometimes you do `COUNTIF(total > 0)`. Why not here?

rviscomi (changes requested):

sql/2020/08_Accessibility/alt_ending_in_image_extension.sql
```diff
@@ -0,0 +1,69 @@
+#standardSQL
+# Alt text ending in an image extension
```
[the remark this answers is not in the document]
obto: Each answers a slightly different question, depending on how the authors want to write about this metric. I added comments to each in the spreadsheet: https://docs.google.com/spreadsheets/d/1UjEBhq0TfYxUpdpq5IuxjeHB4yqhJq4NOKEd6Dwwrdk/edit#gid=704032354

Only metric I could reasonably remove, would be `pct_sites_using`

sql/2020/08_Accessibility/anchors_with_role_button.sql
```diff
@@ -0,0 +1,20 @@
+#standardSQL
+# Anchors with role='button'
+SELECT
+  client,
+  COUNT(0) AS total_sites,
```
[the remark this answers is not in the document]
obto: Are you referring to the `total_sites` stat? Yeah thats a good point, can probably tear it out of almost all the queries

sql/2020/08_Accessibility/common_alt_text_length.sql
obto: ooo i never knew this existed. I'll definitely swap to this instead

sql/2020/08_Accessibility/common_element_attributes.sql
obto: The tough part is this data isn't just for `aria-*` properties, and some of aria properties like `aria-grabbed` only have 0.00018952% usage.

If we're worried about sheet performance, I could remove rows Q-Z in the sheet. So we'd reduce the number of cells by half. Or I could also manually delete non-aria properties until we reach 0.01% of usage like you're suggesting

sql/2020/08_Accessibility/common_video_attributes.sql
obto: Mostly those for a11y, but there are others which may be interesting to the authors like the `mute` attribute.

Since there's not that many results I decided it'd be easiest to give them all the data. Might spark their curiosity and give them ideas for even more interesting queries.

sql/2020/08_Accessibility/skip_links.sql
obto: LGTM.

I do wonder how this affects the performance of the query. Ideally BQ would be able to optimize this only running the `getEarlyHash` function once... but I guess it has no way to know if the function is idempotent

sql/2020/08_Accessibility/title_used_with_alt.sql
obto: Because we're looking at the sheer number of alts overall, and not the individual sites themselves. So I have no need to filter out the sites which do not use alts.

obto (commented):
Thanks for the review. I'll get these fixed up by EOD

sql/2020/08_Accessibility/common_element_attributes.sql
rviscomi: Maybe something like this
```suggestion
  STARTS_WITH(attribute, 'aria-') OR
  total_sites_using > 1000
```

obto:
> > The audit by category will need to be optimized since it timed out after running out of memory. So I'm going to refrain from futzing with it and adding any more queries till we get this PR merged and the authors check out the data :)
> 
> I think I've improved it enough there by excluding really, really large sites (16 sites). Checkout [latest version](https://github.com/HTTPArchive/almanac.httparchive.org/blob/pwa-sql-2020/sql/2020/14_PWA/lighthouse_pwa_audits.sql).

Terrific! I'll check this out :)

sql/2020/08_Accessibility/skip_links.sql
rviscomi: Both queries run in about 45 seconds and consume the same amount of data. It seems BQ is able to optimize the repeated function calls.

sql/2020/08_Accessibility/common_element_attributes.sql
obto: LGTM

sql/2020/08_Accessibility/skip_links.sql
obto: Wow thats really cool

sql/2020/08_Accessibility/alt_ending_in_image_extension.sql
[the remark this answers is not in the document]
obto: Removed `pct_sites_using` and added more inline documentation

obto:
@bazzadp One change I had to make to your lighthouse audit script, is to check if the audit is applicable to a site or not. Otherwise the %'s you get are way off.

So instead of using `COUNT(0) AS total`, I use:

```sql
SAFE_DIVIDE(COUNTIF(audits.score > 0), COUNTIF(audits.score IS NOT NULL)) AS pct,
```

obto:
@rviscomi I believe we are now good to go

rviscomi (approved):
Thanks for making these changes! Let's get these committed and iterate as needed.

PR #1161, opened 2020-08-05
Allow different commas per language

bazzadp:
Noticed we use Western commas even in Japanese and Chinese languages where a different comma is used ("、" instead of ","):

Western commas

This PR allows localized commas:

Localized commas


rviscomi:
I saw these commas in a recent PR and had the _exact same_ thought, that we should update them on the contributors page. Thanks for fixing this!

PR #1164, opened 2020-08-05
Test website on pull requests as well as pushes to main

bazzadp:
In #1124 we added the ability for GitHub Actions to build the chapters, run the site and run some tests. We kept that to only happen on pushes to main for now but since it is working well I want to now do this on each pull request too.

However for pull requests we do not update the timestamps, and also crucially do not commit any changes - they will continue to be done on pushes to main. This is to avoid merge conflicts when timestamps are updated and to keep pull requests smaller and easier to review.

I'd like to keep this check off from generic pushes, as it does take a few mins to run. Think it's find to only run on pull request when code should be ready to commit.

I've also turned off the pytest action since it is now covered by this. I'll keep it around as a manual GitHub action in case we want to run it (e.g. to test any pytest upgrades quickly).

### DIFF
--- a/.github/workflows/generate_chapters.yml
+++ b/.github/workflows/generate_chapters.yml
@@ -37,7 +37,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    if: (github.event_name == 'push' && github.repository == 'HTTPArchive/almanac.httparchive.org') || github.event.pull_request.head.repo.full_name == 'HTTPArchive/almanac.httparchive.org'
+    if: github.repository == 'HTTPArchive/almanac.httparchive.org'
     steps:
     - name: Checkout branch
       uses: actions/checkout@v2
--- a/.github/workflows/generate_chapters.yml
+++ b/.github/workflows/generate_chapters.yml
@@ -126,5 +126,3 @@ jobs:
 
           [1]: https://github.com/peter-evans/create-pull-request
         labels: generate chapters
-    - name: Check outputs
-      run: echo "Pull Request Number - ${{ steps.cpr.outputs.pull-request-number }}"
--- a/.github/workflows/generate_chapters.yml
+++ b/.github/workflows/generate_chapters.yml
@@ -113,7 +113,6 @@ jobs:
         npm run test
     - name: Create Pull Request
       if: github.event_name != 'pull_request'
-      id: cpr
       uses: peter-evans/create-pull-request@v3
       with:
         title: Update datestamps and generate chapters
--- a/.github/workflows/generate_chapters.yml
+++ b/.github/workflows/generate_chapters.yml
@@ -2,22 +2,30 @@
 ## Custom Web Almanac GitHub action ##
 ######################################
 #
-# Updates timestamps, increments CSS, and then runs "npm run generate"
-# on any pushes to main for markdown, templates or generate scripts
+# This generates the chapters and tests the website
 #
-# Then it tests the website
+# For pushed to main it also updates timestamps and increments CSS.
+# Then we open a Pull Request for any changes generated. We do this
+# only on a push, rather than on pull request, and only on main to
+# avoid avoid merge conflicts acorss pull requests (and also because
+# that would require personal access tokens to be set up for pull
+# requests from forks).
 #
-# Then we open a Pull Request for any changes generated.
-#
-# We do it on a push, rather than on pull request, and only on main to
-# avoid needing GitHub Personal Access Tokens and to avoid merge conflicts
-#
-name: Generate Chapters on Push
+name: Generate Chapters
 on:
   workflow_dispatch:
   push:
     branches:
       - main
+    paths:
+      - src/**.js
+      - src/**.css
+      - src/**.html
+      - src/content/**.md
+      - src/**.py
+      - src/**.json
+      - src/requirements.txt
+  pull_request:
     paths:
       - src/**.js
       - src/**.css
@@ -104,6 +112,7 @@ jobs:
         cd src
         npm run test
     - name: Create Pull Request
+      if: github.event_name != 'pull_request'
       id: cpr
       uses: peter-evans/create-pull-request@v3
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/generate_chapters.yml
```diff
@@ -104,6 +112,7 @@ jobs:
         cd src
         npm run test
     - name: Create Pull Request
+      if: github.event_name != 'pull_request'
```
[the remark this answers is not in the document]
bazzadp: I've never checked this so have just removed it - reduces complexity a tiny bit 😀

.github/workflows/generate_chapters.yml
```diff
@@ -104,6 +112,7 @@ jobs:
...
       id: cpr
```
ibnesayeed, line 116: I think this is not *needed* anymore. ;-)

```suggestion
```

bazzadp:
> I think things are becoming more monolithic gradually, for example, bringing the testing step into chapter generation process minimizes the chance of parallel execution and the whole workflow becomes longer.

I don't think we can do parallel execution for the most part - we need to install the requirements, then we need to generate the chapters, and then we need to run the site, and only if that all works do we want to commit the changes done by the generation.

Possibly the pytest could be run at same time as chapter generation but it's tiny and super quick. Other than that, I don't see much opportunity for parallelisation.

> We can leverage `workflow_run` event, [ newly introduced capability](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/), DRY some tasks.

Would this work when we want the results of the test to be presented in the pull request conversation tab? Or would only the workflow kicked off directly from that show?

.github/workflows/generate_chapters.yml
```diff
@@ -2,17 +2,16 @@
...
-# Then we open a Pull Request for any changes generated.
-#
-# We do it on a push, rather than on pull request, and only on main to
-# avoid needing GitHub Personal Access Tokens and to avoid merge conflicts
-#
-name: Generate Chapters on Push
+name: Generate Chapters
 on:
   workflow_dispatch:
   push:
```
ibnesayeed, line 17: I will not limit it to only the `main` branch as contributors might want to work on feature branches in their forks and want to be confident that their change passes testes before creating a PR.

But then the job is limited to this repository only, so I am not too sure exactly how and when when want it to run.

bazzadp:
Btw the linting is in a separate action, as that **is** something that can be run in parallel (well except any generated HTML is not linted until it's committed to `main` but think that's a small risk).

.github/workflows/generate_chapters.yml
bazzadp: I thought you'd say that - so tried to cover that in the initial comment:

> I'd like to keep this check off from generic pushes, as it does take a few mins to run. Think it's find to only run on pull request when code should be ready to commit.

IMHO running this on EVERY push is a lot of processing potentially on unfinished code that may fail. I think sufficient to do this on opening a pull request when code should be "production ready".

It also means we can avoid that annoying doubling up of the Actions on pull requests like we get on the linting action:

Double Actions

It is quite common on a lot of repos I work on to only run tests on pull requests only.

If people want to run this test locally then they can easily do this locally, before they even commit, by running the following:
- `pytest`
- `npm run generate`
- `npm run tests`
IMHO it's better to do that for code changes and the GitHub action should be a safety net to protect the code, to ensure tests pass before merging, rather than something developers would be depending on in their workflow.

ibnesayeed:
> Would this work when we want the results of the test to be presented in the pull request conversation tab? Or would only the workflow kicked off directly from that show?

I think that will require using artifacts for state management. I have not experimented with it as it is a new feature and I have been really busy in other priorities lately. Here is an example they gave in the blog post that sheds some light:

> As an example, as a maintainer you could set up a workflow that takes the artifacts generated by the pull request workflow, do some analysis, and post comments back to the pull request. -- https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

.github/workflows/generate_chapters.yml
bazzadp: > But then the job is limited to this repository only, so I am not too sure exactly how and when when want it to run.

Oh that's a good point forgot about that. It was restricted to this repo to sop opening pull requests for those working on `main` branch in own repo.

I presume that means it won't run on pull requests so change this to this:

```yml
    if: (github.event_name == 'push' && github.repository == 'HTTPArchive/almanac.httparchive.org') || github.event.pull_request.head.repo.full_name == 'HTTPArchive/almanac.httparchive.org'
```

Any way to split that long line into two?

bazzadp:
> > Would this work when we want the results of the test to be presented in the pull request conversation tab? Or would only the workflow kicked off directly from that show?
> 
> I think that will require using artifacts for state management. I have not experimented with it as it is a new feature and I have been really busy in other priorities lately. Here is an example they gave in the blog post that sheds some light:
> 
> > As an example, as a maintainer you could set up a workflow that takes the artifacts generated by the pull request workflow, do some analysis, and post comments back to the pull request. -- https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

That seems designed to solve the problem of updating the pull request meta-data from a fork, not necessarily of chaining test workflows for a pull request. Even if it could be used for that, it seems the added complexity would more than offset any simplification in modularising this to a separate workflow.

But anyway, as you say, something for future - let's merge this now as think it's good to go!

.github/workflows/generate_chapters.yml
ibnesayeed: > Any way to split that long line into two?

YAML allows multi-line strings, but I have never tried when evaluation conditionals in GH Actions context, so I am not sure how robust their parser is.

ibnesayeed:
> But anyway, as you say, something for future - let's merge this now as think it's good to go!

:+1:

.github/workflows/generate_chapters.yml
bazzadp: Actually I've reverted this as according to [this](https://github.community/t/preventing-a-step-from-running-on-a-pr-originating-from-a-fork/18068/2?u=bazzadp) github.repository is set to the destination repo on pull requests so should be fine as it was.

PR #1167, opened 2020-08-05
Add more test cases

bazzadp:
Add more test cases to test script added in #1124 for fuller coverage.

Goes from 92 test cases to 116 test cases by adding the following:

- ebook pages
- robots.txt
- favicon.ico
- no year redirects (e.g. `/en/` -> `/en/2019/`) for each supported language
- unsupported language (`zz`) redirect
- 404s for each supported language and year

Also moves the reading of config files to `shared.js` as now used in two places.

### DIFF
--- a/src/tools/generate/shared.js
+++ b/src/tools/generate/shared.js
@@ -33,6 +33,21 @@ const find_config_files = async () => {
   return await recursive('config', [filter]);
 };
 
+const get_config_files = async () => {
+
+  let configs = {};
+
+  for (const config_file of await find_config_files()) {
+    const re = (process.platform != 'win32') 
+                  ? /config\/([0-9]*).json/ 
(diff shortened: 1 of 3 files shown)


src/tools/generate/shared.js
```diff
@@ -33,6 +33,21 @@ const find_config_files = async () => {
   return await recursive('config', [filter]);
 };
 
+const get_config_files = async () => {
```
ibnesayeed, line 36: I think we should change the function name to something like `get_yearly_configs` because function names `find_*_files` represent files (handle or name), not the content of the files. This one however, returns parsed object loaded from those files. Also, using the term `yearly` in the name will suggest that the object keys are years.

bazzadp: Good idea! Done.

Issue #896, opened 2020-06-23
[Discuss] Interactive data visualizations


rviscomi:
Last year @bkardell built a [Glitch app](https://rainy-periwinkle.glitch.me/) to help slice the Markup chapter data and it was really useful to deep link for specific subsets of the data. And from the data exploration side I thought it was a great way for users to find their own interesting insights. I'd like to take that idea a step further and discuss how we might be able to embed interactive experiences in the chapters themselves.

Our embedded Google Sheets visualizations have nice hover effects but you can't really _explore_ the data at all. I would love to have "immersive" experiences with the data in which readers could dive into a specific area of interest. Contrived example, but in the JS chapter we could have an interactive figure that lets the reader select one or more frameworks and compare their adoption rates visually.

I see this as a bespoke build in close collaboration between the content (authors/reviewers), analysis, and development teams for only the most popular chapters. Last year we had thought about building some kind of pipeline to auto-generate data viz based on every JSON data file, but I don't think we need to get that sophisticated. (though that would be cool)

Should we get into D3.js or are there better drop-in solutions, like an embedded Glitch app or Data Studio dashboard? How can we ensure that whatever we build is accessible?

cc @HTTPArchive/developers @HTTPArchive/data-analysts

rviscomi:
According to the [Web Almanac analytics](https://datastudio.google.com/s/gUPCF0CEsCw), here are the latest page view counts for each 2019 chapter:

Chapter | Views
-- | --
JavaScript | 33,618
CSS | 11,436
SEO | 6,323
Performance | 5,334
Markup | 4,052
HTTP/2 | 3,996
Third Parties | 3,717
Fonts | 3,171
Accessibility | 3,108
PWA | 3,052
CMS | 3,032
Media | 2,445
Security | 2,419
Mobile Web | 2,250
CDN | 1,988
Ecommerce | 1,724
Page Weight | 1,656
Compression | 1,332
Caching | 1,324
Resource Hints | 961

So the top chapters were JavaScript, CSS, SEO, Performance, Markup, etc in that order. Now that each of these 2020 chapters have content teams of authors and reviewers, we can coordinate with them on a bespoke interactive data visualization that will be engaging to readers.

There are some platform questions we can iron out, like the mechanics of the markdown file for embedding the interactive visualization. Would the build system find/replace a placeholder, or would it be an embedded web component or similar markup that a script would "hydrate" onto the page? How do we support static media like print/ebook?

[...]

PR #1179, opened 2020-08-07
Remove run_queries.sh as no longer used

bazzadp:
We added a bash linter in #1177 however the run_queries.sh is throwing up some errors. Not sure if this is even used but  this fixes it in case it is used, or needed in the future.

@rviscomi / @paulcalvano can you test this?

When I try to run `./run_queries.sh 2019/19_Resource_Hints` I get the below produced in the JSON files:

```
BigQuery error in query operation: Access Denied: Project httparchive: User does
not have bigquery.jobs.create permission in project httparchive.
```

So not 100% sure it's actually working. Or can you give me access so I can test it?

Or if it's not used, then I could just change this PR to delete this file instead as it's the only one failing our linting at the moment when I lint the whole repo.


rviscomi:
Let's remove it. The intent behind the script was to generate JSON results for each SQL file to be used for data viz, but we ended up with (and continue to use) Sheets for that. JSON needed for one-off data viz like #896 can be created manually from BQ.

PR #1185, opened 2020-08-09
Better author social media on website

bazzadp:
As discussed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/1142#discussion_r466341440 we made better use of Author's social media for the eBook. I think we should do the same on the website.

Old:

Old Author Bio

New (depending on width):

New Author Bio Large

New Author Bio Extra-Large

New Author Bio Small

This pull request also mades two adds in LinkedIn links to a few missing places


src/static/css/page.css
```diff
@@ -277,14 +277,23 @@
...
+  margin: 0 10px 0 0;
+  margin: 0 0.625rem 0 0;
+  text-decoration: none;
+  white-space: nowrap;
+}
+
+.authors .social svg {
+  width: 1em;
+  height: auto;
+  vertical-align: -0.125rem;
```
rviscomi, line 294: This seems hacky. Why doesn't `middle` work?

src/templates/base/2019/base_chapter.html
```diff
@@ -86,7 +87,7 @@
 {% macro render_byline() %}
   <div class="byline">{{ self.written_by_before() }}
   {% for author in metadata.get('authors') %}
-    <a class="author" href="{{ url_for('contributors', year=year, lang=lang, _anchor=author) }}">{{ config.contributors[author].name if author in config.contributors else author }}</a>{{ self.comma() if not loop.last }}
+    <a class="author" href="#author-{{ author }}">{{ config.contributors[author].name if author in config.contributors else author }}</a>{{ self.comma() if not loop.last }}
```
rviscomi, line 90: I think this should still link to the Contributors page.

src/static/css/page.css
```diff
@@ -277,14 +277,23 @@
 
 .authors .social {
   display: inline-block;
```
rviscomi, line 279: Given the longer length of this section, I think it looks better to always be on its own line rather than to the right of the author name.

```suggestion
  display: block;
```

bazzadp: Rather than this, we should just change from span to div and remove this line completed. It’s only a span for reasons below...

src/static/css/page.css
bazzadp: Yeah this is only needed if putting it on the same line as Author name as middle seems
to break then. Suspect because of two different font sizes between name and author. As you’re happy to always have this on second line then can remove this hack.

src/templates/base/2019/base_chapter.html
bazzadp: Yeah but the contributor page doesn’t have the bio info so it’s actually more useful to the reader to link to bottom of the page.

But agree it is odd that it will like to different place than reviewer or contributor.

Alternatively, we could try to expose the author bios on the contributors page somehow on a separate PR. A modal or something? Also been thinking it would be good to somehow link contributors/reviewers/translators to chapters anyway. Either that or the long standing search functionality I’ve kept open to look at, at some point...

WDYT?

src/static/css/page.css
bazzadp: Removed.

src/static/css/page.css
bazzadp: Resolved by changing to div.

src/templates/base/2019/base_chapter.html
bazzadp: Have revert this so we can merge. Will raise a separate issue to discuss.

src/static/css/page.css
```diff
@@ -276,15 +276,28 @@
...
+  font-size: 0.9em;
 }
 
 .authors .social a {
-  display: inline-block;
-  vertical-align: middle;
-  margin: 0 5px;
-  margin: 0 0.3125rem;
+  margin: 0 10px 0 0;
+  margin: 0 0.625rem 0 0;
```
rviscomi, line 284: Could this be simplified to `margin-right` or are the 0s significant?
```suggestion
  margin-right: 10px;
  margin-right: 0.625rem;
```

rviscomi: I also wonder if it's worth doing this to maximize space before wrapping.

```css
.authors .social a:last-child {
  margin-right: 0;
}
```

src/templates/base/2019/base_chapter.html
```diff
@@ -121,49 +122,53 @@ <h2 id="authors">
     {% endif %}
       {% set authordata = config.contributors[author] if author in config.contributors else None %}
       {% if authordata %}
-      <li>
+      <li id="author-{{ author }}">
```
rviscomi, line 125: Is this still needed?

src/static/css/page.css
bazzadp: Added both suggestions!

src/templates/base/2019/base_chapter.html
bazzadp: Removed

PR #1188, opened 2020-08-10
Refactor Featured Chapter Code to avoid repetition

bazzadp:
There is a lot of repeated code for the featured chapters. This consolidates it into the `base/2019/index.html` file and leaves the `featured_chapter.html` with just the quotes specific to that language.

This will make it easier to auto generate them.


bazzadp:
Merging as just template updates so low risk

PR #1191, opened 2020-08-10
Narrow home page

### LINKED ISSUE #574: Optimize website for wide viewports
As noted [in a recent PR](https://github.com/HTTPArchive/almanac.httparchive.org/pull/573#discussion_r352280309) there are a couple of inconsistencies with the home page, design, notably:

- [ ] - It's full width, but the other pages are constrained (to 1280px or 1324 - note bringing these inline with above PR)
- [x] - The Start Exploring button in the top right, means the other navigation items (TOC, Contributors, and Methodology) are in a different place on the home page than every other page and "jump" as you navigate to/from this - Completed in #684 

Are these items we want to fix? Or happy this should be different? The [original design](https://github.com/HTTPArchive/almanac.httparchive.org/issues/46#issuecomment-515966879) seemed to have a more constrained width (except for the header) though did always have the Start Exploring button in the same place as currently.

Thoughts?

catalinred:
This fixes #574

- Please note that all the `.html` changes are due to wrapping the content into `.container`.
  This is related to removing the `.container` wrapper from `base.html` as the index is based on a CSS grid structure.
- Enhanced the contributors block on wide displays as well.
- Organized a bit the MQ's from index.css and removed some redundant lines


src/static/css/index.css
```diff
@@ -77,14 +55,14 @@ p {
...
-  padding: 0 60px;
 }
 
 .intro {
   -ms-grid-column: 1;
   grid-area: intro;
   width: 100%;
-  max-width: 420px;
+  max-width: 440px;
+  padding-left: 20px;
```
bazzadp, line 65: Not sure why we need this, but need to remove it on mobile as causing everything to be off-center:

Mobile view with padding

src/templates/base/2019/base.html
```diff
@@ -177,9 +177,7 @@
   </header>
   {% endblock %}
 
-  <div class="container">
-    {% block main %}{% endblock %}
-  </div>
+  {% block main %}{% endblock %}
```
bazzadp, line 180: I'll be honest, I'm not loving the fact that we now need to specify the container on every page, just because the home page is slightly different - it's like the exception setting the rule, rather than the other way around.

Do we need to do this? How come the current home page is able to break out of this?

Alternatively, if we really do need to do this, then can't we wrap this in a Jinja2 block so it can be overridden by the two index.html pages, but left alone like it was on every other page?:

```suggestion
  {% block container %}
  <div class="container">
    {% block main %}{% endblock %}
  </div>
  {% endblock %}
```

src/static/css/index.css
catalinred: Good catch ;) I've fixed that.

The left (padding) 20px  magic number is needed to match the .container side padding for the 900 - 1280px screen res scenario. For the right side padding, considering the .intro-image-wrapper image design, I've skipped it.

sides-padding

bazzadp: Thanks. Agree it’s not really needed but is might be less confusing to put the right side in too. Might also stop that red guys butt going right to the edge. WDYT?

src/templates/base/2019/base.html
catalinred: The home is different, it uses a lot of CSS grid features and it does not play well with the `.container` helper. 
Therefore, we do need this rule exception.

I'm not a big fan of these `.html` changes too, mostly because the git history shows me as the author of so many (indented) lines and that's not true actually.

bazzadp: Well if you go with this new `{% block container %}` suggestion your just need three HTML changes:
1. To `base/2019/base.html` to add this
2. To `base/2019/index.html` to override this to remove the `container` div for this template only.
3. To `base/2020/index.html` to override this to remove the `container` div for this template only.

The rest of your html changes to the other files can be reverted as those other files still will have the container div by default as set in `base.html`.

src/static/css/index.css
catalinred: IMO adding a right 20px `padding` would make things a bit worse here, as there's already too much extra space to the right, due to the current year image's design. 

900px - 1280px is an edge case after all, but if it were to be 100% correct, I think we'd need to add a margin-right: -40px or so to the `.intro-image-wrapper` to make things balanced here. The result here would be the red guy butt a bit cut, just as the below contributors number - which might look good I think :)

catalinred:
Not strictly related to this PR, but I've noticed two things:

- the `/table-of-contents` and `/contributors` are adding extra 10px sides padding, I think this is extra, unless this is not intended perhaps for the ebook pages?

- On the `/2019/` page, for the sake of readability and UX, would love to see the intro phrase split into multiple paragraphs, as in `/2020/` perhaps, what do you think? Same for the Featured Chapter text as well.

bazzadp:
>- the `/table-of-contents` and `/contributors` are adding extra 10px sides padding, I think this is extra, unless this is not intended perhaps for the ebook pages?

The ebook reimplements the contents for those two pieces in `base/2019/base_ebook.html` so doesn’t use those two templates at all. So guessing extra?

> - On the `/2019/` page, for the sake of readability and UX, would love to see the intro phrase split into multiple paragraphs, as in `/2020/` perhaps, what do you think? Same for the Featured Chapter text as well.`

Yeah I could go with that. Though then need to look at all languages. Maybe better as a separate follow up PR after we merge this one? And btw still disagree with @rviscom’s decision to put `2020` temporary text in italics 😀

Leave featured chapter alone btw as lots going on for that in other branches/PRs.

src/templates/base/2019/index.html
```diff
@@ -24,70 +24,74 @@
...
-  </section>
-  <section id="methodology" class="methodology-container">
-    <div class="methodology">
-      <h2 class="title title-center">{{ self.methodology_title() }}</h2>
-      <div class="methodology-data">
-        <div class="websites-tested">
-          <div>
-            {{ self.methodology_stat_1_title() }}
+{% block container %}
+  {% block main %}
```
bazzadp, line 28: Do we need the `{% block main %}` declaration? There’s nothing special about it AFAIK so I think overriding the `{% block container %}` should be enough. Should save you some indents and make this PR look a lot smaller.

catalinred: Done.

catalinred:
>  So guessing extra?

I did remove the extra CSS lines.

> better as a separate follow up PR

Sure thing.

rviscomi:
Go ahead and merge. Thanks @catalinred!

PR #1091, opened 2020-07-24
Performance 2020 queries

max-ostapenko:
Progress on #905 

### distribution of LH performance category scores
- [x] Distribution of Performance Score (slow, moderate, fast) on mobile in LH5
- [x] Changes to Performance Score from LH5 to LH6
- [x] Distribution of Performance Score (slow, moderate, fast) on mobile in LH6
- [x] Average/median changes in performance score between versions
- [x] LH5 ([?] 2020) vs LH6 (August 2020)
### LH audits
- [x] all scores and weightings per performance category
### LCP
- [x] % of CrUX origins that meet the "good" threshold (75%+ < 2.5s)
- [x] % of CrUX origins that meet the "NI" threshold (not good or poor)
- [x] % of CrUX origins that meet the "poor" threshold (25%+ >= 4.0s)
- [x] distribution by device
- [x] Segment by device, country and ECT
### FID
- [x] % of CrUX origins that meet the good/NI/poor thresholds
- [x] distribution by device
- [x] Segment by device, country  and ECT
### CLS
- [x] % of CrUX origins that meet the good/NI/poor thresholds
- [x] distribution by device
- [x] Segment by device, country and ECT
### FCP
- [x] % of CrUX origins that meet the fast/moderate/slow thresholds
- [x] Segment by device, country and ECT
- [x] distribution by device
- [x] Compare against 2019
### TTFB
- [x] % of CrUX origins that meet the fast/moderate/slow thresholds
- [x] Segment by device, country and ECT
- [x] distribution by device
- [x] Compare against 2019
### Field Data
- [x] % of website with `new PerformanceObserver` in JS ([source](https://www.chromestatus.com/metrics/feature/timeline/popularity/2188))

Issue #910, opened 2020-06-27
Mobile Web 2020


obto:
# Part II Chapter 12: Mobile Web

## Content team
| Authors | Reviewers | Analysts | Draft | Queries | Results |
| ------- | --------- | -------- | ----- | ------- | ------- |
| @spanicker @mdiblasio | @malchata @obto @cheneytsai  | @obto | [Doc][~draft-doc] | [*.sql][~sql-dir] | [Sheet][~results-sheet] |

**Content team lead:** @spanicker

Welcome chapter contributors! You'll be using this issue throughout the [chapter lifecycle](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle) to coordinate on the content planning, analysis, and writing stages.

The **content team** is made up of the following contributors:

- [Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)
- [Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)
- [Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)

**New contributors:** If you're interested in joining the content team for this chapter, just leave a comment below and the content team lead will loop you in.

_Note: To ensure that you get notifications when tagged, you must be "[watching](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#notifications)" this repository._

### [**0. Form the content team**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#0-create-content-team)
+ [x] **Jul 6th**: Project owners have selected an author to be the content team lead
+ [x] **Jul 13th**: The content team has at least one author, reviewer, and analyst (minimally viable team formed)

### [**1. Plan content**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
+ [x] **Jul 20th:** The content team has completed the chapter outline in the [draft doc][~draft-doc]
+ [x] **Jul 27th**: Analysts have triaged the feasibility of all proposed metrics

### [**2. Gather data**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
[...]

PR #1201, opened 2020-08-12
Mobile web queries

obto:
This PR contains all the queries for #910. There is quite a large amount of data being accumulated from other sources, so don't be alarmed by the small amount of queries.

- [x] Distribution of 2G, 3G and 4G usage
- [x] % of sites who register a service worker
- [x] Core web vitals:
  - These have been done extensively by the performance team. We can use their queries: https://github.com/HTTPArchive/almanac.httparchive.org/pull/1091
- [x] Most commonly used apps/software/libraries being used on mobile sites. Via wappalyzer

### DIFF
--- a/sql/2020/12_Mobile_Web/12_01.sql
+++ b/sql/2020/12_Mobile_Web/12_01.sql
@@ -1,0 +1,22 @@
+#standardSQL
+# 12_01: 4G, 3G and 2G distribution
+SELECT
+  _4G,
+  _3G,
+  _2G
+FROM (
+  SELECT
+    COUNT(0) OVER () AS n,
+    ROUND(_4GDensity * 100, 2) AS _4G,
+    ROUND(_3GDensity * 100, 2) AS _3G,
+    ROUND(_2GDensity * 100, 2) AS _2G,
+    ROW_NUMBER() OVER (ORDER BY _4GDensity DESC) AS row
+  FROM
+    `chrome-ux-report.materialized.metrics_summary`
+  WHERE
+    yyyymm = 202007 AND
+    _4GDensity + _3GDensity + _2GDensity > 0
+  ORDER BY
+    _4G DESC)
+WHERE
+  MOD(row, CAST(FLOOR(n / 1000) AS INT64)) = 0
--- a/sql/2020/12_Mobile_Web/12_01.sql
+++ b/sql/2020/12_Mobile_Web/12_01.sql
@@ -1,22 +1,16 @@
 #standardSQL
 # 12_01: 4G, 3G and 2G distribution
 SELECT
-  _4G,
-  _3G,
-  _2G
-FROM (
-  SELECT
-    COUNT(0) OVER () AS n,
-    ROUND(_4GDensity * 100, 2) AS _4G,
-    ROUND(_3GDensity * 100, 2) AS _3G,
-    ROUND(_2GDensity * 100, 2) AS _2G,
-    ROW_NUMBER() OVER (ORDER BY _4GDensity DESC) AS row
-  FROM
-    `chrome-ux-report.materialized.metrics_summary`
-  WHERE
-    yyyymm = 202007 AND
-    _4GDensity + _3GDensity + _2GDensity > 0
-  ORDER BY
-    _4G DESC)
+  percentile,
+  APPROX_QUANTILES(_4GDensity, 1000)[SAFE_ORDINAL(percentile * 10)] AS _4GDensity,
+  APPROX_QUANTILES(_3GDensity, 1000)[SAFE_ORDINAL(percentile * 10)] AS _3GDensity,
+  APPROX_QUANTILES(_2GDensity, 1000)[SAFE_ORDINAL(percentile * 10)] AS _2GDensity
+FROM
+  `chrome-ux-report.materialized.metrics_summary`,
+  UNNEST([10, 25, 50, 75, 90]) AS percentile
 WHERE
-  MOD(row, CAST(FLOOR(n / 1000) AS INT64)) = 0
+  yyyymm = 202007
+GROUP BY
+  percentile
+ORDER BY
+  percentile
--- a/sql/2020/12_Mobile_Web/12_03.sql
+++ b/sql/2020/12_Mobile_Web/12_03.sql
@@ -1,0 +1,29 @@
+#standardSQL
+# 12_03: Most used apps
+SELECT
+  _TABLE_SUFFIX AS client,
+  total_sites,
+
+  category,
+  app,
+  COUNT(0) AS sites_with_app,
+  ROUND(COUNT(0) * 100 / total_sites, 2) AS pct_sites_with_app
+FROM
+  `httparchive.almanac.technologies_desktop_*`
+JOIN (
+  SELECT
+    _TABLE_SUFFIX,
+    COUNT(0) AS total_sites
+  FROM
+    `httparchive.almanac.summary_pages_desktop_*`
+  GROUP BY
+    _TABLE_SUFFIX
+)
+USING (_TABLE_SUFFIX)
+GROUP BY
+  client,
+  total_sites,
+  category,
+  app
+ORDER BY
+  pct_sites_with_app DESC
--- a/sql/2020/12_Mobile_Web/12_03.sql
+++ b/sql/2020/12_Mobile_Web/12_03.sql
@@ -1,7 +1,6 @@
 #standardSQL
 # 12_03: Most used apps
 SELECT
-  _TABLE_SUFFIX AS client,
   total_sites,
 
   category,
@@ -9,19 +8,14 @@ SELECT
   COUNT(0) AS sites_with_app,
   ROUND(COUNT(0) * 100 / total_sites, 2) AS pct_sites_with_app
 FROM
-  `httparchive.almanac.technologies_desktop_*`
-JOIN (
+  `httparchive.technologies.2020_08_01_mobile`
+CROSS JOIN (
   SELECT
-    _TABLE_SUFFIX,
     COUNT(0) AS total_sites
   FROM
-    `httparchive.almanac.summary_pages_desktop_*`
-  GROUP BY
-    _TABLE_SUFFIX
+    `httparchive.summary_pages.2020_08_01_mobile`
 )
-USING (_TABLE_SUFFIX)
 GROUP BY
-  client,
   total_sites,
   category,
   app
--- a/sql/2020/12_Mobile_Web/ect_distribution.sql
+++ b/sql/2020/12_Mobile_Web/ect_distribution.sql
@@ -1,16 +1,21 @@
 #standardSQL
 # ECT distribution
 SELECT
+  device,
   percentile,
-  APPROX_QUANTILES(_4GDensity, 1000)[SAFE_ORDINAL(percentile * 10)] AS _4GDensity,
-  APPROX_QUANTILES(_3GDensity, 1000)[SAFE_ORDINAL(percentile * 10)] AS _3GDensity,
-  APPROX_QUANTILES(_2GDensity, 1000)[SAFE_ORDINAL(percentile * 10)] AS _2GDensity
+  APPROX_QUANTILES(_4GDensity / (_4GDensity + _3GDensity + _2GDensity + slow2GDensity + offlineDensity), 1000)[OFFSET(percentile * 10)] AS _4GDensity,
+  APPROX_QUANTILES(_3GDensity / (_4GDensity + _3GDensity + _2GDensity + slow2GDensity + offlineDensity), 1000)[OFFSET(percentile * 10)] AS _3GDensity,
+  APPROX_QUANTILES(_2GDensity / (_4GDensity + _3GDensity + _2GDensity + slow2GDensity + offlineDensity), 1000)[OFFSET(percentile * 10)] AS _2GDensity,
+  APPROX_QUANTILES(slow2GDensity / (_4GDensity + _3GDensity + _2GDensity + slow2GDensity + offlineDensity), 1000)[OFFSET(percentile * 10)] AS slow2GDensity,
+  APPROX_QUANTILES(offlineDensity / (_4GDensity + _3GDensity + _2GDensity + slow2GDensity + offlineDensity), 1000)[OFFSET(percentile * 10)] AS offlineDensity
 FROM
-  `chrome-ux-report.materialized.metrics_summary`,
-  UNNEST([10, 25, 50, 75, 90]) AS percentile
+  `chrome-ux-report.materialized.device_summary`,
+  UNNEST(GENERATE_ARRAY(1, 100)) AS percentile
 WHERE
-  yyyymm = 202007
+  date = '2020-08-01'
 GROUP BY
+  device,
   percentile
 ORDER BY
+  device,
   percentile
--- a/sql/2020/12_Mobile_Web/ect_distribution.sql
+++ b/sql/2020/12_Mobile_Web/ect_distribution.sql
@@ -1,5 +1,5 @@
 #standardSQL
-# 12_01: 4G, 3G and 2G distribution
+# ECT distribution
 SELECT
   percentile,
   APPROX_QUANTILES(_4GDensity, 1000)[SAFE_ORDINAL(percentile * 10)] AS _4GDensity,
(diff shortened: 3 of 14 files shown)


obto:
@rviscomi Did we come to a decision of what data source we should use to find the origin rank of various URLs?

I know we looked at a few, but were also considering dropping origin ranking completely.

rviscomi:
TBH I haven't found a publicly available source that I like and has good coverage of our corpus. I'll defer to @spanicker on which is best for this chapter.

sql/2020/12_Mobile_Web/12_03.sql
```diff
@@ -0,0 +1,29 @@
...
+SELECT
+  _TABLE_SUFFIX AS client,
+  total_sites,
+
+  category,
+  app,
+  COUNT(0) AS sites_with_app,
+  ROUND(COUNT(0) * 100 / total_sites, 2) AS pct_sites_with_app
+FROM
+  `httparchive.almanac.technologies_desktop_*`
```
rviscomi, line 12:
```suggestion
  `httparchive.technologies.2020_08_01_mobile`
```

sql/2020/12_Mobile_Web/12_03.sql
```diff
@@ -0,0 +1,29 @@
...
+JOIN (
+  SELECT
+    _TABLE_SUFFIX,
+    COUNT(0) AS total_sites
+  FROM
+    `httparchive.almanac.summary_pages_desktop_*`
```
rviscomi, line 18:
```suggestion
    `httparchive.summary_pages.2020_08_01_mobile`
```

sql/2020/12_Mobile_Web/12_01.sql
```diff
@@ -0,0 +1,22 @@
...
+  _2G
+FROM (
+  SELECT
+    COUNT(0) OVER () AS n,
+    ROUND(_4GDensity * 100, 2) AS _4G,
+    ROUND(_3GDensity * 100, 2) AS _3G,
+    ROUND(_2GDensity * 100, 2) AS _2G,
+    ROW_NUMBER() OVER (ORDER BY _4GDensity DESC) AS row
+  FROM
+    `chrome-ux-report.materialized.metrics_summary`
```
rviscomi, line 15: Should we use `device_summary` instead, which will allow you to segment by phone and tablet devices? I assume you're not interested in desktop data here.

sql/2020/12_Mobile_Web/12_03.sql
```diff
@@ -0,0 +1,29 @@
+#standardSQL
+# 12_03: Most used apps
+SELECT
+  _TABLE_SUFFIX AS client,
```
rviscomi, line 4: I assume you're not interested in desktop data here, so I've suggested querying the mobile tables directly.

rviscomi (changes requested):
Are there only 3 queries for this chapter? 😕

rviscomi:
Is this ready for review?

sql/2020/12_Mobile_Web/12_01.sql
obto: We just want to know how often a site is accessed via 2/3/4G, regardless of the device. Would `device_summary` be best to use for this?

sql/2020/12_Mobile_Web/12_03.sql
obto: Good call

sql/2020/12_Mobile_Web/12_01.sql
rviscomi: In that case `metrics_summary` SGTM.

I'm not sure about the aggregation method though, this seems like you're sampling the distribution of connection densities, but not actually calculating any summary stats. How about calculating the 10,25,50,75,90th percentiles of origins' ECT?

obto:
@rviscomi We'll be using Core web vitals as well in this chapter. Do you want me to copy/paste the queries the Perf chapter already has done here?

Thinking itd be best to re-use the data in their spreadsheet than to query it twice.

rviscomi:
Sure if the analysis is identical it's ok to reuse existing queries/results. Is there a more specific focus on mobile experiences in this chapter?

obto:
Good q. I'll ping @spanicker in our doc

sql/2020/12_Mobile_Web/12_01.sql
obto: ECT?

bazzadp: https://developer.mozilla.org/en-US/docs/Glossary/Effective_connection_type

obto: The main q we want to answer here is: Given a random website, what does the distribution of 2G, 3G and 4G visitors look like?

And then, what does this distribution look like for the top 10k sites? Is 4G used more often or less often.


Thinking about it more now, this question might be a bit flawed though...

obto:
have not heard back, so going to use the existing queries for now

obto:
> How about calculating the 10,25,50,75,90th percentiles of origins' ECT?

Took a stab at changing 12_01 to reflect this, but think I'm missing something... because these numbers don't look right. @rviscomi Am i misunderstanding what 4GDensity means in CrUX?

percentile | _4GDensity | _3GDensity | _2GDensity
-- | -- | -- | --
10 | 0.8659 | 0 | 0
25 | 0.9476 | 0 | 0
50 | 0.9998 | 0 | 0
75 | 1 | 0.0518 | 0
90 | 1 | 0.1311 | 0

rviscomi:
ECT of 4G is a relatively [low bar](https://wicg.github.io/netinfo/#effective-connection-types) (700 Kbps, 270ms), so unfortunately it doesn't give us as much granular information in CrUX as we'd hope. Your results are correct that the vast majority of experiences are effectively 4G speeds.

obto:
@rviscomi Why do the numbers not equate to 1 though? The 90th percentile is over 100% accounted for

rviscomi:
Each are independent distributions, ie the website having the 50th percentile 4G value is not necessarily the same website having the 50th percentile 3G value. Or to put it differently, 50% of websites have a 4G density < 99.98%, 50% of websites have a 3G density = 0.00%, etc.

obto:
@rviscomi This is the last metric I have on my list to add for Mobile web.

1. Since I haven't heard back from Shubhie on if she'd like different things analyzed than the Perf chapter already did, I'd like to give her (and the other reviewers) view only access to that document. I can make a copy of it and add custom analysis should she decide to.

2. The other metrics we're sourcing from external sources are still under review, and I've pinged Shubhie for an update. I'll keep you posted.

With that being said. Once everything looks good here, lets merge and I'll open a new PR should there be additional metrics.

sql/2020/12_Mobile_Web/service_worker.sql
```diff
@@ -0,0 +1,8 @@
+#standardSQL
+# Sites registering a service worker
+SELECT
+  COUNT(0) AS total,
+  COUNTIF(CAST(JSON_EXTRACT_SCALAR(report, '$.audits.service-worker.score') as NUMERIC) = 1) AS score_sum,
+  ROUND(COUNTIF(CAST(JSON_EXTRACT_SCALAR(report, '$.audits.service-worker.score') as NUMERIC) = 1) * 100 / COUNT(0), 2) as score_percentage
```
rviscomi, line 6:
```suggestion
  COUNTIF(CAST(JSON_EXTRACT_SCALAR(report, '$.audits.service-worker.score') AS NUMERIC) = 1) / COUNT(0) AS score_percentage
```

sql/2020/12_Mobile_Web/service_worker.sql
```diff
@@ -0,0 +1,8 @@
+#standardSQL
+# Sites registering a service worker
+SELECT
+  COUNT(0) AS total,
+  COUNTIF(CAST(JSON_EXTRACT_SCALAR(report, '$.audits.service-worker.score') as NUMERIC) = 1) AS score_sum,
```
rviscomi, line 5:
```suggestion
  COUNTIF(CAST(JSON_EXTRACT_SCALAR(report, '$.audits.service-worker.score') AS NUMERIC) = 1) AS score_sum,
```

rviscomi (changes requested):

obto:
We got some new queries to add to the chapter. I'll make the edits requested once all the new queries are added.

rviscomi:
@obto are you still blocked or is this ready for another review?

sql/2020/12_Mobile_Web/mobile_traffic_distribution.sql
```diff
@@ -0,0 +1,14 @@
+#standardSQL
+# Distribution of traffic coming from mobile devices
+SELECT
```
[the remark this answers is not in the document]
obto: If we group by device shouldn't the pct_traffic_from_mobile always be 100%? I ran the query and it's not however, so it feels like I'm missing something.

sql/2020/12_Mobile_Web/ect_distribution.sql
```diff
@@ -0,0 +1,16 @@
+#standardSQL
+# ECT distribution
+SELECT
+  percentile,
+  APPROX_QUANTILES(_4GDensity, 1000)[SAFE_ORDINAL(percentile * 10)] AS _4GDensity,
+  APPROX_QUANTILES(_3GDensity, 1000)[SAFE_ORDINAL(percentile * 10)] AS _3GDensity,
+  APPROX_QUANTILES(_2GDensity, 1000)[SAFE_ORDINAL(percentile * 10)] AS _2GDensity
+FROM
+  `chrome-ux-report.materialized.metrics_summary`,
```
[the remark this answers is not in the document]
obto: I changed the query, but want to make sure I interpret this right. How would you read the results at the 62nd percentile for phones?: https://docs.google.com/spreadsheets/d/1DGLY7UEWOlDL5_2dtS_j2eqMjiV-Rw5Fe2y6K6-ULvM/edit#gid=1072207571

obto:
All the queries have been updated, just have a few questions i've put in the review comments

sql/2020/12_Mobile_Web/ect_distribution.sql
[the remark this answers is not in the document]
rviscomi: If the question is "why does the sum of the values exceed 100%" remember that each distribution is independent, so there's no expectation of totality. If that's what you're looking for, use `AVG`.

It's easier to answer your question and make my point by looking at the 100th percentile: "The origin with the largest 4G phone density has a density of 100%. The origin with the largest 3G phone density has a density of 100%. The origin with the largest 2G phone density has a density of 10.75%. The origin with the largest slow 2G phone density has a density of 19.26%. The origin with the largest offline phone density has a density of 100%."

sql/2020/12_Mobile_Web/mobile_traffic_distribution.sql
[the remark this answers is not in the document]
rviscomi: Yeah you're right that the `device_summary` table is already normalized and grouped by device, so there's no need to do that here. The only thing to do is filter the results for `phone` experiences:

```sql
SELECT
  percentile,
  APPROX_QUANTILES(phoneDensity, 1000)[OFFSET(percentile * 10)] AS pct_traffic_from_mobile,
FROM
  `chrome-ux-report.materialized.device_summary`,
  UNNEST(GENERATE_ARRAY(1, 100)) AS percentile
WHERE
  date = '2020-08-01' AND
  device = 'phone'
GROUP BY
  percentile
ORDER BY
  percentile
```

sql/2020/12_Mobile_Web/ect_distribution.sql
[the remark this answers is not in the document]
obto: Thanks. Think i'll copy paste your explanation directly into a comment in the google sheet

obto:
@rviscomi Existing feedback has been resolved, and the new queries are ready for review. Thanks for the patience :)

Could you give a quick look at the comments left in the data? Want to make sure the authors especially understand the CrUX stats. For example in the Mobile traffic distribution sheet. How would you read the 50th percentile?

Data here: https://docs.google.com/spreadsheets/d/1DGLY7UEWOlDL5_2dtS_j2eqMjiV-Rw5Fe2y6K6-ULvM/edit#gid=699494809

obto:
Also, right now payment_request_api_usage is far too expensive to run... over $100. Unfortunately we can't use blink_features.usage since we want ecommerce only.

Should we make an exception and run as is? Or remove the query?

rviscomi:
> Could you give a quick look at the comments left in the data? Want to make sure the authors especially understand the CrUX stats.

Sure, could you @ me in any specific comments you'd like me to take a look at?

> For example in the Mobile traffic distribution sheet. How would you read the 50th percentile?

"50% of websites have a phone density of 77.61% or less." or "The median website has a phone density of..."

> Also, right now payment_request_api_usage is far too expensive to run... over $100. Unfortunately we can't use blink_features.usage since we want ecommerce only.

Not sure I understand. Can't you also `JOIN` the technologies dataset with blink_features?

obto:
With this last edit the mobile web chapter should now be complete.

> Can't you also JOIN the technologies dataset with blink_features?

You are completely correct. Somehow I never saw the `features` table.

> "50% of websites have a phone density of 77.61% or less." or "The median website has a phone density of..."

That's how I was thought to read it as well, but the median site having 77% of their traffic come from mobile is pretty crazy. ofc that isn't saying 77% of users are from mobile, but its still way higher than I thought.

rviscomi:
> That's how I was thought to read it as well, but the median site having 77% of their traffic come from mobile is pretty crazy. ofc that isn't saying 77% of users are from mobile, but its still way higher than I thought.

Yeah I agree. I think it could be justified by some websites having mobile-only subdomains that get 100% mobile traffic, but the rest of websites having a mix of desktop and mobile traffic. The number of websites in the mobile dataset is also much bigger, so those things combined could explain it. If you're still skeptical it would be worth closer inspection to validate the results.

obto:
@rviscomi Phones have over 2million more entries than desktop in that device summary table, so I changed the query to only look at websites which have entries for both phone and desktop. This gives the result of 65.56% of traffic being due to mobile, which is much more inline with other sources.

If everything looks good to you, I think we're ready to merge and close out this PR

rviscomi:
> so I changed the query to only look at websites which have entries for both phone and desktop

I don't think this is methodologically valid. It's not a bug to have mobile-only websites in the dataset; that's a reflection of real-world web usage.

sql/2020/12_Mobile_Web/mobile_traffic_distribution.sql
```diff
@@ -4,11 +4,27 @@ SELECT
   percentile,
   APPROX_QUANTILES(phoneDensity, 1000)[OFFSET(percentile * 10)] AS pct_traffic_from_mobile,
 FROM
-  `chrome-ux-report.materialized.device_summary`,
+  (
```
rviscomi, line 7: I'd suggest reverting this change. If the results require more explanation, we can do that in the chapter.

obto: > > so I changed the query to only look at websites which have entries for both phone and desktop
> 
> I don't think this is methodologically valid. It's not a bug to have mobile-only websites in the dataset; that's a reflection of real-world web usage.

Not saying its a bug, but the question we want to be asking... is for the average site, how much of their traffic is due to phones? Or in more specific terms, wouldn't people think of this question as: "If my site has both mobile and desktop visitors, what % is due to mobile phones".

Perhaps it'd be worth delving into why there are 2million more sites for phones vs desktop. And what kind of pages these even are.

rviscomi: IMO there's nothing special about a site that has 99% mobile vs one that has 100% mobile, so we shouldn't be omitting those with 100% as that will skew/bias the results unnecessarily.  Given how prevalent mobile-only sites are, that seems like all the more reason to account for them in the results.

I do agree that it would be an interesting topic of discussion for the chapter to explore why there may be more mobile sites, eg due to personal computing changes around the world, or mobile-only subdomains, etc. But that's not necessarily something we can answer objectively in the dataset.

obto:
@rviscomi Mobile web traffic changes have been reverted

rviscomi (approved):

obto:
Yay!

PR #1208, opened 2020-08-13
Fix Contributors on home page on mobile

bazzadp:
Fixes home page on mobile:

Home page on mobile

@catalinred going to merge this as already released the broken one. This seems to fix it, but let me know if there's a better fix here.


bazzadp:
OK wanna raise a PR for that fix instead?

PR #1211, opened 2020-08-13
Update the fix for the home contributors block

catalinred:
Related to #1208 & [#issuecomment-673303028](https://github.com/HTTPArchive/almanac.httparchive.org/pull/1208#issuecomment-673303028)

### DIFF
--- a/src/static/css/index.css
+++ b/src/static/css/index.css
@@ -557,7 +557,7 @@ p {
 
   /* Contributors */
   .contributors-container {
-    padding: 0 60px 340px; 
+    padding: 0 60px 340px;
   }
 
   .rectangles {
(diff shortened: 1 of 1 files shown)


bazzadp:
I quite like the way it sits on the line now:

Mobile view of home page contributors

However, should it also do this for the 600-900 viewport?

And do you not think the top red guy is very close to the "See all contributors" button above it (on both < 600 and 600-900)?

catalinred:
> However, should it also do this for the 600-900 viewport?

There are a some magic numbers involved in there, so this will require extra time and attention.
Also, I'm not that sure it will look as good as on <600px without additional design changes as scaling up that contributors number a bit maybe.
 
> And do you not think the top red guy is very close to the "See all contributors" button above it (on both < 600 and 600-900)?

I've seen that before, I'll add some extra space now that you noticed that too.

Issue #505, opened 2019-11-17
Translate 2019 content to Portuguese


ibrahimcesar:
These are the core templates - without which we cannot release any translated chapters. They are in the [language specific templates directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/templates):
- [ ] base.html - 
- [ ] base_chapter.html- 
- [ ] contributors.html - 
- [ ] contributors.html (2020) - 
- [ ] error.html- 
- [ ] index.html - 
- [ ] index.html (2020) - 
- [ ] table_of_contents.html - 

These are the chapters to be translated, in rough order of popularity. They exist in the [content directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content):
- [ ] javascript - 
- [ ] css - 
- [ ] seo - 
- [ ] performance - 
- [ ] http2 - 
- [ ] markup - 
- [ ] third-parties -
- [ ] fonts - 
- [ ] accessibility -
- [ ] pwa - 
- [ ] cms - 
- [ ] security - 
- [ ] media - @garcaplay
- [ ] mobile-web - 
- [ ] cdn - 
- [x] ecommerce - @JMPerez 
- [ ] page-weight - 
- [ ] caching - 
- [ ] compression - 
- [ ] resource-hints - 

Additionally the following pages need translated too in the [language specific templates directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/templates):
- [ ] accessibility_statement.html - 
- [ ] base_ebook.html- 
- [ ] methodology.html - 

There is no need to translate the chapters HTML pages are they are generated off the markdown combined with the above templates.

Please include "Makes progress on #505 " in all pull requests so a link is created from the PR to this issue.

---

Common notes for writing consistency are here: https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Translators'-Guide. Feel free to edit that and/or add Portuguese-specific extras by editing this comment.

Portuguese-specific extra advice:
- None yet

bazzadp:
There's been no progress on this since the issue was opened so am going to close this. We can reopen if we get translators.

@ibrahimcesar let us know if you plan on actioning this and want to re-open it?

HakaCode:
would love to start at this point with `base.html`

bazzadp:
> would love to start at this point with `base.html`

Go for it! Give me a shout if any questions.

[...]

PR #1213, opened 2020-08-13
Portuguese translation of base template

HakaCode:
I hope I did everything correctly.

Makes progress on #505


src/templates/pt/2019/base_chapter.html
```diff
@@ -0,0 +1,20 @@
...
+{% block prev_next_title %}Navegação do capítulo anterior e seguinte{% endblock %}
+
+{% block discuss_this_chapter %}Discuta este capítulo{% endblock %}
+{% block comment %}Comente{% endblock %}
+{% block comments %}comentários{% endblock %}
+{% block queries %}Ver consultas{% endblock %}
+{% block results %}Ver resultados{% endblock %}
+
+{% block previous %}Capítulo anterior (pressione 'p' ou ','){% endblock %}
+{% block next %}Próximo Capítulo (pressione 'n' ou '.'){% endblock %}
```
bazzadp, line 20: That telltale red mark shows we need a new line at the end of this file (some programs complain about missing new lines hence best practice to always end a text file with one - which is why GitHub highlights it when you don't).

```suggestion
{% block next %}Próximo Capítulo (pressione 'n' ou '.'){% endblock %}

```

HakaCode:
> Just one small correction please!
> 
> BTW, see my [Git Guide](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Git-Guide#recommended-way-to-use-git-for-the-web-almanac) for ways to avoid that long commit trail

Yes, i spent hours trying to learn by reading and putting it into practice even though i have never seen it, sometimes i get lost with some commands, however, i haven’t done so far, i use Google Translate :)

HakaCode:
@bazzadp  could you guide me on this?

```bash
$ git fetch upstream
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

bazzadp:
This suggests you haven’t set up your ssh key. We’re you using https access to the repo (did you have to type password each time you wanted to push changes?)?

If so, then you needed to add `upstream` in the same way using https instead of git.

You can switch this to https, using:

`git remote set-url upstream   https://github.com/HTTPArchive/almanac.httparchive.org.git`

It us worth looking into setting up SSH keys though and using `git@github.com` addresses instead of `https://github.com`. Much better as don’t have to type your password each time.

bazzadp:
This suggests you haven’t set up your ssh key. We’re you using https access to the repo (did you have to type password each time you wanted to push changes?)?

If so, then you needed to add `upstream` in the same way using https instead of git.

You can switch this to https, using:

`git remote set-url upstream   https://github.com/HTTPArchive/almanac.httparchive.org.git`

It us worth looking into setting up SSH keys though and using `git@github.com` addresses instead of `https://github.com`. Much better as don’t have to type your password each time.

HakaCode:
I got it, thank you

PR #1214, opened 2020-08-14
Dropdown styles for language and year switcher

sudheendrachari:
Fixes: #986 
Intent: Redesign of year and language switcher

Major Changes:
- Introduce two new macros 'styled_lang_dropdown' and 'styled_year_dropdown'
- Existing DOM structure is retained and logic to generate those have been moved into
'native_year_dropdown' and 'native_lang_dropdown'
- Existing select-and-option structure will be used for mobile view where native dropdown picker of the
OS will be used

UI/UX details:
- The button which triggers dropdown will look attached to the dropdown that comes up
- Dropdown list will be aligned to the left edge of button if the entire dropdown content can fit into the
viewport
- Else, Right edge of Dropdown list will be aligned to the right edge of button
- Chevron in the button  will animate to a check mark when the list is opened
- On hover of the items in dropdown menu, its background color and font color will change

Testing:
- Manually verified the behaviour of both the switchers 
- Verified that options are keyboard focusable (Once menu is open the list items are Tab-able,
clicking Tab on last option navigates out of the list to the next focusable element)
- Verified that options are reachable via screen reader also

Demo: 
Mouse and Keyboard: https://drive.google.com/file/d/15-BA8r5e8QbaSi5MiuvZ4Q18L0wYzzlE/view
Voice-over: https://drive.google.com/file/d/17wkCvS7lISqwDVHXlBbsZXUDwHdL7LIx/view


src/static/css/2019.css
```diff
@@ -1223,3 +1223,97 @@ p.copyright a {
...
+  border-color: currentColor;
+  height: 7px;
+  width: 7px;
+  top: 50%;
+  margin-top: -7px;
+  right: 23px;
+  -webkit-transform: rotate(135deg);
+  -ms-transform: rotate(135deg);
+  transform: rotate(135deg);
+  transition: all 200ms linear;
```
rviscomi, line 1265: Love this! 😍

src/static/css/2019.css
```diff
@@ -1223,3 +1223,97 @@ p.copyright a {
...
+.nav-dropdown-list.align-right {
+  right: 0;
+}
+
+.nav-dropdown-list li {
+  width: 100%;
+}
+
+.nav-dropdown-list a {
+  background: #44536a;
```
rviscomi, line 1294: Reusing this color from the footer background
```suggestion
  background: #5c687d;
```

src/static/css/2019.css
```diff
@@ -1223,3 +1223,97 @@ p.copyright a {
...
+.nav-dropdown-btn {
+  padding: 20px 45px 20px 20px;
+  font-size: 17px;
+  font-size: 1.0625rem;
+  text-decoration: none;
+  position: relative;
+  display: flex;
+  align-items: center;
+  border: none;
+  border-radius: 25px;
```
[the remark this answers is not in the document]
sudheendrachari: That i believe, isn't possible, since percentages are w.r.t width and height respectively if two values are provided.
Here's what can be done:

- For buttons, in dropdown collapsed state: Use a high value like 999px as border-radius, that way it works for any height

- In dropdown expanded state, for button and any of the list items which need rounded corners use '25px' (or any fixed value for that matter)

That way in collapsed state it looks exactly like other buttons on the page (like 'start exploring')

Screenshot 2020-08-16 at 6 52 55 PM

and in expanded state it looks like..

Screenshot 2020-08-16 at 6 56 15 PM

Is the difference noticeable? For me it looks ok

src/static/css/2019.css
```diff
@@ -1223,3 +1223,97 @@ p.copyright a {
...
+
+.nav-dropdown {
+  position: relative;
+  margin-right: 2vw;
+}
+
+.nav-dropdown-btn {
+  padding: 20px 45px 20px 20px;
+  font-size: 17px;
+  font-size: 1.0625rem;
```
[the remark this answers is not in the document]
sudheendrachari: You mean other links in header like 'Table of contents' etc..?
They seem to use 17px (1.0625rem) font size

Screenshot 2020-08-16 at 7 03 48 PM

src/static/css/2019.css
sudheendrachari: This change looks fine in header:

https://drive.google.com/file/d/1vLmcl3WhgrK0Qd_5XVK8D56CQCjoNlZo/view

but in footer background of list matches with footer so it will look like this

Screenshot 2020-08-16 at 9 36 13 PM

BTW in the video above check if other changes you suggested are as expected, transparent bg for button, and white border, and color changes to (#f7f779) yellow on hover and focus

src/static/css/2019.css
bazzadp: Think 100ms is enough? Seems a little laggy with 200ms.

src/static/css/2019.css
bazzadp: I don't think it even looks fine in header btw. Not enough contrast their either IMHO so think we need the dark background.

src/templates/base/2019/base.html
```diff
@@ -323,81 +356,131 @@
...
 {% endmacro %}
 
+{% macro styled_lang_dropdown(switcher_name) %}
+  <div class="nav-dropdown {{switcher_name}}">
+    <button class="nav-dropdown-btn" aria-expanded="false" aria-label="{{ self.language_switcher() }}" >{{ language }}</button>
+    <ul class="nav-dropdown-list hidden">
+      {% for l in supported_languages | sort(attribute='_local_name') %}
+        {% if l != language %}
+          <li>
+            <a href="{{ url_for(request.endpoint, **get_view_args(lang=l.lang_code)) }}">{{ l }}</a>
```
bazzadp, line 398:
```suggestion
            <a lang="{{ l.lang_code }}" href="{{ url_for(request.endpoint, **get_view_args(lang=l.lang_code)) }}">{{ l }}</a>
```

src/templates/base/2019/base.html
```diff
@@ -323,81 +356,131 @@
...
+  <div class="nav-dropdown {{switcher_name}}">
+    <button class="nav-dropdown-btn" aria-expanded="false" aria-label="{{ self.year_switcher() }}">{{ year }}</button>
+    <ul class="nav-dropdown-list hidden">
+      {% for y in supported_years | sort %}
+        {% if y != year %}
+          <li>
+            <a href="{{ url_for(request.endpoint, **get_view_args(lang=lang,year=y)) }}">{{y}}</a>
+          </li>
+        {% endif %}
+      {% endfor %}
```
bazzadp, line 466: We've lost our Home Page link  when chapter is not supported in future year.

Chapter with 2019 but no 2020 version


```suggestion
      {% endfor %}
      <li>
        <a class="year-home-link" href="{{ url_for('home', lang=lang,year=y) }}"> {{ y }} {{ self.home() }}</a>
      </li>
```

And then also need ... separator style.

src/static/css/2019.css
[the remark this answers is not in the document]
rviscomi: As long as we're hard-coding a border-radius, could we use 29px since it's exactly half the height and wouldn't have any noticeable difference with the closed button?

src/static/css/2019.css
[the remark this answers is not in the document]
rviscomi: I was looking at the buttons in production, which use 14px:

0 0 0 0_8080_en_2019_
almanac httparchive org_en_2019_ (2)

Keeping it consistent with the 17px header links actually makes more sense, so we can resolve this.

src/static/css/2019.css
rviscomi: Could be fine tuned.

src/static/css/2019.css
rviscomi: Per https://github.com/HTTPArchive/almanac.httparchive.org/pull/1214#discussion_r471014005 I'd also suggest using a border color on the dropdown, which should help distinguish it.

src/templates/base/2019/base.html
bazzadp: This looks to be fixed now - thanks!

src/templates/base/2019/base.html
```diff
@@ -278,6 +289,81 @@
           });
         }
 
+
+        var languageYearNavList = document.querySelectorAll('.nav-dropdown-list');
+        for (let i = 0; i < languageYearNavList.length; i++) {
+          const list = languageYearNavList[i];
+          
+        }
+
```
bazzadp, line 298: Does this do anything now? Can we remove?

```suggestion
```

bazzadp (changes requested):
This is looking pretty good now. Just a few more changes required.

src/templates/base/2019/base.html
```diff
@@ -278,6 +289,81 @@
...
+            list.removeEventListener('keydown', trapFocusInList);
+            document.body.removeEventListener('click', closeAnyOtherOpenDropdown, true);
+          }
+        }
+
+        function navBtnKeyDownHandler(e) {
+          if(e.keyCode === 27) {
+            e.target.click();
+          }
+        }
```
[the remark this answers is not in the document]
sudheendrachari: Gotcha, added

src/static/css/2019.css
```diff
@@ -1223,3 +1227,123 @@ p.copyright a {
...
+  background-color: #5c687d;
+  border-top-color: #5c687d;
+  border-bottom-color: white;
+  z-index: 2;
+}
+
+.nav-dropdown-btn:focus,
+.nav-dropdown-btn:hover {
+  color: #f7f779;
+}
```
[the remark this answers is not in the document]
sudheendrachari: Got it, added

src/templates/base/2019/base.html
sudheendrachari: Good catch, thanks, was meant to be removed

src/static/css/2019.css
sudheendrachari: Changed, footer now open upwards

bazzadp (commented):
This is looking really nice now.

A few things I've spotted (some of these might be in the "nit" category!):

More noticable in Chrome 86 (thanks to the `focus-visible` change), Firefox and Safari is that the border is split - the button is yellow, but the button drop down is white which breaks the illusion of it being one menu item. Is it possible to make it all yellow when button is in focus?

Yellow and White border

On a similar note there's a small, one-pixel gap between the borders. Is it possible to close this gap?

1 pixel gap

The keyboard focus ring is a bit hidden when first menu item is focused due to z-index changes:

broken focus ring

Also not loving use of `z-index` to be honest but can understand why you have done it. Not sure if there is a a better way that would solve above and avoid using z-index?

bazzadp:
Maybe we should just accept the white border in the bottom instead of trying to fake this? That would avoid needing to use `z-index` and so also solve the focus-ring problem, and maybe the yellow/white border difference is less noticeable and so acceptable?

White border in drop down

Another thing I've noticed is that pressing up on the top item cycles right to the bottom. This kind of makes sense but acts weird on Year switcher where there is only one item. Should we include the button (i.e. the current selection) as an option in the arrow key cycle up and down?

sudheendrachari:
The one pixel gap, is because of the way we wanted it to look like a single polygon and do not want white border between button and menu list, and hence the cover up with gray bottom border for button (and z-index is required because list is absolute positioned)
@rviscomi What are your thoughts on this pixel gap and also about Barry's idea about changing border color to all white
@bazzadp I am little unsure if we should include button itself in list where focus is trapped and that is my initial reason to not include that arrow down keydown handler to open from button because even though visually we want them to look like single polygon, a11y wise it might be confusing to jump out of the list and jump into it again. Basically i wanted to trap focus only in the list

bazzadp:
> @bazzadp I am little unsure if we should include button itself in list where focus is trapped and that is my initial reason to not include that arrow down keydown handler to open from button because even though visually we want them to look like single polygon, a11y wise it might be confusing to jump out of the list and jump into it again. Basically i wanted to trap focus only in the list

Yeah can see that rational. Ultimately we need to decide if we want this to look and act like a single list or not. At the moment we have a kind of halfway-house. And remember A11Y is not just about blind people - sighted keyboard users could easily be confused if it looks like one list but then doesn't act like one. Personally I think add the white line to clearly differentiate between the menu and the button and solves most of these problems. It's still a small problem for the year selector since it only has one value at the moment - it will be clearer in future when there are more values and maybe we just take the hit until then. Or perhaps add a "Help Contribute" link like the "Help Translate".

src/templates/base/2019/base.html
```diff
@@ -203,6 +203,17 @@
...
           </ul>
         </nav>
+        <nav id="mobile-footer-nav-items" aria-label="{{ self.footer_title() }}" class="nav-items">
+          <ul>
+            {{ self.non_chapter_nav_links() }}
+            <li>
+              {{ year_switcher('mobile') }}
+            </li>
+            <li>
+              {{ language_switcher('mobile') }}
```
bazzadp, line 213: Getting duplicate ids in the generate HTML so can we change this:

```suggestion
              {{ year_switcher('mobile-footer') }}
            </li>
            <li>
              {{ language_switcher('mobile-footer') }}
```

src/templates/base/2019/base.html
```diff
@@ -323,81 +408,141 @@
...
-    {% endfor %}
-    <option disabled="disabled" aria-hidden="true">
-      ────
-    </option>
-    <option value="https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Translators'-Guide">
-      {{ self.help_translate() }}
-    </option>
-  </select>
-</div>
+  {% if switcher_name == "mobile" %}
```
bazzadp, line 411:
```suggestion
  {% if switcher_name == "mobile" or switcher_name == "mobile-footer" %}
```

src/templates/base/2019/base.html
```diff
@@ -323,81 +408,141 @@
...
-              {% if y not in supported_years %}
-              <option value="{{ url_for('home', lang=lang,year=y) }}">
-                {{ y }} {{ self.home() }}
-              </option>
-              {% endif %}
-              {% endfor %}
-            {% endif %}
-        </select>
-      </div>
+      {% if switcher_name == "mobile" %}
```
bazzadp, line 467:
```suggestion
      {% if switcher_name == "mobile" or switcher_name == "mobile-footer" %}
```

bazzadp (changes requested):
Noticed we've got some duplicate `ids` with this, so suggested a couple of changes to solve this. May want to rename `mobile` to `mobile-header` too for consistency.

Could you also merge the latest `main` branch into this branch as there are a couple of changes since this branch was started that will lint the generated HTML which would have caught this?

bazzadp:
Oh and one more thing I noticed is in the footer when the menu opens upwards instead of downwards the bottom border is white instead of yellow:

Bottom white border on menu open in footer

Sorry for discovering these one by one - it's quite a bit change all in all!

sudheendrachari:
Thanks @bazzadp for helping with the rebase issue for this branch
Have addressed duplicate id issue and added yellow border when list is open

bazzadp (approved):
LGTM! Great job and thanks for being so patient through all the feedback!

src/templates/base/2019/base.html
```diff
@@ -278,6 +289,80 @@
...
+        }
+        function trapFocusInList(e) {
+          var list = e.currentTarget;
+          var isInFooter = list.classList.contains('footer-list');
+          if(e.key === "ArrowDown") {
+            var siblingElem = isInFooter ? e.target.parentElement.previousElementSibling : e.target.parentElement.nextElementSibling;
+            var focusableElem = siblingElem ? siblingElem.querySelector('a') : (isInFooter ? lastFocusableElementInList : firstFocusableElementInList);
+            e.preventDefault();
+            focusableElem.focus();
+          } else if (e.key === "ArrowUp") {
```
rviscomi, line 307: After expanding the dropdown list in the footer, I would expect the "Up" key to focus on the option directly above the button. It seems to be scrolling up instead. I need to press "Down" to focus on one of the options before being able to use the "Up" key in this way.

src/templates/base/2019/base.html
```diff
@@ -278,6 +289,80 @@
...
+          if (e.target.classList.contains('dropdown-open')) {
+            return
+          };
+          var openDropdownBtn = document.querySelector('.nav-dropdown-btn.dropdown-open');
+          openDropdownBtn && openDropdownBtn.click();
+        }
+        function trapFocusInList(e) {
+          var list = e.currentTarget;
+          var isInFooter = list.classList.contains('footer-list');
+          if(e.key === "ArrowDown") {
```
rviscomi, line 302: In the language selector in the footer, the "Down" key selects the Spanish option first. Pressing "Down" again selects "Help Translate" next.

In order to traverse all of the language options in order, we need to press Down to select Spanish, then Up, Up, Up.

For the footer list options, which are reversed, I think pressing "Down" at first should select the last option ("Help Translate") and the "Up" key should select the first option ("Spanish").

rviscomi (changes requested):
This is so well done, thank you @sudheendrachari! 🏆 

I just have two small keyboard usability comments about the footer.

src/templates/base/2019/base.html
sudheendrachari: This is something that i had thought about initially but wasn't sure if i should add that functionality because even in footer the chevron icon still points down but because visually the list was longer than available space we've decided to render list above button. so because of that i was unsure if we should just support arrow down key or arrow up as well.
Now that you've also noticed the lack of it makes me believe thats how people might expect it to work, so added that logic as well

bazzadp: LGTM now!

rviscomi (approved):

Issue #1178, opened 2020-08-07
Add Lighthouse testing


bazzadp:
As mentioned in our [Technical Accomplishments page](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Technical-accomplishments) we currently have very good Lighthouse Scores:

![Lighthouse scores for Web Almanac with 99% for Performance, 100% for Accessibility, 100% for Best Practices and 100% for SEO](https://user-images.githubusercontent.com/10931297/89048613-f8e33500-d347-11ea-9e95-fee0a4bd1c09.png)

I'd like to keep it that way and prevent any regressions as we develop the platform.

Can we add Lighthouse testing, either as part of our `npm run test` command or as a separate [continuous integration](https://github.com/GoogleChrome/lighthouse-ci). Preference would be part of `npm run test` so can be run offline while developing.

Now when running it locally we do lose a few points:

![Local Lighthouse scores are lower](https://user-images.githubusercontent.com/10931297/89655482-1a9d6880-d8c2-11ea-952b-56ac7456c9e6.png)

This is because Flask, doesn't gzip responses, doesn't use HTTP/2 and the 127.0.0.1 domain causes some SEO points to be deducted - but that's fine. I think we should still set 90/100/100/85 as performance budgets and alert if anything goes below this, but open to suggestions.

What to test as well? Guessing too slow to test every page, so maybe should pick some key pages (Home Page, and a large chapter like CSS)? We should make it easy to add new pages to the test in case 2020 ends up being different from 2019 pages. Could we also have a "test all pages" option which we could run periodically if we wanted to?

rviscomi:
Could use the PageSpeed Insights API to get Lighthouse scores as well as real-user experience data from CrUX.

bazzadp:
> Could use the PageSpeed Insights API to get Lighthouse scores as well as real-user experience data from CrUX.

That would be good for production monitoring @rviscomi but I'm more thinking of this Issue as a test in pull requests - to stop any recessions getting into production.

bazzadp:
@sudheendrachari I'm super keen on this so are you still planning on working on this? Hoping we didn't put you off with that drop down change! 😀

sudheendrachari:
Haha. Yes, I'm still willing to work on this. Hopefully there should be some progress by next week

bazzadp:
Good stuff - looking forward to it!

bazzadp:
Any more progress here @sudheendrachari ?

[...]

PR #1216, opened 2020-08-15
Fix lighthouse regressions

bazzadp:
We’ve recently lost some Lighthouse points on Accessibility and SEO due to #1185 meaning we’re no longer getting 100% for them:

Lighthouse Score for 2019 JavaScript chapter showing 92 for Performance, 94 for Accessibility, 100 for Best Practices and 99 for SEO

This PR fixes it.

Hopefully when we complete #1178 this won’t be allowed to happen.

### DIFF
--- a/src/templates/base/2019/base_chapter.html
+++ b/src/templates/base/2019/base_chapter.html
@@ -134,7 +134,7 @@ window.discussion_url="https://discuss.httparchive.org/t/{{ metadata.get('discus
                 <div class="social">
                   {% if authordata.twitter %}
                   <a class="twitter" href="https://twitter.com/{{ authordata.twitter }}" aria-labelledby="author-{{ author }}-twitter">
-                    <svg width="22" height="22" role="img" aria-hidden>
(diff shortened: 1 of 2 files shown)


src/templates/base/2019/base_chapter.html
```diff
@@ -134,7 +134,7 @@ <h2 id="authors">
                 <div class="social">
                   {% if authordata.twitter %}
                   <a class="twitter" href="https://twitter.com/{{ authordata.twitter }}" aria-labelledby="author-{{ author }}-twitter">
-                    <svg width="22" height="22" role="img" aria-hidden>
+                    <svg width="22" height="22" role="img">
```
borisschapira, line 137: Do you know why there is `role="img"` here and not on other pictos?

bazzadp: Can you give me an example where it's not set? Just did a search for `<svg` and always seems to be set to me?

It's not necessary for `` elements as they are directly mapped to images, whereas `<svg>`'s may not be. Is that what you mean?

PR #1230, opened 2020-08-19
Remove generated content from being tracked in Git

### LINKED ISSUE #1221: Should we remove generate content from git repo?
@HTTPArchive/developers , question for you all:

We auto generate some content, including:
- Chapters `.html` files (from chapters `.md` files)
- The `sitemap.xml`
- Potentially `featured_chapters.html` (depending on outcome of #1104)
- Ebook `.html` and `.pdfs`

This creates some confusion as searching the code base (e.g. to fix a typo) leads to multiple hits - but only the `.md` files should be edited. It also means some people edit both `.md` and `.html` files when they don't need to.

We could remove these files from the Git repo (and add the relevant `.gitignore` entries to ensure they are not re-uploaded).

The generated HTML files will still be generated, tested and linted as part of our GitHub action so don't think we lose anything with that.

We should also add the `npm run generate` to the `npm run deploy` script as they need to be generated and uploaded to App Engine to actually run the site.

The one downside is developers will now need to run `npm run generate` to generate these `.html` files, to be able to run and be able to develop the website. So a developer just making some CSS changes but nothing else, now has an extra step to complete. Do we see that being a problem?

Thoughts?

bazzadp:
Fixes #1221 by removing all generated content from git, adding `.gitignore` entries to stop it getting added back, and adding deploy script to ensure it's generated before deploying.

I’ve left the PDFs in there as unlikely to be confused for real content and allows us to continue to use the GitHub Action to generate the ebooks rather than make people have to install Prince locally.


src/tools/scripts/deploy.sh
```diff
@@ -0,0 +1,47 @@
...
+if [ "$DEPLOY" != "Y" ] && [ "$DEPLOY" != "y" ]; then
+    echo
+    echo "Cancelling deploy"
+    exit 0
+fi
+
+echo "Deploying"
+gcloud app deploy --project webalmanac --stop-previous-version
+
+read -r -n 1 -p "Are you happy to push changes to production?: [Y/N]" DEPLOY
```
rviscomi, line 35: In what scenario would we want live code to deviate from the `production` branch?

src/tools/scripts/deploy.sh
```diff
@@ -0,0 +1,47 @@
...
+if [ "$(pgrep -f 'python main.py')" ]; then
+  echo "Killing existing server to run afresh"
+  pkill -9 python main.py
+fi
+
+echo "Run and test website"
+./tools/scripts/run_and_test_website.sh
+
+echo "Generate ebooks"
+npm run ebooks
```
rviscomi, line 21: What's the behavior if the deployer doesn't have the ebook prerequisites (Prince installed)? Does the deploy fail or does it proceed without ebooks?

src/tools/scripts/deploy.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+# exit when any command fails instead of trying to continue on
+set -e
+
+echo "Update local production branch"
+git checkout production
```
rviscomi, line 7: Not sure why, but I get nervous running git commands automatically. Should we assume the deployer has already manually updated the branch and maybe exit early from the script if it detects unmerged changes?

src/.gitignore
```diff
@@ -1,3 +1,7 @@
 .idea/
 .vscode/
-node_modules/
\ No newline at end of file
+node_modules/
+templates/*/*/chapters/
```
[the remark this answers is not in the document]
bazzadp: I looked into it and you can’t have different `.gitignores` for each branch [without hacks workarounds](https://stackoverflow.com/questions/29579546/git-excludesfile-for-a-branch/29583813#29583813).

We could update `.gitignore` as part of this but that feels hacky too.

Ultimately we need to decide if storing in git or not. I don’t see a point in half storing it.

Btw I’m still not 100% convinced about not storing it. It’s nice to be able to see HTML differences for some changes. Then again the actual HTML rather than Jinja2 template is what’s important and we don’t store that - though maybe should if we decide to go static site.

For now on balance I’m still for removing it. Not 100% but maybe 90% 😀

src/tools/scripts/deploy.sh
bazzadp: And that’s why I check before pushing it (which I see you asked about later).

Wonder how to check for unmerged changes. Would be nice if there’s a simple way to check that. Will have a google.

src/tools/scripts/deploy.sh
bazzadp: It fails and exits instead of continuing. Prince is now basically a requirement to run this script. We can run individual steps manually if really want to avoid it.

I’m in two kinds here too.

The “Generate Ebooks” GitHub action works quite well and avoid having to do local installs with Git. But at the expense of having to remember to run that first and store the results in Git.

Ultimately I see this as a short term solution and I’d like to love to CI/CD so adding burden of installing PrinceXML for the few deployers we have is OK with me.

Or alternatively we keep PDFs in Git and only regenerate them periodically via the GitHub action if prince not installed?

src/tools/scripts/deploy.sh
```diff
@@ -0,0 +1,47 @@
...
+
+echo "Run and test website"
+./tools/scripts/run_and_test_website.sh
+
+echo "Generate ebooks"
+npm run ebooks
+
+echo "Please test the site locally"
+
+read -r -n 1 -p "Are you ready to deploy?: [Y/N]" DEPLOY
```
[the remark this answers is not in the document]
bazzadp: Oh wasn’t aware that was a convention.

I’m still in two minds as to whether we want so many confirmations or whether it should just run all the way through. It makes me nervous but the old deploy bypassed the gcloud check and worked fine. Maybe should have more confidence in this script and remove the confirmations. WDYT?

src/tools/scripts/deploy.sh
bazzadp: You wouldn’t. I’m just nervous about auto-pushing to Git without confirmation!

src/tools/scripts/deploy.sh
```diff
@@ -0,0 +1,47 @@
...
+read -r -n 1 -p "Are you happy to push changes to production?: [Y/N]" DEPLOY
+if [ "$DEPLOY" != "Y" ] && [ "$DEPLOY" != "y" ]; then
+    echo
+    echo "Exiting"
+    exit 0
+fi
+
+echo "Push production branch"
+git push
+git status
```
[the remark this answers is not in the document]
bazzadp: The `git status`? Just confirmation it’s a clean branch now.

src/tools/scripts/deploy.sh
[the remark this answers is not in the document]
bazzadp: Fixed.

src/.gitignore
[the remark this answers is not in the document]
bazzadp: The script now zips to deploy.zip which can be uploaded to Release folder as a release artifact. Might look to automate this more in the future.

src/tools/scripts/deploy.sh
bazzadp: I've added a check to see if it's a clean branch (with option to continue even if not) and also exit on first error so think this is safe to automate.

src/tools/scripts/deploy.sh
bazzadp: OK change of plan here. I've reverted the change so eBooks are still stored in Git. This allows us to run the GitHub action to generate the ebooks. I've added a step to confirm this is done:

```bash
read -r -n 1 -p "Please confirm you've updated the eBooks via GitHub Actions. [y/N]: " DEPLOY
if [ "$DEPLOY" != "Y" ] && [ "$DEPLOY" != "y" ]; then
    echo
    echo "Cancelling deploy"
    exit 0
fi
echo
```

src/tools/scripts/deploy.sh
bazzadp: I've removed this so now will push to production automatically if all other steps complete successfully.

bazzadp:
@rviscomi I think I've addressed all your feedback and made some improvements too. Could you have another look?

@ibnesayeed I'd value your input here too!

src/tools/scripts/deploy.sh
```diff
@@ -0,0 +1,92 @@
...
+set -e
+
+# Check branch is clean first
+if [ -n "$(git status --porcelain)" ]; then 
+  read -r -n 1 -p "Your branch is not clean. Do you still want to deploy? [y/N]: " DEPLOY
+  if [ "$DEPLOY" != "Y" ] && [ "$DEPLOY" != "y" ]; then
+      echo
+      echo "Cancelling deploy"
+      exit 0
+  fi
```
ibnesayeed, line 13: We can DRY this block of the code using a function that takes the prompt text as an argument because this logic is being repeated below a few times.

src/tools/scripts/deploy.sh
```diff
@@ -0,0 +1,92 @@
...
+
+echo "Deploying"
+echo Y | gcloud app deploy --project webalmanac --stop-previous-version
+
+echo "Push production branch"
+git push
+git status
+
+echo "Please update release on GitHub using tag $TAG_VERSION@production"
+echo "Please upload deploy.zip as the release artifact"
```
ibnesayeed, line 91: If these messages are call to action, then we might want to highlight them using color escapes.

ibnesayeed:
One way to envision the whole production branch is as following:

* Track only the source files in the `main` branch as the source of truth that can be used to generate the site at any historical state
* Add a build process that creates a `dist` folder that contains the complete directory structure of only files needed to serve the site, such as:
    * All the generated files
    * Scripts that are needed at run time, but no other helper scripts to maintain the project
    * Derived files such as bundled JS/CSS (minified or otherwise), but not the pre-transpilation files such as CoffeeScript/TypeScript/SASS
    * Files that are going to be the same in both the source as well as the runtime environment
* Exclude the `dist` folder using the `.gitignore` file
* Force push the content of the `dist` folder to the `production` branch
* Additionally, if maintaining the history of the production branch is not important, it can be purged each time before pushing new content to it to only contain the most recent distribution

bazzadp:
That’s kind of what I’m going for here @ibnesayeed but in a slightly different manner than you’re suggesting.

It’s not quite as easy to separate source code from deployed (`dist`) code as you are suggesting.

For a start some of our source code is deployed as is. Particularly our Python code, the `static` folder and most of the templates. So we’d duplicate that code in the `deploy` folder, which seems a little pointless - why not keep the folder structure we have and use the `.gcloudignore` like I’m suggesting?

Also some of our helper scripts (particularly those in the `tools/generate` folder) are important to see _how_ and _why_ the artifacts were built the way they were. We are not using the release to actually deploy, nor to feed downstream (can’t see anyone becoming dependent on us - but who knows what the future holds!). So I see the release artifact as the snapshot in time as to what was released in case we even need to go back to understand what exactly was in a release for whatever reason. Yes in theory the git tagging does that too, but as there is generated content that will no longer be tracked due to this PR I think it’s important to track what was deployed - not just the source files. Which is the point of that. Maybe the deployer had a different machine setup and so that might explain why the deployed files were as they should have been? Or maybe they didn’t have a clean checkout? So  I think having the source files, the build files, and the generated content all in the `deploy.zip` file will be helpful in case we were need to do some investigation to identify a release issue.

I also don’t know how you can force push a folder that’s in `.gitignore` - is that possible?

I also dislike the idea of `production` diverging from `main`. That way lies merge conflicts and errors IMHO. And while it’s true that `production` is its own branch, there is nothing to stop is throwing it away and resetting or rebasing to main if it gets in a mess. That’s not possible if it diverges - at least not cleanly without losing history. While the most benefit it probably in the latest deploy as you say, why not use GitHub to track them all?

I think storing the released artifacts in the Release section of GitHub is a lot cleaner, gives us the history, and links it to a tagged version in Git so really like that workflow. Plus it gives us a release history and allows other contributors to watch for releases to see when site changes (maybe when their development, or writing, or translation finally goes live!).

Plus for me, the Draft release notes let’s me see when items are piling up and we should cut a release, as well as reminding me what exactly has changed since the last release.

src/tools/scripts/deploy.sh
bazzadp: Good idea. Done.

src/tools/scripts/deploy.sh
bazzadp: Good idea - done here and in a few other places.

src/tools/scripts/deploy.sh
```diff
@@ -0,0 +1,107 @@
...
+if [[ -f deploy.zip ]]; then
+  echo "Removing old deploy.zip"
+  rm -f deploy.zip
+fi
+
+echo "Zipping artifacts into deploy.zip"
+zip -q -r deploy . --exclude @.gcloudignore
+
+echo "Deploying to GCP"
+echo "Y" | gcloud app deploy --project webalmanac --stop-previous-version
```
ibnesayeed, line 89: You may want to use this handy little utility for such cases called, `yes`, to generate a stream of `yes`es. If you want to yield a different string, such as `Y`, this can be done using the optional argument like `yes Y` (ironically, to answer no, you need to use `yes no`).

```suggestion
yes | gcloud app deploy --project webalmanac --stop-previous-version
```

src/tools/scripts/deploy.sh
```diff
@@ -0,0 +1,107 @@
...
+git status
+
+echo "Checking out main branch"
+git checkout main
+
+echo
+echo -e "${GREEN}Successfully deployed!${RESET_COLOR}"
+echo
+echo -e "${AMBER}Please update release on GitHub: https://github.com/HTTPArchive/almanac.httparchive.org/releases${RESET_COLOR}"
+echo -e "Using tag ${TAG_VERSION}@production${RESET_COLOR}"
```
ibnesayeed, line 102: Did you mean this?

```suggestion
echo -e "${AMBER}Using tag ${TAG_VERSION}@production${RESET_COLOR}"
```

src/tools/scripts/deploy.sh
```diff
@@ -0,0 +1,107 @@
...
+echo -e "${AMBER}Please upload deploy.zip as the release artifact${RESET_COLOR}"
+echo
+echo "Have a nice day!"
```
ibnesayeed, line 105: Who knows what time it is? :wink:

```suggestion
echo "Have a nice one!"
```

src/tools/scripts/deploy.sh
bazzadp: Don't need a string - just need one of them and think `echo "Y"` is clearer here than `yes y`

src/tools/scripts/deploy.sh
bazzadp: Change it to "Have a good one!"

bazzadp:
Some observations after the first manual deploy of the site using this process:

Zipping the assets produced a 91MB zip file. That’s a bit bigger than I realised it would be (It’s mostly the images) and it does take a minute or two to upload on my connection.

It looks like GitHub automatically adds the state of the repo as assets at the bottom of a release (in both `zip` and `tar.gz` formats). That contains 90% of what we need and there’s a lot of duplication.

We could either just zip up the generated content, in the script in this PR, to make this quicker and smaller? Or just skip the whole thing because the zipped versions of the repo are just so handy that they are auto generated and make releases a matter of deploying and then sending the auto generated draft release note live.

What do you two think?

Ultimately I’d like to run the deploy through GitHub Actions but still some work to do to enable that.

src/.gcloudignore
```diff
@@ -6,7 +6,11 @@
 # For more information, run:
 #   $ gcloud topic gcloudignore
 #
+# Note we are also using this for zip ignore in deploy.xh, so have to be a little
```
rviscomi, line 9:
```suggestion
# Note we are also using this for zip ignore in deploy.sh, so have to be a little
```

src/README.md
```diff
@@ -152,28 +152,26 @@ _Make sure you have generated the ebooks PDFs first in the main branch, by runni
...
-git push
-```
-
-  - Check out the `production` branch
-  - Run `git status` to ensure you don't have any uncommitted changes locally
-  - Merge any remote changes (both origin/production and origin/main branches)
-  - Push the merge-commit back up to origin/production
-
-4. Browse the website locally as one final QA test, then deploy the changes live:
+3. Deploy the site:
```
rviscomi, line 155: Is "_Make sure you have generated the ebooks PDFs first in the main branch, by running the Generate Ebooks GitHub Action_" above still true?

bazzadp: Yes that is still the case as discussed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/1230#discussion_r475244806

PR #1232, opened 2020-08-19
Config updates for new chapters and other misc cleanups

bazzadp:
@HakaCode , @chengxicn , @borisschapira , @garcaplay , @ksakae1216 could you approve (or suggest better wording!) your language templates for the three new chapters this year:

- Capabilities - this is about adding new app-like capabilities to websites
- Privacy
- Jamstack (left this as is for all the languages as kind of a term)

### DIFF
--- a/src/templates/en/2019/base.html
+++ b/src/templates/en/2019/base.html
@@ -154,7 +154,10 @@ Unless otherwise noted, the metrics in all of the 20 chapters of the Web Almanac
     "cdn": "CDN",
     "page-weight": "Page Weight",
     "resource-hints": "Resource Hints",
-    "http2": "HTTP/2"
+    "http2": "HTTP/2",
+    "privacy": "Privacy",
+    "capabilities": "Capabilities",
+    "jamstack": "JAMstack"
   }
 %}
 
@@ -165,7 +168,7 @@ Unless otherwise noted, the metrics in all of the 20 chapters of the Web Almanac
   "designers": "Designers",
   "developers": "Developers",
   "editors": "Editors",
-  "leads": "Leads",
+  "leads": "Project Leads",
   "reviewers": "Reviewers",
   "translators": "Translators", 
   }
--- a/src/templates/en/2019/base.html
+++ b/src/templates/en/2019/base.html
@@ -157,7 +157,7 @@ Unless otherwise noted, the metrics in all of the 20 chapters of the Web Almanac
     "http2": "HTTP/2",
     "privacy": "Privacy",
     "capabilities": "Capabilities",
-    "jamstack": "JAMstack"
+    "jamstack": "Jamstack"
   }
 %}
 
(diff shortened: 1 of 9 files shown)


src/templates/fr/2019/base.html
```diff
@@ -154,7 +154,10 @@
     "cdn": "CDN",
     "page-weight": "Poids des pages",
     "resource-hints": "Indices de Ressources",
-    "http2": "HTTP/2"
+    "http2": "HTTP/2",
+    "privacy": "Intimité",
```
borisschapira, line 158:
```suggestion
    "privacy": "Vie Privée",
```

src/templates/en/2019/base.html
```diff
@@ -154,7 +154,10 @@
     "cdn": "CDN",
     "page-weight": "Page Weight",
     "resource-hints": "Resource Hints",
-    "http2": "HTTP/2"
+    "http2": "HTTP/2",
+    "privacy": "Privacy",
+    "capabilities": "Capabilities",
+    "jamstack": "JAMstack"
```
rviscomi, line 160: @ahmadawais is this the correct capitalization?

borisschapira (changes requested):
We tend to translate Privacy by "Vie Privée" (literally: (the) private life).
Other than that, it's okay by me.

src/templates/en/2019/base.html
bazzadp: Yeah there seems some difference of opinion here:

Jamstack:
https://jamstack.org/
https://www.netlify.com/jamstack/

JAMstack:
https://jamstack.wtf/

Would be good to get agreement on what we are going to use.

borisschapira: Since you ask, as the term has been coined by Netlify, I would go with them on Jamstack, but I don't think it really matters.

bazzadp: Apparently the official word is Jamstack: https://github.com/jamstack/jamstack.org/issues/279

Will change this PR for not but feel free to disagree before it's merged! @rviscomi / @ahmadawais we should go through and update all other references in the repo if we're in agreement on this.

ahmadawais: Used to be JAMstack but after a healthy discussion and lots of confusion it's now Jamstack. While I don't like the later I think that's the way to go. So go ahead with Jamstack.

chengxicn (approved):
Chinese chapters are correct and proper.Approved

bazzadp:
Gonna merge this as planning a release. @garcaplay / @c-torres if you spot anything wrong with the Spanish translations then let me know. Think the others are all approved.

garcaplay:
Sorry, didn't see this PR on time neither. As soon as I can I will read those sections and if anything doesn't makes sense or whatever will add a PR with the correction.

PR #1236, opened 2020-08-20
Streamline figure markdown

bazzadp:
Makes progress on #1138 

At present this does not:
- change the anchor links as the ids (will need to think how we handle 2019 links if we do change these and update all the 2019 chapters to the new format - which I'd like to do for consistency).
- do much for the tables (though `id`s are automated and `fig-caption` is automated a little).
- add the menu or query links suggested in #1138 

It definitely does clear up a lot of the repetition and makes it a lot easier to add figures and have consistency in the mark up used for figures.

At the moment I've only moved two of the 2019 chapters to show how they work. Can add the other chapters to this PR once it's been reviewed, or open a new PR for them.


src/content/en/2019/css.md
```diff
@@ -31,67 +31,76 @@ Color is an integral part of theming and styling on the web. Let's take a look a
...
 
-<figure>
-  <a href="/static/images/2019/css/fig1.png">
-    <img src="/static/images/2019/css/fig1.png" alt="Figure 1. Popularity of color formats." aria-labelledby="fig1-caption" aria-describedby="fig1-description" width="600" height="371" data-width="600" data-height="371" data-seamless data-frameborder="0" data-scrolling="no" data-iframe="https://docs.google.com/spreadsheets/d/e/2PACX-1vQO5CabwLwQ5Lj1_9bbEFnFM1qEqCorymaBHrcaNiMSJ7sYDKHUI5iish5VAS-SxN447UTW-1-5-OjE/pubchart?oid=1946838030&amp;format=interactive">
-  </a>
-  <div id="fig1-description" class="visually-hidden">Bar chart showing the adoption of HSL, HSLA, RGB, RGBA, and hex color formats. Hex is used on 93% of desktop pages, RGBA on 83%, RGB on 22%, HSLA 19%, and HSL 1%. Desktop and mobile adoption is similar for all color formats except HSL, for which mobile adoption is 9% (9 times higher).</div>
-  <figcaption id="fig1-caption">Figure 1. Popularity of color formats.</figcaption>
-</figure>
+{{ figure_markup(
```
rviscomi, line 34: WDYT about leaving all of the 2019 contents as-is and only using these function for 2020? Not sure if it's worth the maintenance cost to edit everything.

src/content/en/2019/css.md
```diff
@@ -31,67 +31,76 @@ Color is an integral part of theming and styling on the web. Let's take a look a
...
+  id=1,
```
rviscomi, line 35: Figure numbering is a huge part of #1138 and I see in the PR description that it's purposely not included. Do you plan to iterate on it? If not should we continue the discussion of pros/cons or technical implementation in the issue?

src/templates/base/2019/base.html
```diff
@@ -385,6 +385,41 @@
   {% endmacro %}
 {% endif %}
 
+{%- macro figure_markup(
```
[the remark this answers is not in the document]
bazzadp: OK will remove

src/templates/base/2019/base.html
```diff
@@ -385,6 +385,41 @@
...
+  caption="",
+  alt=caption,
+  description="",
+  sheets_url="",
+  width=600,
+  height=371,
+  data_width=width,
+  data_height=height,
+  big_number="",
+  really_big_number=""
```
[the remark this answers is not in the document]
bazzadp: Yeah could get out of hand I suppose. Though at the moment we only have those two classes. Will have another look.

src/content/en/2019/css.md
```diff
@@ -31,67 +31,76 @@ Color is an integral part of theming and styling on the web. Let's take a look a
...
+  image="/static/images/2019/css/fig1.png",
```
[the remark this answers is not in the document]
bazzadp: Good idea - will have a look

src/content/en/2019/css.md
bazzadp: I like have the consistency. I have some regexs worked out to make this relatively easy. Also have a feeling we may forget how this works, or may stop testing it between now and 2020 figures being ready. Having old chapters as examples would be good.

Also makes translations easier as there is a lot of duplication and confusing syntax.

Also means any new features we add (like your menu) are automatically backported. Though I do agree that adding extra meta-data (e.g. links to SQL or specific sheets) isn't worth the effort to dd to the old years so may not benefit that much from it.

src/content/en/2019/css.md
bazzadp: Potentially. Either as part of this PR (which might make sense to avoid going at all the chapters again shortly after merging this). However I wanted feedback on what's been done so far and am still mulling over the figures in my head.

While I agree maintaining figure numbers automatically would be ideal, I think there's still huge value in what's in this PR so far even without that. The syntax is hugely simplified and now very easy to make changes consistently (e.g. change "Figure 2" to "Figure 2.2").

I'll be honest though and admit I'm still not 100% convinced on the automatic numbering. I think it would be possible to generate the numbers for the `<figure>`'s with "id=XXX" place holders and replacing with regex (much like we do now), or perhaps petty through some DOM queries. However I am less convinced about in the text, nor moving the anchors to names.

I still think naming things is hard. Yes we ideally want descriptive name for image files too which should give us those unique names but I still think writing `In Figure {{ fig_id("fig_dist_websites_performance") }} above, you can see how FCP...` is much more painful than writing `In Figure 1 above` - even if that means dealing with the consequences that some figures might need to be renumbered as we add the Figures just before launch (and only very rarely afterwards).

And, whether we like it or not, I think people may well refer to them externally ("In the Web Almanac in Figure 1.2 they showed" - though admittedly can't find an example of this!), or (by taking screenshots](https://www.stefanjudis.com/blog/highlights-of-the-http-archive-web-almanac/) so we should try to be consistent in figure numbering after launch as much as possible. I remember in #592 noticing the numbering change but that's much less likely no we don't see the templates - especially now we've taken that further with #1230 .

And, this one is probably an edge case, but I do flick between languages (e.g. /en/2019/cdn#fig-1 and /ja/2019/cdn#fig-1) so having consistency is helpful to me, rather than language specific anchors. I think the short anchor names are handy.

While on the subject, I also think the language thing is still a complexity to be solved. They can either use English (not very i18n friendly - though take your point that chapter names are not internationalised) or use their own tags - where we have to ensure consistency between references in text and figures, as well as include the English anyway for the image link.

So, in summary, I've still concerns to work through and automatic figure numbering may or may not be added. We should continue the discussion (either here or in #1138 ) while I play with how to do it technically. BUT, either way, I think there's huge value in this PR already as is.

src/templates/base/2019/base.html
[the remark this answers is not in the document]
bazzadp: Fixed.

src/content/en/2019/css.md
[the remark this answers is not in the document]
bazzadp: Fixed

src/templates/base/2019/base.html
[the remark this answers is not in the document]
bazzadp: Fixed

src/templates/base/2019/base.html
```diff
@@ -385,6 +385,43 @@
...
+    {%- endif %}
+  </a>
+  <button hidden="" class="fig-description-button" aria-expanded="false" aria-controls="fig{{ id }}-description" data-show-text="{{ show_description(id) }}" data-hide-text="{{ hide_description(id) }}">{{ show_description(id) }}</button>
+  <div id="fig{{ id }}-description" class="visually-hidden">{{ description|safe }}</div>
+  {%- elif content != "" %}
+    {%- if classes != "" %}
+  <div class="{{ classes }}">{{ content }}</div>
+    {%- else %}
+  <div>{{ content }}</div>
+    {%- endif %}
```
ibnesayeed, line 419: I am not sure if the linter will be happy with an empty class value like `class=""` (which is perfectly valid), if it is then we can avoid conditional in the template. Another possibility is to always have a default class value.

```suggestion
  <div class="{{ classes }}">{{ content }}</div>
```

bazzadp: Yeah good idea. Think will always be a class for these so will just go with this simple one.

src/content/en/2019/css.md
bazzadp: We could compromise and auto generate the figure numbers (like we do now) but leave the figure numbers hardcoded in the text, and still use the `figX`-style anchors?

That would leave us in same place as now, but with the figure markup drastically simplified (which is see as the main aim of this PR).

Of course that means there’s still a risk that the text references and figures get out of sync, but that’s a present risk anyway and an acceptable risk given how infrequently I expect these to change, and makes the writing simpler.

WDYT?

src/templates/base/2019/base.html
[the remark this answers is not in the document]
bazzadp: I reverted this back. If's just easier to regex for it as "figure(" is a bit generic.

src/content/en/2019/css.md
bazzadp: OK I've had another stab at this and brought it closer to what you want:

Figure Ids are now auto-generated. They were already, so removing this actually was a reversion in that respect. I was a little uncomfortable with this anyway but think I've mitigated against breakage concerns with the addition of a a new (optional) `figure_lock` value in the chapter YAML metadata:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/bead9c5941a35b8872c207dfe1299d049e05f5ba/src/content/en/2019/css.md#L11-L14

https://github.com/HTTPArchive/almanac.httparchive.org/blob/bead9c5941a35b8872c207dfe1299d049e05f5ba/src/content/en/2019/javascript.md#L11-L14

If this config is set, then `npm run generate` errors when the site is built if the figure count does not match that. So we SHOULD set this just before publishing the chapter and "lock" in the figure count:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/bead9c5941a35b8872c207dfe1299d049e05f5ba/src/tools/generate/generate_figure_ids.js#L73-L82

I think this is a good safety check and hopefully alleviates both my concerns on figure ids changing, and therefore (hopefully) your concerns on using simple figure ids as anchor links.

Given that, I have kept the `fig-12` anchors. I like 'em! - short and sweet and no concerns about dupes or character encoding, or having to add them to `big-number` and `table` figures. As I say I think once we start labelling figures we should try our best to avoid changing the numbering and if we do control that, then link breakage is less of a concern.

And finally I've added the ability to reference figures in text:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/bead9c5941a35b8872c207dfe1299d049e05f5ba/src/content/en/2019/css.md#L89-L99

As you can see the `ref="figref1"` is an optional argument added to the `figure_markup` call, and then references to that in the writing are automatically replaced with the appropriate figure id. In this case the text becomes:

> Unsurprisingly, In Figure 2.5 above...

Let me know your thoughts!

rviscomi: The outcomes I'm hoping to achieve in this PR are abstracting away figure markup/a11y best practices and ergonomic improvements to the authoring experience. I think we hit the first goal with `figure_markup`. But on the ergonomics side I'm afraid that these changes have expanded in complexity to the point that it's actually more cumbersome to use than the current implementation. WDYT about reverting the latest `figure_lock` and `figref` changes and we can merge the `figure_markup` improvements? We can iterate on the ergonomic design asynchronously in #1138.

rviscomi:
BTW, while testing this locally I had to resolve a merge conflict. I figured I'd push those [changes](https://github.com/HTTPArchive/almanac.httparchive.org/pull/1236/commits/0cbabd6f3cd160c9fbee28d4660007e2277fa7dd) to your branch so you don't have to repeat it.

src/content/en/2019/css.md
bazzadp: I do think the `figref` is more bother than its worth and happy to remove that. Was never particularly happy with that anyway as you may have noticed 😉

However I do like the `figure_lock` stuff. The more we abstract away the figure numbering (which we have done almost completely in this) the more easy it is to accidentally change figure numbering by removing a figure and then that breaks all references in chapters and anchor links - like we almost did in #592. This is actually more likely now we no longer can see this as we stopped including the generated templates in the main PRs ages ago - and as of more recently don’t even store them in Git at all! Would you be up for keeping those? We only need to “lock” the chapters once published so authors don’t need to bother about this. WDYT?

rviscomi: I don't see the benefit of `figure_lock` TBH. Contrary to my goal in #1138 to make figure management more automated/abstract, it creates another thing that needs to be manually kept in sync.

bazzadp: Fair enough - it is gone!

src/content/en/2019/css.md
rviscomi: Will this create a discrepancy in figure numbering between CSS/JS and the rest of the 2019 chapters? The former will use the new chapter number prefixes (eg "Figure 1.1").

rviscomi (approved):
LGTM, sorry for the delay.

Just one comment about figure numbering consistency for 2019.

src/content/en/2019/css.md
bazzadp: Just committed the other chapters there - for English anyway (will do the other languages now). Have some regexes that makes this easy and even spotted a few longstanding typos as part of that when the regexes didn't work!

The only question is that two of our chapters (Markdown and Performance) had different figure anchors (`#fig1` instead of `#fig-1` like the other 18 chapters use). Previously, when we automated this a little, we catered for this:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/4b23bd53f348f52dee73a0714d6c50ebf67d6738/src/tools/generate/generate_figure_ids.js#L10-L16

I think we shouldn't complicate the logic by catering for it this time by adding "if 2019 and (markdown or performance)" and should just be consistent across the chapters despite the fact that this might be that anchor links for these chapters will change. You OK with that? And don't let the fact you authored one of these chapters sway you too much 😉

rviscomi (approved):
Would you also be up for documenting the `figure_markup` interface for authors, either in the wiki or a README?

bazzadp:
All chapters updated. Merging and yes I'll document in the wiki. Who owns this? Analysts? Authors? Editors?

PR #1249, opened 2020-08-24
Fix percentiles in FR translation of JavaScript chapter

borisschapira:
Using Voice Over, I noticed that using 'p25' is, in fact, read as 'page 25'.
This fix this, so that screen reader read the content in a more intelligible way.

Makes progress on #539


bazzadp:
Hmmm... I agree VoiceOver is reading it out wrong, but not loving the idea of stopping using these common terms for percentile. And it's a much bigger issue than just the French JavaScript chapter. Seems to affect the CDN, JavaScript and Media chapters in various languages.

Then again despite "p10" being common phrasing to statisticians, "10th percentile" does read easier for those less familiar with dealing with percentiles, so maybe this is another case of an accessibility "bug", highlighting a real accessibility issue for all? So maybe we should accept this and make similar changes elsewhere too? Might be a bit more effort for charts [like in the CDN chapter](http://127.0.0.1:8080/en/2019/cdn#fig-9) but can be worked around (e.g. having a "percentile" group header and just the numbers below)

@rviscomi do you have an opinion here?

rviscomi:
No strong opinions but if assistive technologies are interpreting it incorrectly I agree there's a bug. I'm ok with fixing that by either changing it for everyone (p10 -> 10th percentile) or special casing the ATs (`<span aria-label="10th percentile">p10</p>`).

bazzadp:
Annoying the `aria-label` workaround doesn't work on `<span>`'s, `<div>`'s or most other static elements. It does appear to work for `h1`-`h6` on voice over (as [suggested by this blog post](http://www.davidmacd.com/blog/does-aria-label-override-static-text.html)) but that's not consistent with other screen readers.

So let's fix in text to "10th percentile". Will merge this as this is the only French chapter affected and open separate PRs for other languages.

PR #1253, opened 2020-08-25
Display Ebook download size on link

bazzadp:
The Ebook is quite large (approx 17Mb) so we should make people aware of this before they click on the download link.

Previous:
Ebook downlink without size

With this change:
Ebook downlink with size

### DIFF
--- a/src/server/helpers.py
+++ b/src/server/helpers.py
@@ -87,7 +87,7 @@ def get_chapter_nextprev(config, chapter_slug):
     return prev_chapter, next_chapter
 
 
-def get_ebook_size(lang, year):
+def get_ebook_size_in_mb(lang, year):
     ebook_file = STATIC_DIR + '/pdfs/web_almanac_%s_%s.pdf' % (year, lang)
     if os.path.isfile(ebook_file):
         size = os.path.getsize(ebook_file)
--- a/src/server/helpers.py
+++ b/src/server/helpers.py
@@ -91,10 +91,8 @@ def get_ebook_size(lang, year):
     ebook_file = STATIC_DIR + '/pdfs/web_almanac_%s_%s.pdf' % (year, lang)
     if os.path.isfile(ebook_file):
         size = os.path.getsize(ebook_file)
-        # Convert to MB with decimal point precision (e.g. 17.1MB)
-        # Note we use 1000 rather than 1024 as that's worst case and how
-        # MacOS reports it - Windows seems to use 1000 / 1024
-        size = int(round(size / 1024 / 1000, 0))
+        # Convert to MB
+        size = int(round(size / 1024 / 1024, 0))
         return size
 
     return 0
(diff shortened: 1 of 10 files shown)


src/server/helpers.py
```diff
@@ -87,8 +87,15 @@ def get_chapter_nextprev(config, chapter_slug):
...
 
 
-def ebook_exists(lang, year):
-    return os.path.isfile(STATIC_DIR + '/pdfs/web_almanac_%s_%s.pdf' % (year, lang))
+def get_ebook_size(lang, year):
+    ebook_file = STATIC_DIR + '/pdfs/web_almanac_%s_%s.pdf' % (year, lang)
+    if os.path.isfile(ebook_file):
+        size = os.path.getsize(ebook_file)
+        # Convert to MB
+        size = int(round(size / 1024 / 1024, 0))
```
ibnesayeed, line 95: This line has an inherent assumption that the size is going to be more than half a megabyte, because otherwise it will return 0 which will be a falsy and will prevent the download link from appearing.

I will suggest returning the size as is from here and for display purposes wire a helper function that can convert the number into a human readable one, something like:

```py
def humanize(size):
    for unit in ['B','KB','MB','GB','TB']:
        if size < 1024.0:
            return "%.1f%s" % (size, unit)
        size /= 1024.0
    return "%.1f%s" % (size, 'BP')
```

bazzadp: Given that the current size is 17MB, and that we only generate ebooks when all chapters are translated I think that's a fair assumption and don't think we need the complexity of your function.

Additionally, not specifying the unit allows translations to format this as they please (e.g. English might use 17MB, but another language might prefer "17 MB" (with a space) or even "MB17").

So would prefer to leave it as is.

ibnesayeed: I was not aware that ebooks are generated only after all the chapters with their assets are in. I was afraid that if someone is testing the generation of ebooks with very little data in place, it is not unlikely to have a file size less than half a megabyte, which will be rounded down to zero and will show strange results.

I would at least suggest the function name to be more explicit about the returned result, so may be we can change it to `get_ebook_size_in_mb` which will otherwise not be obvious without looking into the function body.

bazzadp: Seems a little verbose to me but ok. Done!

bazzadp: BTW in case you're curious, we enable ebooks here:
https://github.com/HTTPArchive/almanac.httparchive.org/blob/f93b839af84fc1fafa07172a111dbf446665f79f/src/config/2019.json#L6

ibnesayeed: > Seems a little verbose to me but ok. Done!

That's a fair point. Another possibility is to leave the function name without `_in_mb` suffix, but keep the suffix in the `ebook_size_in_mb` template variable so that translators know how to format the phrase.

bazzadp: If we're gonna have it in one place, then might as well have it in both I say.

Issue #1254, opened 2020-08-25
Voiceover reading percentiles incorrectly


bazzadp:
As noted in #1249 Voiceover is reading "p10" as "page 10" so need to change this to "10th percentile" or equivalent.

Affects the following chapters:
- [x] en cdn - #1256 
- [x] en javascript - #1256 
- [x] en media - #1256 
- [x] es javascript - #1264 
- [x] es media - #1264 
- [x] fr javascript - fixed in #1249 
- [x] ja cdn #1265
- [x] ja javascript #1265
- [x] ja media #1265

The CDN chapters will require some restructuring for table headers.

PR #1256, opened 2020-08-25
Voiceover percentiles for English chapters

bazzadp:
Makes progress on #1254 

Note this adds another heading to the CDN tables:

Old:
Percentile's marked p10 - p90

New:
Percentile's marked 10 - 90 with a percentile heading

### DIFF
--- a/src/content/en/2019/cdn.md
+++ b/src/content/en/2019/cdn.md
@@ -486,11 +486,15 @@ In contrast, the median TLS negotiation for the majority of CDN providers is bet
     <thead>
       <tr>
         <td></td>
-        <th scope="col">p10</th>
-        <th scope="col">p25</th>
-        <th scope="col">p50</th>
-        <th scope="col">p75</th>
-        <th scope="col">p90</th>
+        <th colspan="5" scope="col">Percentile</th>
+      </tr>
+      <tr>
+        <td></td>
+        <th scope="col">10</th>
+        <th scope="col">25</th>
+        <th scope="col">50</th>
+        <th scope="col">75</th>
+        <th scope="col">90</th>
       </tr>
     </thead>
     <tbody>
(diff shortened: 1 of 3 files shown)


rviscomi:
By separating the "percentile" text from the numbers, is this any clearer to assistive technology users? 🤔

bazzadp:
> By separating the "percentile" text from the numbers, is this any clearer to assistive technology users? 🤔

In so much as any large table is clear if just listening to it! But it does stop reading as "Page X" so that's a plus 😁

We already use this in other chapters [like the compression chapter](http://127.0.0.1:8080/en/2019/compression#fig-1)

bazzadp:
> By separating the "percentile" text from the numbers, is this any clearer to assistive technology users? 🤔

Oops - looks like I never actually add the note on testing for this that I did!

On VoiceOver it's not as nice when you go into the cells as it only announces the `colgroup` heading

Old: "Page 10 1, Page 25 1, Page 50 1, Page 75 1, Page 90 2"
New "Percentile 1, 1, 1, 1, 2"

So voiceover doesn't announce sub headers and only announces the `colgroup` heading when first going into that (column 1). Ideally it should announce "Percentile 10 1, Percentile 25 1..." though on other hand that could also be repetitive and annoying. Perhaps better would be "Percentile 10 1, 25 1...".

Either way, that seems like a VoiceOver bug and I still think this change is better than what it announces currently as removes the incorrect "Page".

bazzadp:
Updated to better implementation [after discussion on the A11Y slack](https://web-a11y.slack.com/archives/C042TSFGN/p1598616792300300).

I still think using the p10 - p90 headings in tables should be discouraged in future as not everyone will be familiar what they mean, and we can't add a "Percentile" `colgroup` without removing the individual column headings from being read in VoiceOver (and maybe other ATs too).

However this change solves the immediate problem of VoiceOver reading this out wrong so think this is good to merge now.

borisschapira (approved):

PR #1271, opened 2020-09-04
GitHub Action to test generated template changes

bazzadp:
Since removing the generated content in #1230 we've lost the ability to track changes to our templates from library upgrades. This makes it difficult to know whether any upgrades from dependabot are safe to commit or if there is an impact that should be investigated further before upgraded. This was noted in https://github.com/HTTPArchive/almanac.httparchive.org/issues/1221#issuecomment-675275804

This PR adds a GitHub action which runs on package.json and package-lock.json changes (and on demand) which compares the generated templates folder with those in main and adds a comment to the PR if there are template changes. This will allow for easier review of library upgrades.

Note it also locks the package.json file further as upgrades are handled by dependabot and without this more stricter locking, changes may not be seen.

I do not run this on every commit as we're assuming any other template changes are intentional due to adding new content or enhancing our code, and this could be quick noisy otherwise.

### DIFF
--- a/.github/workflows/test-template-changes.yml
+++ b/.github/workflows/test-template-changes.yml
@@ -26,7 +26,7 @@ jobs:
   build:
     name: Build and Test Template Changes
     runs-on: ubuntu-latest
-    if: github.repository == 'bazzadp/almanac.httparchive.org'
+    if: github.repository == 'HTTPArchive/almanac.httparchive.org'
     steps:
     - name: Checkout branch
       uses: actions/checkout@v2.3.2
--- a/.github/workflows/test-template-changes.yml
+++ b/.github/workflows/test-template-changes.yml
@@ -1,0 +1,51 @@
+######################################
+## Custom Web Almanac GitHub action ##
+######################################
+#
+# This Action generated the dynamic templates. It also then takes a copy
+# of the templates folder, and then checks out main branch and regenerates the
+# templates and compares the two. If run from a pull request it adds a comment
+# showing the diffs.
+#
+# Note it does not run a full test of the website as that will be completed by
+# the test_website.yml GitHub Action.
+#
+# This Action is useful to check impact of dependency upgrades (e.g. if
+# prettier version change results in different HTML being generated), since we
+# do not store generated templates in Git and so cannot otherwise see these
+# changes. Changes can then be accepted or investigated further before merging.
+#
+name: Test Template Changes
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - '**/package.json'
+      - '**/package-lock.json'
+jobs:
+  build:
+    name: Build and Test Template Changes
+    runs-on: ubuntu-latest
+    if: github.repository == 'bazzadp/almanac.httparchive.org'
+    steps:
+    - name: Checkout branch
+      uses: actions/checkout@v2.3.2
+      with:
+        fetch-depth: 0
+    - name: Setup Node.js for use with actions
+      uses: actions/setup-node@v1.4.2
+      with:
+        node-version: 12.x
+    - name: Test Template Changes
+      id: test-template-changes
+      run: ./src/tools/scripts/test_template_changes.sh
+    - name: 'Comment PR'
+      uses: actions/github-script@0.3.0
+      if: github.event_name == 'pull_request' && env.PR_COMMENT != ''
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        script: |
+          const { issue: { number: issue_number }, repo: { owner, repo }  } = context;
+          // Unescape any escaped comments so code blocks appear correctly
+          const pr_comment = process.env.PR_COMMENT.replace(/%0A/g,'\n').replace(/%0D/g,'\r').replace(/%25/g,'%');
+          github.issues.createComment({ issue_number, owner, repo, body: pr_comment });
--- a/.github/workflows/test-template-changes.yml
+++ b/.github/workflows/test-template-changes.yml
@@ -28,7 +28,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.repository == 'HTTPArchive/almanac.httparchive.org'
     steps:
-    - name: Checkout branch
+    - name: Checkout Code
       uses: actions/checkout@v2.3.2
       with:
         fetch-depth: 0
--- a/.github/workflows/test-template-changes.yml
+++ b/.github/workflows/test-template-changes.yml
@@ -2,7 +2,7 @@
 ## Custom Web Almanac GitHub action ##
 ######################################
 #
-# This Action generated the dynamic templates. It also then takes a copy
+# This Action generates the dynamic templates. It also then takes a copy
 # of the templates folder, and then checks out main branch and regenerates the
 # templates and compares the two. If run from a pull request it adds a comment
 # showing the diffs.
--- a/src/tools/scripts/test_template_changes.sh
+++ b/src/tools/scripts/test_template_changes.sh
@@ -58,6 +58,7 @@ cat "${DIFF_FILENAME}"
 
 NUM_DIFFS=$(wc -l < "${DIFF_FILENAME}")
 if [ "${NUM_DIFFS}" -ne "0" ]; then
+  # Weird syntax is to make is Mac compatible: https://stackoverflow.com/a/1252191/2144578
   ESCAPED_OUTPUT=$(sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/%0A/g' -e 's/\r/%0D/g' -e 's/\%/%25/g' "${DIFF_FILENAME}")
   PR_COMMENT="Please note, that the following diffs happen in the templates on this branch compared to main:%0A\`\`\`%0A${ESCAPED_OUTPUT}%0A\`\`\`%0A"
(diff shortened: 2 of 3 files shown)


bazzadp:
Oh and btw you can see an example of what is produced here: https://github.com/bazzadp/almanac.httparchive.org/pull/143

src/tools/scripts/test_template_changes.sh
```diff
@@ -0,0 +1,69 @@
...
+
+NUM_DIFFS=$(wc -l < "${DIFF_FILENAME}")
+if [ "${NUM_DIFFS}" -ne "0" ]; then
+  ESCAPED_OUTPUT=$(sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/%0A/g' -e 's/\r/%0D/g' -e 's/\%/%25/g' "${DIFF_FILENAME}")
+  PR_COMMENT="Please note, that the following diffs happen in the templates on this branch compared to main:%0A\`\`\`%0A${ESCAPED_OUTPUT}%0A\`\`\`%0A"
+  echo "::set-env name=PR_COMMENT::${PR_COMMENT}"
+fi
+
+echo "Removing templates backup"
+rm -rf "${TEMP_TEMPLATES_DIRECTORY}"
```
ibnesayeed, line 67: Should we also remove `DIFF_FILENAME` explicitly? I know it is in the TEMP space, but perhaps cleaning it would be no harm.

.github/workflows/test-template-changes.yml
```diff
@@ -0,0 +1,51 @@
...
+    paths:
+      - '**/package.json'
+      - '**/package-lock.json'
+jobs:
+  build:
+    name: Build and Test Template Changes
+    runs-on: ubuntu-latest
+    if: github.repository == 'HTTPArchive/almanac.httparchive.org'
+    steps:
+    - name: Checkout branch
```
ibnesayeed, line 31:
```suggestion
    - name: Checkout Code
```

src/tools/scripts/test_template_changes.sh
```diff
@@ -0,0 +1,69 @@
...
+set -e
+
+echo "Differences:"
+cat "${DIFF_FILENAME}"
+
+NUM_DIFFS=$(wc -l < "${DIFF_FILENAME}")
+if [ "${NUM_DIFFS}" -ne "0" ]; then
+  ESCAPED_OUTPUT=$(sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/%0A/g' -e 's/\r/%0D/g' -e 's/\%/%25/g' "${DIFF_FILENAME}")
+  PR_COMMENT="Please note, that the following diffs happen in the templates on this branch compared to main:%0A\`\`\`%0A${ESCAPED_OUTPUT}%0A\`\`\`%0A"
+  echo "::set-env name=PR_COMMENT::${PR_COMMENT}"
```
ibnesayeed, line 63: Since the message is printed in the next step, I think it would be more suitable to populate as an output variable using `::set-output`.

src/tools/scripts/test_template_changes.sh
```diff
@@ -0,0 +1,69 @@
...
+set +e
+diff -r templates "${TEMP_TEMPLATES_DIRECTORY}" > "${DIFF_FILENAME}"
+set -e
+
+echo "Differences:"
+cat "${DIFF_FILENAME}"
+
+NUM_DIFFS=$(wc -l < "${DIFF_FILENAME}")
+if [ "${NUM_DIFFS}" -ne "0" ]; then
+  ESCAPED_OUTPUT=$(sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/%0A/g' -e 's/\r/%0D/g' -e 's/\%/%25/g' "${DIFF_FILENAME}")
```
ibnesayeed, line 61: In one of my scripts I did this using a custom `escapify` function. https://github.com/ibnesayeed/repo-attrs/blob/26b8384fe72cfecbd0cb56ed99643cfc85a92302/entrypoint.sh#L4-L9

src/tools/scripts/test_template_changes.sh
bazzadp: Yeah I never really get the difference between explicitly using outputs versus env variables and which is better or worse.

I did try to add this as an output but got a syntax error. See here: https://github.com/bazzadp/almanac.httparchive.org/actions/runs/239234777 Any ideas what I did wrong here?

src/tools/scripts/test_template_changes.sh
bazzadp: Good idea. Done.

src/tools/scripts/test_template_changes.sh
bazzadp: Nice. But since for this script it's in a file anyway I think using `sed` is just as good and potentially gives us more powerful options should they be needed.

src/tools/scripts/test_template_changes.sh
ibnesayeed:
```
outputs:
PR_COMMENT:
    description: The comment to put in pull request
```

This is not how you set output. This syntax is for shareable actions' metadata file, not the inline one.

There are JavaScript functions around input, output, and env variable, but when were are using the Shell scripts, we use `echo` command with [specific syntax and prefixes](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-output-parameter) to achieve this. For example:

```
echo "::set-env name=action_state::yellow"
echo "::set-output name=action_fruit::strawberry"
```

These will set (case-insensitive) `action_state` environment variable and `action_fruit` output variable. Output variables are also handled by environment variables internally (using a specific prefix).

src/tools/scripts/test_template_changes.sh
```diff
@@ -0,0 +1,71 @@
...
+if [ "${NUM_DIFFS}" -ne "0" ]; then
+  # Weird syntax is to make is Mac compatible: https://stackoverflow.com/a/1252191/2144578
+  ESCAPED_OUTPUT=$(sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/%0A/g' -e 's/\r/%0D/g' -e 's/\%/%25/g' "${DIFF_FILENAME}")
+  PR_COMMENT="Please note, that the following diffs happen in the templates on this branch compared to main:%0A\`\`\`%0A${ESCAPED_OUTPUT}%0A\`\`\`%0A"
+  echo "::set-env name=PR_COMMENT::${PR_COMMENT}"
+fi
+
+echo "Removing templates backup"
+rm -rf "${TEMP_TEMPLATES_DIRECTORY}"
+rm -f "${DIFF_FILENAME}"
```
ibnesayeed, line 69: This could have been added in the previous `rm` command as it can take multiple arguments, but there is no problem keeping it separate either.

.github/workflows/test-template-changes.yml
```diff
@@ -0,0 +1,51 @@
...
+# do not store generated templates in Git and so cannot otherwise see these
+# changes. Changes can then be accepted or investigated further before merging.
+#
+name: Test Template Changes
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - '**/package.json'
+      - '**/package-lock.json'
```
ibnesayeed, line 24: Is globbing pattern necessary here? If were are watching two specific files and not something nested deep, we might as well specify their paths.

src/tools/scripts/test_template_changes.sh
bazzadp: I did that as well in the script if you [look at the commit](https://github.com/bazzadp/almanac.httparchive.org/commit/418337ff0448c9bcfc21536e7b79ad687d1218b3#diff-7497f9dd00232279bf97574072bbe5ebR59). Do I not also need to set the fact the there is an output for this step in the yml file?

And btw if they are env variables under the hood then what’s benefit of using output variables as opposed to env variables?

ibnesayeed: > I did that as well in the script if you look at the commit.

You should also escape multi-line text before adding it to `set-env` or `set-output` etc. ([see one of my codes for example](https://github.com/ibnesayeed/repo-attrs/blob/master/entrypoint.sh)).

> Do I not also need to set the fact the there is an output for this step in the yml file?

That is not for inline steps, but when you publish reusable actions, then you specify it in the metadata file.

> And btw if they are env variables under the hood then what’s benefit of using output variables as opposed to env variables?

I am not too sure about any specific advantages of using an output variable instead of an environment variable. It changes the context where the information is available and feels more systematic. That said, if it gives headaches in getting it right then we can move on with environment variables and call it a day.

bazzadp: Does seem outputs are for published actions, where you can show the outputs in the YAML and where there’s a higher risk of env variables clashes.

So yeah let’s leave as is since it‘s all working and doesn’t appear to be an advantage.

.github/workflows/test-template-changes.yml
bazzadp: No real difference at the moment to be honest. No real harm to glob as only looks at track files (so won’t include node_modules dependencies) and who knows if someone will a separate package.json for a particular subfolder/module in future.

src/tools/scripts/test_template_changes.sh
bazzadp: Quite long env variables to think two lines is just clearer.

ibnesayeed (approved):
LGTM!

PR #1276, opened 2020-09-06
Lint all files on Super Linter upgrade

bazzadp:
At the minute I run a manual full lint on the code base for Super Liunter upgrades. I'd like to automate this.
Also remove out of date comments.

@ibnesayeed can you think of any way to avoid the duplication required in the two `if` statements? Or this the cleanest we can get?

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -32,7 +32,7 @@ jobs:
         run: |
           echo "::set-env name=VALIDATE_ALL_CODEBASE::false"
       - name: Lint Code Base
-        uses: github/super-linter@v3.9.3
+        uses: github/super-linter@v3.9.4
         env:
           DEFAULT_BRANCH: main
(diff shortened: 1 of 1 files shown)


ibnesayeed:
> @ibnesayeed can you think of any way to avoid the duplication required in the two if statements? Or this the cleanest we can get?

One possibility to avoid complex conditionals is to create a duplicate linter workflow for comprehensive linting and configure it to run in the two situations when we want complete linting and for the shallow linting keep the existing workflow after slimming it down further. Obviously, it won't be DRY anymore, but sometimes it is better to keep things simple to understand. That said, if we do not anticipate more complexities in the two conditionals going forward then the current approach is okay too.

ibnesayeed (approved):
LGTM!

bazzadp:
Thanks for reviewing. Let’s keep as is and revisit if it starts to grow more.

PR #1277, opened 2020-09-07
Few small perf fixes for home page

bazzadp:
Noticed a few minor perf issues with home page:
- Height and width were wrong way round for home page logo
- Compressed star character (again!) to stop Lighthouse complaining
- Logo had a font weight of 600 even though we only have 400 and 700 fonts. Seems to be same as 700 so using that instead to avoid browser having to manually scale


bazzadp:
Nice. Image action has finally fixed webp optimisation!

PR #1288, opened 2020-09-12
Tidy up 2019 sql

bazzadp:
Big tidy up to:
- Add date to 2019 SQL queries
- Move to `count(0)` consistently
- Add newline characters at end of SQL files

Probably should have done this before people started writing 2020 queries!

A lot of this was manual copy and past so there's a good chance one or two errors got in there. Still, it's better than it was, and I think it's a good thing to do as the 2019 queries - as referenced on our website will no longer give accurate data AND will cost a lot as the data set is getting bigger. Especially relevant until 2020 Almanac and Queries go live.

Don't expect you to review every, single change but just adding you as reviewers so 1) you're aware of this and 2) you can confirm you're happy to make this change.


rviscomi (approved):
Huge improvement, thank you!! 🚀

PR #1292, opened 2020-09-14
Upgrade to Python 3.8

bazzadp:
Looks like [Google App Engine now officially supports Python3.8](https://cloud.google.com/appengine/docs/standard/python3/runtime) - previously it was marked as beta so didn't want to upgrade to it.

To be honest I've been using Python 3.8 all along but let's make it official and run on 3.8 in the cloud too.

Deployed test version here and all good: https://20200914t192537-dot-webalmanac.uk.r.appspot.com/

### DIFF
--- a/.github/workflows/generate_chapters.yml
+++ b/.github/workflows/generate_chapters.yml
@@ -44,10 +44,10 @@ jobs:
       uses: actions/setup-node@v1.4.2
       with:
         node-version: 12.x
-    - name: Set up Python 3.7
+    - name: Set up Python 3.8
       uses: actions/setup-python@v2.1.2
       with:
-        python-version: '3.7'
+        python-version: '3.8'
     - name: Run the website
       run: ./src/tools/scripts/run_and_test_website.sh
     - name: Create Pull Request
--- a/.github/workflows/generate_ebooks.yml
+++ b/.github/workflows/generate_ebooks.yml
@@ -26,11 +26,10 @@ jobs:
       uses: actions/setup-node@v1.4.2
       with:
         version:  12.x
-    # Install Python 3.7 as used by Google Cloud Platform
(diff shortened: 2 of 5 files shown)


rviscomi:
What are the benefits of upgrading?

bazzadp:
Nothing I really want to use to be honest. This more to keep on top of things and because I’ve been using 3.8 locally (I presume you have too since we never specified a version in our virtual env?) so been meaning to align.

Full set of changes here: https://docs.python.org/3/whatsnew/3.8.html

rviscomi:
Oh ok. Is this a no-op or could there be any compatibility / local development risks? Upgrades can sometimes break things so I'm just trying to get a sense for the risk vs reward.

bazzadp:
Well I want to do it in part to avoid any local development issues/deployment risks! As I say I've always used 3.8 because we never specified a minor versions so it always used the latest version 3:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/cb7b2c2f4b5be3a0fb3dd4e6807675cb603b2c78/src/README.md#L17-L20

Similar the DockerFile just uses v3 (which means v3.8) for anyone using that.

So basically this brings the server side inline with what (I assume) we're all developing on.

Out of interest what are you running? Looks like 3.8 was only released in October 2019 so guess old timers on the project like you may still be running 3.7? Even if you are and chose to stay on it, this PR updates our GitHub Actions to 3.8 so any PR will be tested on 3.8 anyway.

Additionally the GitHub Action tests the whole website (including running `pytest`) and completed with no issues. The `no-promote` test deploy showed no issues either..

So any upgrade always carries some level of risk, but think this is as close to no-op as we can get!

rviscomi:
Yeah I'm on 3.7.3

rviscomi (approved):

PR #1296, opened 2020-09-16
Compress Images on pushes to main and at other times

bazzadp:
The [Calibre-Image Github Action](https://github.com/calibreapp/image-actions) was not able to compress images on pull requests from forks, due to security restrictions in GitHub Actions themselves.

However they have recently merged a pull request (from yours truly!) that allows it to run in Compress Only mode and then open a new Pull Request (much like we do for our timestamp updates) for those image optimisations, meaning we no longer have to accept unoptimised images from PRs from forks and remember to optimise them ourselves.

I've also added the ability to run this on-demand, and also weekly just in case anything slips through, or they add any improvements.

This will be useful as the Analysts start to submit the images for the 2020 chapters which should start happening soon.

### DIFF
--- a/.github/workflows/calibreapp-image-action.yml
+++ b/.github/workflows/calibreapp-image-action.yml
@@ -37,12 +37,8 @@ jobs:
     name: calibreapp/image-actions
     runs-on: ubuntu-latest
     # Only run on main repo on and PRs that match the main repo
-    if: |
-      (github.event_name != 'pull_request' &&
-       github.repository == 'HTTPArchive/almanac.httparchive.org')
-      ||
-      (github.event_name == 'pull_request' &&
-       github.event.pull_request.head.repo.full_name == github.repository)
+    if: github.repository == 'HTTPArchive/almanac.httparchive.org' &&
+      (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
     steps:
       - name: Checkout branch
         uses: actions/checkout@v2.3.2
--- a/.github/workflows/calibreapp-image-action.yml
+++ b/.github/workflows/calibreapp-image-action.yml
@@ -64,4 +64,6 @@ jobs:
           body: ${{ steps.calibre.outputs.markdown }}
           labels: performance
       - name: Check outputs
+        # If it's not a pull request then commit any changes as a new PR
+        if: steps.cpr.outputs.pull-request-number != ''
         run: echo "Pull Request Number - ${{ steps.cpr.outputs.pull-request-number }}"
--- a/.github/workflows/calibreapp-image-action.yml
+++ b/.github/workflows/calibreapp-image-action.yml
@@ -4,35 +4,66 @@
 # 
 # This runs the calibre GitHub action to compress our images
 #
-# This Action has a few limitations:
-#
-# 1) This currently only works on pull requests from this repo
-#    and fails silently for PRs from forks. We've added a PR
-#    to handle this better (https://github.com/calibreapp/image-actions/pull/54)
-#    and after that can probably combine to the generate_chapters.yml.
-#
-# 2) It seems to only do a basic level of optimisations and better results can
-#    be achieved from https://tinypng.com/ or https://squoosh.app/ but not as automated
-#    so still good to have this.
-#
-# 3) Currently doesn't handle WebP (https://github.com/calibreapp/image-actions/issues/31)
+# This Action has a few different methoids for running:
+# 1) For Pull Requests from and to the main repo it can run and update the PR
+# 2) For pushes to main, it will run and create a new PR. This captures PR from
+#    Forks.
+# 3) It also runs every sunday at 11pm in case anything else sneaks in or any
+#    improvements are added to Calibre. This will also open a PR if necessary.
+# 4) Finally it can be kicked off manually and it will open a PR if necessary.
 #
 name: Compress images
 on:
+  schedule:
+    # Run a full compression every Sunday at 11pm to start the week right
+    - cron: '00 23 * * 0'
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - '**.jpg'
+      - '**.jpeg'
+      - '**.png'
+      - '**.webp'
   pull_request:
     paths:
       - '**.jpg'
+      - '**.jpeg'
       - '**.png'
       - '**.webp'
 jobs:
   build:
-    # Only run on Pull Requests within the same repository, and not from forks as they don't work
-    if: github.event.pull_request.head.repo.full_name == github.repository
     name: calibreapp/image-actions
     runs-on: ubuntu-latest
+    # Only run on main repo as PRs can't be updated on forks
+    if: github.repository == 'HTTPArchive/almanac.httparchive.org'
+    env:
+      COMPRESS_ONLY: false
     steps:
-      - uses: actions/checkout@v2.3.2
-      - name: calibreapp/image-actions
-        uses: docker://calibreapp/github-image-actions
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Set COMPRESS_ONLY variable to true
+      # If it's not a pull request then run in compressOnly mode
+      if: github.event_name != 'pull_request'
+      run: |
+        echo "::set-env name=COMPRESS_ONLY::true"
+    - name: Checkout branch
+      uses: actions/checkout@v2.3.2
+    - name: calibreapp/image-actions
+      id: calibre
+      uses: calibreapp/image-actions@master
+      with:
+        githubToken: ${{ secrets.GITHUB_TOKEN }}
+        compressOnly: ${{ env.COMPRESS_ONLY }}
+    - name: Create Pull Request
+      # If it's not a pull request then commit any changes as a new PR
+      if: |
+        github.event_name != 'pull_request' &&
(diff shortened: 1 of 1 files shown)


.github/workflows/calibreapp-image-action.yml
```diff
@@ -4,35 +4,67 @@
...
+    if: github.repository == 'HTTPArchive/almanac.httparchive.org'
+    env:
+      COMPRESS_ONLY: false
     steps:
-      - uses: actions/checkout@v2.3.2
-      - name: calibreapp/image-actions
-        uses: docker://calibreapp/github-image-actions
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Set COMPRESS_ONLY variable to true
```
ibnesayeed, line 44: All the steps should be indented one level to make it consistent with the rest of the file, YAML parsers would consider it no different than without the indentation in this case, but I am suggesting this change just for the sake of consistency.

.github/workflows/calibreapp-image-action.yml
```diff
@@ -4,35 +4,67 @@
...
+      run: |
+        echo "::set-env name=COMPRESS_ONLY::true"
+    - name: Checkout branch
+      uses: actions/checkout@v2.3.2
+    - name: calibreapp/image-actions
+      id: calibre
+      uses: calibreapp/image-actions@master
+      with:
+        githubToken: ${{ secrets.GITHUB_TOKEN }}
+        compressOnly: ${{ env.COMPRESS_ONLY }}
```
ibnesayeed, line 56: We can conditionally set the `compressOnly` parameter inline, so no need to initialize an environment variable `COMPRESS_ONLY` with the default value `false` and then setting it to true in a conditional step.

```suggestion
        compressOnly: ${{ github.event_name != 'pull_request' }}
```

.github/workflows/calibreapp-image-action.yml
```diff
@@ -4,35 +4,67 @@
...
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Set COMPRESS_ONLY variable to true
+      # If it's not a pull request then run in compressOnly mode
+      if: github.event_name != 'pull_request'
+      run: |
+        echo "::set-env name=COMPRESS_ONLY::true"
+    - name: Checkout branch
+      uses: actions/checkout@v2.3.2
+    - name: calibreapp/image-actions
```
ibnesayeed, line 51: Choose a better name for the step.

.github/workflows/calibreapp-image-action.yml
```diff
@@ -4,35 +4,67 @@
...
+      with:
+        githubToken: ${{ secrets.GITHUB_TOKEN }}
+        compressOnly: ${{ env.COMPRESS_ONLY }}
+        jpegProgressive: true
+    - name: Create Pull Request
+      # If it's not a pull request then commit any changes as a new PR
+      if: |
+        github.event_name != 'pull_request' &&
+        steps.calibre.outputs.markdown != ''
+      id: cpr
```
ibnesayeed, line 63: Do we use this `id` anywhere?

.github/workflows/calibreapp-image-action.yml
```diff
@@ -4,35 +4,67 @@
...
+    - name: Set COMPRESS_ONLY variable to true
+      # If it's not a pull request then run in compressOnly mode
+      if: github.event_name != 'pull_request'
+      run: |
+        echo "::set-env name=COMPRESS_ONLY::true"
+    - name: Checkout branch
+      uses: actions/checkout@v2.3.2
+    - name: calibreapp/image-actions
+      id: calibre
+      uses: calibreapp/image-actions@master
```
ibnesayeed, line 53: Version pin the action.

ibnesayeed (changes requested):

rviscomi:
Not excited about more auto-generated PRs to be honest! But I understand that we're limited in how much we can do in the manual PRs themselves. 😅

.github/workflows/calibreapp-image-action.yml
bazzadp: Done

.github/workflows/calibreapp-image-action.yml
bazzadp: Clever! Almost bordering on too clever at the cost of readability but think it's just this side of it so let's do it.

.github/workflows/calibreapp-image-action.yml
bazzadp: Done.

.github/workflows/calibreapp-image-action.yml
bazzadp: We do now :-)

.github/workflows/calibreapp-image-action.yml
bazzadp: This one I'm going to push back on. Unfortunately this action doesn't using versioning and pushes everything immediately to a Docker build that's used no matter what version number is used. So going to leave it off here so we don't assume it is versioned.

.github/workflows/calibreapp-image-action.yml
ibnesayeed: I would have not suggested it, but I though that too many things were done up front just to set this one variable. :-)

.github/workflows/calibreapp-image-action.yml
```diff
@@ -4,35 +4,64 @@
...
+        id: cpr
+        uses: peter-evans/create-pull-request@v3.3.0
+        with:
+          title: Auto Compress Images
+          branch-suffix: timestamp
+          commit-message: Compress Images
+          body: ${{ steps.calibre.outputs.markdown }}
+          labels: performance
+      - name: Check outputs
+        run: echo "Pull Request Number - ${{ steps.cpr.outputs.pull-request-number }}"
```
ibnesayeed, line 67: The step this relies on is conditional while this one is not.

.github/workflows/calibreapp-image-action.yml
ibnesayeed: You are right, the last time they released anything was in Feb!

bazzadp:
> Not excited about more auto-generated PRs to be honest! But I understand that we're limited in how much we can do in the manual PRs themselves. 😅

Yeah the workflow for auto updating in-repo pull requests is nice but pity they haven't figured out how to do this for forks yet. This is the best alternative we can get I think because the alternative (uncompressed images, or manually compressing every so often) is worse.

Will need to see how annoying the weekly run gets. In theory it shouldn't pick up anything after first run or two and is just there as a fail safe, but in practice it might compress things slightly differently and crate noise. Noise that leads to better performance though! But noise none-the-less. We can turn it off if it gets too annoying.

.github/workflows/calibreapp-image-action.yml
bazzadp: Good spot - fixed!

.github/workflows/calibreapp-image-action.yml
```diff
@@ -4,35 +4,69 @@
...
-    if: github.event.pull_request.head.repo.full_name == github.repository
     name: calibreapp/image-actions
     runs-on: ubuntu-latest
+    # Only run on main repo on and PRs that match the main repo
+    if: |
+      (github.event_name != 'pull_request' &&
+       github.repository == 'HTTPArchive/almanac.httparchive.org')
+      ||
+      (github.event_name == 'pull_request' &&
+       github.event.pull_request.head.repo.full_name == github.repository)
```
bazzadp, line 45: @ibnesayeed can you think of a more elegant way of writing this?

bazzadp: Went with this instead:

```yml
    # Only run on main repo on and PRs that match the main repo
    if: github.repository == 'HTTPArchive/almanac.httparchive.org' &&
      (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
```

.github/workflows/calibreapp-image-action.yml
```diff
@@ -4,35 +4,67 @@
...
+        uses: peter-evans/create-pull-request@v3.3.0
+        with:
+          title: Auto Compress Images
+          branch-suffix: timestamp
+          commit-message: Compress Images
+          body: ${{ steps.calibre.outputs.markdown }}
+          labels: performance
+      - name: Check outputs
+        # If it's not a pull request then commit any changes as a new PR
+        if: steps.cpr.outputs.pull-request-number != ''
```
ibnesayeed, line 69: I think we can do something like below and will not need the `id` property in  the previous step.

```suggestion
        if: ${{ success() }}
```

.github/workflows/calibreapp-image-action.yml
ibnesayeed: I guess, they ought to provide better means like shorthand properties to distinguish common patters to skip jobs or tasks. ;-)

.github/workflows/calibreapp-image-action.yml
bazzadp: I'd prefer to leave it as is. If there is nothing to commit the previous step completes successfully. Of course that shouldn't happen in this case since we only call the previous step if there is something, but the principal still applies - we only want to show this step if there is a PR number which is what the current if statement does.

In fact this whole if statement is a bit superfluous anyway since in other actions we just print it out even if it's not been set but probably better to check it's there.

ibnesayeed (approved):
LGTM!

PR #1301, opened 2020-09-19
Add DISTINCT to ServiceWorker queries

bazzadp:
There's some duplication in the Service Workers table. This adds a DISTINCT for next year - though no need to rerun for this year.

### DIFF
--- a/sql/util/manifests.sql
+++ b/sql/util/manifests.sql
@@ -1,5 +1,5 @@
-SELECT DISTINCT
-  date,
+SELECT
+  DISTINCT date,
   client,
   page,
   url,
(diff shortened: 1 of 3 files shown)


sql/util/manifests.sql
```diff
@@ -1,4 +1,4 @@
-SELECT
+SELECT DISTINCT
   date,
```
rviscomi, line 2: formatting nit
```suggestion
SELECT
  DISTINCT date,
```

rviscomi (approved):

sql/util/manifests.sql
bazzadp: Fixed in all 3 queries

PR #1319, opened 2020-09-27
Translate featured_chapters to Portuguese

HakaCode:
Makes progress on #505


bazzadp (approved):
LGTM! Thanks.

As you translate each's chapter you should check you've translated it consistently and also enable the chapter in line 1 of this file.

HakaCode:
> LGTM! Thanks.
> 
> As you translate each's chapter you should check you've translated it consistently and also enable the chapter in line 1 of this file.

I understood, I ended up seeing how the website would look with the new language "Portuguese", and I understood the logic used, thanks for the help with the tips.

PR #1331, opened 2020-09-30
Fixes as I gathered the real data

Tiggerito:
A strange issue where the query could not coerce a string array.

Some more defaulting values to zero.

Corrections in some JavaScript json references

Better sorting

Changed the video percentile query to just include pages with videos so that the data had some value.

### DIFF
--- a/sql/2020/07_SEO/pages_wpt_bodies_hreflang_by_device_and_http_header_value.sql
+++ b/sql/2020/07_SEO/pages_wpt_bodies_hreflang_by_device_and_http_header_value.sql
@@ -11,7 +11,9 @@ CREATE TEMPORARY FUNCTION get_wpt_bodies_info(wpt_bodies_string STRING)
 RETURNS STRUCT<
     hreflangs ARRAY<STRING>
 > LANGUAGE js AS '''
-var result = {};
+var result = {
+  hreflangs: []
+};
 
 try {
     var wpt_bodies = JSON.parse(wpt_bodies_string);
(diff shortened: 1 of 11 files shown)


sql/2020/07_SEO/pages_wpt_bodies_hreflang_by_device_and_http_header_value.sql
```diff
@@ -11,15 +11,17 @@ CREATE TEMPORARY FUNCTION get_wpt_bodies_info(wpt_bodies_string STRING)
...
+};
 
 try {
     var wpt_bodies = JSON.parse(wpt_bodies_string);
 
     if (Array.isArray(wpt_bodies) || typeof wpt_bodies != 'object') return result;
 
     if (wpt_bodies.hreflangs && wpt_bodies.hreflangs.http_header && wpt_bodies.hreflangs.http_header.values) {
-        result.hreflangs = wpt_bodies.hreflangs.http_header.values;
+        result.hreflangs = wpt_bodies.hreflangs.http_header.values.map(v=> v); // seems to fix a coercion issue!
```
rviscomi, line 24: 🤨

rviscomi (approved):

sql/2020/07_SEO/pages_wpt_bodies_hreflang_by_device_and_http_header_value.sql
Tiggerito: Strange, it's as if the array from the map was better than the original array. I tested that values was always an array of string?

rviscomi: Strange indeed, but whatever works

PR #1338, opened 2020-10-04
Check if audit is applicable

obto:
Quick update to the LH metric to check if the audit was applicable to a site or not. Only applicable sites should be counted towards the total, or you end up with incorrect %'s of sites meeting each audit.

### DIFF
--- a/sql/2020/14_PWA/lighthouse_pwa_audits.sql
+++ b/sql/2020/14_PWA/lighthouse_pwa_audits.sql
@@ -24,7 +24,7 @@ return results;
 SELECT
   audits.id AS id,
   COUNTIF(audits.score > 0) AS num_pages,
-  count(0) AS total,
+  COUNT(0) AS total,
   COUNTIF(audits.score IS NOT NULL) AS total_applicable,
   SAFE_DIVIDE(COUNTIF(audits.score > 0), COUNTIF(audits.score IS NOT NULL)) AS pct,
   APPROX_QUANTILES(audits.weight, 100)[OFFSET(50)] AS median_weight,
--- a/sql/2020/14_PWA/lighthouse_pwa_audits.sql
+++ b/sql/2020/14_PWA/lighthouse_pwa_audits.sql
@@ -24,8 +24,8 @@ return results;
 SELECT
   audits.id AS id,
   COUNTIF(audits.score > 0) AS num_pages,
-  COUNT(0) AS total,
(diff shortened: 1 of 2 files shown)


sql/2020/14_PWA/lighthouse_pwa_audits.sql
```diff
@@ -24,8 +24,8 @@ return results;
 SELECT
   audits.id AS id,
   COUNTIF(audits.score > 0) AS num_pages,
-  COUNT(0) AS total,
-  COUNTIF(audits.score > 0) / COUNT(0) AS pct,
+  COUNTIF(audits.score IS NOT NULL) AS total_applicable,
```
bazzadp, line 27:
```suggestion
  count(0) AS total,
  COUNTIF(audits.score IS NOT NULL) AS total_applicable,
```

bazzadp (changes requested):
Thanks - good spot. And just [after I'd tweeted out about this](https://twitter.com/tunetheweb/status/1312774697520697345?s=20) :-(

One suggestion is to keep total number of pages. This is good for identifying LH5 scores (some still creeping in there even in September).

sql/2020/14_PWA/lighthouse_pwa_audits.sql
bazzadp: Gentle ping @obto !

obto: LGTM. Added to the PR

sql/2020/14_PWA/lighthouse_pwa_audits.sql
```diff
@@ -24,8 +24,9 @@ return results;
 SELECT
   audits.id AS id,
   COUNTIF(audits.score > 0) AS num_pages,
-  COUNT(0) AS total,
-  COUNTIF(audits.score > 0) / COUNT(0) AS pct,
+  count(0) AS total,
```
bazzadp, line 27:
```suggestion
  COUNT(0) AS total,
```

bazzadp (approved):
Small capitalisation issue but will change that and merge.

I was more letting you know this in case you wanted to change https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2020/08_Accessibility/lighthouse_a11y_audits.sql

Will add to this PR and merge.

PR #1339, opened 2020-10-04
Fix incorrectly escaped symbols

bazzadp:


rviscomi:
@bazzadp can you clarify when HTML characters need to be escaped? If we want `<head>` to appear as a code block, for example, it seems like we should use < and > characters around both `code` and `head` tags. Is that right?

bazzadp:
Yeah need to do some more digging here. I'm surprised that we don't need to escape and looks like Jinja does convert this:

```
<code><head></code>
```

To this:
```
<code>&lt;head&gt;</code>
```

Which really surprises me. You'd think it would either escape all or none!

Anyway noticed our chapters are wrong at present so thought best to fix.

Will dig some more...

PR #1344, opened 2020-10-05
Http2 query fixes

gregorywolf:
Progress on #921


sql/2020/22_HTTP_2/avg_percentage_of_resources_loaded_over_HTTP_by_version_per_site.sql
```diff
@@ -0,0 +1,29 @@
...
+  FROM 
+    `httparchive.almanac.requests`
+  WHERE
+    date = '2020-09-01'
+  GROUP BY
+    client,
+    page
+)
+GROUP BY 
+  client
```
rviscomi, line 29: Minor reorganization to make the protocol variable and reshape the results, also adding an "other" bucket to account for everything else.

```suggestion
# standardSQL
# Average percentage of resources loaded over HTTP .9, 1, 1.1, 2, and QUIC per site
SELECT
  client,
  protocol,
  AVG(num_requests / total_requests) AS avg_pct_requests_per_page
FROM (
  SELECT 
    client,
    page,
    IF(JSON_EXTRACT_SCALAR(payload, '$._protocol') IN ('http/0.9', 'http/1.0', 'http/1.1', 'HTTP/2', 'QUIC'), JSON_EXTRACT_SCALAR(payload, '$._protocol'), 'other') AS protocol,
    COUNT(0) AS num_requests,
    SUM(COUNT(0)) OVER (PARTITION BY client, page) AS total_requests
  FROM 
    `httparchive.almanac.requests`
  WHERE
    date = '2020-09-01'
  GROUP BY
    client,
    page,
    protocol)
GROUP BY 
  client,
  protocol
ORDER BY
  avg_pct_requests_per_page DESC
```

sql/2020/22_HTTP_2/tls_adoption_by_http_version.sql
```diff
@@ -1,22 +1,36 @@
...
+    MAX(IFNULL(JSON_EXTRACT_SCALAR(payload, '$._tls_version'), JSON_EXTRACT_SCALAR(payload, '$._securityDetails.protocol'))) AS tls_version
+  FROM
+    `httparchive.almanac.requests`
+  WHERE
+    date = '2020-09-01' AND
+    url like 'https://%'
+  GROUP BY
     client,
-    JSON_EXTRACT_SCALAR(payload, '$._protocol') as http_version,
-    JSON_EXTRACT_SCALAR(payload, '$._tls_version') AS tls_version
```
rviscomi, line 14: @gregorywolf can you clarify what needed to be fixed/improved with this old query?

One thing I see is falling back from `tls_version` to `securityDetails.protocol`. Curious about the motivation for everything else.

rviscomi (changes requested):

sql/2020/22_HTTP_2/avg_percentage_of_resources_loaded_over_HTTP_by_version_per_site.sql
```diff
@@ -0,0 +1,29 @@
...
+FROM (
+  SELECT 
+    client,
+    page,
+    COUNT(0) AS num_requests, 
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="http/0.9",1,0)) AS http_0_9, 
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="http/1.0",1,0)) AS http_1_0, 
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="http/1.1",1,0)) AS http_1_1,
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="HTTP/2",1,0)) AS http_2,
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="QUIC",1,0)) AS quic
```
bazzadp, line 19: We had a better check for HTTP/3 / QUIC:

```suggestion
    SUM(IF(REGEXP_CONTAINS(JSON_EXTRACT_SCALAR(payload, "$._protocol"), r"(?i)(quic|^h3|^http/3)"),1,0)) AS quic    
```

sql/2020/22_HTTP_2/avg_percentage_of_resources_loaded_over_HTTP_by_version_per_site.sql
```diff
@@ -0,0 +1,29 @@
...
+    COUNT(0) AS num_requests, 
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="http/0.9",1,0)) AS http_0_9, 
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="http/1.0",1,0)) AS http_1_0, 
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="http/1.1",1,0)) AS http_1_1,
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="HTTP/2",1,0)) AS http_2,
+    SUM(IF(JSON_EXTRACT_SCALAR(payload, "$._protocol") ="QUIC",1,0)) AS quic    
+  FROM 
+    `httparchive.almanac.requests`
+  WHERE
+    date = '2020-09-01'
```
bazzadp, line 23: Why September?

```suggestion
    date = '2020-08-01'
```

sql/2020/22_HTTP_2/tls_adoption_by_http_version.sql
```diff
@@ -1,22 +1,36 @@
...
+FROM
+  `httparchive.almanac.requests` req,
+    (SELECT
+    client,
+    NET.HOST(url) AS domain,
+    MAX(IFNULL(JSON_EXTRACT_SCALAR(payload, '$._tls_version'), JSON_EXTRACT_SCALAR(payload, '$._securityDetails.protocol'))) AS tls_version
+  FROM
+    `httparchive.almanac.requests`
+  WHERE
+    date = '2020-09-01' AND
```
bazzadp, line 19: Again why September?

```suggestion
    date = '2020-08-01' AND
```

bazzadp (changes requested):

sql/2020/22_HTTP_2/tls_adoption_by_http_version.sql
bazzadp: @rviscomi I worked with @gregorywolf on this one (and also pulled in @tomvangoethem for Security). We've noticed the vast majority of request do NOT have a TLS Version set.

Run this query for example:

```sql
# standardSQL
#standardSQL
SELECT
  client,
  JSON_EXTRACT_SCALAR(payload, '$._protocol') AS protocol,
  JSON_EXTRACT_SCALAR(payload, '$._tls_version') AS tls_version,
  JSON_EXTRACT_SCALAR(payload, '$.securityDetails.protocol') AS tls_version2,
  COUNT(0) AS freq,
  COUNTIF(firstHtml) AS freq_firstHtml,
  COUNTIF(firstReq) AS freq_firstReq
FROM
  `httparchive.sample_data.requests`
WHERE
  date = '2020-08-01' AND
  url like 'https://%'
GROUP BY
  client,
  protocol,
  tls_version,
  tls_version2
ORDER BY
  freq desc,
  client,
  protocol
```

and you get this:

TLS version

I suspect that the TLS version is only set for the first request, and working with @tomvangoethem we saw that those firstHTML and firstReq certainly had a lot less missing TLS Versions (but not 0) as can be seen with above table.

Now we could just ignore all this and just measure the non-null versions. If they are due to first requests then we'll pick it up for the first request. That would work though there's questions as to what you're measuring. If you have 20 requests over a site that uses one domain (www.example.com) and try to compare it to a similar site that uses a shared set of domains (www.example.com, assets1.example.com, assets2.example.com and assets3.example.com) and counts that 4 times? Or should we measure just home pages and first request? It really depends if you want to measure by site or by request.

So I suggested we keep it at request level, but run a subquery to figure out the max TLS for each domain and then use that for the request. That seems to have eradicated most of the missing TLS values. There's still some for HTTP/2 and QUIC that I suspect might be due to connection coalescing where the first connection is not with the same domain. Anyway small minority.

WDYT?

BTW @tomvangoethem is waiting to make the same change in the Security chapter depending on what we agree here.

sql/2020/22_HTTP_2/avg_percentage_of_resources_loaded_over_HTTP_by_version_per_site.sql
rviscomi: I've added the results for this new query to the sheet: https://docs.google.com/spreadsheets/d/1M1tijxf04wSN3KU0ZUunjPYCrVsaJfxzuRCXUeRQ-YU/edit#gid=119448240

sql/2020/22_HTTP_2/tls_adoption_by_http_version.sql
bazzadp: Oh and we also checked and a very small number (190 out of 46,224 domains in the sample set) have multiple TLS versions per domain. I've a few theories on that: changes through the month (the crawl takes over a month), or distributed load balancers, or connection coalescing. Still think taking the MAX is a pretty good proxy for what is supported by that domain and the differences are small.

rviscomi: Does this do what you want?

```sql
# standardSQL
# Distribution of TLS versions
SELECT
  client,
  tls_version,
  COUNT(0) AS freq,
  ROUND(COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client), 4) pct
FROM (
  SELECT
    client,
    page,
    MAX(IFNULL(JSON_EXTRACT_SCALAR(payload, '$._tls_version'), JSON_EXTRACT_SCALAR(payload, '$._securityDetails.protocol'))) AS tls_version
  FROM
  WHERE
    date = '2020-08-01' AND
    STARTS_WITH(url, 'https') AND
    NET.REG_DOMAIN(url) = NET.REG_DOMAIN(page)
  GROUP BY
    client,
    page)
GROUP BY
  client,
  tls_version
ORDER BY
  freq / total DESC
```


client | tls_version | freq | total | pct
-- | -- | -- | -- | --
desktop | TLS 1.2 | 2,458,172 | 4,502,570 | 54.59%
mobile | TLS 1.2 | 2,576,607 | 4,921,113 | 52.36%
mobile | TLS 1.3 | 2,336,815 | 4,921,113 | 47.49%
desktop | TLS 1.3 | 2,035,713 | 4,502,570 | 45.21%
desktop | TLS 1.0 | 8,192 | 4,502,570 | 0.18%
mobile | TLS 1.0 | 7,211 | 4,921,113 | 0.15%
desktop |   | 403 | 4,502,570 | 0.01%
mobile |   | 410 | 4,921,113 | 0.01%
desktop | TLS 1.1 | 90 | 4,502,570 | 0.00%
mobile | TLS 1.1 | 69 | 4,921,113 | 0.00%
mobile | QUIC | 1 | 4,921,113 | 0.00%

rviscomi: I'm a bit skeptical of the use of `MAX` here, as it's being done on strings and any use of QUIC would be superseded by TLS.

tomvangoethem: > Does this do what you want?

I don't think so; this gets the TLS usage of the pages, not of the requests (speaking for security chapter & based on the proposed query).

> I'm a bit skeptical of the use of MAX here, as it's being done on strings and any use of QUIC would be superseded by TLS.

The idea would be that a single host would only have a single TLS version; for the sample data that's largely true (with the exceptions of 190 out of 46,224 hosts). For the full dataset that may indeed be different (more connections -> potentially more chances that QUIC was used), getting the most common value might be more representative? Not entirely sure how do this cleanly though

tomvangoethem: one way of getting the most common value instead of using `MAX` (based on the @rviscomi's query):
```sql
# standardSQL
# Distribution of TLS versions
CREATE TEMP FUNCTION getMostCommonValue(arr ARRAY<STRING>) AS (
  (SELECT x FROM UNNEST(arr) as x GROUP BY x ORDER BY COUNT(0) DESC LIMIT 1)
);

SELECT
  client,
  tls_version,
  COUNT(0) AS freq,
  ROUND(COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client), 4) pct
FROM (
  SELECT
    client,
    page,
    getMostCommonValue(ARRAY_AGG(IFNULL(JSON_EXTRACT_SCALAR(payload, '$._tls_version'), JSON_EXTRACT_SCALAR(payload, '$._securityDetails.protocol')) IGNORE NULLS)) AS tls_version
  FROM
    `httparchive.sample_data.requests`
  WHERE
    date = '2020-08-01' AND
    STARTS_WITH(url, 'https') AND
    NET.REG_DOMAIN(url) = NET.REG_DOMAIN(page)
  GROUP BY
    client,
    page)
GROUP BY
  client,
  tls_version
ORDER BY
  freq / total DESC
```

rviscomi: I don't see any valid way to attribute requests from one TLS version to a different TLS version. Having the number of requests as the numerator and denominator in your metrics is less helpful because pages that make many requests are overrepresented.

Here's a different approach that counts the number of pages having at least one request of the given TLS version:

```sql
# standardSQL
# Distribution of TLS versions
SELECT
  client,
  tls_version,
  COUNT(DISTINCT page) AS freq,
  total,
  COUNT(DISTINCT page) / total AS pct
FROM (
  SELECT
    client,
    page,
    IFNULL(JSON_EXTRACT_SCALAR(payload, '$._tls_version'), JSON_EXTRACT_SCALAR(payload, '$._securityDetails.protocol')) AS tls_version
  FROM
  WHERE
    date = '2020-08-01' AND
    STARTS_WITH(url, 'https') AND
    NET.REG_DOMAIN(url) = NET.REG_DOMAIN(page))
JOIN (
  SELECT
    _TABLE_SUFFIX AS client,
    COUNT(0) AS total
  FROM
    `httparchive.summary_pages.2020_08_01_*`
  GROUP BY
    client)
USING
  (client)
WHERE
  tls_version IS NOT NULL
GROUP BY
  client,
  tls_version,
  total
ORDER BY
  pct DESC
```


client | tls_version | freq | total | pct
-- | -- | -- | -- | --
desktop | TLS 1.2 | 2,571,656 | 5,593,642 | 45.97%
mobile | TLS 1.2 | 2,693,314 | 6,347,919 | 42.43%
mobile | TLS 1.3 | 2,336,815 | 6,347,919 | 36.81%
desktop | TLS 1.3 | 2,035,713 | 5,593,642 | 36.39%
mobile | QUIC | 41,494 | 6,347,919 | 0.65%
desktop | QUIC | 34,114 | 5,593,642 | 0.61%
desktop | TLS 1.0 | 9,323 | 5,593,642 | 0.17%
mobile | TLS 1.0 | 8,224 | 6,347,919 | 0.13%
desktop | TLS 1.1 | 109 | 5,593,642 | 0.00%
mobile | TLS 1.1 | 86 | 6,347,919 | 0.00%

tomvangoethem: Agreed that pages that send a lot of requests would bias results based on all requests. But instead of the pages, shouldn't we look at TLS usage of the request hosts? By just looking at the pages, we'd miss all the subresources. So for instance, something like this:

```sql
# standardSQL
# Distribution of TLS versions
CREATE TEMP FUNCTION getMostCommonValue(arr ARRAY<STRING>) AS (
  (SELECT x FROM UNNEST(arr) as x GROUP BY x ORDER BY COUNT(0) DESC LIMIT 1)
);

SELECT
  client,
  tls_version,
  COUNT(0) AS freq,
FROM (
  SELECT
    client,
    NET.HOST(url) AS host,
    getMostCommonValue(ARRAY_AGG(IFNULL(JSON_EXTRACT_SCALAR(payload, '$._tls_version'), JSON_EXTRACT_SCALAR(payload, '$._securityDetails.protocol')) IGNORE NULLS)) AS tls_version
  FROM
    `httparchive.sample_data.requests`
  WHERE
    date = '2020-08-01' AND
    STARTS_WITH(url, 'https')
  GROUP BY
    client,
    host)
WHERE tls_version IS NOT NULL
GROUP BY
  client,
  tls_version
ORDER BY
  pct DESC
```

client | tls_version | freq | total | pct
-- | -- | -- | -- | --
desktop_10k | TLS 1.2 | 13629 | 21574 | 0.6317326411421155
mobile_10k | TLS 1.2 | 12212 | 20274 | 0.6023478346650883
mobile_10k | TLS 1.3 | 8023 | 20274 | 0.39572851928578473
desktop_10k | TLS 1.3 | 7916 | 21574 | 0.36692314823398536
desktop_10k | TLS 1.0 | 24 | 21574 | 0.0011124501715027348
mobile_10k | TLS 1.0 | 21 | 20274 | 0.0010358094110683633
mobile_10k | QUIC | 18 | 20274 | 8.878366380585972E-4
desktop_10k | QUIC | 4 | 21574 | 1.8540836191712246E-4
desktop_10k | TLS 1.1 | 1 | 21574 | 4.6352090479280616E-5

It seems that these results are not deterministic; probably because of the cases where different TLS versions have been observed the same number of times

Perhaps it would then be better to include all TLS versions of a host:

```sql
# standardSQL
# Distribution of TLS versions
SELECT
  client,
  tls_version,
  COUNT(0) AS freq,
FROM (
  SELECT
    client,
    NET.HOST(url) AS host,
    ARRAY_AGG(DISTINCT IFNULL(JSON_EXTRACT_SCALAR(payload, '$._tls_version'), JSON_EXTRACT_SCALAR(payload, '$._securityDetails.protocol')) IGNORE NULLS) AS tls_versions
  FROM
    `httparchive.sample_data.requests`
  WHERE
    date = '2020-08-01' AND
    STARTS_WITH(url, 'https')
  GROUP BY
    client,
    host),
  UNNEST (tls_versions) AS tls_version
GROUP BY
  client,
  tls_version
ORDER BY
  pct DESC
```
client | tls_version | freq | total | pct
-- | -- | -- | -- | --
desktop_10k | TLS 1.2 | 13679 | 21840 | 0.6263278388278388
mobile_10k | TLS 1.2 | 12290 | 20549 | 0.5980826317582364
mobile_10k | TLS 1.3 | 8111 | 20549 | 0.394715071293007
desktop_10k | TLS 1.3 | 8024 | 21840 | 0.3673992673992674
mobile_10k | QUIC | 127 | 20549 | 0.0061803494087303516
desktop_10k | QUIC | 112 | 21840 | 0.005128205128205128
desktop_10k | TLS 1.0 | 24 | 21840 | 0.001098901098901099
mobile_10k | TLS 1.0 | 21 | 20549 | 0.0010219475400262787
desktop_10k | TLS 1.1 | 1 | 21840 | 4.578754578754579E-5

bazzadp: Hmmm the more I think about this the more I think @rviscomi 's right and we should just look at the number of sites that support each version of TLS.

Looking at the number of requests (rather than number of sites) sounds sensible, as a proxy for measuring traffic, but it's not really because our stats are not weighted so each site gets equal weight, whether they get a lot of traffic (think Google, Facebook, Twitter...etc.), or only a little (think barrysawesomesite.com) so if you really want to look at requests we need to look at weighted requests – which we can't do.

So maybe we should just look at `firstHTML` and when we run that the vast majority of those pages do have a TLS version (unsurprisingly really as would expect those to be the first request as well which is the one I suspect has the TLS version set).

So for HTTP/2 query we should add `WHERE firstHTML` and remove all that complexity I just added, and for Security (which already had this) you don't need to make any changes and I've just completely wasted your time! Sorry! But interesting discussion I think...

What do you all think?

tomvangoethem: For the Security chapter there are two queries for the TLS version, [one for the pages](https://github.com/HTTPArchive/almanac.httparchive.org/blob/security-sql-2020/sql/2020/11_Security/tls_versions_pages.sql) (which has `firstHtml`) and [one for the requests](https://github.com/HTTPArchive/almanac.httparchive.org/blob/security-sql-2020/sql/2020/11_Security/tls_versions_requests.sql) (which doesn't).

I would opt to keep the one for the pages, and either adjust the one for the requests to group by host (I think this makes most sense - "of all the web servers, XX% support TLS 1.3"), or get the TLS version of all the requests (as in [last year's edition](https://almanac.httparchive.org/en/2019/security#protocol-versions)), or just drop it

rviscomi: Something like this?

```sql
# standardSQL
# Distribution of TLS versions
SELECT
  client,
  tls_version,
  COUNT(DISTINCT page) AS freq,
  total,
  COUNT(DISTINCT page) / total AS pct
FROM (
  SELECT
    client,
    page,
    IFNULL(JSON_EXTRACT_SCALAR(payload, '$._tls_version'), JSON_EXTRACT_SCALAR(payload, '$._securityDetails.protocol')) AS tls_version
  FROM
  WHERE
    date = '2020-08-01' AND
    STARTS_WITH(url, 'https') AND
    firstHtml)
JOIN (
  SELECT
    _TABLE_SUFFIX AS client,
    COUNT(0) AS total
  FROM
    `httparchive.summary_pages.2020_08_01_*`
  GROUP BY
    client)
USING
  (client)
GROUP BY
  client,
  tls_version,
  total
ORDER BY
  pct DESC
```

(counts null TLS versions and only looks at the firstHtml response)


client | tls_version | freq | total | pct
-- | -- | -- | -- | --
desktop | TLS 1.2 | 2,172,718 | 5,593,642 | 38.84%
mobile | TLS 1.2 | 2,272,091 | 6,347,919 | 35.79%
desktop | TLS 1.3 | 1,793,182 | 5,593,642 | 32.06%
mobile | TLS 1.3 | 1,928,250 | 6,347,919 | 30.38%
mobile |   | 536,517 | 6,347,919 | 8.45%
desktop |   | 398,528 | 5,593,642 | 7.12%
desktop | TLS 1.0 | 6,153 | 5,593,642 | 0.11%
mobile | TLS 1.0 | 5,376 | 6,347,919 | 0.08%
mobile | QUIC | 207 | 6,347,919 | 0.00%
desktop | QUIC | 138 | 5,593,642 | 0.00%
desktop | TLS 1.1 | 62 | 5,593,642 | 0.00%
mobile | TLS 1.1 | 42 | 6,347,919 | 0.00%

bazzadp: SGTM

bazzadp: Should your summary_pages request also only look at https urls?

bazzadp: Are at least add the scheme so we can see if those large null percentages are over http?

rviscomi: Updated so the denominator is only HTTPS pages. @gregorywolf please update the `tls_adoption` query with this code and commit. The results in the sheet will match this query.

```sql
# standardSQL
# Distribution of TLS versions
SELECT
  client,
  tls_version,
  COUNT(DISTINCT page) AS freq,
  total,
  COUNT(DISTINCT page) / total AS pct
FROM (
  SELECT
    client,
    page,
    IFNULL(JSON_EXTRACT_SCALAR(payload, '$._tls_version'), JSON_EXTRACT_SCALAR(payload, '$._securityDetails.protocol')) AS tls_version
  FROM
  WHERE
    date = '2020-08-01' AND
    STARTS_WITH(url, 'https') AND
    firstHtml)
JOIN (
  SELECT
    _TABLE_SUFFIX AS client,
    COUNT(0) AS total
  FROM
    `httparchive.summary_pages.2020_08_01_*`
  WHERE
    STARTS_WITH(url, 'https')
  GROUP BY
    client)
USING
  (client)
GROUP BY
  client,
  tls_version,
  total
ORDER BY
  pct DESC
```

gregorywolf:
@rviscomi @bazzadp Pretty sure that all of the suggested changes have been addressed

sql/2020/22_HTTP_2/tls_adoption_by_http_version.sql
rviscomi: Doesn't look like this query was updated

rviscomi (changes requested):

sql/2020/22_HTTP_2/tls_adoption_by_http_version.sql
bazzadp: @rviscomi you forgot the HTTP version!

@gregorywolf I believe below is the query you want to commit. I've run this against the full table and saved the results in the sheet. Note I've also updated the protocol query to also include `'http/2+quic/46'` and (for future proofing) `'HTTP/3'`. I think we should do the same to your other query in this PR that also looks for HTTP version.

```sql
# standardSQL
# Distribution of TLS versions
SELECT
  client,
  protocol,
  tls_version,
  COUNT(DISTINCT page) AS freq,
  total,
  COUNT(DISTINCT page) / total AS pct
FROM (
  SELECT
    client,
    page,
    IF(JSON_EXTRACT_SCALAR(payload, '$._protocol') IN ('http/0.9', 'http/1.0', 'http/1.1', 'HTTP/2', 'QUIC', 'http/2+quic/46', 'HTTP/3'), JSON_EXTRACT_SCALAR(payload, '$._protocol'), 'other') AS protocol,
    IFNULL(JSON_EXTRACT_SCALAR(payload, '$._tls_version'), JSON_EXTRACT_SCALAR(payload, '$._securityDetails.protocol')) AS tls_version
  FROM
  WHERE
    date = '2020-08-01' AND
    STARTS_WITH(url, 'https') AND
    firstHtml)
JOIN (
  SELECT
    _TABLE_SUFFIX AS client,
    COUNT(0) AS total
  FROM
    `httparchive.summary_pages.2020_08_01_*`
  WHERE
    STARTS_WITH(url, 'https')
  GROUP BY
    client)
USING
  (client)
GROUP BY
  client,
  protocol,
  tls_version,
  total
ORDER BY
  pct DESC
```

bazzadp (approved):

gregorywolf:
ok. All changes are now made. Sorry for the confusion!!!!!

bazzadp:
As discussed ran the other query too (with the two extra protocols) and saved to the Sheet.

You happy to merge this now @rviscomi ?

rviscomi (approved):

PR #1358, opened 2020-10-12
2020 graphics

bazzadp:
This adds the three new hero images prepared by our designer in appropriate format:

- Capabilities
- JAMStack
- Privacy

It also moves all the hero images from `src/static/images/2019/` folder to a new `src/static/images/hero-images` folder and adds redirects in case the old locations are references. This is done these hero images are shared between 2019 and 2020 chapters and other chapter images are not so would no longer be able to use the `chapter_image_dir` both both this and figures. Also three of the chapters do not exist in 2019 so would have involved creating a folder for those.

Finally I've also added the PSD files. These are quite large - 750MB which doubles the size of our repo 😔 What do you think @rviscomi ? I think they should be in Git and not in a Google Drive that's easy to lose hold of, but am concerned it dramatically increases our repo size.


rviscomi:
> Finally I've also added the PSD files. These are quite large - 750MB which doubles the size of our repo 😔 What do you think @rviscomi ? I think they should be in Git and not in a Google Drive that's easy to lose hold of, but am concerned it dramatically increases our repo size.

PSDs are technically "source" files so keeping them in "source control" makes sense, but these are just prohibitively big and they would only be relevant to designers. We should try to keep our repo footprint small. I just did a `git pull` in another branch and had to wait a minute while this branch synced with my local copy. I'd suggest not having them.

bazzadp:
> PSDs are technically "source" files so keeping them in "source control" makes sense, but these are just prohibitively big and they would only be relevant to designers. We should try to keep our repo footprint small. I just did a `git pull` in another branch and had to wait a minute while this branch synced with my local copy. I'd suggest not having them.

And it's gone! Hopefully with a squash commit and deleting og this branch there's no permanent damage done.

src/server/routes.py
```diff
@@ -115,7 +115,13 @@ def ebook(lang, year):
     return render_template('%s/%s/ebook.html' % (lang, year), config=config)
 
 
-# Redirect requests for the older image URLs to new URLs
+# Redirect requests for the older hero image URLs to new URLs
+@app.route('/static/images/2019/<folder>/<regex("hero.*"):image>')
+def redirect_hero_images(folder, image):
+    return redirect("/static/images/hero-images/%s/%s" % (folder, image)), 301
```
rviscomi, line 121: My only concern with this is if we want to update a hero image for a chapter. We shouldn't necessarily overwrite and lose all older images. Having them versioned by year helps with that. If a 2020 chapter is reusing its 2019 image, it'd be ok to reference 2019 in the image file path rather than duplicate all images into 2019 and 2020 directories. WDYT?

bazzadp: Yeah I did consider that too and can change to do that (would also make this PR a lot smaller as less file renames and no python changes!) however felt a little weird for the following reasons:
- The 2019 chapter images folder would contain the hero image and the figures, but for 2020 it would be figures only.
- We'd need to add 2019 chapter image folders for the three new chapters which just feels a little icky.
- We'd still need to change the hero image folders to stop them using the `chapter_image_dir` variable (at least for 2020 chapters)

But we are reusing the 2019 design, and reference that in other places (e.g. 2019.css) so maybe that's OK.

BTW there are other ways to handle new versions of hero images (e.g. `hero_xl_v2.png`, or even put them in a new `hero-images-v2` folder).

Anyway, not strongly wedded to the current implementation, so if you still think better to have the hero images under the 2019 folder then happy to change it to that.

rviscomi: > Anyway, not strongly wedded to the current implementation, so if you still think better to have the hero images under the 2019 folder then happy to change it to that.

+1 thank you! I hope you didn't spend a lot of time renaming all ~100 images and you used a script 🙈

bazzadp: Just copied them and then deleted the originals and git was clever enough to see they ere a rename!

That's them all reverted now. I've left the hero image references as direct hardcodes to make it easier for when we copy these templates to 2020 and want the still to refer to 2019.

rviscomi (approved):

PR #1367, opened 2020-10-18
Remove use of set-env as being deprecated by GitHub Actions

bazzadp:
GitHub are [deprecating the `set-env` command](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) and started noticing warnings in our Action logs.

This PR makes some changes so they aren't used anymore.

### DIFF
--- a/.github/workflows/test-template-changes.yml
+++ b/.github/workflows/test-template-changes.yml
@@ -37,7 +37,6 @@ jobs:
       with:
         node-version: 12.x
     - name: Test Template Changes
-      id: test-template-changes
       run: ./src/tools/scripts/test_template_changes.sh
     - name: 'Comment PR'
       uses: actions/github-script@0.3.0
--- a/.github/workflows/test-template-changes.yml
+++ b/.github/workflows/test-template-changes.yml
@@ -41,11 +41,11 @@ jobs:
       run: ./src/tools/scripts/test_template_changes.sh
     - name: 'Comment PR'
       uses: actions/github-script@0.3.0
-      if: github.event_name == 'pull_request' && steps.test-template-changes.outputs.PR_COMMENT != ''
+      if: github.event_name == 'pull_request' && env.PR_COMMENT != ''
       with:
         github-token: ${{ secrets.GITHUB_TOKEN }}
         script: |
           const { issue: { number: issue_number }, repo: { owner, repo }  } = context;
           // Unescape any escaped comments so code blocks appear correctly
-          const pr_comment = steps.test-template-changes.outputs.PR_COMMENT.replace(/%0A/g,'\n').replace(/%0D/g,'\r').replace(/%25/g,'%');
+          const pr_comment = process.env.PR_COMMENT.replace(/%0A/g,'\n').replace(/%0D/g,'\r').replace(/%27/g,"'").replace(/%22/g,'"').replace(/%25/g,'%');
           github.issues.createComment({ issue_number, owner, repo, body: pr_comment });
--- a/.github/workflows/test-template-changes.yml
+++ b/.github/workflows/test-template-changes.yml
@@ -41,11 +41,11 @@ jobs:
       run: ./src/tools/scripts/test_template_changes.sh
     - name: 'Comment PR'
       uses: actions/github-script@0.3.0
-      if: github.event_name == 'pull_request' && env.PR_COMMENT != ''
+      if: github.event_name == 'pull_request' && steps.test-template-changes.outputs.PR_COMMENT != ''
       with:
         github-token: ${{ secrets.GITHUB_TOKEN }}
         script: |
           const { issue: { number: issue_number }, repo: { owner, repo }  } = context;
           // Unescape any escaped comments so code blocks appear correctly
-          const pr_comment = process.env.PR_COMMENT.replace(/%0A/g,'\n').replace(/%0D/g,'\r').replace(/%25/g,'%');
+          const pr_comment = steps.test-template-changes.outputs.PR_COMMENT.replace(/%0A/g,'\n').replace(/%0D/g,'\r').replace(/%25/g,'%');
           github.issues.createComment({ issue_number, owner, repo, body: pr_comment });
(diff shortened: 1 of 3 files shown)


.github/workflows/test-template-changes.yml
```diff
@@ -41,11 +41,11 @@ jobs:
       run: ./src/tools/scripts/test_template_changes.sh
     - name: 'Comment PR'
       uses: actions/github-script@0.3.0
-      if: github.event_name == 'pull_request' && env.PR_COMMENT != ''
+      if: github.event_name == 'pull_request' && steps.test-template-changes.outputs.PR_COMMENT != ''
```
ibnesayeed, line 44: I wonder, if object attribute names with hyphens (such as `steps.test-template-changes`) are accessible with dot-notation? In some situations such attributes with hyphens or spaces are accessed using square bracket notation. Unless I am absolutely sure, I would change them to the corresponding underscore-form.

.github/workflows/test-template-changes.yml
```diff
@@ -41,11 +41,11 @@ jobs:
...
       with:
         github-token: ${{ secrets.GITHUB_TOKEN }}
         script: |
           const { issue: { number: issue_number }, repo: { owner, repo }  } = context;
           // Unescape any escaped comments so code blocks appear correctly
-          const pr_comment = process.env.PR_COMMENT.replace(/%0A/g,'\n').replace(/%0D/g,'\r').replace(/%25/g,'%');
+          const pr_comment = steps.test-template-changes.outputs.PR_COMMENT.replace(/%0A/g,'\n').replace(/%0D/g,'\r').replace(/%25/g,'%');
```
ibnesayeed, line 50: Same as above!

src/tools/scripts/test_template_changes.sh
```diff
@@ -61,7 +61,7 @@ if [ "${NUM_DIFFS}" -ne "0" ]; then
   # Weird syntax is to make is Mac compatible: https://stackoverflow.com/a/1252191/2144578
   ESCAPED_OUTPUT=$(sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/%0A/g' -e 's/\r/%0D/g' -e 's/\%/%25/g' "${DIFF_FILENAME}")
   PR_COMMENT="Please note, that the following diffs happen in the templates on this branch compared to main:%0A\`\`\`%0A${ESCAPED_OUTPUT}%0A\`\`\`%0A"
-  echo "::set-env name=PR_COMMENT::${PR_COMMENT}"
+  echo "::set-output name=PR_COMMENT::${PR_COMMENT}"
```
ibnesayeed, line 64: Do we want to just set the output here or store in the `$GITHUB_ENV`?

ibnesayeed (commented):
The other day when I saw GitHub's changelog, I thought about reporting it as a ticket, but then I has some other priorities and after that I forgot about it.

.github/workflows/test-template-changes.yml
bazzadp: Tested this and no issue. I could see where having dots in the name might cause issues but can't see why dashes would.

Anyway discovered other issues when testing so stopped using output variables and went back to env variables so this is a moot point.

.github/workflows/test-template-changes.yml
bazzadp: Same answer :-)

src/tools/scripts/test_template_changes.sh
bazzadp: I thought output would be better as it would prevent any name clashes (PR_COMMENT isn't that unique and not hard to assume we might use that again) but after spending far too long trying to get it to work with escaping line numbers I've given up and gone back to env variables. It seems github script doesn't like outputs with line breaks.

.github/workflows/test-template-changes.yml
ibnesayeed: > I could see where having dots in the name might cause issues but can't see why dashes would.

Well, there are situations where a dash will be considered as a separate token and be treated as minus sign. Try the following in a web browser's DevTools console:

```
o = {"a": "b", "c-d": "e"}
>> {a: "b", c-d: "e"}

o["a"]
>> "b"

o.a
>> "b"

o["c-d"]
>> "e"

o.c-d
>> Uncaught ReferenceError: d is not defined
    at <anonymous>:1:5
```

src/tools/scripts/test_template_changes.sh
```diff
@@ -59,9 +59,9 @@ cat "${DIFF_FILENAME}"
 NUM_DIFFS=$(wc -l < "${DIFF_FILENAME}")
 if [ "${NUM_DIFFS}" -ne "0" ]; then
   # Weird syntax is to make is Mac compatible: https://stackoverflow.com/a/1252191/2144578
-  ESCAPED_OUTPUT=$(sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/%0A/g' -e 's/\r/%0D/g' -e 's/\%/%25/g' "${DIFF_FILENAME}")
+  ESCAPED_OUTPUT=$(sed -e ':a' -e 'N' -e '$!ba' -e 's/\%/%25/g' -e 's/\n/%0A/g' -e 's/\r/%0D/g' -e "s/'/%27/g" -e 's/"/%22/g' "${DIFF_FILENAME}")
```
ibnesayeed, line 62: There are some replacements that can be done in Bash without sed. [See an example in the escapify function of a code I wrote a while ago](https://github.com/ibnesayeed/repo-attrs/blob/master/entrypoint.sh).

ibnesayeed (commented):
LGTM!

ibnesayeed (approved):
Forgot to mark it approved.

.github/workflows/test-template-changes.yml
bazzadp: Interesting. Still doesn't matter as not using this code but removed the id anyway in https://github.com/HTTPArchive/almanac.httparchive.org/pull/1367/commits/d9f104ed6eeddce0a1585dea4343b715782c1437 as it wasn't being used.

PR #1369, opened 2020-10-18
Markup 2020 script tag types

Tiggerito:
More changes for #899

### DIFF
--- a/sql/2020/03_Markup/pages_almanac_by_device_and_script_type_frequency.sql
+++ b/sql/2020/03_Markup/pages_almanac_by_device_and_script_type_frequency.sql
@@ -1,0 +1,40 @@
+#standardSQL
+# pages almanac metrics grouped by device and element attribute use (frequency)
+
+CREATE TEMP FUNCTION AS_PERCENT (freq FLOAT64, total FLOAT64) RETURNS FLOAT64 AS (
+  ROUND(SAFE_DIVIDE(freq, total), 4)
+);
+
+CREATE TEMPORARY FUNCTION get_almanac_attribute_info(almanac_string STRING)
+RETURNS ARRAY<STRING> LANGUAGE js AS '''
+try {
+    var almanac = JSON.parse(almanac_string); 
+
+    if (Array.isArray(almanac) || typeof almanac != 'object') return [];
+
+    if (almanac.scripts && almanac.scripts.nodes && almanac.scripts.nodes.map) {
+      return almanac.scripts.nodes.map(n => {if (n.type) return n.type.toLowerCase().trim(); else return "NOT_SET"; }); 
+    }
+    
+} catch (e) {
+    
+}
+return []; 
+''';
+
+SELECT
+  _TABLE_SUFFIX AS client,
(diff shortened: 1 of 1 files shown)


sql/2020/03_Markup/pages_almanac_by_device_and_script_type_frequency.sql
```diff
@@ -0,0 +1,40 @@
+#standardSQL
+# pages almanac metrics grouped by device and element attribute use (frequency)
+
+CREATE TEMP FUNCTION AS_PERCENT (freq FLOAT64, total FLOAT64) RETURNS FLOAT64 AS (
+  ROUND(SAFE_DIVIDE(freq, total), 4)
+);
```
bazzadp, line 6: Seems a bit overkill to have a whole function for one line of code?

sql/2020/03_Markup/pages_almanac_by_device_and_script_type_frequency.sql
```diff
@@ -0,0 +1,40 @@
...
+  `httparchive.pages.2020_08_01_*`,
+  UNNEST(get_almanac_attribute_info(JSON_EXTRACT_SCALAR(payload, '$._almanac'))) AS type_name
+GROUP BY
+  client,
+  type_name
+ORDER BY
+  freq DESC,
+  client
+LIMIT
+  1000
```
bazzadp, line 40: Think we might have talked about this before but personally I think we should have a percentage limit than an absolute one. WDYT?

bazzadp (commented):
Changes look fine to me, but made two small comments. Let me know what you think and can merge this if you wanna keep as is.

sql/2020/03_Markup/pages_almanac_by_device_and_script_type_frequency.sql
Tiggerito: More about consistency. And a copy/paste. There was mention of making a function like this global so everyone uses the same pattern. I personally think we could drop the ROUND as it can cause accuracy issues and is not needed if people format in the sheets.

sql/2020/03_Markup/pages_almanac_by_device_and_script_type_frequency.sql
Tiggerito: Just trying to get these things done. I can't work out how to get that pct to work as a way to limit the output?

bazzadp: Oh yeah looks like it doesn't like using this particular pct:
```
Column pct contains an analytic function, which is not allowed in HAVING clause at [34:3]
```

This works using the `freq` column instead:

```
GROUP BY
  client,
  type_name
HAVING
  freq > 1000
ORDER BY
  freq DESC,
  client
```

Tiggerito: Done

bazzadp (approved):

PR #1371, opened 2020-10-19
Stop over linting

bazzadp:
As per https://github.com/github/super-linter/issues/671#issuecomment-706131556 adding a small bit of config so we only lint the actual changes in a branch, and not the full differences between two branches if main goes ahead of a branch.


ibnesayeed (approved):
LGTM!

PR #1372, opened 2020-10-19
Add CodeQL GitHub Action

bazzadp:
GiHub have recently launched [Code Scanning via CodeQL](https://github.blog/2020-09-30-code-scanning-is-now-available/) which scans source code for Security Vulnerabilities, which is an additional step that our linters may not cover.

Now, as a static site, we hopefully don't have too much to worry about on the security front, but we still wouldn't want our server, or our users client side code compromised so I think we should add it.

It has already identified a few minor things that linting didn't pick up (I'm running it with both security and quality checks btw). I've included most of those fixes in this PR and also marked a couple as "won't fix" - the tool makes this very easy.

@rviscomi it's also identified a few issues in the CSS utils that I'll leave you to consider whether to fix or close off: https://github.com/HTTPArchive/almanac.httparchive.org/security/code-scanning?query=ref%3Arefs%2Fheads%2Fcodeql

Results will normally appear in the Pull Request but also in the Security tab of GitHub.

### DIFF
--- a/.github/workflows/code-static-analysis.yml
+++ b/.github/workflows/code-static-analysis.yml
@@ -1,0 +1,77 @@
+# Runs the CodeQL static code analysis
+
+name: "Static Code Analysis"
+
+on:
+  push:
+    paths:
+      - '**.js'
+      - '**.py'
+      - '**.html'
+  pull_request:
+    paths:
+      - '**.js'
+      - '**.py'
+      - '**.html'
+  schedule:
+    #        ┌───────────── minute (0 - 59)
+    #        │ ┌───────────── hour (0 - 23)
+    #        │ │ ┌───────────── day of the month (1 - 31)
+    #        │ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
+    #        │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+    #        │ │ │ │ │
+    #        │ │ │ │ │
+    #        │ │ │ │ │
+    #        * * * * *
+    - cron: '30 1 * * 0'
+  workflow_dispatch:
+
+jobs:
+  CodeQL-Build:
+    # CodeQL runs on ubuntu-latest, windows-latest, and macos-latest
+    runs-on: ubuntu-latest
+    # Skip duplicate job for local branches
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          # Must fetch at least the immediate parents so that if this is
+          # a pull request then we can checkout the head of the pull request.
+          # Only include this option if you are running this workflow on pull requests.
+          fetch-depth: 2
+
+      # If this run was triggered by a pull request event then checkout
+      # the head of the pull request instead of the merge commit.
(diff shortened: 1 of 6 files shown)


bazzadp:
Note as the first commit of this code the results do not appear in this pull request:

Missing analysis for base commit

But in future they will.

rviscomi:
Can you exclude `sql/**.js`?

bazzadp:
Can do but did you have a look at them? Might be worth fixing some of them and might be useful to catch real errors.

Or can just mark those warnings as “won’t fix”. It only blocks on errors and not warnings and even then you can triage errors to say “won’t fix” too. Once an issue is triaged the the same line of code shouldn’t flag again.

As with all of these things if the hassle of dealing with it is more effort than the usefulness then we shouldn’t enable it, but I was pleasantly surprised how little noise it came with - even with adding quality checks - and the things it spotted in the other code, while not super critical, are things that it was right to flag.

So have a look and let me know if not useful and you want to exclude.

rviscomi:
Yeah I did look. The JS it's flagging is a library to be used in other places, so the warnings like "unused function" aren't helpful because they _are_ used by the SQL UDFs. The other warnings pointing out a redundant use of `var && var.prop` when `var` is always truthy are valid, but this isn't production code and we care more about the query output than JS perfection.

bazzadp:
Closed em for you. I think worth keeping in case the flag anything that is an issue. Yes the code is not in production but we obviously still want the code to work correctly! If it ends up being really noisy and annoying for these utility scripts then can revisit and exclude this path.
Agree with that?

rviscomi:
Yeah if it catches things that will break the JS that's ok, but warnings meant to shave bytes of JS are unhelpful noise for SQL-bound code.

.github/workflows/code-static-analysis.yml
```diff
@@ -0,0 +1,77 @@
...
+    #        * * * * *
+    - cron: '30 1 * * 0'
+  workflow_dispatch:
+
+jobs:
+  CodeQL-Build:
+    # CodeQL runs on ubuntu-latest, windows-latest, and macos-latest
+    runs-on: ubuntu-latest
+    # Skip duplicate job for local branches
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
```
ibnesayeed, line 34: I am wondering, will this condition allow `schedule` and `workflow_dispatch` events?

ibnesayeed (commented):
Should we use `let` instead of `var` in JS files?

.github/workflows/code-static-analysis.yml
bazzadp: Good spot. Fixed.

bazzadp:
> Should we use `let` instead of `var` in JS files?

At the moment we don't use ES6 syntax (including `let`) in client side JavaScript for better browser support.

`Let` in for loops for example is only partially supported in for loops and basically acts like `var` so can cause bugs if people assume it's acting like `let` should.

We don't have masses of IE11 usage (about 1.12% since launch - and some of that is probably me testing on it 😀) or other older browsers, and maybe we should stop living in the past and move on, but that's what we've gone with previously so think any change there should be discussed outside of this PR. At the moment the changes in this PR basically are using `var`, which the browser will use anyway.

ibnesayeed (approved):
LGTM!

PR #1376, opened 2020-10-20
Highlight H4, H5 and H6 in table of contents as well

### LINKED ISSUE #1363: Highlight H4s in sidebar TOC
The green "you are here" indicator in the Table of Contents sidebar appears to only highlight H1-3 content. As you scroll through subsections of H4s, the indicator doesn't highlight those nested headings. It only seems to be updated when a new H1-3 comes into view. This causes unexpected behavior, like when scrolling up, a long subsection doesn't update the indicator:

1. Go to https://almanac.httparchive.org/en/2019/performance#conclusion
2. Scroll up into the FID section above
3. Indicator is stuck on "Conclusion" until the FID H3 becomes visible

image

Should the indicator also highlight H4s?

SaptakS:
Adds 'article h4' also in selectors to be marked for IntersectionObserver API. This will highlight individual H4s as well. If we want to avoid doing that, then we can check if the Intersection node is a h4 and activate the parent h3 node instead of itself. Let me know which seems more desireable.

Fixes #1363

### DIFF
--- a/src/static/js/chapter.js
+++ b/src/static/js/chapter.js
@@ -351,7 +351,7 @@ function indexHighlighter() {
   }, options);
 
   // Add an intersection observer to each heading
-  var all_headings = document.querySelectorAll('article h1, article h2, article h3');
+  var all_headings = document.querySelectorAll('article h1, article h2, article h3, article h4, article h5, article h6');
   for (var index = 0; index < all_headings.length; ++index) {
     var heading = all_headings[index];
     observer.observe(heading);


bazzadp:
LGTM! I wonder should we include h5 and h6? Don't have any I don't think but any harm to include to future proof?

src/static/js/chapter.js
```diff
@@ -351,7 +351,7 @@ function indexHighlighter() {
   }, options);
 
   // Add an intersection observer to each heading
-  var all_headings = document.querySelectorAll('article h1, article h2, article h3');
+  var all_headings = document.querySelectorAll('article h1, article h2, article h3, article h4');
```
bazzadp, line 354: As suggested let's future proof this:

```suggestion
  var all_headings = document.querySelectorAll('article h1, article h2, article h3, article h4, article h5, article h6');
```

bazzadp (changes requested):

src/static/js/chapter.js
SaptakS: I think that's a good idea.

bazzadp (approved):

rviscomi:
Thanks @SaptakS! 🚀

PR #1401, opened 2020-11-03
Lighthouse ci action

### LINKED ISSUE #1178: Add Lighthouse testing

bazzadp:
Fixes #1178 

Adds Automatic Lighthouse checks to our Test Website action to test for regressions as good changes.

Note we disable a few checks:
- Time-based performance checks are disabled as too unreliable
- Optimisations we don't currently perform (e.g. minifying CSS and JS and removing unused CSS and JS)
- Turns off items that don't work on dev server (e.g. Canonical is incorrect, Compression is not used, Crawling is disabled...etc.).

This means we should only get real failures.

I also added a weekly run against production action. This looks at Lighthouse scores - though again not Performance due to variability as really want to avoid introducing noisy checks that create false alerts sometimes.

As well as avoiding #1216 (which is where this whole thing came from!) it's also identified another couple of issues meaning we have less than perfect scores:
- The Spanish CSS chapter has an empty link
- Links with lints don't have enough spacing on mobile

I think this shows the value of automating these checks!

Monitoring performance needs some more thought. We could do several runs to average, and set thresholds, or leverage the Page Speed Insights API to get CrUX data or something similar. Anyway think this PR has enough value even without explicit performance monitoring.

Interested to hear your thoughts!


src/tools/scripts/set_lighthouse_urls.sh
```diff
@@ -0,0 +1,105 @@
...
+    # Then back to the pull request changes
+    git checkout --progress --force "${COMMIT_SHA}"
+
+    # Transform the files to http://127.0.0.1:8080 URLs
+    LIGHTHOUSE_URLS=$(echo "${CHANGED_FILES}" | sed 's/src\/content/http:\/\/127.0.0.1:8080/g' | sed 's/\.md//g' | sed 's/\/base\//\/en\//g' | sed 's/src\/templates/http:\/\/127.0.0.1:8080/g' | sed 's/\.html//g' | sed 's/_/-/g' | sed 's/\/2019\/accessibility-statement/\/accessibility-statement/g' )
+    if [ "${LIGHTHOUSE_URLS}" = "" ]; then
+        LIGHTHOUSE_URLS="${BASE_URLS}"
+    else
+        LIGHTHOUSE_URLS=$( echo -e "${LIGHTHOUSE_URLS}\n${BASE_URLS}")
+    fi
```
ibnesayeed, line 74: It should be possible to avoid this conditional by always concatenating the two lists and then removing any empty lines, which will take care of the condition when either of the lists is empty or has additional leading or trailing new lines.
```suggestion
    LIGHTHOUSE_URLS=$(echo -e "${LIGHTHOUSE_URLS}\n${BASE_URLS}" | sed '/^$/d')
```

src/tools/scripts/set_lighthouse_urls.sh
```diff
@@ -0,0 +1,105 @@
...
+
+else
+
+    # Else test every URL (except PDFs) in the sitemap
+    LIGHTHOUSE_URLS=$(grep loc templates/sitemap.xml | grep -v static | sed 's/ *<loc>//g' | sed 's/<\/loc>//g' | sed 's/https:\/\/almanac.httparchive.org/http:\/\/127.0.0.1:8080/g')
```
ibnesayeed, line 79: Many sed instances can be removed in favor or a single grep as noted above.

src/tools/scripts/set_lighthouse_urls.sh
```diff
@@ -0,0 +1,105 @@
...
+cat > "${LIGHTHOUSE_CONFIG_FILE}" << END_CONFIG
+{
+  "ci": {
+    "collect": {
+      "url": [
+${LIGHTHOUSE_URLS}
+      ]
+    },
+${LIGHTHOUSE_CONFIG}
+END_CONFIG
```
ibnesayeed, line 105: This looks very brittle to me as it is dependent on how exactly certain files are formatted. How about something more programmatic like:

```suggestion
echo ${LIGHTHOUSE_URLS} | jq '. | split("\n") | map(select(length > 0))' -Rs | jq '{ci: {collect: {url: .}}} > "${LIGHTHOUSE_CONFIG_FILE}"
```

ibnesayeed (changes requested):

.github/workflows/test_website.yml
```diff
@@ -39,3 +39,19 @@ jobs:
...
+      uses: treosh/lighthouse-ci-action@v3
+      id: LHCIAction
+      with:
+        # For dev, turn off all timing perf audits (too unreliable) and a few others that don't work on dev
+        configPath: .github/lighthouse/lighthouse-config-dev.json
+        uploadArtifacts: true # save results as an action artifacts
+        temporaryPublicStorage: true # upload lighthouse report to the temporary storage
+    - name: Show Lighthouse outputs
+      run: |
+        echo '${{ steps.LHCIAction.outputs.manifest }}' | jq -r '.[] | (.summary|tostring) + " - " + .url'
```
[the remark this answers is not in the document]
bazzadp: Ah yeah only have that comment to give it a heading in log. But good idea will add.

rviscomi:
Thanks @bazzadp for the PR. Excited about LH integration. I'll defer to @ibnesayeed's review.

src/tools/scripts/set_lighthouse_urls.sh
```diff
@@ -0,0 +1,105 @@
...
+http://127.0.0.1:8080/en/2019/
+http://127.0.0.1:8080/en/2019/javascript
+http://127.0.0.1:8080/en/2020/
+END
+)
+
+if [ "${production}" == "1" ]; then
+    
+    # Get the production URLs from the production sitemap
+    LIGHTHOUSE_URLS=$(curl -s https://almanac.httparchive.org/sitemap.xml | grep "<loc" | grep -v static | sed 's/ *<loc>//g' | sed 's/<\/loc>//g')
```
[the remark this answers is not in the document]
bazzadp: Agree with changing `static` to `/static/`. Don't agree the other change is simpler! I'm having trouble getting my head round it to be honest and I think my unix is pretty decent!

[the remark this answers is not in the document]
ibnesayeed: Explaining `grep -oP '<loc>\K\S+(?=</loc>)'`:

`-o`: Only output the matching part, not the whole line.
`-P`: Use Perl style RegEx to avoid unnecessary escaping and having access to more powerful means of capture groups.
`\K`: What follows this will be recorded as a capture group.
`\S`: Anything that is not a white-space character.
`(?=</loc>)`: A look ahead pattern to declare when to end the previous capture group.

The problem with multiple sed instances is that it is very difficult to ensure they will be working in the same proximity and position as each instance is acting independently. Though, that is not an issue here because we already know the kind of input we are dealing with. Moreover, invoking multiple programs many times can accumulate over time to increase the processing time.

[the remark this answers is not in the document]
ibnesayeed: If you do not like the `\K` notation, you can use a combination of look behind and look ahead to isolate the point of interest like this:

```
grep -oP '(?<=<loc>)\S+(?=</loc>)'
```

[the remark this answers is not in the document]
ibnesayeed: > I'm having trouble getting my head round it to be honest and I think my unix is pretty decent!

It is less about Unix knowledge and more about RegEx. As they say, [Regular Expressions: Now You Have Two Problems](https://blog.codinghorror.com/regular-expressions-now-you-have-two-problems/)! :laughing:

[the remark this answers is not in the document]
bazzadp: Yes put having two simple `sed` commands with simple regexs, versus one complex `grep` command with complex, difficult to understand and maintain, regex using capture groups and look ahead patterns is not simplifying it it my point.

Using pipes and keeping commands simple is a huge part of Unix philosophy precisely because it keeps things simple.

So thanks for the feedback but going to change `static` to `/static/` but keep the rest as it is.

src/tools/scripts/set_lighthouse_urls.sh
```diff
@@ -0,0 +1,105 @@
...
+    # First checkout main to get list of differences
+    git pull --quiet
+    git checkout main
+    # Then get the changes
+    CHANGED_FILES=$(git diff --name-only "main...${COMMIT_SHA}" --diff-filter=d content templates | grep -v base.html | grep -v ejs | grep -v base_ | grep -v toc.html | grep -v sitemap)
+    # Then back to the pull request changes
+    git checkout --progress --force "${COMMIT_SHA}"
+
+    # Transform the files to http://127.0.0.1:8080 URLs
+    LIGHTHOUSE_URLS=$(echo "${CHANGED_FILES}" | sed 's/src\/content/http:\/\/127.0.0.1:8080/g' | sed 's/\.md//g' | sed 's/\/base\//\/en\//g' | sed 's/src\/templates/http:\/\/127.0.0.1:8080/g' | sed 's/\.html//g' | sed 's/_/-/g' | sed 's/\/2019\/accessibility-statement/\/accessibility-statement/g' )
```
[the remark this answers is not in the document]
bazzadp: I do agree it's a good use case. However, I just don't think changing the default and well-known pattern of using `s//g` is a good idea just because sed supports it. So I vote to keep it as is despite the extra escaping being needed. Think our JS developers will find it less confusing just in case any of them touch this in future.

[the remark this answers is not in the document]
ibnesayeed: I am okay with keeping them escaped. However, if we chose to go with using a different delimiter instead of the well-known one (which is not the default by the way, nothing is a default delimiter in sed or else it would work without specifying one), it will not be a bad idea to add a comment saying:

```
# Using alternate sed delimiter as 's|find|replace|' instead of 's/find/replace/' to avoid unnecessary forward-slash escaping
```

This will serve both as a reminder and education note.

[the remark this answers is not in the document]
bazzadp: I more meant the default regex separator in most (all?) programming languages is `/` and only `sed` allows that to change AFAIK.

src/tools/scripts/set_lighthouse_urls.sh
bazzadp: This is a good suggestion for robustness. Added.

src/tools/scripts/set_lighthouse_urls.sh
bazzadp: I'm gonna go with simplicity again over robustness. This runs on every commit and we're in total control of this so happy to accept the risk.

src/tools/scripts/set_lighthouse_urls.sh
```diff
@@ -0,0 +1,105 @@
...
+
+elif [ "${RUN_TYPE}" == "pull_request" ] && [ "${COMMIT_SHA}" != "" ]; then
+
+    # If this is part of pull request then get list of files as those changed
+    # Uses similar logic to GitHub Super Linter
+    # First checkout main to get list of differences
+    git pull --quiet
+    git checkout main
+    # Then get the changes
+    CHANGED_FILES=$(git diff --name-only "main...${COMMIT_SHA}" --diff-filter=d content templates | grep -v base.html | grep -v ejs | grep -v base_ | grep -v toc.html | grep -v sitemap)
```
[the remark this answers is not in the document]
bazzadp: Arguable as to whether that's simpler to understand. Will leave as is if you don't mind.

src/tools/scripts/set_lighthouse_urls.sh
bazzadp: Same answer as above. Will keep with simplicity over single commands.

src/tools/scripts/set_lighthouse_urls.sh
ibnesayeed: Testing its functionality:

```
$ BASE_URLS=$(cat <<-END
http://127.0.0.1:8080/en/2019/
http://127.0.0.1:8080/en/2019/javascript
http://127.0.0.1:8080/en/2020/
END
)

$ echo "${BASE_URLS}" | jq -Rs '. | split("\n") | map(select(length > 0))' | jq '{ci: {collect: {url: .}}}'
{
  "ci": {
    "collect": {
      "url": [
        "http://127.0.0.1:8080/en/2019/",
        "http://127.0.0.1:8080/en/2019/javascript",
        "http://127.0.0.1:8080/en/2020/"
      ]
    }
  }
}
```

ibnesayeed (commented):
I am okay with the other changes except the last one where I think using `jq` for building and formatting the config file would be better than manually doing it, which is brittle as it introduces tight coupling between the config file format and the shell script.

src/tools/scripts/set_lighthouse_urls.sh
bazzadp: Ok but is there a way to insert it into the existing config?

The intention of this but is to have most of the config in the separate JSON file (as easier to maintain) and then insert the URLs (which will be dynamic) in the right place above those other config. It seems your suggestion just overwrites the whole config file with the URLs only and therefore loses all the other config.

Is there a way for jq to read a JSON file and then insert the url section at a specific path but maintain the rest of the JSON? Because if that could be done without too much complication then I would be open to that.

ibnesayeed: I did not realize that there were other things in the config file and you were attempting to insert an object inside. Yes, it is possible with `jq`, I will provide the code for that in a few.

ibnesayeed: Test:

```
$ LIGHTHOUSE_URLS=$(cat <<-END
http://127.0.0.1:8080/en/2019/
http://127.0.0.1:8080/en/2019/javascript
http://127.0.0.1:8080/en/2020/
END
)

$ echo "${LIGHTHOUSE_URLS}" | jq -Rs '. | split("\n") | map(select(length > 0))' | jq -s '.[0] * {ci: {collect: {url: .[1]}}}' lighthouse-config-prod.json -
{
  "ci": {
    "assert": {
      "assertions": {
        "categories:accessibility": [
          "error",
          {
            "minScore": 1
          }
        ],
        "categories:best-practices": [
          "error",
          {
            "minScore": 1
          }
        ],
        "categories:seo": [
          "error",
          {
            "minScore": 1
          }
        ]
      }
    },
    "collect": {
      "url": [
        "http://127.0.0.1:8080/en/2019/",
        "http://127.0.0.1:8080/en/2019/javascript",
        "http://127.0.0.1:8080/en/2020/"
      ]
    }
  }
}
```

ibnesayeed: I have updated the suggestion above to reflect this change. Please give it a try and see if it works. If you do not want to do everything in one step, it is possible to create a temporary JSON file with the list of URIs and then merge it into the config JSON. However, I like the single command approach better as it does not create unnecessary temporary files that we will need to remove after merging.

bazzadp: I get this on MacOS:

```
jq: error (at <stdin>:5): array (["http://12...) and object ({"ci":{"col...) cannot be multiplied
```

But had some other problems with jq on MacOS so will commit it and see how linux in GitHub Actions likes it.

bazzadp: Same thing on linux. Trying to debug it now...

ibnesayeed: I think it is because you might be using jq version prior to `1.4` which did not have `*` support.

ibnesayeed: > `jq: error (at <stdin>:5): array (["http://12...) and object ({"ci":{"col...) cannot be multiplied`

How are you testing it? The way I formatted things in my examples above, I have already wrapped the array inside of the object at the nested level, so there should not be the situation of merging array with object.

bazzadp: Ah works now. Didn't like the fact I was reading the file and also writing to it at the same time - which the Linter picked up to be fair.

This fixed it: https://github.com/HTTPArchive/almanac.httparchive.org/pull/1401/commits/7e5686f3db889f5a607f07a5754b016a6ce95de8

This is awesome @ibnesayeed thanks very much for your jq mastery!

ibnesayeed: Here is another way of doing the same, but without using `*` operator:

```
$ echo "${LIGHTHOUSE_URLS}" | jq -Rs '. | split("\n") | map(select(length > 0))' | jq -s '.[0].ci.collect.url = .[1] | .[0]' lighthouse-config-prod.json -
```

bazzadp: Original way worked fine with that last linting commit. Example run here: https://github.com/HTTPArchive/almanac.httparchive.org/runs/1349670806?check_suite_focus=true

ibnesayeed: > Didn't like the fact I was reading the file and also writing to it at the same time

Oops! That's an oversight on my end. I wish `jq` had a solution like `sed -i`. :smile:

ibnesayeed (approved):
LGTM!

However, I wish there was a way in lighthouse to supply list of URIs separately, without mixing it into the config file. Please see if there are any configuration options that can point to a file for list of URIs. That said, this PR can be merged, and if a better approach is found, we can make necessary changes later.

Issue #1195, opened 2020-08-10
Contributor profile pages?


rviscomi:
I'd like to pitch an idea for discussion: year-agnostic profile pages for each contributor.

For those who contribute to multiple annual editions, some of their metadata is duplicated across the config files, like name and social links. Authors also have bios in the year's [base template](https://github.com/HTTPArchive/almanac.httparchive.org/blob/0d9238dc7e5528d04222cc415d8ded0db3cea232/src/templates/en/2019/base.html#L171-L201). It may be useful to deduplicate the metadata about a contributor that doesn't change over time, like their name, social links, and bio, from the metadata that is year-specific, like the teams and chapters.

So for example, I could add basic metadata about myself to `/src/config/contributors/rviscomi.json` (or `.yaml`):

```json
{
  "name": "Rick Viscomi",
  "avatar_url": "https://avatars0.githubusercontent.com/u/1120896?v=4&s=200",
  "github": "rviscomi",
  "twitter": "rick_viscomi",
  "bio":  "Rick Viscomi is[...]"
}
```

Then in [`2019.json`](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/src/config/2019.json) and [`2020.json`](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/src/config/2020.json) I only need to map my contributor ID `rviscomi` to the teams I contributed to those years. We could also get info about which chapters I contributed to from the YAML in the chapter markdown files.

All of this contributor info could be made publicly available via endpoints like `/<lang>/contributors/<id>`, eg `https://almanac.httparchive.org/en/contributors/rviscomi`. I imagine the page being somewhat similar to [dev.to/rick_viscomi](https://dev.to/rick_viscomi):

![image](https://user-images.githubusercontent.com/1120896/89833087-7835ed00-db2e-11ea-8b02-88d99e1c99fb.png)

- profile photo
- name
- bio
- social links
- snippets of chapters I've contributed to (as an author, reviewer, analyst, editor, or translator)
- badges representing teams I've been on and the number of years I've been a contributor (_gotta catch 'em all!_)

Right now only authors have bios and they're only visible on the chapter pages. This proposal would enable all contributors to have bios on their personal profile pages and it'd be easier to go back through all of their contributions over the years.

Curious to hear if this is something contributors want, whether there are any unforeseen technical complexities, or if there are any other ideas for ways we could utilize a page like this to showcase contributors' work.

[...]

PR #1413, opened 2020-11-04
Move Author bios to Meta data

bazzadp:
Since we merged #1104 which had a prettier re-org this is easier.

A few issues/reasons not to do this:

1. Quotes still need unescaping which looks a little funny.
2. Some people Authored multiple chapters (e.g. Paul) so need two bios.
3. Can't easily reuse this data if we ever implement contributor bio pages as mentioned in #1195 
4. Requires another bulk update to all the chapters 😞

Other than that, seems like a good thing to do.


bazzadp:
Will convert rest of chapters if happy to continue with this. So just a Draft PR to allow easy review until I do that.

rviscomi:
> Some people Authored multiple chapters (e.g. Paul) so need two bios.

Should be rare. Not sure if we have that problem this year. One thing to watch out for is consistency across chapters translated by different people. One cool thing would be some middleware that can support something like "bio: see chapter X" and the author's bio in chapter X gets reused.

> Can't easily reuse this data if we ever implement contributor bio pages as mentioned in #1195

I think a middleware solution could be useful here. During the chapter generation script, extract the metadata and generate a new metadata file with all author info. Or if we go the bio page route maybe we have to maintain a standalone contributor bio metadata file anyway, and not couple it with the md at all.

> Requires another bulk update to all the chapters 😞

Greatly appreciate your efforts ❤️

PR #1424, opened 2020-11-05
Remove dotted line on Help Translate menu

### LINKED ISSUE #1386: Overflow of the language picker divider
image

The divider in the language picker appears to be exceeding the width of the select button, even though the options may be wider than the button itself.

This seems to be an edge case when content is only available in English, which would only affect the 2020 chapters.

bazzadp:
Fixes #1386 

image

### DIFF
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -1347,3 +1347,12 @@ p.copyright a {
   width: max-content;
   min-width: 100%;
 }
+
+.nav-dropdown-list .help-translate {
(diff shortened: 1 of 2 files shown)


bazzadp:
Actually figured it should still be differentiated some way so put it in italics:

image

bazzadp:
Actually went another way so edited first comment.

Let's just remove this dotted line if no other languages but keep it if there are. But also like the italics so let that in:

With languages

No languages

rviscomi (approved):
LGTM!

PR #1436, opened 2020-11-06
Add npm run start command

bazzadp:
Adds an `npm run start` command which can be used to install, generate and start the website.


rviscomi:
Is there a way to detect the OS in the script and run the appropriate commands?

bazzadp:
> Is there a way to detect the OS in the script and run the appropriate commands?

Done. Though it does require an initial `npm install` to be run as easiest way is using [`run-script-os`](https://www.npmjs.com/package/run-script-os) - I tried writing my own node script for this but [not the easiest to not buffer exec output in node](https://stackoverflow.com/questions/9777215/running-a-shell-command-from-node-js-without-buffering-output) so led to a long delay when seemingly nothing was happening.

src/tools/scripts/run_and_test_website.bat
```diff
@@ -0,0 +1,32 @@
...
+rem #
+
+
+echo "Installing and testing python environment"
+python -m pip install --upgrade pip
+pip install -r requirements.txt
+pytest
+
+echo "Installing node modules"
+CALL npm install
```
rviscomi, line 19:
```suggestion
call npm install
```

rviscomi (approved):

PR #1437, opened 2020-11-07
Add option to generate individual watchers and a watcher

bazzadp:
The Chapter generation with `npm run generate` can take a while to complete. This can be frustrating if you just want to one chapter generated (e.g. if you are writing or editing it).

This PR adds the ability to pass arguments to `npm run generate`, for example:
```
npm run generate en/2020/markup
```

Or even patterns (note patterns must be in quotes to prevent OS attempting to match to files):
```
npm run generate ".*/2019/css"
npm run generate "en/.*/css"
npm run generate ".*/2020/.*"
```

It also adds a file watcher, which monitors the `content` directory and automatically regenerates a chapter when it sees it being modified:
```
npm run watch
```

### DIFF
--- a/src/tools/generate/chapter_watcher.js
+++ b/src/tools/generate/chapter_watcher.js
@@ -1,0 +1,23 @@
+var watch = require('node-watch');
+let { generate_chapters } = require('./generate_chapters');
+const { exec } = require("child_process");
+ 
+watch('content', { filter: /\.md$/, recursive: true }, async function(evt, name) {
+  console.log('File modified: %s', name);
+  if (evt != 'remove') {
+    await generate_chapters(name)
+      /*
+      const command = 'npm run generate ' + name;
+      exec (command, (err, stdout, stderr) => {
+        if (err) {
+          // some err occurred
+          console.error(err)
+        } else {
+          // the *entire* stdout and stderr (buffered)
+          console.log(`stdout: ${stdout}`);
+          console.log(`stderr: ${stderr}`);
+        }
+      });
+      */
+  }
+});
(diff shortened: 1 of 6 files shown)


src/tools/generate/chapter_watcher.js
```diff
@@ -0,0 +1,23 @@
+var watch = require('node-watch');
+let { generate_chapters } = require('./generate_chapters');
+const { exec } = require("child_process");
+ 
+watch('content', { filter: /\.md$/, recursive: true }, async function(evt, name) {
+  console.log('File modified: %s', name);
+  if (evt != 'remove') {
+    await generate_chapters(name)
+      /*
```
rviscomi, line 9: Can the comment block be removed?

rviscomi (approved):
This is what I've always wanted! 🤩

src/tools/generate/chapter_watcher.js
bazzadp: Oops yes! Earlier iteration. It's gone!

rviscomi:
What's the recommended way to stop the server after it's been started in a background process? ctrl+c won't work anymore.

bazzadp:
`pkill -9 python main.py` will do it. That’s what the `deploy.sh` script does to ensure you’re running the latest code before you deploy.

bazzadp:
Could add this as first thing in “npm run start” script too so it kills old versions.

bazzadp:
Or even add an `npm run stop` script.

Or both?

rviscomi:
Both 👍

bazzadp:
Done - https://github.com/HTTPArchive/almanac.httparchive.org/pull/1443

PR #1443, opened 2020-11-07
Add kill scripts

bazzadp:
Add `npm run stop` command

Also adds the same comment to `npm run start` to ensure clean start.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -45,6 +45,11 @@ npm run start
 
 4. In your web browser, enter the following address: http://127.0.0.1:8080
 
+To stop the server run the following:
+
+```
+npm run stop
+```
 
 ## Generating chapters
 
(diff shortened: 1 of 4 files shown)


rviscomi (approved):
What's up with the github-actions diff?

bazzadp:
It’s why I raised #1444 - it’s randomly decided to replace the 2019 Markup chapter with the 2020 one in the test run. Not sure why yet (and nothing to do with this PR btw).

FYI this GitHub Action runs if there are any changes to package.json - it allows us to see if any dependency upgrades change the outputted HTML and then can see if this is a problem or not. A successful run means we can merge that update without fear. Love all the automatic testing we have this year!

I only expected it to be run on dependabot PRs to be honest but luckily it’s run on this and highlighted a bug in our code supporting multiple years (again nothing to do with this PR - it’s just one of the first ones since we merged the first 2020 chapter).

bazzadp:
Suspect it’s the lack of year in the middle line here:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/9fa81d0706b9f93280c59d437825c12b20f789e0/src/tools/generate/generate_ebooks.js#L54-L56

Will test, and submit a fix ASAP.

bazzadp:
Merge fix and reran and no repeat. So think we're good. Merging this.

PR #1448, opened 2020-11-08
2020 chapter metadata

rviscomi:
Placeholders for all chapters except Capabilities, whose markdown file is already being reviewed.

Content team leads must still review the metadata and fill in `TODO`s and confirm that all of the contributors listed are accurate and sorted in order of their contributions.

This PR also includes a fix for generate_table_of_contents.js, which expects at least some headings in the markdown. With empty markdown sections, this needed to be changed to handle null content.


bazzadp (commented):
The only slight issue is that the chapters will now generate and be in the sitemap. Which has two implications:
1. Google will see them and crawl them. Maybe no bad thing to get these registered in Google early, but think we should maybe set the descriptions the same as 2019 for now for that reason. I can do that.
2. People can go to them if they know the URL. Think that's fine as long as we're not signposting it in ToC (which we are not) or elsewhere. May also want to remove todo chapters from featured chapters. I can look at that.

bazzadp:
Oh and btw these stubs mean they are not being generated in featured-chapters for now (as the featured quotes and stats are not set) but will be if author sets them, then they will be included even if early merge. So better to exclude them.

src/tools/generate/generate_chapters.js
```diff
@@ -66,20 +77,24 @@ const generate_chapters = async (chapter_match) => {
...
-          featured_quotes[language][year] = {};
+        if ( sitemap_languages[year].includes(language) ) {
+          sitemap.push({ language, year, chapter, metadata });
         }
-        featured_quotes[language][year][chapter] = chapter_featured_quote;
-      }
-      if ( sitemap_languages[year].includes(language) ) {
-        sitemap.push({ language, year, chapter, metadata });
+        ebook_chapters.push({ language, year, chapter, metadata, body, toc });
       }
```
bazzadp, line 97: This is all just in indentation change - apart from the first `if` statement. So looks a lot more than it is.

bazzadp (approved):
I made a few changes:
- I filter out "todo" chapters from featured_chapters, sitemap and ebook so they won't be included until ready (as a reminder, a chapter can be "todo" (JSON config) and/or "unedited" (metadata) as two distinct flags (one in config and one in meta data). "Todo" means not ready for publishing, while "unedited" means publish but display a note that it's unedited).
- Merged main including change #1451 to fix lighthouse errors (I merged #1451 without review as very simple and wanted to have a clean run of tests here).
- Added meta descriptions based on last year's ones. While I think they should be reviewed by Authors in case the chapter covers different things than last year, I also think there should be a placeholder text to give an example - for consistency and to allow authors to see the sort if thing to write. Let's discuss if you disagree.

rviscomi:
Thanks! These changes LGTM.

PR #1441, opened 2020-11-07
Add Abby, add ToC and remove unused index.html

bazzadp:
Bit of clean up here:
- Add @AbbyTsai to both configs - let us know if you want Twitter, Github, Website or LinkedIn info here
- Add 2020 ToC for Trad Chinese
- Remove unused index.html file

PR #1452, opened 2020-11-08
Update home page to reference 2020 chapters now we have some ready

### LINKED ISSUE #1405: Point 2020 home page links to 2020 chapters
In #1383 we basically filled out the 2020 home page but pointed some content to 2019 content until it's ready:
- [ ] Primary CTA
- [ ] Featured Quotes

Once 2020 chapters start becoming ready we should update this.

bazzadp:
Closes #1405 

Now we have some chapters we can stop referring to the 2019 edition in the 2020 home page.

Only slight issue is that we will see English Featured Quotes in translations until at least one chapter is Translated. Alternative is to drop back to 2019 featured quote for now?

### DIFF
--- a/src/templates/base/2019/index.html
+++ b/src/templates/base/2019/index.html
@@ -4,11 +4,7 @@
 {% block image_height %}562{% endblock %}
 {% block image_width %}715{% endblock %}
 
-{% if featured_chapters_exists(lang, year) or featured_chapters_exists('en', year) %}
-{% set year_with_chapters = true %}
-{% else %}
-{% set year_with_chapters = false %}
-{% endif %}
+{% set year_with_chapters = featured_chapters_exists(lang, year) or featured_chapters_exists('en', year) %}
 
 {% block breadcrumb %}
 <script type="application/ld+json">
--- a/src/templates/base/2019/index.html
+++ b/src/templates/base/2019/index.html
@@ -32,7 +32,7 @@
       <h1 class="title title-lg title-alt">{{ self.intro_title() }}</h1>
       <h2>{{ self.intro_sub_title() }}</h2>
       {{ self.mission() }}
-      {% if year < "2020" %}
+      {% if year < "2021" %}
       <a href="{{ url_for('table_of_contents', year=year, lang=lang) }}" class="btn">
         {{ self.start_exploring() }}
       </a>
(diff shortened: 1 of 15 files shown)


src/templates/base/2019/index.html
```diff
@@ -32,7 +32,7 @@
       <h1 class="title title-lg title-alt">{{ self.intro_title() }}</h1>
       <h2>{{ self.intro_sub_title() }}</h2>
       {{ self.mission() }}
-      {% if year < "2020" %}
+      {% if year < "2021" %}
```
rviscomi, line 35: Is there a way to use the template inheritance and macro overrides to solve this kind of pattern? I'd like to see year-specific templates extend previous years' templates and be the ones to explicitly declare which parts they want to be different.

src/templates/base/2019/index.html
```diff
@@ -84,16 +84,16 @@ <h3>{{ localizedChapterTitles[featured_chapter] }}</h3>
         </div>
         {% endif %}
       </div>
-      <a href="{{ url_for('chapter', year=2019, chapter=featured_chapter, lang=lang) }}" class="btn">
+      <a href="{{ url_for('chapter', year=year, chapter=featured_chapter, lang=lang) }}" class="btn">
```
rviscomi, line 87: For things that are year-agnostic, should they be in the base/index.html template?

bazzadp: Now we only have `templates/base/2019/` templates and no `templates/base/2020/` templates at all (since I got rid of the `2020/index.html` one in #1441 ), the `templates/base/2019/` basically are the year-agnostic ones.

There definitely is an argument to move these up a folder and remove the `templates/base/2019/` folder completely. But not as part of this PR. Will do that once there are no PRs open with base template changes.

src/templates/base/2019/index.html
bazzadp: Will have a look. Also did think about removing it completely but thought it might be useful for next year, but yeah you're right - must be a better way to do this.

bazzadp: I went a different way with this. - see https://github.com/HTTPArchive/almanac.httparchive.org/pull/1452/commits/7e354c593c11d37e4928b670241f0ea82c5077ef

Means a few more changes but sets us up nicely for 2021

src/templates/base/2019/index.html
```diff
@@ -4,6 +4,12 @@
 {% block image_height %}562{% endblock %}
 {% block image_width %}715{% endblock %}
 
+{% if featured_chapters_exists(lang, year) or featured_chapters_exists('en', year) %}
+{% set year_with_chapters = true %}
+{% else %}
+{% set year_with_chapters = false %}
+{% endif %}
```
rviscomi, line 11: Are these the same thing?
```suggestion
{% set year_with_chapters = featured_chapters_exists(lang, year) or featured_chapters_exists('en', year) %}
```

bazzadp: Yup this works. Thanks!

rviscomi (approved):

Issue #1402, opened 2020-11-03
Translate 2019 Content to Hindi


bazzadp:
These are the core templates - without which we cannot release any translated chapters. They are in the [language specific templates directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/templates):
- [x] base.html (2019) - @PraveenPal4232 
- [x] base.html (2020) - @PraveenPal4232 
- [ ] base_chapter.html- @PraveenPal4232 
- [x] contributors.html (2019) - @PraveenPal4232 
- [x] contributors.html (2020) - @PraveenPal4232 
- [x] error.html- @PraveenPal4232 
- [x] index.html (2019) - @PraveenPal4232 
- [x] index.html (2020) - @PraveenPal4232 
- [x] table_of_contents.html (2019) - @PraveenPal4232 
- [x] table_of_contents.html (2020) - @PraveenPal4232 

These are the chapters to be translated, in rough order of popularity. They exist in the [content directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content):
- [ ] javascript - 
- [ ] css - 
- [ ] seo - 
- [ ] performance - 
- [ ] http2 - 
- [ ] markup - 
- [ ] third-parties - 
- [ ] fonts - 
- [ ] accessibility - 
- [ ] pwa - 
- [ ] cms - 
- [ ] security - 
- [ ] media - 
- [ ] mobile-web - 
- [ ] cdn - 
- [ ] ecommerce - 
- [ ] page-weight - 
- [ ] caching - 
- [ ] compression - 
- [ ] resource-hints - 

Additionally the following pages need translated too in the [language specific templates directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/templates):
- [ ] accessibility_statement.html - 
- [ ] base_ebook.html- 
- [ ] methodology.html 2019 - 
- [ ] methodology.html 2020 - 

Please include "Makes progress on #1402 " in all pull requests so a link is created from the PR to this issue.

There is no need to translate the chapters HTML pages in `src/templates/en/2019/chapters/` nor the `src/templates/en/2019/ebook.html` file as they are generated off the markdown combined with the above templates.

---

Common notes for writing consistency are here: https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Translators'-Guide. Feel free to edit that and/or add Hindi-specific extras by editing this comment.

Hindi specific extra advice:

*

[...]

PR #1458, opened 2020-11-09
Translation of base.html to Hindi

PraveenPal4232:
Translation of base.html 2019 to Hindi
For issue #1402

### DIFF
--- a/src/templates/hi/2019/base.html
+++ b/src/templates/hi/2019/base.html
@@ -1,6 +1,6 @@
 {% extends "base/2019/base.html" %}
 
-{% block description %}The Web Almanac वेब रिपोर्ट की एक वार्षिक स्थिति है जो वेब समुदाय की विशेषज्ञता को HTTP संग्रह के डेटा और रुझानों के साथ जोड़ती है।{% endblock %}
+{% block description %}Web Almanac वेब रिपोर्ट की एक वार्षिक स्थिति है जो वेब समुदाय की विशेषज्ञता को HTTP संग्रह के डेटा और रुझानों के साथ जोड़ती है।{% endblock %}
 
 {% block twitter_image_alt %}{{ year }} Web Almanac{% endblock %}
 
@@ -28,7 +28,7 @@
 {% block page_navigation %}पेज नेविगेशन{% endblock %}
 
 {% block help_translate %}अनुवाद में मदद करें{% endblock %}
-{% block translation_not_available %}अंग्रेजी - अंग्रेजी में उपलब्ध नहीं है{% endblock %}
+{% block translation_not_available %}अंग्रेजी - हिन्दी में उपलब्ध नहीं है{% endblock %}
 {% block language_switcher %}भाषा परिवर्तन करें{% endblock %}
 {% block year_switcher %}वर्ष परिवर्तन करें{% endblock %}
 
@@ -85,10 +85,14 @@
 {% block accessibility_statement %}अभिगम्यता विवरण{% endblock %}
 
 {% block featured_chapter %}विशेष अध्याय{% endblock %}
+{% block featured_chapter_last_year %}विशेष अध्याय<br/>{{ year|int -1 }} Web Almanac से {% endblock %}
 
 {# Check if read_chapter already defined in child template as macros can't be overridden #}
 {% if not read_chapter %}
 {% macro read_chapter(chapter) %}अध्याय <span class="featured-chapter-name">{{ chapter }}</span> पढ़ें{% endmacro %}
+{% endif %}
+{% if not read_last_years_chapter %}
+{% macro read_last_years_chapter(chapter) %}अध्याय {{ year|int-1 }} <span class="featured-chapter-name">{{ chapter }}</span> पढ़ें{% endmacro %}
 {% endif %}
 
 {% block contributors_description %}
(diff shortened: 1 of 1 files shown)


src/templates/hi/2019/base.html
```diff
@@ -0,0 +1,188 @@
+{% extends "base/2019/base.html" %}
+
+{% block description %}The Web Almanac वेब रिपोर्ट की एक वार्षिक स्थिति है जो वेब समुदाय की विशेषज्ञता को HTTP संग्रह के डेटा और रुझानों के साथ जोड़ती है।{% endblock %}
```
bazzadp, line 3: Should "The" be translated?

src/templates/hi/2019/base.html
```diff
@@ -0,0 +1,188 @@
...
+{% endblock %}
+
+{% block read_last_years_almanac %}{{ year | int - 1 }} Web Almanac को पढ़िए{% endblock %}
+
+{% block http_archive_link %}HTTP Archive होम{% endblock %}
+
+{% block page_navigation %}पेज नेविगेशन{% endblock %}
+
+{% block help_translate %}अनुवाद में मदद करें{% endblock %}
+{% block translation_not_available %}अंग्रेजी - अंग्रेजी में उपलब्ध नहीं है{% endblock %}
```
bazzadp, line 31: According to Google Translate this Translates as "English - not available in English"

It should be "In English - not available in Hindi" and is added after chapters which have not been translated yet.

For example [in Spanish it says](https://almanac.httparchive.org/es/2019/table-of-contents#part-1):

> (en inglés - no disponible en español)

Which means "In English - not available in Spanish"

src/templates/hi/2019/base.html
```diff
@@ -0,0 +1,188 @@
...
+{% block footer_title %}फुटर नेविगेशन{% endblock %}
+{% block copyright %}© Web Almanac. <a href="https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/LICENSE">Apache 2.0</a> के तहत लाइसेंस प्राप्त है।{% endblock %}
+{% block accessibility_statement %}अभिगम्यता विवरण{% endblock %}
+
+{% block featured_chapter %}विशेष अध्याय{% endblock %}
+
+{# Check if read_chapter already defined in child template as macros can't be overridden #}
+{% if not read_chapter %}
+{% macro read_chapter(chapter) %}अध्याय <span class="featured-chapter-name">{{ chapter }}</span> पढ़ें{% endmacro %}
+{% endif %}
```
bazzadp, line 92: New changes we're about to add in #1452 

I used Google Translate so can you check these?

```suggestion
{% block featured_chapter %}विशेष अध्याय{% endblock %}
{% block featured_chapter_last_year %}विशेष अध्याय<br />2019 Web Almanac से{% endblock %}

{# Check if read_chapter already defined in child template as macros can't be overridden #}
{% if not read_chapter %}
{% macro read_chapter(chapter) %}अध्याय <span class="featured-chapter-name">{{ chapter }}</span> पढ़ें{% endmacro %}
{% endif %}
{% if not read_last_years_chapter %}
{% macro read_last_years_chapter(chapter) %}{{ year|int -1}} अध्याय <span class="featured-chapter-name">{{ chapter }}</span> पढ़ें{% endmacro %}
{% endif %}
```

bazzadp (changes requested):
Looks very good! A few things I spotted though.

Also if you merge latest main into this branch then that will solve the code errors being highlighted (they are completely unrelated to your changes in case you were worried!)

src/templates/hi/2019/base.html
PraveenPal4232: "The" article doesn't make sense in Hindi Literature
Here we have two options we can ignore "The" as I did and also we can remove "The" from here.

Let's take an Example

The Taj Mahal (In English)
ताजमहल (In Hindi)

src/templates/hi/2019/base.html
PraveenPal4232: Ok, got it :)

src/templates/hi/2019/base.html
bazzadp: Will take your guidance on this as to what reads best. Here's what we said in the [Translator's Guide](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Translators'-Guide):

> English product names (Web Almanac, HTTP Archive) should remain in English but can be adjusted slightly to local version (e.g. Le Web Almanac rather than The Web Almanac). For some languages this rule may be ignored if there are language specific character sets (e.g. Japanese might use katakana, hiragana, or kanj rather than English if that makes more sense).

PraveenPal4232: Yes, I have read that. According to my "Web Almanac" would be best here.
Because "The" article doesn't make sense in Hindi Literature.

bazzadp: Sounds good to me!

PraveenPal4232: Great will do it :)

src/templates/hi/2019/base.html
PraveenPal4232: Will do that in my next commit :)

bazzadp (approved):
LGTM - let me know if you're happy for me to merge or are still adding changes?

PraveenPal4232:
> LGTM - let me know if you're happy for me to merge or are still adding changes?

Of course, please do that. I am done. Thank

bazzadp:
Done. Thank you for this contribution!

That `base.html` is a tricky one as includes lots of snippets and macros used by the site. It's a key one though so good to get it done.

Looking forward to getting the rest of the core templates so we can more on to the more interesting chapters!

Issue #1445, opened 2020-11-07
Support code formatting for headings in chapter outline sidebar


rviscomi:
![image](https://user-images.githubusercontent.com/1120896/98449729-f7be6280-2103-11eb-92a7-1f45d9ced99c.png)

The heading "`viewport` specifications" includes a code block for the word "viewport". This block is formatted in the heading itself, but not its entry in the chapter outline sidebar, making it appear as if the word is plain and should be capitalized for consistency with other headings.

Headings in the outline should retain the code formatting.

bazzadp:
HTML is not supported in ToC as discussed in #656 .

Will have another look to see if anything's changed since then or if I can think of a better way of supporting it than I came up with last time.

rviscomi:
Good find, yeah it'd be nice to revisit this to see if we can support similar use cases.

bazzadp:
Actually it was supported then, but just messed up the anchor link. So not exactly the same. But almost certainly related. Anyway will have a look if no one grabs it before me.

bazzadp:
Turns out this was easier than I thought. Shown doesn't support HTML in Headings for ToC (so #656 is still a problem) but it does support Markdown so we can solve this.

PR coming up...

PR #1464, opened 2020-11-09
Update headers with code or English language

### LINKED ISSUE #656: i18n - `lang` attribute usage in titles are incompatible with the chapter toc
In some headings, I need to keep english words (because we usually don't translate them in French). For example, `Le <i lang="en">lazy loading</i> natif`.

If I do that in the markdown template, in a title:

```
## Le <i lang="en">lazy loading</i> natif
```

The script that generates the chapter table of content doesn't keep it and generates a strange anchor: 

```
<li>
    <a href="#le-i-langenlazy-loadingi-natif">Le lazy loading natif</a>
</li>
```

To fix the content, we would need to amend this line, in order to get the `innerHTML` instead of the textContent.

https://github.com/HTTPArchive/almanac.httparchive.org/blob/05862f757558393ee8ce7fb6efca737e603b831a/src/tools/generate/generate_table_of_contents.js#L23

and authorize HTML content in toc.html for the title…

To fix the anchor, I think it's relative to the markdown generation in `generate_chapters.js`, I don't think we could fix this easily.

bazzadp:
Fixes #656 as raised a long time ago by @borisschapira 

With the addition of Markdown support in Headings (#1445) in pull request #1461 I wanted to have another look at this old issue complaining about weird ids when using HTML in headers - particularly `<i lang="en">...</i>` for accessible translations.

As a reminder of the issue:

While Showdown does support (and correctly remove) markdown in headings to create the ids, it will not do this for HTML.

So this:
```
### Le <i lang="en">lazy loading</i> natif {le-lazy-loading-natif}
```
Gets an id of: `le-i-langenlazy-loadingi-natif`. It's still a valid id (it's correctly stripped the `<` and `>` symbols which would make this invalid) but looks very messy.

However there is another option (`customizedHeaderId`) which when turned on, allows you to optionally specify the header id after the header like this:

```
### Le <i lang="en">lazy loading</i> natif {le-lazy-loading-natif}
```

The risks of this are low: if it is not used the auto generated id will be used as normal, and duplicate ids are checked for during HTML linting. In fact we already have one chapter (French Accessibility) that's already using this incorrectly so we're already getting this wrong. This just offers a way of doing it right.

So I think we should enable this as an option for translators, which this PR does as well as updating the headings I could find from a quick look - also set code headings for caching chapter in all languages.

Let me know your thoughts.

### DIFF
--- a/src/content/en/2019/caching.md
+++ b/src/content/en/2019/caching.md
@@ -258,7 +258,7 @@ The same data for mobile is shown below. As can be seen, the cacheability of con
 }}
 
 
-## Cache-Control vs Expires
+## `Cache-Control` vs `Expires`
 
 In HTTP/1.0, the `Expires` header was used to indicate the date/time after which the response is considered stale. Its value is a date timestamp, such as:
 
@@ -282,7 +282,7 @@ HTTP/1.1 introduced the `Cache-Control` header, and most modern clients support 
   )
 }}
 
-## Cache-Control directives
+## `Cache-Control` directives
 
 The HTTP/1.1 [specification](https://tools.ietf.org/html/rfc7234#section-5.2.1) includes multiple directives that can be used in the `Cache-Control` response header and are detailed below. Note that multiple can be used in a single response.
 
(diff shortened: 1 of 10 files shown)


rviscomi (approved):
LGTM but I think the original suggestion to automatically ignore markup/down in the heading is still worth pursuing.

bazzadp:
Showdown sets the ids and ignores Markdown for them. However it doesn’t for HTML.

I went as far as looking at the code involved to submit a PR to showdown and it’s not trivial. Plus [the core maintainer has taken a break from the project due to Covid](https://github.com/showdownjs/showdown/issues/833) so even if I got a fix it’s unlikely to be accepted and released any time soon.

Cleaning it up after Showndown has converted is a bit flakey as discussed in #656 

So think this is the best we’ve got for now! Plus I deem it low effort, low risk, and no worse than status quo if people don’t use it.

Will leave open for opinions from @borisschapira if he gets a chance to look and then merge later today.

PR #1477, opened 2020-11-10
Add chapter links heading and link

bazzadp:
I often find myself want to direct people to the chapter links at the bottom of each chapter:

Chapter Links screenshot

I have to say "Below the Conclusion" or "Above the Authors"

This adds a header that can be linked to directly:

Chapter Links with header screenshot

Hover over it to get the link as usual:

Chapter Links with header hover screenshot

@HTTPArchive/translators if any of you get a chance to review Google Translate's work that would be very much appreciated!


rviscomi:
Does this make it show up in the TOC? I don't think it does but just checking.

src/templates/en/2019/base_chapter.html
```diff
@@ -10,6 +10,7 @@
 
 {% block prev_next_title %}Previous and next chapter navigation{% endblock %}
 
+{% block chapter_links %}Chapter Links{% endblock %}
```
rviscomi, line 13: Is this technically an appendix?

bazzadp:
> Does this make it show up in the chapter outline/TOC? I don't think it does but just checking.

No it doesn't. It's the same as Authors.

There's an argument maybe it should? But think the actual ToC entries are more important so I would leave as is.

bazzadp:
They are also smaller in size to reflect lesser importance compare to the ToC entries - compare Conclusion to Chapter Links/Author here:

Headers

Despite the fact they are smaller visually, they are both `<h2>` as you shouldn't skip headers (they used to be `<h4>` IIRC correct changed but kept smaller styling) which is a slight inconsistency.

Maybe we should have an `<h2>Appendices</h2>` with `<h3>Chapter Links</h3>` and `<h3>Authors</h3>` underneath them? However I also don't like double headings and can't think of what to put between "Appendices" and "Chapter Links" headings in that case to avoid that. Plus the reuse of "Appendices" terminology pointed out above - would be especially confusing in the ebook.

rviscomi:
Changing direction from "Appendix", what about naming this section more as a CTA? For example "Explore the results" or "Continue the discussion" etc.

bazzadp:
> Changing direction from "Appendix", what about naming this section more as a CTA? For example "Explore the results" or "Continue the discussion" etc.

Oh quite like that! Prefer  "Explore the results" as "Continue the discussion" only really applies to the first link.

bazzadp:
Changed:

Explore the results

rviscomi (approved):

PR #1493, opened 2020-11-11
Add ability to run the linter locally

### LINKED ISSUE #1486: Add instructions and on how to run super-linter locally
https://github.com/HTTPArchive/almanac.httparchive.org/pull/1393#pullrequestreview-527958173

We can run a super-liner running in github actions locally.
I add the how to run to README.md.

https://github.com/github/super-linter/blob/master/docs/run-linter-locally.md

Optionally:
 - Do you have an idea to exclude unwanted folders such as node_modules from the target?

bazzadp:
Fixes #1486 

@MSakamaki I hope you don't mind but I went ahead and completed this. You can run it by doing the following:

```
npm run lint
```

Or to provide the files to directories to lint on the command line:

```
npm run lint tools/generate templates/base
```

Could you test locally and let me know how you get on?


bazzadp:
Note above warning is an expected change. We've added an HTMLHint hint to these pages to allow them to be linted.

bazzadp:
@ibnesayeed can I ask for your review on this too since you're good at Docker and shell scripting?

ibnesayeed (commented):
I think its good the way it is implemented, but if I were to do it, I would take a different approach here. Rather than having platform specif scripts that set things up and then execute a docker container, I would have written a custom entrypoint shell script and kept it in the repo and use it when running the container while mounting the current working directory (i.e., the repo). The script will then take necessary steps to copy stuff in temporary folders and run the underlying linter script from that context (something like the following).

```
$ docker container run -it --rm -v "$PWD":/app -w /app/src --entrypoint=./linter-entrypoint.sh github/super-linter
```

And the `linter-entrypoint.sh` can be something like:

```
#!/usr/bin/env bash

# set all the necessary environments variables
export ENV1=val1
export ENV2=val2

# Copy files/folders from the mounted repo at /app into /tmp/
# Remove unnecessary files and move things around as needed
# Change working directory to the folder inside of /tmp/

# Call the underlying linter script that is the default entrypoint in the docker image
/action/lib/linter.sh "$@"
```

bazzadp:
That probably would have been simpler and saved me having to figure out how the heck to do things in stupid bat files!

My only thing against this is I don't like is the dependency this creates on the internal workings of the super-linter. If they change anything internally, then our script stops working.

I'm hoping all this copying of files is a short term workaround and once they fix this to exclude the files as they should, we can just lint in the checked out repo. At that point the shell scripts become a lot simpler and this is less of an issue.

So I'm comfortable leaving as is for now. But thanks for the suggestion.

ibnesayeed:
> That probably would have been simpler and saved me having to figure out how the heck to do things in stupid bat files!

When I read the description of the PR that things are running in a container and then saw the batch file, I said to myself, why would we do this? One motivation of running thing in containers is to forget about platform specific hacks. By the way, the bat file looks and feels right, but I have ZERO experience of Windows shell.

> My only thing against this is I don't like is the dependency this creates on the internal workings of the super-linter. If they change anything internally, then our script stops working.

This is a valid concern, but I would not worry too much about it because such things don't really change that often. The `ENTRYPOINT ["/action/lib/entrypoint.sh"]` instruction in the Dockerfile has been there unchanged since their initial commit (as per the git blame).

> So I'm comfortable leaving as is for now. But thanks for the suggestion.

I am sure you have put a lot of efforts in this, so if it is working, it can be left as is for now. However, adding a comment somewhere to point to this discussion or creating a ticket to refactor this in the future would be a good idea.

ibnesayeed:
@bazzadp I know you do not play a lot of Docker game, so I think this will be a an interesting read for you to know how [CMD and ENTRYPOINT interact in Docker](https://docs.docker.com/engine/reference/builder/#understand-how-cmd-and-entrypoint-interact), unless you have read it already.

bazzadp:
OK @ibnesayeed you've convinced me. Wasn't aware you could run other scripts within someone else's Docker container by overriding the default EntryPoint.

Have updated the PR. Could you have another look?

src/README.md
```diff
@@ -96,6 +96,31 @@ npm run watch
...
+First up pull the Super-Linter Docker image (this takes some time to download but only need to do this once or when you want to upgrade the version of the super-linter):
+
+```
+docker pull github/super-linter:latest
+```
+
+Then to run the linting do this:
+
+```
+npm run lint
```
ibnesayeed, line 114: Is is a fair assumption that a contributor will have NPM installed? If not, then we might want to surface the Docker command here which this is a wrapper of.

src/tools/scripts/run_linter_locally.sh
```diff
@@ -0,0 +1,70 @@
...
+export VALIDATE_CSS=true
+export VALIDATE_HTML=true
+export VALIDATE_JAVASCRIPT_ES=true
+export VALIDATE_JSON=true
+export VALIDATE_MD=true
+export VALIDATE_PYTHON_PYLINT=true
+export VALIDATE_PYTHON_FLAKE8=true
+export VALIDATE_YAML=true
+
+echo "Starting linting"
```
ibnesayeed, line 69: How does the linter know that want it to lint files under the `/tmp/` folder and not under the `app` folder? Do we need to change the working directory to `/tmp/lint/src`?

src/package.json
```diff
@@ -19,6 +19,9 @@
     "watch": "node ./tools/generate/chapter_watcher",
     "ebooks": "node ./tools/generate/generate_ebook_pdfs",
     "deploy": "./tools/scripts/deploy.sh",
+    "lint": "run-script-os",
+    "lint:darwin:linux": "docker container run -it --rm -v $PWD/..:/app -w /app/src --entrypoint=./tools/scripts/run_linter_locally.sh github/super-linter",
```
ibnesayeed, line 23: It is generally a good idea to place double quotes around `"$PWD"` when used in commands like these to avoid unintentional argument split in case any parent directory name has spaces in it.

src/README.md
bazzadp: I think this is a fair prerequisite for most developers. For those working in a pure Docker environment (like I know you like to do) with no local install they can always get the full command from `package.json` and set up an alias. I think I'd rather keep the README clean.

Maybe I could add a piece under the Docker section but leave this as is?

ibnesayeed:
> Wasn't aware you could run other scripts within someone else's Docker container by overriding the default EntryPoint.

Yes, you can do so by a few different ways:

* By overriding the entrypoint at runtime using the `--entrypoint` flag
* By creating a `Dockerfile` that inherits from the original one and overriding the `ENTRYPOINT` directive

src/README.md
ibnesayeed: Yes, I am okay with this.

src/tools/scripts/run_linter_locally.sh
bazzadp: The Linter always lints files under `/tmp/lint`. Normally you would map that directly to your git directory (which is what I originally had), but it's only because we override that that this is even an issue.

Also note the Linter needs to be able to see the `.github` directory to get our linter rule configurations so definitely can't change it to `src`.

src/README.md
bazzadp: Done!

src/package.json
bazzadp: Done

src/tools/scripts/run_linter_locally.sh
ibnesayeed: Got it! I didn't know they have a magic location in place as the default. It perhaps is a good idea that linting is happening elsewhere, in case the process yields any files or changes the state of existing ones. Especially, when running in Docker and proper UID mapping is not in place, the process may create files under the mounted folder with root permissions.

ibnesayeed (approved):
LGTM!

src/tools/scripts/run_linter_locally.sh
bazzadp: Yeah will you'd think. But not sure it's that safe - I managed to delete my whole repo by mapping `/tmp/lint/` to the actual directory and then my script deleted that - oops! Ah well was time for a clean up anyway and everything was checked in on GitHub so luckily didn't lose anything!

 So just because it's in Docker doesn't mean it can't delete or overwrite the files. Still got to trust what we're doing, when adding this to our toolset (and it's already running in a GitHub Action). Plus that's what Git is for - to store the main copy offline!

bazzadp:
@MSakamaki I'm gonna go ahead and merge this since we've had a thorough review from @ibnesayeed . Would love to hear your thoughts on it when you next get a chance to try it out.

MSakamaki:
Thanks for implementation @bazzadp!
let us help solve your developer issues more.

PR #1494, opened 2020-11-11
2020 JS SQL

rviscomi:
Progress on #897 

- Histogram of uncompressed JS response sizes broken down by desktop/mobile and 1P/3P

The histogram bins are 5 KB wide, which can be adjusted as needed. Most of the interesting data is < 5 KB so maybe it's worth breaking that down more and ignoring the tail? @tkadlec let me know what you'd like to do.


rviscomi:
Merging to keep it moving and we can reopen new PRs with tweaks as needed.

Issue #1229, opened 2020-08-19
Increase pytest coverage


bazzadp:
@Super-Fly did some excellent work in #949 and #1176 to refactor out Python code.

We should look to increase our pytest coverage to test some of these function/helper files.

- [x] - Add Coverage reports - @bazzadp - #1555 
- [x] - `server/init.py` (69%) - @bazzadp - #1853
- [x] - `server/config.py` (96%) - @AbbyTsai - #1605
- [x] - `server/errors.py` (65%) - @bazzadp - #1853
- [x] - `server/helpers.py` (31%) - @bazzadp - #1576
- [x] - `server/language.py` (78%) - @bazzadp - #1853
- [x] - `server/routes.py` (58%) - @bazzadp - #1607 
- [x] - `server/validate.py` (60%) - @bazzadp - #1523

Latest coverage report (will update this as we merge each PR):

```
---------- coverage: platform darwin, python 3.8.0-final-0 -----------
Name                            Stmts   Miss  Cover   Missing
-------------------------------------------------------------
server/__init__.py                 39      1    97%   20
server/config.py                   69      0   100%
server/csp.py                       1      0   100%
server/errors.py                   23      7    70%   11, 16-17, 27-28, 33-34
server/feature_policy.py            1      0   100%
server/helpers.py                 126      0   100%
server/language.py                 33      1    97%   10
server/routes.py                   72      0   100%
server/tests/__init__.py            0      0   100%
server/tests/config_test.py        37      0   100%
server/tests/helpers_test.py      184      0   100%
server/tests/routes_test.py        69      0   100%
server/tests/validate_test.py      52      0   100%
server/validate.py                 79      5    94%   39-41, 66-67
-------------------------------------------------------------
TOTAL                             785     14    98%

Required test coverage of 95% reached. Total coverage: 98.22%
================================== 105 passed, 21 warnings in 6.79s ===
```

bazzadp:
Adding a coverage checker in #1555 and Editing the first comment to add the tasks that need completing.

Would love to get to 100% coverage and think that's easily achievable.

We can use `# pragma no cover` if we absolutely have to to exclude lines we don't need to test (e.g. Flask calls), but ideal we add at least one test for every line and [Flask gives guidance on how to test Flask apps](https://flask.palletsprojects.com/en/1.1.x/testing/).

rviscomi:
[...]

Issue #1222, opened 2020-08-17
Should we change to a static site?


bazzadp:
@HTTPArchive/developers , another question for you.

This question has come up (again) with some chats with you and so I'm opening this issue for discussion.

_Note: this has been discussed before in #634 and we closed this issue. I've decided to open a new issue, rather than reopen that last one, as that last one got a little sidetracked on other GitHub issues._

The Web Almanac is basically a static site, but has a full Flask-based Python engine using Jinja2 templates to display these pages. Some have suggested we should move to a purely static site, or alternative static site generators (e.g. [11yy](https://www.11ty.dev/) or some other Jamstack), which would potentially simplify our site, potentially lead to a performance boost and also potentially lead to other hosting opportunities (e.g. hosting directly on GitHub).

From my point of view, I can see the arguments for and against, which I've tried to summarise below:

On the one hand, I would say it has the following areas for potential improvement:
- There is indeed a (slight) performance issue for the first run of the site as the Python server is launched. This is noticeable when you run Lighthouse for a little used page, and then rerun Lighthouse and get a better score. We still score over around 90 for most of our pages, which is still a very good score and it's questionable whether users would notice the performance improvement. For a "hot" server, the performance impact seems really difficult (impossible?) to notice, and either way the results are cached for 3 hours on the Google Cloud Platform CDN at which point it is a static site.
- The server is built on different technologies (Python, Flask, Jinja2), than the front end (HTML, CSS and JavaScript) and the build scripts (JavaScript and Bash shell scripts). It may be better to consolidate our tech stack on JavaScript to allow a greater movement between front end and back end developers.
- We don't yet have CI/CD to allow pushing to production just by merging a change (though @rviscomi have talked about this recently and we believe this is possible to do given our improved automation and testing completed this year).

On the other hand, what we have is working well, we've had no major issues or items that we can't support on the site and functionality has been relatively easy to add (personally I had no experience of Python, Flask or Jinja2 before working on this project but have contributed a huge number of changes to that code). The refactor of Python in #949 and #1176 has helped improved the Python code no end, and the templating system we have (although seemingly [quite convoluted at first](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Developers'-Guide#jinja2ejs-templates)) does allow us great support of other languages.

[...]

PR #1508, opened 2020-11-13
Update timestamps at release time

### LINKED ISSUE #1224: Keeping track of Last Updated timestamps
We have a need to track the last update time for a page.

This is done for a few different reasons (mostly to do with SEO):
- We have [Article Structured Data](https://developers.google.com/search/docs/data-types/article#article-types) on each page and the `dateModified` is [recommended](https://developers.google.com/search/docs/data-types/article#article-types) - and basically shows as an error if you don't have it.
- Our [sitemap lists it](https://almanac.httparchive.org/sitemap.xml) which Search Engines can use to prioritise which pages to crawl.
- We even talked [in the past about exposing this to the users](https://github.com/HTTPArchive/almanac.httparchive.org/issues/518#issuecomment-555663257)
- We may use this if we ever implement the RSS feed discussed in #1079 

Now it took us a while to get this working, and we use a [GitHub Action](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/.github/workflows/generate_chapters.yml) which runs on merges to `main` branch, runs [some fancy `sed` commands](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/src/tools/scripts/update_timestamps_versions.sh) and then opens a pull request to commit the changes. See #1186 as an example.

It also updates the CSS and JS hash to give it a unique value based off of timestamp for cache-busting reasons (e.g. `static/css/2019.css?v=20200730202229`), but I'm less worried about that.

This may sound a bit over the top and convoluted but has several advantages:
- It's automated so doesn't depend on anyone updating the timestamp when making a change
- Doing it after merging avoids merge conflicts if two PRs are altering the same code so it doesn't matter which order they are merged in.
- It doesn't depend on any time settings on people's machines.
- Timestamps should only go forward.

However the big disadvantage is that it produces a second pull request for every commit that changes which is irritating, noisy, and pollutes our git history. We also tried to turn on branch protections in GitHub so tests must pass, but PRs coming from GitHub Actions don't kick off other Github Actions (to avoid recursions!) so they don't run so must be force pushed 😞

**Do we even need to do this?**

My experience is yes - it's good to have a last updated time for an article for all the reasons given above and more. Setting this all the same to the deploy time is just wrong to me (especially as we launch 2020 and future years and likely update past years content less, and less). We also have had indexing problems in the past (see #884) and have still not cleaned these up two months later (though only one 2019 chapter remains unindexed at this time - ironically the French SEO chapter 😀 ). So, despite being small, Google doesn't crawl our entire site like it could and we appear to have a very small crawl budget - so we should prioritise recently updated content.

**Is there a better way to handle the timestamps?**

In the past we tried to do it based on GitHub commit time (see #317), but as soon as we commit that obviously updates that and problems ensue. Perhaps if we do #1221 and stop generating the site, then we should tackle this at the same time and  just grab the last committed timestamp direct from the `origin/main` branch, update the deployment, but NOT commit it back to Git? It feels a little flakey basing this on Git commit timestamps though but, on the other hand, it's not super critical if it ever goes wrong.

Any other ideas?

bazzadp:
Fixes #1224 

This moves the tracking of timestamps from within the `template` html files and content `md` files to a JSON file that only needs to be updated once, on release night. This avoids the repeated pull requests every time you check something in.

At moment, to make review easier, I haven't removed the timestamp from all the templates and content md's. Will look to that once approved (though might leave 2020 content for now to avoid confusing the authors).


bazzadp:
Above is all. These are expected diffs.

rviscomi:
Nice work! I've been wanting this for a while.

I've only skimmed the PR so far but one thing I noticed is that the cache-busting hashes are rendered at load time. We used to render/bake them into the HTML at build time. Why do the work on the web server when it's static?

bazzadp:
> I've only skimmed the PR so far but one thing I noticed is that the cache-busting hashes are rendered at load time. We used to render/bake them into the HTML at build time. Why do the work on the web server when it's static?

I presume by "build time" you mean when we rerun the "Generate Timestamps and Ebook" action and not for every `npm run generate`? The former could work, but the later would lead to needless iterations and merge conflicts - remember the hashes are needed for the index.html, methodology.html...etc. which are not rendered from `md` files.

Even if we did this as part of "Generate Timestamps and Ebook" action, we'd need to commit the template changes for those HTML pages, which could lead to needlessly iterating the templates and littering our git history (avoiding this is the what spurn me into action btw - accepting the additional PRs was never that burdensome to me). Consider `page.css` which is loaded by chapters, methodology, ebook and accessibility statement (as an aside maybe we should move some of the ToC css out of this into 2019.css to avoid this!). Anyway if you make an edit to `page.css` that's 4 templates you're submitting even if they have no other changes.

The last updated times are also "static" so by that rational, why not do them at build time too? Because it would lead to the same issues.

To do bother of them at build time, without these issues, would be if we moved more towards a static site as discussed in #1222 

At the end of the day the hash is another look up - same as timestamp. The values are already pre-calculated and loaded at server start up, so think the additional work is minimal.

src/server/helpers.py
```diff
@@ -195,6 +199,15 @@ def accentless_sort(value):
     return sorted(value, key=lambda i: strip_accents(i[1]).lower())
 
 
+def get_file_date_info(file, type):
+    timestamps_config = get_timestamps_config()
+    value = timestamps_config.get(file, {}).get(type)
+    # Default Published and Last Updated to today
+    if value is None and (type == "published_date" or type == "last_updated_date"):
+        value = datetime.datetime.utcnow().isoformat()
+    return value
```
rviscomi, line 208: Is this the same thing?

```suggestion
    # Default Published and Last Updated to today
    today = datetime.datetime.utcnow().isoformat()
    return timestamps_config.get(file, {}).get(type, today)
```

src/server/routes.py
```diff
@@ -11,7 +12,8 @@
 def home(lang, year):
     config = get_config(year)
     ebook_size_in_mb = get_ebook_size_in_mb(lang, year)
-    return render_template('%s/%s/index.html' % (lang, year), config=config, ebook_size_in_mb=ebook_size_in_mb)
+    return render_template('%s/%s/index.html' % (lang, year), config=config,
+                           get_file_date_info=get_file_date_info, ebook_size_in_mb=ebook_size_in_mb)
```
rviscomi, line 16: Why does the home page need to know how many MB the ebook is?

src/server/helpers.py
```diff
@@ -67,7 +69,8 @@ def render_error_template(error, status_code):
         elif os.path.isfile(TEMPLATES_DIR + '/%s/%s/error.html' % (DEFAULT_LANGUAGE.lang_code, DEFAULT_YEAR)):
             lang = DEFAULT_LANGUAGE.lang_code
             year = DEFAULT_YEAR
-    return render_template('%s/2019/error.html' % lang, lang=lang, year=year, error=error), status_code
+    return render_template('%s/2019/error.html' % lang, lang=lang, year=year, error=error,
+                           get_file_date_info=get_file_date_info), status_code
```
rviscomi, line 73: Rather than adding the function kwarg to each template, I think this qualifies for the [global Jinja variables list](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/src/server/__init__.py#L55):

Alternatively, what about baking the published/modified dates directly into the overridden [`render_template`](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/src/server/helpers.py#L9) function, similar to how we make supported years/languages globally available? That way there are no functions to call and no arguments to manage, eg templates can simply render `{{ last_modified_date }}`.

src/templates/base.html
```diff
@@ -5,9 +5,9 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     {% block head %}
       <title>{% block title %}The Web Almanac{% endblock %}</title>
-      <link rel="stylesheet" href="/static/css/normalize.css?v=20200630133215">
+      <link rel="stylesheet" href="/static/css/normalize.css?v={{ get_file_date_info('static/css/normalize.css', 'hash') }}">
```
rviscomi, line 8: The filename is duplicated in the `href` attr and `get_file_date_info` arg. Could we create a function like this to handle the versioning?

```py
def get_versioned_filename(path):
  version = get_file_date_info(path, 'hash')
  return '/%s?v=%s' % (path, version)
```

_Side note: why isn't the leading slash included in the config keys?_

That would make it possible to do this:

```suggestion
      <link rel="stylesheet" href="{{ get_versioned_filename('static/css/normalize.css') }}">
```

rviscomi (changes requested):

src/config/last_updated.json
```diff
@@ -0,0 +1,1080 @@
+{
+  "en/2019/": {
+    "published_date": "2019-11-04T12:00:00",
+    "last_updated_date": "2020-06-21T00:00:00.000Z",
```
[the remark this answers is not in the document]
bazzadp: Doesn't really matter but fixed. Can't remember why this was but suspect the way I parsed the published date to get the initial values.

Confirmed the script definitely generates Z values for both now, so updated all non-Z values to Z.

.github/workflows/generate_timestamps_ebooks.yml
```diff
@@ -5,7 +5,7 @@
 # Generate the ebooks when this GitHub Action is run manually
 # This should be run everytime before release
 #
-name: Generate Ebooks
+name: Generate Timestamps and Ebooks
```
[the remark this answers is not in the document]
bazzadp: Done.

src/server/helpers.py
bazzadp: Seems to work - thanks!

src/server/helpers.py
bazzadp: Good idea! Fixed.

We should do this for config also.

src/templates/base.html
bazzadp: Good call.

No reason for no leading slash other than the fact they are relative paths to from where the generate script is called. Have added leading slash to the `static` ones as confusing to web developers if not there. I've left it off templates as they are relative to template dir.

src/server/helpers.py
bazzadp: Needed a slight varient:

```
    # Default Published and Last Updated to today
    today = datetime.datetime.utcnow().isoformat()
    if type == 'hash':
        return timestamps_config.get(file, {}).get(type)
    else:
        return timestamps_config.get(file, {}).get(type, today)
```

src/server/routes.py
bazzadp: Every page now does since we show it in the ToC menu. Moved it to the `render_template` function as per below.

bazzadp: Added the ebook file size generation to JSON it's loaded at start up and doesn't need to be calced on the fly any more.

src/server/helpers.py
```diff
@@ -195,6 +192,29 @@ def accentless_sort(value):
     return sorted(value, key=lambda i: strip_accents(i[1]).lower())
 
 
+def get_file_date_info(file, type):
```
rviscomi, line 195: Is is possible to get tests for this, or am I pushing my luck? 😆

src/server/helpers.py
```diff
@@ -195,6 +192,29 @@ def accentless_sort(value):
...
+        return timestamps_config.get(file, {}).get(type, today)
+    else:
+        return timestamps_config.get(file, {}).get(type)
+
+
+def get_versioned_filename(path):
+    version = get_file_date_info(path, 'hash')
+    if version:
+        return '%s?v=%s' % (path, version)
+    else:
```
rviscomi, line 209: Should this ever happen? If not, should it fail more loudly as an exception? For example, if it means that the frontend is trying to render a non-existent static resource that will inevitably 404, an exception will help us catch it during testing.

src/server/helpers.py
```diff
@@ -42,8 +45,14 @@ def render_template(template, *args, **kwargs):
         if os.path.isfile(year_lang_template):
             template_supported_years.append(y)
 
+    date_published = get_file_date_info(template, "date_published")
+    date_modified = get_file_date_info(template, "date_modified")
+    ebook_size_in_mb = get_ebook_size_in_mb(lang, year)
```
rviscomi, line 50: I'm still confused why every page needs to know the ebook size in MB. Isn't this only needed by the TOC page?

src/server/helpers.py
bazzadp: Yeah I’ve been lazy on Python tests. Will see what I can add easily here. And if a huge effort then will wait until #1229 is tackled.

src/server/helpers.py
bazzadp: When you add a new CSS file but haven’t run the pre-release script then if won’t exist in config yet. So that’s one instance. In prod hopefully shouldn’t happen.

src/server/helpers.py
bazzadp: The ebook is now linked in the ToC menu which appears on every page. The size is needed for two reasons:
1. To decide whether to show the ebook or not. We don’t show in French 2019 for example as no ebook yet. Nor in any 2020 language - including English. Could have another flag but decided why have two when want size anyway.

2. As we link to the PDF (and I know you’ve raised #1517 to discuss this but personally I think we should leave as is) it’s rude to ask them to download a huge file without making them away of the size. And it doesn’t take up much space. So the size shows in the ToC menu in every page.

src/server/helpers.py
bazzadp: Added loads of tests for the new functions in https://github.com/HTTPArchive/almanac.httparchive.org/pull/1508/commits/d6883427afe28c5b7c17a8a33d16c7a602123bda

src/server/helpers.py
bazzadp: OK added an exception logging and confirmed it still returns fine so won't interrupt development.

src/server/helpers.py
bazzadp: Updated this to test all ebooks that are configured. Yeah!

rviscomi (approved):
Ship it! 🚀

PR #1393, opened 2020-10-30
Add table of contents dropdown list

MSakamaki:
I have add #985  features.

Issue #985, opened 2020-07-08
Table of Contents dropdown list


rviscomi:
I often find myself wanting to jump to a specific chapter. There are a few ways to do this currently:

- navigate via direct URL
- navigate to the TOC page and click the link
- move sequentially previous/next from chapter to chapter

I'd like to propose a new UI that works from any page to navigate directly to any given chapter. The "Table of Contents" link in the header could reveal a dropdown list of chapters on hover/focus. See the "Reports" list on the HTTP Archive website, for example:

![image](https://user-images.githubusercontent.com/1120896/86953921-24954580-c123-11ea-9ecb-a61049deb50f.png)

The contents of the list would be the chapters in order and grouped by section, as they appear on the TOC page. These chapters should be generated by the config and not hardcoded, to make it applicable for any year. It must also be fully accessible and follow best practices.

Some design work is needed. The list of chapters is long so we would need to design it with scrolling in mind. And we may want a mobile-specific design or, for example, rely on native UI for radio inputs.

Any @HTTPArchive/developers interested in building this?

MSakamaki:
I want to get to work on this.
I'll try to fix the table of contents to a dropdown list first.

rviscomi:
Great, thanks @MSakamaki! See #986 for the related style changes.

MSakamaki:
@rviscomi 
I implemented the dropdown as advised.
some refactoring and will submit a PR today.

PC Header
![pc-header](https://user-images.githubusercontent.com/2183053/97654112-9a903480-1aa5-11eb-96b2-8db14ea1e818.png)

PC Footer
![pc-footer](https://user-images.githubusercontent.com/2183053/97654120-9ebc5200-1aa5-11eb-817d-450a4d44de54.png)

Mobile Header
![mobile-header](https://user-images.githubusercontent.com/2183053/97654130-a1b74280-1aa5-11eb-9de9-0a013ddd4b02.png)

Mobile Header (select)
![mobile-header-select](https://user-images.githubusercontent.com/2183053/97654135-a4b23300-1aa5-11eb-9b76-796c5f7d8e5d.png)

Mobile Footer
![mobile-footer](https://user-images.githubusercontent.com/2183053/97654146-a7ad2380-1aa5-11eb-8e5a-214a5d56caed.png)

bazzadp:
Fixed in #1393

rviscomi:
[...]

PR #1512, opened 2020-11-13
Misc fixes for ToC dropdown

bazzadp:
Misc fixes from ToC drop don added in #1393 for #985

I was running the pages through an HTML validator for another issue and it was complaining about some invalid HTML as we are using an `<a>` element inside of a `<button>` which is not allowed. I've pulled it out as a separate element and use JS to show/hide whichever as appropriate. Wish the HTML linter found these :-(

I've also hidden the dropdown arrows until JS kicks in, and done same for mobile menu.

I also fixed some of the issues discussed in https://github.com/HTTPArchive/almanac.httparchive.org/issues/985#issuecomment-727235918:
- Open ToC menu changes from tick to up arrow
- Current chapter should not be clickable
- Highlight current page with tick and yellow colouring
- Opacity change as was too light.

### DIFF
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -1414,8 +1414,8 @@ div.js-show,
   flex-direction: column;
 }
 
-.nav-dropdown-list-chapter.ebook small {
-  padding: 0 0 24px 20px;
+.nav-dropdown-list-chapter.ebook {
+  padding: 0 0 24px 24px;
 }
 
 .nav-dropdown-list-todo,
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -972,6 +972,10 @@ p.copyright a {
   display: none;
 }
 
+button.js-unhide {
+  display: none;
+}
+
 #skiptocontent a {
   padding: 6px;
   position: absolute;
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -659,13 +659,6 @@ header .cta {
   transform: rotate(135deg);
 }
 
-.js-switcher-menu::after {
-  height: 0;
-  width: 0;
-  margin-top: 0;
-  border: 0;
-}
-
 .table-of-contents-switcher select,
 .language-switcher select,
 .year-switcher select {
@@ -979,8 +972,7 @@ p.copyright a {
   display: none;
 }
 
-button.js-unhide,
-select.js-unhide {
+div.js-show {
   display: none;
 }
 
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -1431,7 +1431,9 @@ div.js-show,
 }
 
 @media (prefers-reduced-motion: reduce) {
-  *, ::before, ::after {
+  *,
+  ::before,
+  ::after {
     animation-delay: -1ms !important;
     animation-duration: 1ms !important;
     animation-iteration-count: 1 !important;
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -1291,7 +1291,8 @@ div.js-show,
   outline: revert;
 }
 
-.nav-dropdown-btn::after {
+.nav-dropdown-btn::after,
+.nav-dropdown-list-current::after {
   content: "";
   position: absolute;
   pointer-events: none;
@@ -1309,10 +1310,26 @@ div.js-show,
   transition: all 200ms linear;
 }
 
-.nav-dropdown-btn.dropdown-open::after {
+.nav-dropdown-btn.dropdown-open::after,
+.nav-dropdown-list-current::after {
   width: 17px;
   right: 16px;
   margin-top: -9px;
+}
+
+.table-of-contents .nav-dropdown-btn.dropdown-open::after {
+  height: 7px;
+  width: 7px;
+  margin-top: -2px;
+  right: 23px;
+  -webkit-transform: rotate(-45deg);
+  -ms-transform: rotate(-45deg);
+  transform: rotate(-45deg);
+}
+
+.nav-dropdown-list-current {
+  position: relative;
+  color: #f7f779;
 }
 
 .nav-dropdown-list {
@@ -1403,7 +1420,7 @@ div.js-show,
 
 .nav-dropdown-list-todo,
 .nav-dropdown-list-current {
-  opacity: 0.3;
+  opacity: 0.5;
 }
 
 .nav-dropdown-list .help-translate {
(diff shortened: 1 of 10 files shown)


rviscomi:
The ebook part is extremely large. Do we need all that text? I'll file a new issue to discuss linking to /ebook vs the PDF.

image

src/static/css/2019.css
```diff
@@ -1304,12 +1310,28 @@ p.copyright a {
...
 }
 
+.table-of-contents .nav-dropdown-btn.dropdown-open::after {
+  height: 7px;
+  width: 7px;
+  margin-top: -2px;
+  right: 23px;
+  -webkit-transform: rotate(-45deg);
+  -ms-transform: rotate(-45deg);
+  transform: rotate(-45deg);
```
rviscomi, line 1327: BTW do we support `prefers-reduced-motion` yet? If/When we do, it's worth disabling the transition on this property.

src/templates/base/2019/base.html
```diff
@@ -439,11 +465,11 @@
   {% block scripts %}
     <script nonce="{{ csp_nonce() }}">
       (function() {
-        document.querySelectorAll('.js-enable-script-remove-tag').forEach(element => {
-          var text = element.textContent;
-          var parentEl = element.parentElement;
-          parentEl.removeChild(element);
-          parentEl.innerText = text;
+        document.querySelectorAll('.js-hide').forEach(element => {
```
rviscomi, line 468: Is this for the 0.00001% of users with JS disabled? Can this be solved with `<noscript>`?

src/static/css/2019.css
bazzadp: We do in JavaScript:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/e68274c79677de4a298abc0c6e3dae859ed16f52/src/static/js/chapter.js#L280-L287

Added a CSS override in https://github.com/HTTPArchive/almanac.httparchive.org/pull/1512/commits/65eca17ee0ef5f66198886baf6c2684a8da4dce5 as based on [this web.dev article](https://web.dev/prefers-reduced-motion/#(bonus)-forcing-reduced-motion-on-all-websites).

src/templates/base/2019/base.html
bazzadp: It's for those where JavaScript doesn't execute properly - either because they don't have JS enabled, or there's a syntax error, or unsupported JS syntax, or whatever. `<no script>` doesn't cover all of those. Can't remember who said it, but every browser is a no JS browser until it's downloaded parsed and executed.

bazzadp:
I've shorted the ebook considerably in https://github.com/HTTPArchive/almanac.httparchive.org/pull/1512/commits/442ad003352e994c6d0a5a338fadeafcc83a36ca

Ebook link

Technically we should have the Prince wording, but think the fact we have it on the ToC page and the watermark is surely enough.

src/templates/en/2019/base.html
```diff
@@ -140,6 +140,7 @@
 {% block appendices %}Appendices{% endblock %}
 
 {% block ebook_title %}Ebook{% endblock %}
+{% block ebook_download_short %}Ebook PDF ({{ ebook_size_in_mb }}MB){% endblock %}
```
rviscomi, line 143: Nit: Should we try to use this pattern more to avoid retranslations for the same phrases?
```suggestion
{% block ebook_download_short %}{{ ebook_title }} PDF ({{ ebook_size_in_mb }}MB){% endblock %}
```

src/static/css/2019.css
```diff
@@ -972,6 +972,11 @@ p.copyright a {
   display: none;
 }
 
+div.js-show,
+#menu.js-show {
```
rviscomi, line 976: Are the element/ID important for specificity or can we use the plain `.js-show` selector? I'd also suggest combining with the `.hidden` rule above:

```css
.hidden,
.js-show {
  display: none;
}

**Review by rviscomi — APPROVED** [review]

File: src/templates/en/2019/base.html (hunk 3 above)
Comment (bazzadp) [reply src/templates/en/2019/base.html:143]: That actually becomes:

```
{% block ebook_download_short %}{{ self.ebook_title() }} PDF ({{ ebook_size_in_mb }}MB){% endblock %}
```

I'm torn on this myself. Yes it's reuse but a bit more confusing for translators to read and ensure they understand and translate appropriately (e.g. change ordering to "PDF Ebook" for some languages).

Think clearer to leave as is to be honest.

File: src/static/css/2019.css (hunk 4 above)
Comment (bazzadp) [reply src/static/css/2019.css]: They were needed for specificity - but changed the code a bit since I wrote this and no longer needed. Hurray!

Good call in merging too.

Fixed.

File: src/templates/base/2019/base.html [hunk 5]
```diff
@@ -278,7 +305,7 @@
             </li>
           </ul>
         </nav>
-        <nav aria-label="{{ self.menu_title() }}" id="menu" aria-labelledby="menu-btn">
+        <nav aria-label="{{ self.menu_title() }}" id="menu" aria-labelledby="menu-btn" class="js-show">
```
MSakamaki, line 308: @bazzadp 
I prefer to go to the footer navigation item rather than hide the hamburger menu if javascript is disabled.

Options:
However, it can be annoying to return to the top after moving to the footer.
For mobile, do you think it would be nice to have a link in the footer that takes you to the top of the page as well?

MSakamaki (commented):
Thanks for addressing the rest of the issues, @bazzadp !
I play, and comment on one thing that caught my attention.

src/templates/base/2019/base.html
bazzadp: Good idea - done.

I'll leave the scroll to top link for now as a bit out of scope for this and as Rick says the non-JS version should be used by very few users! I know iOS has easy scroll to top by tapping on menu in browser. Not sure if Android does but from a quick Google it appears not!

bazzadp:
I also changed the ToC option to look like the button even when a link (consistent with link buttons on home page) and to continue to still show the Year and Lang menus but disabled. This should avoid any layout shift in case there is any delay with the JS kicking in.

@rviscomi to your earlier point about the `.hidden` class I got annoyed fighting specificity again and changed this to a `!important`. Don’t like to use them much but feel this is one case where it’s warranted and useful.

Think we’re good to merge now but will have another look in a bit and then merge and release.

MSakamaki (approved):
👍

PR #1518, opened 2020-11-14
Allow HTML in Big Issues

bazzadp:
As noticed in https://github.com/HTTPArchive/almanac.httparchive.org/issues/539#issuecomment-727274501 we've a bug:

image


rviscomi (approved):
LGTM to fix this specific case, but I think it's worth reevaluating why `&nbsp;` characters are needed in general. If to prevent "%" from wrapping, can't we use CSS `white-space: nowrap`?

bazzadp:
It's a French thing as they use spaces between figure and % and figure and units. Yes in this case we should use no wrap but in general text we can't so it's used all over their translations.

PR #1523, opened 2020-11-15
Support language redirects

### LINKED ISSUE #1516: Language code for Traditional Chinese
We are currently using `zh-CHT` as the language code for Traditional Chinese. The [Nu Validator says this is not a valid `lang` code](https://validator.w3.org/nu/?doc=https%3A%2F%2Falmanac.httparchive.org%2Fen%2F2020%2F).

Googling around [it appears](https://stackoverflow.com/questions/732109/chinese-language-codes) that `zh-CHT` was used in the past, `zh-HANT` is official way, but `zh-TW` is the most common.

Well we have access to a database of crawl stats to confirm this and [the Markup chapter has even done the work for us](https://docs.google.com/spreadsheets/d/1Ta7amoUeaL4pILhWzH-SCzMX9PsZeb1x_mwrX2C4eY8/edit#gid=2047285366)! Here's the `zh` usage on mobile (in lowercase):

client | freq | html_lang | %
-- | -- | -- | --
mobile | 36,357 | zh-tw | 0.57%
mobile | 8,304 | zh-cn | 0.13%
mobile | 4,844 | zh-hant | 0.08%
mobile | 4,012 | zh | 0.06%
mobile | 2,719 | zh-hant-tw | 0.04%
mobile | 1,545 | zh-hk | 0.02%
mobile | 1,021 | zh_tw | 0.02%
mobile | 564 | zh-hans | 0.01%
mobile | 399 | zh-cmn-hans | 0.01%
mobile | 198 | zh_cn | 0.00%
mobile | 125 | zh-hant-hk | 0.00%
mobile | 98 | zh_hk | 0.00%
mobile | 96 | zh-cmn-hant | 0.00%
mobile | 56 | zh-hans-cn | 0.00%
mobile | 48 | zh-hans-tw | 0.00%
mobile | 21 | zh-cmn-hans-cn | 0.00%
mobile | 11 | zh-cmn | 0.00%
mobile | 10 | zh-cmn-hant-tw | 0.00%
mobile | 9 | zh-en | 0.00%
mobile | 9 | zh-sg | 0.00%
mobile | 8 | zh-hang-tw | 0.00%
mobile | 8 | zh-cht | 0.00%

So it does appear that `zh-CHT` is not commonly used and `zh-TW` is indeed the preferred usage, with `zh-HANT` a distant second (`zh-CN` is for Simplified Chinese so not the same). `zh-TW` is also what [MDN uses](https://developer.mozilla.org/zh-TW/) for example.

@AbbyTsai what's your thoughts here as you picked this lang code. Should we change to `zh-TW`? This would mean it's more likely to be auto selected by going to https://almanac.httparchive.org and I would prefer to have "valid" HTML.

bazzadp:
Support language redirects in our URLs so if you try `/en-GB/` it redirects to `/en/`

This also allows support of different Chinese Dialects (including plain old `zn` which redirects to simplified Chinese - `zh-CN` same as MDN does).

It also allows us to move a language like we need to resolve #1516


src/server/validate.py
```diff
@@ -76,9 +78,24 @@ def validate_lang_and_year(lang, year):
...
     if lang is not None and lang not in supported_langs:
         logging.debug('Unsupported language set: %s.' % lang)
+
+        # Check it's not just a simple case issue
+        if lang.lower() in supported_langs:
+            return (lang.lower(), year)
+
+        # Handle lookups for special cases (e.g. Chinese)
+        if lang.lower() in LANGUAGE_MAPPING:
+            return (LANGUAGE_MAPPING.get(lang.lower()), year)
```
SaptakS, line 91: Shouldn't we add a check that the `LANGUAGE_MAPPING.get(lang.lower())` is in `supported_langs` just to avoid any possible errors.

src/server/language.py
```diff
@@ -25,6 +25,21 @@ def lang_code(self):
         return self._lang_code
 
 
+# Allow maping of one language to another for when country agnostic lookip
```
SaptakS, line 28:
```suggestion
# Allow maping of one language to another for when country agnostic lookup
```

src/server/validate.py
bazzadp: Good idea in case someone enters bad config. Added.

SaptakS (approved):
looks good to me

Issue #906, opened 2020-06-27
Security 2020


obto:
# Part II Chapter 11: Security

## Content team
| Authors | Reviewers | Analysts | Draft | Queries | Results |
| ------- | --------- | -------- | ----- | ------- | ------- |
| @nrllh @tomvangoethem  |  @cqueern @bazzadp @edmondwwchan | @AAgar @tomvangoethem | [Doc][~draft-doc] | [*.sql][~sql-dir] | [Sheet][~results-sheet] |

**Content team lead:** @nrllh

Welcome chapter contributors! You'll be using this issue throughout the [chapter lifecycle](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle) to coordinate on the content planning, analysis, and writing stages.

The **content team** is made up of the following contributors:

- [Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)
- [Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)
- [Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)

**New contributors:** If you're interested in joining the content team for this chapter, just leave a comment below and the content team lead will loop you in.

_Note: To ensure that you get notifications when tagged, you must be "[watching](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#notifications)" this repository._

### [**0. Form the content team**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#0-create-content-team)
+ [x] **Jul 6th**: Project owners have selected an author to be the content team lead
+ [x] **Jul 13th**: The content team has at least one author, reviewer, and analyst (minimally viable team formed)

### [**1. Plan content**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
+ [x] **Jul 20th:** The content team has completed the chapter outline in the [draft doc][~draft-doc]
+ [x] **Jul 27th**: Analysts have triaged the feasibility of all proposed metrics

### [**2. Gather data**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
[...]

PR #1534, opened 2020-11-16
Security SQL: final queries

tomvangoethem:
Progress on #906

### DIFF
--- a/sql/2020/11_Security/cryptominer_share.sql
+++ b/sql/2020/11_Security/cryptominer_share.sql
@@ -1,38 +1,19 @@
 #standardSQL
---share of cryptominers
-WITH
-  total AS (
-  SELECT
-    COUNT(0) AS ttl
-  FROM
-    `httparchive.technologies.2020_08_01_desktop`
-  WHERE
-    category = 'Cryptominers'
-    OR category='Cryptominer')
+# Share of cryptominers
 SELECT
   app,
-  freq,
-  ROUND((freq/(
-      SELECT
-        ttl
-      FROM
-        total)),2) AS pct
-FROM (
-  SELECT
-    SUM(freq) AS freq,
-    app
-  FROM (
-    SELECT
-      COUNT(0) AS freq,
-      app
-    FROM
-      `httparchive.technologies.2020_08_01_desktop`
-    WHERE
-      category = 'Cryptominers'
-      OR category='Cryptominer'
-    GROUP BY
-      app)
-  GROUP BY
-    app)
+  _TABLE_SUFFIX AS client,
+  SUM(COUNT(0)) OVER (PARTITION BY _TABLE_SUFFIX) AS total_cryptominers,
+  COUNT(0) AS freq,
+  COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY _TABLE_SUFFIX) AS pct
+FROM
+  `httparchive.technologies.2020_08_01_*`
+WHERE
+  category = 'Cryptominers'
+  OR category = 'Cryptominer'
+GROUP BY
+  client,
+  app
 ORDER BY
-  pct DESC
+   client,
+   pct DESC
--- a/sql/2020/11_Security/cryptominer_share.sql
+++ b/sql/2020/11_Security/cryptominer_share.sql
@@ -5,7 +5,7 @@ WITH
   SELECT
     COUNT(0) AS ttl
   FROM
-    `httparchive.technologies.2020_10_01_desktop`
+    `httparchive.technologies.2020_08_01_desktop`
   WHERE
     category = 'Cryptominers'
     OR category='Cryptominer')
--- a/sql/2020/11_Security/cryptominer_share.sql
+++ b/sql/2020/11_Security/cryptominer_share.sql
@@ -1,0 +1,38 @@
+#standardSQL
+--share of cryptominers
+WITH
+  total AS (
+  SELECT
+    COUNT(0) AS ttl
+  FROM
+    `httparchive.technologies.2020_10_01_desktop`
+  WHERE
+    category = 'Cryptominers'
+    OR category='Cryptominer')
+SELECT
+  app,
+  freq,
+  ROUND((freq/(
+      SELECT
+        ttl
+      FROM
+        total)),2) AS pct
+FROM (
+  SELECT
+    SUM(freq) AS freq,
+    app
+  FROM (
+    SELECT
+      COUNT(0) AS freq,
+      app
+    FROM
+      `httparchive.technologies.2020_08_01_desktop`
+    WHERE
+      category = 'Cryptominers'
+      OR category='Cryptominer'
+    GROUP BY
+      app)
+  GROUP BY
+    app)
+ORDER BY
+  pct DESC
--- a/sql/2020/11_Security/cryptominer_usage.sql
+++ b/sql/2020/11_Security/cryptominer_usage.sql
@@ -1,0 +1,24 @@
+#standardSQL
+SELECT
+  month,
+  client,
+  freq
+FROM (
+  SELECT
+    LEFT(_TABLE_SUFFIX, 10) AS month,
+  IF
+    (ENDS_WITH(_TABLE_SUFFIX, '_desktop'),
+      'desktop',
+      'mobile') AS client,
+    COUNT(0) AS freq
+  FROM
+    `httparchive.technologies.*`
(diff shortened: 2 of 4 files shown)


tomvangoethem:
I'm guessing that the failed checks are unrelated to this particular pull request? The reported error is:

```
Error: Unable to process command '::add-path::/opt/hostedtoolcache/node/12.19.0/x64/bin' successfully.
Error: The `add-path` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

bazzadp:
Just merged a fix (#1535) into `main`. Can you update your branch?

tomvangoethem:
Yep, that fixed it 👍

sql/2020/11_Security/cryptominer_share.sql
```diff
@@ -0,0 +1,38 @@
+#standardSQL
+--share of cryptominers
+WITH
+  total AS (
+  SELECT
+    COUNT(0) AS ttl
+  FROM
+    `httparchive.technologies.2020_10_01_desktop`
```
bazzadp, line 8: Why October?

```suggestion
    `httparchive.technologies.2020_08_01_desktop`
```

Can we also not inline this query like we usually do instead of a separate WITH query?

sql/2020/11_Security/cryptominer_share.sql
```diff
@@ -0,0 +1,38 @@
...
+      app
+    FROM
+      `httparchive.technologies.2020_08_01_desktop`
+    WHERE
+      category = 'Cryptominers'
+      OR category='Cryptominer'
+    GROUP BY
+      app)
+  GROUP BY
+    app)
```
bazzadp, line 36: I don't understand why you need this double inner loop?

This seems to give same results:

```suggestion
  SELECT
    COUNT(0) AS freq,
    app
  FROM
    `httparchive.technologies.2020_08_01_desktop`
  WHERE
    category = 'Cryptominers'
    OR category='Cryptominer'
  GROUP BY
    app)
```

I'm also a little cynical on these stats. Are we really saying only 348 sites out of the 5.5 Million desktop sites we crawl have cryptominers on them? That seems very low.

Also why only querying desktop?

sql/2020/11_Security/version-evolution-top-technologies.sql
```diff
@@ -0,0 +1,50 @@
...
+      COUNT(0) AS num
+    FROM
+      `httparchive.technologies.*`
+    WHERE
+      REGEXP_CONTAINS(_TABLE_SUFFIX, r'^20(20|19).*')
+    GROUP BY
+      category,
+      app
+    ORDER BY num DESC
+    LIMIT 20
```
bazzadp, line 34: This will mean you're limiting it to the top technologies for 2019 and 2020 rather than the top technologies as of now (or Aug 2020). That may be fine, but thought I'd point it out just to make sure you get that.

bazzadp (changes requested):

rviscomi:
@tomvangoethem please see @bazzadp's feedback. We should merge this ASAP.

tomvangoethem:
Will fix it tomorrow!

sql/2020/11_Security/version-evolution-top-technologies.sql
tomvangoethem: That's as intended; since we're looking at the software over the period of a year, I thought it made sense to look at those that have been most popular during this period.

tomvangoethem:
Updated cryptominer query. However, as @bazzadp mentioned, the number looks very low. Looking into that right now.

tomvangoethem:
Coinhive: shut down
CoinImp: potentially outdated URL pattern in Wappalyzer. Based on the [documentation](https://www.coinimp.com/documentation), it should be `^https://www\.hostingcloud\.racing/.+\.(js|wasm)` - I get 167 pages with this for desktop; Wappalyzer reports 175 (sets could be mutually exclusive though)
WebMinePool: some 67 desktop pages include `https://webminepool.com/lib/base.js` - not considered by Wappalyzer
JSECoin: shut down in April 2020

From the paper "[Is Cryptojacking Dead after Coinhive Shutdown?](https://arxiv.org/pdf/2001.02975.pdf)":
>  The results revealed that 99% of sites no longer continue cryptojacking.

So it seems that cryptojacking is not financially viable any more (because of hard fork that caused a 50+% drop in hash rate, and the drop in value of XMR), which makes me think that the results are more or less representative

sql/2020/11_Security/cryptominer_share.sql
tomvangoethem: Updated query; low number of sites can be explained based on several of the key cryptojacking services shutting down, and low financial benefit as compared to some time ago.

bazzadp: OK but your history still stills low usage.

Also worth remembering we only look at home pages so an XSS on an inner page (more likely?) means we couldn't pick that up.

sql/2020/11_Security/cryptominer_usage.sql
```diff
@@ -0,0 +1,21 @@
...
+    `httparchive.technologies.*`
+  WHERE
+    category = 'Cryptominers'
+    OR category='Cryptominer'
+  GROUP BY
+    _TABLE_SUFFIX)
+ORDER BY
+  client,
+  month,
+  freq DESC
```
bazzadp, line 21: I still don't see need of inner loop. This gives same result:

```sql
#standardSQL
SELECT
  LEFT(_TABLE_SUFFIX, 10) AS month,
  IF(ENDS_WITH(_TABLE_SUFFIX, '_desktop'), 'desktop', 'mobile') AS client,
  COUNT(0) AS freq
FROM
  `httparchive.technologies.*`
WHERE
  category = 'Cryptominers'
  OR category='Cryptominer'
GROUP BY
  _TABLE_SUFFIX
ORDER BY
  client,
  month,
  freq DESC
```

tomvangoethem: Indeed, those seem exactly the same. Updated.

sql/2020/11_Security/version-evolution-top-technologies.sql
```diff
@@ -0,0 +1,50 @@
...
+      app,
+      COUNT(0) AS num
+    FROM
+      `httparchive.technologies.*`
+    WHERE
+      REGEXP_CONTAINS(_TABLE_SUFFIX, r'^20(20|19).*')
+    GROUP BY
+      category,
+      app
+    ORDER BY num DESC
```
bazzadp, line 33:
```suggestion
    ORDER BY
      num DESC
```

sql/2020/11_Security/version-evolution-top-technologies.sql
```diff
@@ -0,0 +1,50 @@
...
+  ) AS top ON (tech.category = top.category AND tech.app = top.app)
+  WHERE
+    REGEXP_CONTAINS(info, r'\d+\.\d+')
+    AND REGEXP_CONTAINS(_TABLE_SUFFIX, r'^20(20|19).*')
+  GROUP BY
+    tech.category,
+    tech.app,
+    info,
+    _TABLE_SUFFIX)
+WHERE pct > 0.01
```
bazzadp, line 44:
```suggestion
WHERE
  pct > 0.01
```

bazzadp (approved):
LGTM. A few very minor nits.

This good to merge?

sql/2020/11_Security/cryptominer_share.sql
tomvangoethem: Since 2018_12_15 (at the time when HTTPArchive started crawling CrUX list I believe), there seems to be a steady decline in the number of cryptomining inclusions (although the number of visited home pages has been increasing). It could also be that people are not removing inclusions and thus are still being counted as using cryptomining (e.g. CoinHive domain is shut down, so the requests for it get NXDOMAIN).
I don't think we can claim that cryptojacking is dead, but several things point in the direction that it's less fruitful and thus less prominent.

tomvangoethem:
Yep, good to merge I think. That should be it for Security SQL 🥳

bazzadp:
Great stuff. Now just to write the thing 😁

PR #1538, opened 2020-11-17
Address Markup TODOs, light editing

j9t:
Progress on #899


src/content/en/2020/markup.md
```diff
@@ -452,8 +452,7 @@ The source of these elements appears to be mixed, as in some are unknown while o
 <figcaption>{{ figure_link(caption="Frequency and average use of standard heading elements.", sheets_gid="277662548", sql_file="pages_wpt_bodies_by_device_and_percentile_and_heading_level.sql") }}</figcaption>
 </figure>
 
-{# TODO(authors): Expand on this to include more discussion and interpretation of the results. #}
-Wait.
+You expected to see only the heading elements you know from the specifications, but there’s more:
```
bazzadp, line 455: Suggest rewording slightly:

```suggestion
You might have expected to only see the standard `<h1>` to `<h6>` elements, but some sites actually use more levels:
```

src/content/en/2020/markup.md
```diff
@@ -699,10 +698,9 @@ As a rule of thumb and for [usability reasons](https://www.nngroup.com/articles/
 
 <p class="note">Within the latest Safari and Firefox versions, setting <code>target="_blank"</code> on <code>a</code> elements implicitly provides the same <code>rel</code> behavior as setting <code>rel="noopener"</code>. This is already <a href="https://chromium-review.googlesource.com/c/chromium/src/+/1630010">implemented in Chromium</a> as well and will land in Chrome 88.</p>
 
-{# TODO(editors, authors): The following two sections each read like conclusions. Consider merging. #}
-## Status and trends
+## Conclusion
 
-We've sprinkled some observations throughout the chapter, and you'll have made your own observations. At the end of this 2020 analysis, here are some things that stood out for us.
+At the end of this 2020 analysis, here are some things that stood out for us.
```
bazzadp, line 703: I think this is a little short, from a layout perspective, as we then launch straight into that big figure.

Conclusion

I'd prefer a bit more of an introductory paragraph to the conclusion so that sentence doesn't look so lonesome.

I think we want to keep the big figure and our convention is to list big figures first and then talk about them. And even without that it would be a little short IMHO.

Here's what I suggested previously, which is a little longer and retains some of your original intent (even if not your exact wording!), but possibly that's still a little short. Or am I over thinking this?

```suggestion
We've touched on some observations throughout the chapter, but as a reflection on the state of markup in 2020, here are some things that really stood out for us:
```

image

What do you think?

bazzadp (commented):
Some small suggestions I'd like your opinions on.

src/content/en/2020/markup.md
j9t: LGTM! (Would leave out “really” but that’s a nit, I’m not attached to this.)

j9t (commented):
@bazzadp, what do you think, do we want to merge this update (and take it from there)?

src/content/en/2020/markup.md
bazzadp: Removed the `really`.

bazzadp:
Let's merge this, if you're happy to commit my revised other suggestion? Still quite a few TODOs to do (mostly with the authors) but every step forward is a step in the right direction 😀

j9t (commented):
Yes, LGTM! Thanks for fleshing this out with me.

j9t:
@bazzadp, when would this PR be available live? Am I missing something?

bazzadp:
Hey @j9t there is currently no auto release on merge (though I want this but have some barriers to close first!). So releases are done manually at the moment and we normally wait until there’s enough to warrant a release.

As there’s still a few ToDos here I hadn’t planned on releasing this change in the hope those will come soon and we can remove the Unedited tag. Are we planning on making more changes to address them? If not it’ll be picked up on the next release. You can watch the repo for releases as we send out a notification when we do releases.

j9t:
Hi @bazzadp—I was looking at doing several rounds (editing and releasing) to get the TODOs out of the way as that seemed easier to coordinate and faster, however I didn’t raise that :) So yes, if we could roll this out that would be great, and I’d then move on to the next edits, some of which require a larger group again.

I haven’t been looking at the “Unedited” flag again; from my view any update should be “ready” (including the one that’s live) so I think it’s completely up to you and team to decide when to lift that.

bazzadp:
I've released that there now as had some templating changes that no harm to get out well ahead of the release.

However, as I say, there is a bit of overhead to releasing - particularly at the moment as I'm having severe Internet problems (that last release took 15 mins waiting for the upload!). So we wouldn't normally do a release for every small wording change, while we so busy trying to get the other chapters as far as yours. As I say, I'd like to get to the stage where that happens and think we're not far off with the automated testing we have, but needs some more work and have other priorities at the moment. But once a change is merged into `main` you're good to iterate on the next fix so no need to wait for each fix to be released. If you want a test version to see what changes look like then give us a shout as that's a bit easier.

For the unedited flag, yes Rick or I will remove once we close out those final TODOs one way or another. It's just a warning to the users that this is a pre-release version and may change slightly - which I think is still the case here. We don't have a `semi-edited`, or `mostly-edited` flag unfortunately for you! 😁 I think as long as we do that by 9th December launch that's fine, as that's when we hope for a big uplift in traffic with the launch. Though sooner the better would be good, as imagine it'll only get more manic with the other chapters between now and then!

PR #1543, opened 2020-11-17
Start website in full in Docker

bazzadp:
We recently added the `npm run start` command in #1436 which does a full clean install, generates the chapters, runs the website tests and has a file watcher (added in #1437 and #1457 ) to regenerate chapters as they are saved.

This brings the same functionality to our Dockerfile.

I also moved the start of the website to earlier (before the tests and `npm run generate`) so you don't have to wait for them to complete to have old website up.

Note the `pkill` command on MacOS accepts files without quote, but not on linux so had to change slightly.

### DIFF
--- a/src/tools/scripts/run_and_test_website.sh
+++ b/src/tools/scripts/run_and_test_website.sh
@@ -48,19 +48,17 @@ fi
 
 if [ "$(pgrep -f 'python main.py')" ]; then
   echo "Killing existing server to run a fresh version"
-  pkill -9 python main.py
+  pkill -9 -f "python main.py"
+fi
+
+if [ "$(pgrep -f 'node ./tools/generate/chapter_watcher')" ]; then
+  echo "Killing existing watcher to run a fresh version"
+  pkill -9 -f "node ./tools/generate/chapter_watcher"
 fi
 
 echo "Installing and testing python environment"
 python -m pip install --upgrade pip
 pip install -r requirements.txt
-pytest
-
-echo "Installing node modules"
-npm install
-
-echo "Building website"
-npm run generate
 
 echo "Starting website in background mode for tests"
 python main.py background &
@@ -68,6 +66,15 @@ python main.py background &
 sleep 2
 # Check website is running as won't have got feedback as backgrounded
 pgrep -f "python main.py"
+
+# Run pytests
+pytest
+
+echo "Installing node modules"
+npm install
+
+echo "Building website"
+npm run generate
 
 echo "Testing website"
 npm run test
@@ -84,7 +91,7 @@ if [ "${debug}" == "1" ]; then
 
   if [ "$(pgrep -f 'python main.py')" ]; then
     echo "Killing server to run a fresh version in debug mode"
-    pkill -9 python main.py
+    pkill -9 -f "python main.py"
   fi
 
   echo "Starting website in foreground mode so it reloads on file changes"
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -23,4 +23,4 @@ RUN     pip install -r requirements.txt
 RUN     npm install
 COPY    . ./
 
-CMD     ["python", "main.py"]
+CMD     ["npm", "run", "start"]
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -23,4 +23,4 @@ RUN     pip install -r requirements.txt
 RUN     npm install
 COPY    . ./
 
-CMD     ["npm", "run", "start"]
+CMD     ["./tools/scripts/run_and_test_website.sh", "-d"]
--- a/src/README.md
+++ b/src/README.md
@@ -269,7 +269,7 @@ root@[CID]:/app# exit
 6. To customize the image use `PYVER`, `NODEVER`, and `SKIPGC` build arguments to control which versions of Python and Node are used and whether Google Cloud SDK is installed.
 
 ```
-docker image build --build-arg PYVER=3.7 --build-arg NODEVER=14.x --build-arg SKIPGC=false -t webalmanac:custom .
+docker image build --build-arg PYVER=3.8 --build-arg NODEVER=14.x --build-arg SKIPGC=false -t webalmanac:custom .
 ```
 
 7. If you want to run the GitHub Super-Linter without `npm` being installed you need to call the command directly as given in `package.json`.
@@ -277,11 +277,11 @@ docker image build --build-arg PYVER=3.7 --build-arg NODEVER=14.x --build-arg SK
 This will depend on your operating system but for MacOS/Linux this would be:
 
 ```
-docker container run -it --rm -v "$PWD/..":/app -w /app/src --entrypoint=./tools/scripts/run_linter_locally.sh github/super-linter
+docker container run -it --rm -v /app/node_modules -v "$PWD/..":/app -w /app/src --entrypoint=./tools/scripts/run_linter_locally.sh github/super-linter
 ```
 
 And for Windows:
 
 ```
-docker container run --rm -v %cd%\\..:/app -w /app/src --entrypoint=./tools/scripts/run_linter_locally.sh github/super-linter
+docker container run --rm -v /app/node_modules -v %cd%\\..:/app -w /app/src --entrypoint=./tools/scripts/run_linter_locally.sh github/super-linter
 ```
--- a/src/tools/scripts/stop_website.sh
+++ b/src/tools/scripts/stop_website.sh
@@ -9,5 +9,7 @@
 # It is useful as sometimes these are backgrounded so this offers a quick way of stopping them.
 #
 
-pkill -9 python main.py
-pkill -9 node ./tools/generate/chapter_watcher
+pkill -9 -f "python main.py"
+pkill -9 -f "./tools/generate/chapter_watcher"
+
+exit 0


sudheendrachari:
Trying this change locally on a fresh checkout...

sudheendrachari:
Did a fresh checkout
and ran 
`docker image build -t webalmanac .`
and then
`docker container run --rm -it -v "$PWD":/app -p 8080:8080 webalmanac`

I see this error now 

Screenshot 2020-11-18 at 10 46 30 AM

bazzadp:
Looks like you are missing some recently added dependencies. Is your branch up to date?

> `git checkout -b auto_gen_chapters_for_docker origin/auto_gen_chapters_for_docker`

That will have created a new branch, from the current branch rather than checked out the remote version of this branch. Can you try the following to delete that branch and check it out again:

```
git checkout -d auto_gen_chapters_for_docker
git checkout main
git pull
git checkout auto_gen_chapters_for_docker
docker image build -t webalmanac .
docker container run --rm -it -v "$PWD":/app -p 8080:8080 webalmanac
```

sudheendrachari:
> That will have created a new branch, from the current branch rather than checked out the remote version of this branch. Can you try the following to delete that branch and check it out again:

Forgot to mention that i did pull from origin/auto_gen_chapters_for_docker and was able to see your commit locally with message "Start website in full in Docker" earlier as well. But anyway tried deleting that branch and checking out like you mentioned.

But i end up seeing the same error. The error seems to be happening because `npm install` is not the first command that the docker is doing, may be? To validate that I just ran `npm install` and run docker container run command i see it running 

Screenshot 2020-11-18 at 4 42 37 PM

sudheendrachari:
BTW with this change will it also run `npm run generate` as part of docker command?

ibnesayeed:
Sorry @sudheendrachari, that was an oversight from my end when I added Docker support in the beginning. What is happening here is, the `node_modules` folder is created inside the docker image at build time, but when we mount current working directory at run time, we overwrite that. This means, those who have a proper `node_modules` folder in their local checkout will not see the issue, but that means they either install npm dependencies without Docker at least once or copy the data from the container outside on host machine (neither of these are good idea when running in Docker). To fix this issue, run the following command:

```
$ docker container run --rm -it -v /app/node_modules -v "$PWD":/app -p 8080:8080 webalmanac
```

In this case we are explicitly creating a volume `/app/node_modules` of the modules folder inside of the container before mounting the host folder, so that subfolder is not affected by the mounted volume.

src/tools/scripts/run_and_test_website.sh
```diff
@@ -48,19 +48,17 @@ fi
...
+  pkill -9 -f "python main.py"
+fi
+
+if [ "$(pgrep -f 'node ./tools/generate/chapter_watcher')" ]; then
+  echo "Killing existing watcher to run a fresh version"
+  pkill -9 -f "node ./tools/generate/chapter_watcher"
 fi
 
 echo "Installing and testing python environment"
 python -m pip install --upgrade pip
```
ibnesayeed, line 60: This line is causing `pip` to be installed again every time a container is run, which means the version in the Docker image is not the latest.

```
$ docker container run --rm -it -v /app/node_modules -v "$PWD":/app -p 8080:8080 webalmanac

> almanac.httparchive.org@0.0.1 start /app
> run-script-os


> almanac.httparchive.org@0.0.1 start:darwin:linux /app
> ./tools/scripts/run_and_test_website.sh -d

Installing and testing python environment
Collecting pip
  Downloading pip-20.2.4-py2.py3-none-any.whl (1.5 MB)
     |████████████████████████████████| 1.5 MB 3.6 MB/s 
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 20.1
    Uninstalling pip-20.1:
      Successfully uninstalled pip-20.1
Successfully installed pip-20.2.4
```

bazzadp:
Ah thanks for explaining @ibnesayeed !

 I want to do a fresh `npm install` anyway each time it's run in case dependences change but had hoped this would be quick to run in most cases if your image was recently built (as most modules are already installed).

I've done two things in [my last commit](https://github.com/HTTPArchive/almanac.httparchive.org/pull/1543/commits/2ba573104b4ff5a133fe7c908279b186f039015c):

1. Changed the README to reference the additional command line arguments needed
2. Changed the `npm run start` to the actual script called so even if people don't do that it will work (though will be slower as will have to do a full `npm install` rather than just a check for updates).

Let me know how that goes now?

src/tools/scripts/run_and_test_website.sh
bazzadp: If an image was built a while ago `pip` could be out of date. By doing this command it will try to install the latest and if that was what the container was built with, then no download and no issue:

```
Installing and testing python environment
Requirement already up-to-date: pip in /usr/local/lib/python3.9/site-packages (20.2.4)
Requirement already satisfied: flask==1.1.2 in /usr/local/lib/python3.9/site-packages (from -r requirements.txt (line 1)) (1.1.2)
Requirement already satisfied: flask-talisman==0.7.0 in /usr/local/lib/python3.9/site-packages (from -r requirements.txt (line 2)) (0.7.0)
...
```

And if it wasn't then time to update your container to avoid this slower way of starting. 😁

Either way it will work - just will be faster if container was built with latest pip.

Does that make sense? Can you see any issue with that?

ibnesayeed: I would prefer if we do not install things in the image at runtime that are needed for the lifecycle of a contain, but make sure the image has them already. This way the image is more self-contained, portable, and reproducible. If a dependency or package is old, it needs to be escalated in the image building time and a fresh image should be built so that at run time we have everything we need.

ibnesayeed:
Ideally, I would separate environment setup steps like installing dependencies and packages in a separate shell script than the one that is responsible for running the site and testing. This way, if one runs the command on bare host machine, both can be run, but when running things in a container, we do not perform the setup steps at runtime, assuming that the image has those things sorted out (that's one of the appeals of containers).

src/tools/scripts/run_and_test_website.sh
bazzadp: Going to Resolve this and move to the general conversation as two different threads on basically the same topic.

bazzadp:
> I would prefer if we do not install things in the image at runtime that are needed for the lifecycle of a contain, but make sure the image has them already. This way the image is more self-contained, portable, and reproducible.

Agreed. But I think we should check again at run time to ensure developers are not developing on old dependencies.

> If a dependency or package is old, it needs to be escalated in the image building time and a fresh image should be built so that at run time we have everything we need.

What do you mean "in the image building time"? What if a developer builds the Docker Image and then uses that for the next year without needing to rebuild it? My concern is they are potentially developing based on very old dependencies.

By having the Docker image built with the best dependencies at build time, but then checking and updating if necessary at run time, we get the best of both worlds - the image is self-contained, portable **and** also up to date.

Maybe you lose slightly on the reproducibility front, as might change between runs, but you gain in reproducibility with other developers, GitHub Actions and what will be run in production! And that is much more important to me than keeping the Docker container on old versions for reproducibility reasons.

bazzadp:
(I also should point out this was more a by-product of reusing the `run_and_test_website.sh` script used by GitHub Actions, and more recently those developing outside of Docker rather than an intentional choice, but it's still the right choice in my opinion).

ibnesayeed:
> By having the Docker image built with the best dependencies at build time, but then checking and updating if necessary at run time, we get the best of both worlds

This is similar to those self-updating software that check for update each time they are started and instead of telling the users whether or not they want to upgrade, they just self update and restart again.

This monolithic approach sounds helpful, but actually comes with the lack of separation of concerns, resulting in repeating yourself (anti-DRY) as package installation and environment configuration is happening in many different places, which may cause inconsistencies as the project grows, resulting is a maintenance nightmare.

It seems like most of the time checking for dependencies results in a lot of skips because those are already met, but knowing that everything is up to date itself has a cost that adds up when doing it at each run and hitting upstream repositories (like PyPI or NPM). It sure is possible that a developer might be running a little older versions of certain dependencies, but our code should not be as brittle that it breaks if the latest and greatest nightly versions are not installed. If the image is years old, then an attempt to run things will break anyway and that would lead developers to rebuild their image.

That said academic discussion aside, I am okay with this PR for now.

bazzadp:
> This is similar to those self-updating software that check for update each time they are started and instead of telling the users whether or not they want to upgrade, they just self update and restart again.

Disagree. Other than `pip` this is updating software to the versions matching in our GitHub repo. We have already controlled and accepted that those updates are fine to accept through the PR process managed by dependabot that you suggested.

If it was updating to latest version of our dependencies I could see your point but it's not - it's upgrading to the _accepted versions_ of dependencies used by this project and no further. We could remove `pip` itself from this process but I don't see why we couldn't go to latest for that to avoid warnings and other issues since we're doing this anyway.

> This monolithic approach sounds helpful, but actually comes with the lack of separation of concerns, resulting in repeating yourself (anti-DRY) as package installation and environment configuration is happening in many different places, which may cause inconsistencies as the project grows, resulting is a maintenance nightmare.

Massively disagree. The whole reason this happens here is because we are reusing the start script used by GitHub actions and those with direct installs. So package installations and environment configuration to be consolidated in ONE place, and REDUCES inconsistencies as the project grows, AVOIDING a maintenance nightmare.

> It seems like most of the time checking for dependencies results in a lot of skips because those are already met, but knowing that everything is up to date itself has a cost that adds up when doing it at each run and hitting upstream repositories (like PyPI or NPM).

I'm presuming starting the development server is done once at the start of a session and not for each run or code change. Again I'd prefer that "hit" for the consistency.

> It sure is possible that a developer might be running a little older versions of certain dependencies, but our code should not be as brittle that it breaks if the latest and greatest nightly versions are not installed. If the image is years old, then an attempt to run things will break anyway and that would lead developers to rebuild their image.

Couldn't disagree more. Many things and dependencies change that can cause small issues. And our code is not that brittle and we have good tests (that do these fresh installs), but that's no reason not to ensure what we have a consistent development environment when it is fired up.

ibnesayeed (approved):
LGTM!

I must say, we did not have these many disagreements ever before! :wink:

bazzadp:
Thanks. Debate is good and value your opinion and challenges! Think we're both clearer now on how it works, and why (even if we wouldn't approach it in the exact same way!).

@sudheendrachari if you can confirm it's all good for you now then I'll merge.

sudheendrachari:
Sure, will try

sudheendrachari (approved):
Works like a charm!
I really like the way server starts quickly and the slow part like `npm run generate` run later

bazzadp:
> I really like the way server starts quickly and the slow part like `npm run generate` run later

Yeah that had been annoying me for a while hence why I switched it as part of this.

All good. Merging now!

PR #1551, opened 2020-11-19
Move sitemap to new timestamps

bazzadp:
Just realised we didn't move the sitemap over to the new last updated timestamps from #1508


rviscomi:
I haven't reviewed this but since it's an urgent bugfix, go ahead with the merge.

Issue #1558, opened 2020-11-20
Code cleanup to move away from 2019 as a base


bazzadp:
Currently we have 2019 as a "base" for a lot of our code, and 2020 is an extension of that. I'd like to clean this up properly, to make it more obvious what's base and what's 2019. This will allow greater multi-year support, avoid confusion and also allow Translations to support just 2020, whereas currently they must support both 2019 and 2020.

Tasks to do:
- [x] - Move `src/templates/base/2019/` to `src/templates/base/` - #1559
- [x] - Separate language templates from, for example, `src/templates/en/2019/` to `src/templates/en/`, leaving behind smaller 2020-like files in the 2019 directory. - #1568 
- [x] - Rename `2019.css` to `almanac.css` - #1918 
- [ ] - Decide what to do with the Chapter Hero Images (currently in 2019 image dir). Duplicate in 2020? Fallback to 2019 if not exists in 2020 (how will this scale)? Move to separate Hero-Images dir (but then how to override a chapter in future - or is that next year's problem)? - #1920 

None of these are particularly difficult but opening this issue to track and to allow smaller PRs to handle them. Would be good to get a good few of them done now while waiting on chapters.

Issue #1468, opened 2020-11-09
2020 chapter metadata still uses 2019


rviscomi:
```html
<title>Capabilities | 2019 | The Web Almanac by HTTP Archive</title>
<meta property="og:title" content="Capabilities | 2019 | The Web Almanac by HTTP Archive">
<meta property="og:image" content="https://almanac.httparchive.org/images/2019/capabilities/hero_lg.jpg">
<meta name="twitter:title" content="Capabilities | 2019 | The Web Almanac by HTTP Archive">
<meta name="twitter:image" content="https://almanac.httparchive.org/images/2019/capabilities/hero_lg.jpg">
```

These will cause UX confusion for readers but also 404 images for social scrapers like Twitter:

![image](https://user-images.githubusercontent.com/1120896/98609375-203c8d00-22bb-11eb-8c2d-4da573a5894f.png)

rviscomi:
I've fixed the title issues in #1469 and it appears that the 2020 hero illustrations do actually live in the [2019 directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/static/images/2019/capabilities). But it's still 404ing because it seems like the server thinks the "images" subdirectory is an invalid language: https://almanac.httparchive.org/images/2019/capabilities/hero_lg.jpg

![image](https://user-images.githubusercontent.com/1120896/98610522-c2f60b00-22bd-11eb-98d1-bf7d6431e0b3.png)

I'll try to get the 2019 chapter image directory served again, but we should keep this issue open to avoid confusion about 2020 content served from 2019 directories, especially for chapters that didn't exist in 2019.

rviscomi:
The 404 was because the chapter hero image URLs were missing the `/static` part of the path. Fixed in #1470 and also added dynamic year to chapter metadata in non-English localized templates.

bazzadp:
> I'll try to get the 2019 chapter image directory served again, but we should keep this issue open to avoid confusion about 2020 content served from 2019 directories, especially for chapters that didn't exist in 2019.

Ah think I misunderstood https://github.com/HTTPArchive/almanac.httparchive.org/pull/1358#discussion_r504239477. Thought you were asking to keep in 2019 directory but now see you mean to have them in 2020 directory with either copies of 2019 for reused chapters, or fallback to there. Yes let's leave that open until that's fixed.

> I've force-pushed these changes to the production server without a deploy, so they will be picked up in the next formal deployment.

[...]

PR #1559, opened 2020-11-20
Move base templates out of 2019 folder

bazzadp:
Makes progress on #1558 

Moves templates in `src/templates/base/2019/` to `src/templates/base/` as they are not 2019 specific.


rviscomi:
Assuming we get to a point when new annual editions are redesigned at least slightly with a new color palette, layout variations, and new illustrations, could these changes make it harder? I had thought that keeping the base templates theme-agnostic would be better for extensibility. For example, if 2021 is a completely different look and feel, how much of the core template infrastructure would need to be rewritten?

bazzadp:
Those are good questions, and that's why the base templates were originally segregated by year. However, we don't have 2020 templates and I now feel so much 2019 references in 2020 site is counter productive and confusing. It also means the 2020 site is built slightly differently than the 2019 site.

We could rename `2019` to `v1` instead of dropping the folder completely. However, I think there's a balance to be struck between how much we design for the (unknown) future, and how much we have the code reflect the current reality.

Would the 2021 Almanac, or some future one, need a complete template rewrite? Or just different CSS? Or would we just have `base_v2.html` different and the rest is basically the same? Or would we do a complete site redesign and so update 2019 and 2020 to latest design? Or would any future design move away from Flask to some other technology so folder structure would change anyway? These are questions we don't know the answer to and so having a more confusing structure now, on the chance we might use that in future, isn't the right way to go in my opinion.

Also nothing is set in stone – the relative ease with which we're moving this, and the great testing coverage we have, means this decision can easily be reverse if we need to segregate the designs again in the future.

It's a similar discussion for the 2020 chapter hero images. Currently they are in the 2019 chapter images folder and 19 out of the 22 images are reused. So should we leave them in 2019 folder? Copy to 2020 and increase our repo size for no real gain? Have some sort of soft link if Git supports that? Or move them to a year-agnostic `hero-images` folder? I originally proposed that later one, and you were concerned that if a chapter wanted to override a hero-image in the future then how would that work. Which is a fair point, but I'm still thinking that's a problem to deal with in the future rather than suffering with either complexity or duplicated images in the present. So, after some thought, I'm still thinking a `hero-images` folder is the right way to go. And we can always move to a `hero-image/v1/` and a `hero-images/v2` folder in the future if we need to.

rviscomi:
The way you're using v1 and v2 is how I'm thinking of the years themselves, ie 2019 and 2020 themselves are version numbers. So the 2020 Markup chapter is using the 2019 version of the hero image. That's ok. But it's not the same as the 2020 Capabilities chapter using the 2019 version of the hero image, because it never existed then. I can live with the first scenario but the second seems broken. The same goes for the templates; 2020 is an extension of the 2019 theme. 2019 is an extension of the base theme. Having the base be the lowest common denominator that any theme would need is IMO the right way to use abstract templates.

You made a lot of other points that I want to address, but I thought I'd throw that out there first.

bazzadp:
> So the 2020 Markup chapter is using the 2019 version of the hero image. That's ok. But it's not the same as the 2020 Capabilities chapter using the 2019 version of the hero image, because it never existed then.

Yes but then you're left with either copying all the 2019 images to 2020 chapter and duplicating, or having a "if 2020 exists, then use that, else fallback to 2019" (that won't scale to more years), or having chapter meta data to say which folder to use. All this complexity for a problem we don't currently have - we only have one set of hero-images. Of course that might change in future but, as I say, who knows if we'll even be on this stack in the future.

As to the templates we currently have the following for the 2019 home page:

The 2019 home page is `en/2019/index.html`
which extends `base/2019/index.html`
which extends `en/2019/base.html`
which extends `base/2019/base.html`
which extends `base.html`

Now ignoring the fact we have to read 5 templates for now (there's good reasons for that even if it seems over the top), that's what we have for 2019. But crucially not for 2019.

The 2019 home page is `en/2020/index.html`
which extends `base/2019/index.html`
which extends `en/2020/base.html`
which extends `en/2019/base.html`
which extends `base/2019/base.html`
which extends `base.html`

So there is an extra level of templating in there (6 rather than 5) and three 2019 templates in the 2020 build path. So a 2019 page is built one way, and a 2020 page the other. This could lead to issues if we only test one version.

Additionally, `en/2020/base.html` and `en/2019/base.html` are massively different files because the 2019 version includes LOTs more data and the 2020 version only overrides a few of these. And of course we NEED the 2019 templates before we can launch a 2020 language.

I would much rather having this:

The 2019 home page is `en/2019/index.html`
which extends `base/index.html`
which extends `en/2019/base.html` (not sure we even need this - haven't looked closely enough yet)
which extends `en/base.html`
which extends `base/base.html`
which extends `base.html`

The 2019 home page is `en/2020/index.html`
which extends `base/index.html`
which extends `en/2020/base.html` (not sure we even need this - haven't looked closely enough yet)
which extends `en/base.html`
which extends `base/base.html`
which extends `base.html`

That way:

- Each year is built the same, which is better for consistency and testing.
- Each year template only contains the stuff needed by that year, rather than loading and overriding for some years.
- We can launch 2020 in Scottish (for example), without having to have any 2019 templates (this is currently not possible).
- There is no confusion as to why 2019 templates are needed by 2020 pages.

2019, 2020 and 2021 (assuming it's the same design) are all exactly the same.

rviscomi:
Can we defer this to after the 2020 launch though?

bazzadp:
My preference would be to get this out of the way in the next week giving us a clear week or even two before launch to monitor for issues while there’s many eyes on it.

The reduces risk of something not working properly because it’s still hardcoded incorrectly to 2019 (e.g. ebook, or first chapter translation). And, as I say, it means any new translators can start translating the 2020 Almanac rather than have to do the 2019 templates first.

Delaying until after launch means not doing it until late January realistically and then there being a lot less eyes on it when we do do it.

rviscomi:
Whose eyes, developers to do code reviews or users to spot errors? I'd think it's better to make big changes that could cause user-facing issues _after_ a big launch to avoid any risk of breaking things.

bazzadp:
OK I'll make one last ditch effort to persuade you to take this now, and then will drop it until next year if you still don't agree.

> Whose eyes, developers to do code reviews or users to spot errors?

The Web Almanac contributors now. I think we have a unique opportunity over the next 2 weeks that we have a lot of eyes (Authors, Reviewers and Contributors) on their chapters before launch. The staging process we use helps enormously with this and is a very good idea. After launch dies down we know from experience last year, it will quickly die down to almost no one.

As to the code review of this change, I think you'll find there is very little to do, if you agree with the approach in this discussion, despite the apparent size of this change. If you look at the changes 99% of them are a simple switch from `base/2019/` to `base/`. It's then pretty obvious whether a chapter works or not for each year. Then the automated testing then takes care of the rest.

I'm not sure anyone, apart from myself, understands the logic of the template inheritance at present, that can review it form that perspective - which is a large part of what I'm trying to fix here!

>  I'd think it's better to make big changes that could cause user-facing issues after a big launch to avoid any risk of breaking things.

I don't think this is a big change, or I wouldn't suggest it this close to the launch. Yes it's undeniable a lot of files have changed but, as per above, the actual changes themselves are very minor and mostly covered by our automated testing (which tests every URL for every language returns a 200 and also some negative testing still results in a 404 or a redirect).

I also am doing this simplification to __reduce__ the chance if things breaking - like they did in #1468 

Finally we're seeing a lot of interest in new languages. We've launched Traditional Chinese, Hindi and soon Russian in last few weeks. There is a good chance the launch will add further interest, and this simplification, makes it easier to launch years in isolation (so they can concentrate on 2020 translations).

rviscomi:
After having given this more thought, TBH I still don't see the benefit/urgency. But you've spent more time thinking about this so I'll defer to your judgment :)

rviscomi (approved):

PR #1560, opened 2020-11-20
2020 CSS SQL updates

rviscomi:
Progress on #898 

cc @LeaVerou


sql/2020/01_CSS/custom_property_value_types.sql
```diff
@@ -0,0 +1,159 @@
...
+            incrementByKey(ret.values, value);
+
+            // What type of value is it?
+            let type;
+
+            let call = null;
+            // NOTE(rviscomi): Excluding XL values to avoid UDF timeouts.
+            if (value.length < 1000) {
+              call = extractFunctionCalls(value).filter(c => c.pos[0] === 0 && c.pos[1] === value.length)[0];
+            }
```
rviscomi, line 82: @LeaVerou FYI this `extractFunctionCalls` was causing the UDF to timeout on the full dataset. I suspect it was a problem with some values being parsed incorrectly because capping the length to 1000 seems to have fixed it. LMK if this is ok or if there's a better way.

rviscomi: Results in [this sheet](https://docs.google.com/spreadsheets/d/1sMWXWjMujqfAREYxNbG_t1fOJKYCA6ASLwtz4pBQVTw/edit?ts=5fad577d#gid=1306054386)

LeaVerou: Thanks! Yeah, that seems like a reasonable heuristic. Indeed, it seems common for Rework to parse multiple rules and declarations as one value, either due to syntax errors in the CSS, or due to parser bugs.

LeaVerou (approved):

sql/2020/01_CSS/custom_property_value_types.sql
rviscomi: As this is a custom metric, the parsing bug would have been in [css-variables.js](https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/css-variables.js), not Rework. Rework has definitely had similar bugs though.

LeaVerou: There is no logic to parse declarations there, so it must be a syntax error in the CSS.

rviscomi: Gotcha. Thanks for clarifying.

Based on your review of the results, do you think there are any more SQL changes needed? If not I'll go ahead and merge this, otherwise I'll keep it open to catch any remaining changes. I think I've responded to all of your comments, though some still have open questions.

PR #1564, opened 2020-11-21
Add support for EditorConfig

lex111:
I think this needs to be added as it helps with basic text formatting, which is sometimes difficult to keep track of manually.

EditorConfig is probably supported by all editors, I use VS Code and it works perfectly there.

> EditorConfig is a standardized opt-in way to help collaborators auto-set their editor on a per-project basis for various formatting options, e.g. spacing and tabs. Some editors (e.g. WebStorm) honor it out of the box; others (e.g. VSC) require installing a small extension. EditorConfig files also affect how code renders in certain engines, including GitHub's web interface.

https://editorconfig.org/


bazzadp:
I can't get this to work for Visual Studio Code. I probably being stupid but might need to add to README how to use it.

I tried saving an existing file which doesn't follow the rules, and expect it to change the spaces - no change.

Then I find out [you need to install this via node](https://stackoverflow.com/questions/46846128/editorconfig-for-vs-code-not-working). OK so I do that. Still doesn't work.

Then uninstall the Extension, restart VS Code, reinstall the extension. Still doesn't work.

Then I find out [open files might be an issue](https://github.com/editorconfig/editorconfig-vscode/issues/283). So I close all files, restart VSCode, open file and edit. Doesn't work.

Then I try to create a new file. Doesn't work.

What am I doing wrong?

lex111:
Are you sure you have this extension installed in your VS Code? https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig
To make EditorConfig work, I only had to do that.

bazzadp:
OK it is working. I presumed it would auto format on save, but looks like you need to explicitly choose "Format Document" to change existing documents - and even then seems to not work on Python files. That's probably a good thing as avoids massive changes to existing code not following these convention.

.editorconfig
```diff
@@ -0,0 +1,10 @@
+# editorconfig.org
+root = true
+
+[*.html]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
```
bazzadp, line 10: These seem to match our convention:

```suggestion
[*]
indent_style = space
end_of_line = lf
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true

[*.{html,md,js,css,sql}]
indent_size = 2

[*.py]
indent_size = 4
```

bazzadp (changes requested):
I still had to have editorconfig npm installed. Could you add a section to the `src/README.md` to note that? Something along the lines of:

> The Web Almanac uses a specific "style guide" for code, including 2 spaces for indentation (except Python which uses 4 spaces).
>
> An `.editorconfig` file exists for those using [EditorConfig](https://editorconfig.org/) to help enforce those styles. This may require installing the `editorconfig` node module via npm (we suggest installing globally - `npm install -g editorconfig` ) and then installing the extension for your IDE (for example there is a [Visual Studio Code extension](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig)).

Let's also add to `src/.gcloudignore` and `src/.dockerignore` as no point including this in those.

lex111:
I've added a new section to the README file, but I'm not sure about the its heading, so feel free to change it. Also I moved the `.editorconfig` file to the `src` folder.

src/.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.{html,md,js,css,sql}]
+indent_size = 2
+
+[*.py]
+indent_size = 4
```
rviscomi, line 15: @bazzadp any reason not to prefer 2 spaces for consistency with every other language?

bazzadp: When I looked at this when setting up the linters I found claims that [4 spaces are wildly more common in Python than two](https://stackoverflow.com/a/1125719).

The linters were therefore set up with the default of 4 spaces so this has been enforced for a while. While pylint can be configured in our config file to be two, it doesn’t look like flake8 can except on command line and the [GitHub super linter doesn’t allow that yet](https://github.com/github/super-linter/issues/987).

One of the things I like about this PR is that it offers the option of enforcing the two standards easily.

bazzadp:
> Also I moved the `.editorconfig` file to the `src` folder.

Actually you love it back? Want these for the SQL files too and they don’t live under src.

lex111:
Ah, yes, it is makes sense, so moved it back to root of repository.

bazzadp (approved):
I’m good with this. Thanks for this @lex111 !

@rviscomi any further comments or happy to merge?

src/.editorconfig
bazzadp: Also spaces much more important in Python so I actually think it’s a good thing for them to be more obvious there.

rviscomi (approved):

PR #1567, opened 2020-11-21
Improve queries with CDNs, update tables dates

denar90:
- update CDNs to detect Cloudflare proxies of original deploy service. Was able to detect AWS clodufront, Verce, Netlify proxied via Cloudflare. Kudos to @ahmadawais 
- update table dates

### DIFF
--- a/sql/2020/17_JAMstack/ssg_compared_to_2019.sql
+++ b/sql/2020/17_JAMstack/ssg_compared_to_2019.sql
@@ -8,13 +8,13 @@ SELECT
   total,
   COUNT(0) / total AS pct
 FROM
-  `httparchive.technologies.2020_08_01_*`
+  `httparchive.technologies.2020_10_01_*`
 JOIN (
   SELECT
     _TABLE_SUFFIX,
     COUNT(0) AS total
   FROM
-    `httparchive.summary_pages.2020_08_01_*`
+    `httparchive.summary_pages.2020_10_01_*`
   GROUP BY
     _TABLE_SUFFIX)
 USING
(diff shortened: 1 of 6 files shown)


ahmadawais:
Excellent work @denar90. 🎉

@rviscomi can you please merge the PR so we cam have the final data required to write the several sections in the content?

ahmadawais (approved):

rviscomi:
Is there something wrong with the August dataset? Per the [2020 Methodology](https://almanac.httparchive.org/en/2020/methodology#dataset), all chapters are standardized on the August dataset, unless there was some kind of data loss, for example a bug in Lighthouse that got fixed in September.

denar90:
No, just wanted to get fresh data, I'll revert the change.

sql/2020/17_JAMstack/ssg_compared_to_2019.sql
```diff
@@ -35,13 +35,13 @@ SELECT
   total,
   COUNT(0) / total AS pct
 FROM
-  `httparchive.technologies.2019_07_01_*`
+  `httparchive.technologies.2019_08_01_*`
```
rviscomi, line 38: 2019 used the July dataset. These changes should also be reverted.

rviscomi:
Oh ok, yeah I understand that August data may feel a bit stale by now, but think about the Almanac more like a snapshot in time, where each chapter looks at the same snapshot from different angles. Then next year we'll take another snapshot, compare, and so on. It's also the reason why we made a big push in July to preplan all of the analysis, because we needed to lock in any pipeline changes (eg custom metrics) before the August crawl began. Thanks for reverting, this is looking better now, just need to revert the 2019 changes as well.

sql/2020/17_JAMstack/ssg_compared_to_2019.sql
```diff
@@ -35,13 +35,13 @@ SELECT
...
 JOIN (
   SELECT
     _TABLE_SUFFIX,
     COUNT(0) AS total
   FROM
-    `httparchive.summary_pages.2019_07_01_*`
+    `httparchive.summary_pages.2019_08_01_*`
```
rviscomi, line 44:
```suggestion
    `httparchive.summary_pages.2019_07_01_*`
```

rviscomi (approved):

denar90:
Thank you

PR #1574, opened 2020-11-25
Syntax Highlighting Capabilities

bharatagsrwal:
Syntax Highlighting Capabilities for HTML, CSS, JS and SQL added

please check `generate_chapters.js`, `2019.css` and `package.json` file.

### DIFF
--- a/src/package.json
+++ b/src/package.json
@@ -43,6 +43,8 @@
     "node-fetch": "2.6.1",
     "xml-js": "1.6.11",
     "run-script-os": "1.1.3",
-    "node-watch": "0.7.0"
-  }
+    "node-watch": "0.7.0",
+    "rainbow-code": "^2.1.7"
+  },
+  "dependencies": {}
 }
--- a/src/package.json
+++ b/src/package.json
@@ -44,7 +44,6 @@
     "xml-js": "1.6.11",
     "run-script-os": "1.1.3",
     "node-watch": "0.7.0",
-    "rainbow-code": "^2.1.7"
-  },
-  "dependencies": {}
+    "rainbow-code": "2.1.7"
+  }
 }
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -1449,7 +1449,6 @@ p.copyright a {
 /* Syntax Highlighting */
 pre {
   margin: 0;
-  border: 1px solid #ccc;
   word-wrap: break-word;
   padding: 6px 10px;
   line-height: 19px;
@@ -1464,23 +1463,14 @@ pre code {
 }
 
 code {
-  border: 1px solid #eaeaea;
   margin: 0px 2px;
   padding: 0px 5px;
-  font-size: 12px;
+  font-size: 1rem;
 }
 
 pre,
 code {
-  font-family: Consolas, 'Liberation Mono', Courier, monospace;
   color: #333;
-  background: #f8f8f8;
-  border-radius: 3px;
-}
-
-pre,
-pre code {
-  font-size: 13px;
 }
 
 pre .comment {
@@ -1513,6 +1503,10 @@ pre .css-value,
 pre .vendor-fix,
 pre .support.namespace {
   color: #333;
+}
+
+pre .function.call{
+  color: navy;
 }
 
 pre .constant.numeric,
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -9,10 +9,10 @@
     local('Lato Italic'),
     local('Lato-Italic'),
     /* url(https://fonts.gstatic.com/s/lato/v16/S6u8w4BMUTPHjxsAUi-qNiXg7eU0.woff2) */
-    url('/static/fonts/Lato-Italic-ext.woff2')
-    format('woff2');
+    url('/static/fonts/Lato-Italic-ext.woff2') format('woff2');
   unicode-range: U+0100-024F, U+0259, U+1E00-1EFF, U+2020, U+20A0-20AB, U+20AD-20CF, U+2113, U+2C60-2C7F, U+A720-A7FF;
 }
+
 /* latin */
 @font-face {
   font-family: 'Lato';
@@ -23,10 +23,10 @@
     local('Lato Italic'),
     local('Lato-Italic'),
     /* url(https://fonts.gstatic.com/s/lato/v16/S6u8w4BMUTPHjxsAXC-qNiXg7Q.woff2) */
-    url('/static/fonts/Lato-Italic.woff2')
-    format('woff2');
+    url('/static/fonts/Lato-Italic.woff2') format('woff2');
   unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
 }
+
 /* latin-ext */
 @font-face {
   font-family: 'Lato';
@@ -37,10 +37,10 @@
     local('Lato Bold Italic'),
     local('Lato-BoldItalic'),
     /* url(https://fonts.gstatic.com/s/lato/v16/S6u_w4BMUTPHjxsI5wq_FQftx9897sxZ.woff2) */
-    url('/static/fonts/Lato-BoldItalic-ext.woff2')
-    format('woff2');
+    url('/static/fonts/Lato-BoldItalic-ext.woff2') format('woff2');
   unicode-range: U+0100-024F, U+0259, U+1E00-1EFF, U+2020, U+20A0-20AB, U+20AD-20CF, U+2113, U+2C60-2C7F, U+A720-A7FF;
 }
+
 /* latin */
 @font-face {
   font-family: 'Lato';
@@ -51,10 +51,10 @@
     local('Lato Bold Italic'),
     local('Lato-BoldItalic'),
     /* url(https://fonts.gstatic.com/s/lato/v16/S6u_w4BMUTPHjxsI5wq_Gwftx9897g.woff2) */
-    url('/static/fonts/Lato-BoldItalic.woff2')
-    format('woff2');
+    url('/static/fonts/Lato-BoldItalic.woff2') format('woff2');
   unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
 }
+
 /* latin-ext */
 @font-face {
   font-family: 'Lato';
@@ -65,10 +65,10 @@
     local('Lato Regular'),
     local('Lato-Regular'),
     /* url(https://fonts.gstatic.com/s/lato/v16/S6uyw4BMUTPHjxAwXiWtFCfQ7A.woff2) */
-    url('/static/fonts/Lato-Regular-ext.woff2')
-    format('woff2');
+    url('/static/fonts/Lato-Regular-ext.woff2') format('woff2');
   unicode-range: U+0100-024F, U+0259, U+1E00-1EFF, U+2020, U+20A0-20AB, U+20AD-20CF, U+2113, U+2C60-2C7F, U+A720-A7FF;
 }
+
 /* latin */
 @font-face {
   font-family: 'Lato';
@@ -79,10 +79,10 @@
     local('Lato Regular'),
     local('Lato-Regular'),
     /* url(https://fonts.gstatic.com/s/lato/v16/S6uyw4BMUTPHjx4wXiWtFCc.woff2) */
-    url('/static/fonts/Lato-Regular.woff2')
-    format('woff2');
+    url('/static/fonts/Lato-Regular.woff2') format('woff2');
   unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
 }
+
 /* latin-ext */
 @font-face {
   font-family: 'Lato';
@@ -93,10 +93,10 @@
     local('Lato Bold'),
     local('Lato-Bold'),
     /* url(https://fonts.gstatic.com/s/lato/v16/S6u9w4BMUTPHh6UVSwaPGQ3q5d0N7w.woff2) */
-    url('/static/fonts/Lato-Bold-ext.woff2')
-    format('woff2');
+    url('/static/fonts/Lato-Bold-ext.woff2') format('woff2');
   unicode-range: U+0100-024F, U+0259, U+1E00-1EFF, U+2020, U+20A0-20AB, U+20AD-20CF, U+2113, U+2C60-2C7F, U+A720-A7FF;
 }
+
 /* latin */
 @font-face {
   font-family: 'Lato';
@@ -107,10 +107,10 @@
     local('Lato Bold'),
     local('Lato-Bold'),
     /* url(https://fonts.gstatic.com/s/lato/v16/S6u9w4BMUTPHh6UVSwiPGQ3q5d0.woff2) */
-    url('/static/fonts/Lato-Bold.woff2')
-    format('woff2');
+    url('/static/fonts/Lato-Bold.woff2') format('woff2');
   unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
 }
+
 /* latin-ext */
 @font-face {
   font-family: 'Lato';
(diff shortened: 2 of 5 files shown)


src/package.json
```diff
@@ -43,6 +43,8 @@
     "node-fetch": "2.6.1",
     "xml-js": "1.6.11",
     "run-script-os": "1.1.3",
-    "node-watch": "0.7.0"
-  }
+    "node-watch": "0.7.0",
+    "rainbow-code": "^2.1.7"
```
bazzadp, line 47: We do hard versions and then dependabot notifies us of any updates. This gives us visibility of dependency updates (though on flip side also means we get an email and automatci PR when one does happen and know that annoys some people!)

```suggestion
    "rainbow-code": "2.1.7"
```

src/package.json
```diff
@@ -43,6 +43,8 @@
...
+  },
+  "dependencies": {}
```
bazzadp, line 49: Not sure why this is needed?

```suggestion
  }
```

src/static/css/2019.css
```diff
@@ -9,10 +9,10 @@
     local('Lato Italic'),
     local('Lato-Italic'),
     /* url(https://fonts.gstatic.com/s/lato/v16/S6u8w4BMUTPHjxsAUi-qNiXg7eU0.woff2) */
-    url('/static/fonts/Lato-Italic-ext.woff2')
-    format('woff2');
+    url('/static/fonts/Lato-Italic-ext.woff2') format('woff2');
```
bazzadp, line 12: Seems like a lot of needless change in this file. Presume you're using a code formatter?

I'm not averse to any of the changes per se, but it does make it difficult to see the actual changes this PR is introducing. Is it possible to only include those?

src/static/css/2019.css
```diff
@@ -1449,3 +1460,106 @@ p.copyright a {
...
+pre .support.value {
+  font-weight: normal;
+  color: #d14
+}
+
+pre .regexp {
+  color: #009926
+}
+
+/* Syntex Highlighting */
```
bazzadp, line 1565: Do we need this last line?

```suggestion
```

src/tools/generate/generate_chapters.js
```diff
@@ -2,7 +2,10 @@ const fs = require('fs-extra');
 const showdown = require('showdown');
 const ejs = require('ejs');
 const prettier = require('prettier');
+const rainbow = require('rainbow-code');
```
bazzadp, line 5: Could we move all this code into it's own file? Like we do for generate_header_links, generate_figure_ids...etc.

src/tools/generate/generate_chapters.js
```diff
@@ -116,10 +119,39 @@ const generate_chapters = async (chapter_match) => {
 
 };
 
+const generate_syntex_snippet = (element, body, type) => {
```
bazzadp, line 122: Syntax not syntex

```suggestion
const generate_syntax_snippet = (element, body, type) => {
```

src/tools/generate/generate_chapters.js
```diff
@@ -116,10 +119,39 @@ const generate_chapters = async (chapter_match) => {
...
+  });
+  css_codes.forEach(element => {
+    body = generate_syntex_snippet(element, body, "css");
+  });
+  js_codes.forEach(element => {
+    body = generate_syntex_snippet(element, body, "javascript");
+  });
+  sql_codes.forEach(element => {
+    body = generate_syntex_snippet(element, body, "sql");
+  });
```
bazzadp, line 151: Could this all be moved to function so we could just call:

```
body = generate_syntax_snippet(element, body, "html");
body = generate_syntax_snippet(element, body, "css");
body = generate_syntax_snippet(element, body, "js");
body = generate_syntax_snippet(element, body, "sql");
```

And it does the `querySelectorAll` and `forEach` in the function? This would avoid having to repeat the same code 4 times.

bazzadp (changes requested):
Not tired it out yet, but looks like the way I think we should implement it so good stuff.

However, I do have some comments for you to look at. After that will give it a trial run.

src/static/css/2019.css
bazzadp: I see a lot of these changes are also causing our linter issues so definitely need to be reverted.

bharatagsrwal: I have installed linter as suggest but I don't know why that is not working, I am trying to resolve that, and soon make a commit.

bazzadp:
Also I see [Lighthouse is complaining as we have some invisible text](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1606319555708-58531.report.html#accessibility) so might need to look at the CSS we've chosen for these styles.

Lighthouse errors

P.S. I love our automated testing to catch these things!! 🎉

src/tools/generate/generate_chapters.js
bharatagsrwal: Damn 😅

src/static/css/2019.css
bazzadp: The linter just checks things meet our preferred style - it doesn't change them to make them meet that style so it's just an alerting system. you need to change them manually.

You must have some auto code formatter in your editor that's that's doing these changes?

bharatagsrwal: @bazzadp How do I fix these?

bazzadp: I think easiest is to take the old version [from here](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/src/static/css/2019.css) and then apply you changes again (presume they are just the ones at the bottom?).

But be careful your code editor doesn't reformat them again. I presume you have some plugin that's doing this? If so might need to figure out what is doing that.

Do a `git diff` before committing that update anyway to make sure only the changes you intended to include are included and everything else is reverted back.

bharatagsrwal:
@bazzadp, Changes are done as per Review

bazzadp:
Looking good from a brief scan of the code! Will spend some time testing this today and let you know how I get on.

One comment so far is that the Methodology section now uses different Syntax highlighting for its SQL snippet. We should move that to Rainbow format, and remove its custom code CSS, so we don’t need to maintain two and so we can auto-generate it in future. But let’s merge this first and then look at that separately as there are quite a lot of Methodology files to change with all the various translations, so this initial PR for the chapters is easier to read and review.

src/static/css/2019.css
```diff
@@ -1449,3 +1445,103 @@ p.copyright a {
     transition-delay: 0s !important;
   }
 }
+
+/* Syntax Highlighting */
+pre {
+  margin: 0;
+  border: 1px solid #ccc;
```
bazzadp, line 1452: Let's remove this border to minimize change (and also doesn't look great on Methodology page):

```suggestion
```

src/static/css/2019.css
```diff
@@ -1449,3 +1445,103 @@ p.copyright a {
...
+  margin: 0px 2px;
+  padding: 0px 5px;
+  font-size: 12px;
+}
+
+pre,
+code {
+  font-family: Consolas, 'Liberation Mono', Courier, monospace;
+  color: #333;
+  background: #f8f8f8;
```
bazzadp, line 1477: Let's remove this border to minimize change (and also doesn't look great on Methodology page):

```suggestion
```

bazzadp (changes requested):
This works great @bharatagsrwal ! Brilliant work.

Just a few small changes and I think this is good to merge.

src/tools/generate/generate_syntax_highlighting.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {String} body HTML body which was parsed from markdown.
+ * @returns body with highlighting syntax snippet.
+ */
+
+const generate_syntax_highlighting = (body) => {
+  const dom = new JSDOM(body);
+  body = generate_syntax_highlighting_for_language(dom, body, "html", "html");
+  body = generate_syntax_highlighting_for_language(dom, body, "css", "css");
+  body = generate_syntax_highlighting_for_language(dom, body, "js", "javascript");
+  body = generate_syntax_highlighting_for_language(dom, body, "sql", "sql");
```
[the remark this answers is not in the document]
bharatagsrwal: Yeah, we can use this also, I will change this and will update it soon,

src/static/css/2019.css
bharatagsrwal: cool, I will change these things as soon as possible, Thanks @bazzadp for suggestions

src/tools/generate/generate_chapters.js
```diff
@@ -3,6 +3,7 @@ const showdown = require('showdown');
 const ejs = require('ejs');
 const prettier = require('prettier');
 
+
```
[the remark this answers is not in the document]
bharatagsrwal: I thought it will help us to make a difference between dependencies and our custom functions and modules.

But anyway I will remove it.

src/static/css/2019.css
bharatagsrwal: I think we should have the background to make it different from others, as I have checked removed border looks fine but when I remove the background it gets the same as earlier, and when we have background it attracts focus, it shows there is something different over there. 

@bazzadp what are your opinions?

bharatagsrwal: image
Without background

image
with background

bazzadp: I pondered this myself. In some ways it helps differentiate it. In other ways, because the code snippets are small it's not really needed.

However the reason I ultimately suggested removing it are because of the knock on effects to other places:
- The Methodology SQL code snippet. Look at that before and after. Definitely looks worse with background because we have the floating card. Now maybe we should replace the floating card with this instead but I was trying to minimise the changes in this as launch date is fast approaching. And the next point:
- In the Table of Contents for chapters with code in headers. For example the 2019 Caching chapter.

Do you see what I mean?

src/tools/generate/generate_chapters.js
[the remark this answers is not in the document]
bazzadp: Possibly, but inconsistent with the code elsewhere as not used this convention so I'd prefer to keep it as is, unless changing everywhere (beyond the scope of this PR).

src/static/css/2019.css
bazzadp: Methdology SQL before and after:

image

image

Table of Contents on Caching chapter before and after:

image

image

bharatagsrwal: okay cool, I will remove the background, will push newly updated

src/tools/generate/generate_syntax_highlighting.js
```diff
@@ -0,0 +1,43 @@
...
+ * Generate Syntax highlighting for a particular Languaage (ex: html, css, javascript etc..)
+ *
+ * @param {object} dom Object of class JSDOM (Parsed HTML).
+ * @param {String} body HTML body parsed from convertor
+ * @param {String} language language which rainbow syntax highlighting to use
+ * @param {String} [alias] an optional alias for a language (e.g. js for javascript)
+ * @returns body with highlighting syntax snippet of a particular language.
+ */
+const generate_syntax_highlighting_for_language = (dom, body, language, alias="") => {
+  const query_selector = alias ? `code.language-${alias}` : `code.language-${language}`;
```
bazzadp, line 15: My suggestion for this line was actually a tiny bit different:

```suggestion
  const query_selector = alias ? `code.language-${language}, code.language-${alias}` : `code.language-${language}`
```

That way it will do both `javascript` and `js` in the one go!

bharatagsrwal: Okay, yeah make sense

rviscomi:
I haven't taken a close look at this PR but one request I have is to try to keep the syntax highlighting color scheme as close to the manual one used on the Methodology page. Those colors were chosen by the site designer to match our overall palette, so it'd be great to continue to use those for consistency. Do we have the ability to override the default theme colors of the syntax highlighter tool?

bazzadp:
Yup. The syntax highlighter tool is merely used to apply the HTML spans with classes. The CSS for those classes is in 2019.css (though my understanding is it was based on one of the suggested styles from the tool).

@bharatagsrwal could you look at the Methodology page and see what changes we need to do to bring this inline?

@rviscomi as I mentioned above, after we merge this we’ll look at changing the Methodology chapter to use the same spans and classes and remove the CSS from that page so it uses the same one. That should result in no visual change to Methodology page but consistent code. But as there are a lot of Methodology pages will land this first, and then do that next. Maybe as part of that we should make the SQL a function call to avoid repeating it so much in eat translation even though it’s not translated? Anyway, as I say, let’s land this first.

bharatagsrwal:
> Yup. The syntax highlighter tool is merely used to apply the HTML spans with classes. The CSS for those classes is in 2019.css (though my understanding is it was based on one of the suggested styles from the tool).
> 
> @bharatagsrwal could you look at the Methodology page and see what changes we need to do to bring this inline?
> 
> @rviscomi as I mentioned above, after we merge this we’ll look at changing the Methodology chapter to use the same spans and classes and remove the CSS from that page so it uses the same one. That should result in no visual change to Methodology page but consistent code. But as there are a lot of Methodology pages will land this first, and then do that next. Maybe as part of that we should make the SQL a function call to avoid repeating it so much in eat translation even though it’s not translated? Anyway, as I say, let’s land this first.

okay, we just need to change the colors in CSS and we are good to do, so I will change those and make it the same as in the Methodology page

bharatagsrwal:
image

image

How about these? I have changed to those colors which are on the methodology page

bazzadp:
> image
> 
> image
> 
> How about these? I have changed to those colors which are on the methodology page

@rviscomi what do you think about something else for strings? Something like #e74c3c ?

Strong string highlighting

Without making this a least a bit more different you kind of lose the benefit of having syntax highlighting!

We don't have any string syntax highlighting in Methodology, which is why it's subtlety have not been noticed before, but comments and numbers are more visually different, so I think strings should be too:

Methodology SQL high lighting

rviscomi:
+1 for more contrast, that was my first impression of https://github.com/HTTPArchive/almanac.httparchive.org/pull/1574#issuecomment-735426933 as well.

The table name in the query (with backticks) is also technically a string if we wanted to keep it consistent.

src/static/css/2019.css
```diff
@@ -1449,3 +1445,93 @@ p.copyright a {
...
+
+pre .attribute,
+pre .variable {
+  color: #000;
+}
+
+pre .string,
+pre .support.value {
+  font-weight: normal;
+  color: #2c4a73;
```
bazzadp, line 1532:
```suggestion
  color: #e74c3c;
```

src/static/css/2019.css
```diff
@@ -1449,3 +1445,93 @@ p.copyright a {
...
+pre code {
+  border: 0;
+  padding: 0;
+  margin: 0;
+  border-radius: 0;
+}
+
+code {
+  margin: 0px 2px;
+  padding: 0px 5px;
```
bazzadp, line 1467: Let's lose these as no border now so don't need padding:

```suggestion
```

bazzadp (changes requested):
OK let's land this thing! Two small changes.

bharatagsrwal:
> if we wanted to keep it consiste

yes, but it seems like it is not a string, the color is almost the same as the default color, so I think @bazzadp #e74c3c this color is great for strings

bazzadp:
> +1 for more contrast, that was my first impression of [#1574 (comment)](https://github.com/HTTPArchive/almanac.httparchive.org/pull/1574#issuecomment-735426933) as well.
> 
> The table name in the query (with backticks) is also technically a string if we wanted to keep it consistent.

Looks like the Rainbow library we're using is clever enough to pick this up! This is hat the SQL looks like when run through that:

SQL syntax highlighting

Though interesting enough it doesn't recognise `#` as a comment in SQL so will need to manually apply the comment span for that.

bharatagsrwal:
> > +1 for more contrast, that was my first impression of [#1574 (comment)](https://github.com/HTTPArchive/almanac.httparchive.org/pull/1574#issuecomment-735426933) as well.
> > The table name in the query (with backticks) is also technically a string if we wanted to keep it consistent.
> 
> Looks like the Rainbow library we're using is clever enough to pick this up! This is what the SQL looks like when run through that:
> 
> SQL syntax highlighting
> 
> Though interesting enough it doesn't recognise `#` as a comment in SQL so will need to manually apply the comment span for that.

you did by Firstly converting using rainbow and pasting code in HTML file?

bazzadp:
> you did by Firstly converting using rainbow and pasting code in HTML file?

Nope. I copied the raw SQL from rendered Methodology page and then temporarily put it in the capabilities chapter page in a sql block and generated the chapter to see what it would look like. Did it perfectly (apart from that comment) first time. Will submit a PR to rainbow to add `#` as a SQL comment but in meantime as it's HTML we can add the span manually when we move the Methodology page over to this.

bazzadp:
PR to add `#` comments to Rainbow: https://github.com/ccampbell/rainbow/pull/250. Hopefully will be accepted so we get this in next version.

@bharatagsrwal you happy to accept those two small changes I suggested so we can merge this?

bharatagsrwal:
> PR to add `#` comments to Rainbow: [ccampbell/rainbow#250](https://github.com/ccampbell/rainbow/pull/250). Hopefully will be accepted so we get this in next version.
> 
> @bharatagsrwal you happy to accept those two small changes I suggested so we can merge this?

Yap, I am doing the changes, so what color should be for String, is it that red one or the same as of now?

bharatagsrwal:
@bazzadp 
image


https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1606681395869-34825.report.html#accessibility

when I tried to change color to #e74c3c

bazzadp:
Orange is always a tough one. 

#B31107 is the best I can find after plugging it into: https://contrast-finder.tanaguru.com/result.html?foreground=%23FFFFFF&background=%23e74c3c&ratio=7&isBackgroundTested=true&algo=Rgb&distanceSort=asc

bharatagsrwal:
> #B31107

image


looks good

bazzadp (approved):

bazzadp:
@bharatagsrwal just realised you're not in the [contributors json](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/src/config/2020.json). Could you submit a PR with your details adding yourself to the `developers` team so you appear on our [contributors page](https://almanac.httparchive.org/en/2020/contributors)? Contributors are listed alphabetically by first name.

PR #1576, opened 2020-11-25
Pytest tests for helpers.py

bazzadp:
Makes progress on #1229 

Moves pytest coverage of [helpers.py](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/src/server/helpers.py) from 31% to 100% and overall pytest coverage from 67% to 87%

Mostly the code itself is untouched, but did identify a couple of edge cases that weren't handled the best. Fixed one in #1575 as could cause issues in prod if new language added, but rest are edge cases I left for this PR.

```
Name                            Stmts   Miss  Cover   Missing
-------------------------------------------------------------
server/__init__.py                 39     12    69%   19-21, 37-45
server/config.py                   75      3    96%   103-106
server/csp.py                       1      0   100%
server/errors.py                   23      8    65%   11, 16-17, 22, 27-28, 33-34
server/feature_policy.py            1      0   100%
server/helpers.py                 126      0   100%
server/language.py                 33      1    97%   10
server/routes.py                   72     30    58%   12, 18, 24-26, 32, 38-42, 48, 58-61, 78-82, 90-92, 98, 103, 109-112, 118
server/tests/__init__.py            0      0   100%
server/tests/helpers_test.py      182      0   100%
server/tests/validate_test.py      52      0   100%
server/validate.py                 79     33    58%   25-53, 59-72, 81-82, 104, 108-109
-------------------------------------------------------------
TOTAL                             683     87    87%
```

@SaptakS since you reviewed my last Python change could you have a look over this?
And would gladly take any other feedback from anyone else with Python experience.

### DIFF
--- a/src/server/helpers.py
+++ b/src/server/helpers.py
@@ -108,7 +108,7 @@ def get_chapter_nextprev(config, chapter_slug):
         if found and next_chapter:
             break
 
-    if found is False:
+    if not found:
         return None
 
     return prev_chapter, next_chapter
--- a/src/server/helpers.py
+++ b/src/server/helpers.py
@@ -71,12 +71,9 @@ def render_error_template(error, status_code):
     if lang not in supported_langs:
         lang = DEFAULT_LANGUAGE.lang_code
 
-    if not (os.path.isfile(TEMPLATES_DIR + '/%s/%s/error.html' % (lang, year))):
-        if os.path.isfile(TEMPLATES_DIR + '/%s/%s/error.html' % (lang, DEFAULT_YEAR)):
-            year = DEFAULT_YEAR
-        elif os.path.isfile(TEMPLATES_DIR + '/%s/%s/error.html' % (DEFAULT_LANGUAGE.lang_code, DEFAULT_YEAR)):
+    if not (os.path.isfile(TEMPLATES_DIR + '/%s/error.html' % (lang))):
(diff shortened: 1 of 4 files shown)


src/server/helpers.py
```diff
@@ -108,6 +108,9 @@ def get_chapter_nextprev(config, chapter_slug):
         if found and next_chapter:
             break
 
+    if found is False:
```
ibnesayeed, line 111: Alternatively:

```suggestion
    if not found:
```

bazzadp: Good idea. Thanks.

bazzadp:
@ibnesayeed addressed your comment. Any reason not to merge this if you've had a look through it?

ibnesayeed (approved):
LGTM!

As a future task, I think there are opportunities to use `@pytest.mark.parametrize()` decorator in many places where tests have a lot of similarities with slight differences. This will make tests DRY and more readable while allowing us to test more with less code.

bazzadp:
> LGTM!
> 
> As a future task, I think there are opportunities to use `@pytest.mark.parametrize()` decorator in many places where tests have a lot of similarities with slight differences. This will make tests DRY and more readable while allowing us to test more with less code.

Thanks for approving! Yeah there's definitely scope for `@pytest.mark.parametrize()`, though I do find it a bit less readable personally so need to strike the balance.

But do think this PR (and the coverage we added in previous one) gives us good basis for finish out the pytests and getting 100% coverage.

PR #1577, opened 2020-11-25
CSS 2020 queries

LeaVerou:
Progress on #898


sql/2020/01_CSS/all_functions.sql
```diff
@@ -0,0 +1,54 @@
...
+'''
+OPTIONS (library="gs://httparchive/lib/css-utils.js");
+
+SELECT
+  *
+FROM (
+  SELECT
+    client,
+    prop,
+    COUNT(DISTINCT page) AS pages,
```
rviscomi, line 39: Are you interested in this as % of pages as well, or is the absolute number useful enough for context? Not super important to calculate in the query as opposed to the sheet though.

LeaVerou: I've been adding computed columns to the sheets with the total # of pages and pct_pages. Doing it in the query would be preferable, if it doesn't increase complexity too much.

LeaVerou:
I also just added a query for all Blink features. Quite enlightening results! I wonder if it would be best to use those numbers instead of the Rework ones for the metrics that could be served by those.

sql/2020/01_CSS/vendor_prefix_summary.sql
```diff
@@ -0,0 +1,113 @@
...
+
+				for (let pseudo of pseudos) {
+					let type = "pseudo_" + (pseudo.indexOf("::") === 0? "elements" : "classes");
+					incrementByKey(ret[type], pseudo);
+					ret.total[type]++;
+				}
+			}
+
+			if (rule.declarations) {
+				walkDeclarations(rule, ({property, value}) => {
```
rviscomi, line 30: The `// -prefix-function()` section is causing this query to timeout. I think a value check might prevent that. The query takes a very long time to run and I'm logging off for the night but wanted to give that suggestion for you to see in the morning.
```suggestion
				walkDeclarations(rule, ({property, value}) => {
				if (value.length > 1000) {
					return;
				}
```

rviscomi (changes requested):

sql/2020/01_CSS/vendor_prefix_summary.sql
LeaVerou: Still no results. Also note that this is happening with the `parsed_css_1k` table, so I think it may actually be returning no results due to a bug with the logic.

sql/2020/01_CSS/all_features.sql
```diff
@@ -0,0 +1,29 @@
...
+  SUM(feature_freq) * 100 / total as pct_pages
+FROM (
+  SELECT
+    client,
+    feature,
+    COUNT(DISTINCT url) AS feature_freq
+  FROM
+    `httparchive.blink_features.features`
+  WHERE
+    yyyymmdd = '20200801'
```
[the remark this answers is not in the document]
LeaVerou: I respectfully disagree. It's not that big a table (4000 rows, slightly less if we exclude unnamed features), and it also helps us cross-check other stats to find anomalies. E.g. I'm still unsure why popular properties like `color` have higher usage via Blink features compared to the AST queries. It also helps easily measure some stats we may not have planned to query (e.g. `display: contents`), and frankly, is interesting in its own right.

sql/2020/01_CSS/vendor_prefix_summary.sql
rviscomi: Confirmed that the suggested change produces results:


client | prop | pages | freq | total | pct
-- | -- | -- | -- | -- | --
desktop | total | 5036308 | 2934010328 | 5868020656 | 0.5
mobile | total | 5779675 | 3899424875 | 7798849750 | 0.5
mobile | properties | 5760913 | 3265650025 | 7798849750 | 0.4187348301
desktop | properties | 5015245 | 2453285776 | 5868020656 | 0.4180772223
mobile | keywords | 4884489 | 264043378 | 7798849750 | 0.03385670791
desktop | keywords | 4116709 | 184325267 | 5868020656 | 0.03141182995
desktop | functions | 3540529 | 141897399 | 5868020656 | 0.02418147572
mobile | functions | 4155633 | 172619077 | 7798849750 | 0.02213391494
desktop | pseudo_elements | 4228723 | 97160938 | 5868020656 | 0.01655770211
mobile | pseudo_elements | 4860030 | 127033773 | 7798849750 | 0.01628878323
desktop | pseudo_classes | 3385086 | 37292161 | 5868020656 | 0.006355151624
mobile | pseudo_classes | 3843642 | 44044055 | 7798849750 | 0.005647506544
desktop | media | 2569576 | 20048787 | 5868020656 | 0.003416618341
mobile | media | 3316548 | 26034567 | 7798849750 | 0.003338257286

sql/2020/01_CSS/vendor_prefix_summary.sql
```diff
@@ -0,0 +1,113 @@
...
+
+SELECT
+  *
+FROM (
+  SELECT
+    client,
+    prop,
+    COUNT(DISTINCT page) AS pages,
+    COUNT(0) AS freq,
+    SUM(COUNT(0)) OVER (PARTITION BY client) AS total,
```
rviscomi, line 103: This double-counts the `total` "property". See https://github.com/HTTPArchive/almanac.httparchive.org/pull/1577/files/4c28d6db9fb721f397f1600ea9a359f9b83e43c3#diff-c1fb1712782c56177cb3b4af70616e7bdce3ea485c02d97fb889301bc4d2413eR47 for an example of how to omit it from the sum of frequencies.

rviscomi (changes requested):

sql/2020/01_CSS/all_features.sql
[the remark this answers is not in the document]
rviscomi: Fine. Can you at least query the summary table?

```sql
SELECT
  client,
  feature,
  num_urls AS freq,
  total_urls AS total,
  pct_urls AS pct_pages
FROM
  `httparchive.blink_features.usage`
WHERE
  yyyymmdd = '20200801'
ORDER BY
  pct_pages DESC
```

sql/2020/01_CSS/vendor_prefix_summary.sql
rviscomi: Here's the full query that I ran to get those results:

```sql
#standardSQL
CREATE TEMPORARY FUNCTION getPrefixStats(css STRING) RETURNS ARRAY<STRING> LANGUAGE js AS '''
try {
  function compute() {
    let ret = {
      total: {},
      pseudo_classes: {},
      pseudo_elements: {},
      properties: {},
      functions: {},
      keywords: {},
      media: {}
    };

    ret.total = Object.fromEntries(Object.keys(ret).map(k => [k, 0]));

    walkRules(ast, rule => {
      // Prefixed pseudos
      if (rule.selectors) {
        let pseudos = rule.selectors.flatMap(r => r.match(/::?-[a-z]+-[\\w-]+/g) || []);

        for (let pseudo of pseudos) {
          let type = "pseudo_" + (pseudo.indexOf("::") === 0? "elements" : "classes");
          incrementByKey(ret[type], pseudo);
          ret.total[type]++;
        }
      }

      if (rule.declarations) {
        walkDeclarations(rule, ({property, value}) => {
          if (value.length > 1000) {
            return;
          }

          // Prefixed properties
          if (/^-[a-z]+-.+/.test(property)) {
            incrementByKey(ret.properties, property);
            ret.total.properties++;
          }

          // -prefix-function()
          for (let call of extractFunctionCalls(value, {names: /^-[a-z]+-.+/})) {
            incrementByKey(ret.functions, call.name);
            ret.total.functions++;
          }

          // Prefixed keywords
          if (!matches(property, /(^|-)(transition(-property)?|animation(-name)?)$/)) {
            for (let k of value.matchAll(/(?<![-a-z])-[a-z]+-[a-z-]+(?=$|\\s|,|\\/)/g)) {
              incrementByKey(ret.keywords, k);
              ret.total.keywords++;
            }
          }
        });
      }

      // Prefixed media features
      if (rule.media) {
        let features = rule.media
                  .replace(/\\s+/g, "")
                  .match(/\\(-[a-z]+-[\\w-]+(?=[:\\)])/g);

        if (features) {
          features = features.map(s => s.slice(1));

          for (let feature of features) {
            incrementByKey(ret.media, feature);
            ret.total.media++;
          }
        }
      }
    });

    ret.total.total = sumObject(ret.total);

    for (let type in ret) {
      ret[type] = sortObject(ret[type]);
    }

    return ret;
  }


  let ast = JSON.parse(css);
  let stats = compute(ast).total;
  return Object.entries(stats).flatMap(([prop, freq]) => {
    return Array(freq).fill(prop);
  });
}
catch (e) {
  return [];
}
'''
OPTIONS (library="gs://httparchive/lib/css-utils.js");

SELECT
  *
FROM (
  SELECT
    client,
    prop,
    COUNT(DISTINCT page) AS pages,
    COUNT(0) AS freq,
  FROM
    UNNEST(getPrefixStats(css)) AS prop
  WHERE
    date = '2020-08-01'
  GROUP BY
    client,
    prop)
WHERE
  pages >= 1000
ORDER BY
  pct DESC
```

sql/2020/01_CSS/vendor_prefix_summary.sql
LeaVerou: Why not just divide by 2?

rviscomi: It may not be clear to someone reading these queries, but omitting the magic total property from the total calculation would help to explain the intention.

LeaVerou:
@rviscomi I just pushed repetition.sql, could you take a look before I run it on `parsed_css`? It seems to return reasonable results on parsed_css_1k

sql/2020/01_CSS/all_features.sql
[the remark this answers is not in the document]
rviscomi: Ping

sql/2020/01_CSS/vendor_prefix_summary.sql
rviscomi: Could you update the PR with the working query?

rviscomi (changes requested):
`repetition.sql` LGTM, nice work

Just a couple of outstanding comments, then this is good to go

sql/2020/01_CSS/all_features.sql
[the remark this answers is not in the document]
LeaVerou: Oh nice! Done. Should I re-run the query?

LeaVerou:
Done!

sql/2020/01_CSS/all_features.sql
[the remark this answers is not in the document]
rviscomi: The results should be identical, so no need to rerun. This is mostly just to make the query less expensive/complex for anyone else who wanted to run it.

sql/2020/01_CSS/vendor_prefix_summary.sql
rviscomi: PTAL this is the last blocker

sql/2020/01_CSS/vendor_prefix_summary.sql
```diff
@@ -0,0 +1,117 @@
...
+    COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client) AS pct
```
rviscomi, line 105:
```suggestion
    SUM(COUNT(IF(prop = 'total', NULL, 0))) OVER (PARTITION BY client) AS total,
    COUNT(IF(prop = 'total', NULL, 0)) / SUM(COUNT(IF(prop = 'total', NULL, 0))) OVER (PARTITION BY client) AS pct
```

rviscomi (changes requested):

rviscomi (approved):

PR #1592, opened 2020-11-29
Accessibility 2020 queries: Find the pages with the longest alts

obto:
Find the pages with the longest alt tags so we can verify the crawler isn't broken and see them for ourselves

Progress on #907

### DIFF
--- a/sql/2020/08_Accessibility/README.md
+++ b/sql/2020/08_Accessibility/README.md
@@ -1,5 +1,5 @@
 ## Files and their purpose
 
-### pages_with_longest_alts.sql
+### [pages_with_longest_alts.sql](./pages_with_longest_alts.sql)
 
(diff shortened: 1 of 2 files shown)


rviscomi:
WDYT about adding a note to the README file in this SQL directory with more info on what this file is intended to be used for, eg debugging rather than expert interpretation.

obto:
Great idea. Were you thinking something like the edit I just made?

sql/2020/08_Accessibility/README.md
```diff
@@ -0,0 +1,5 @@
+## Files and their purpose
+
+### pages_with_longest_alts.sql
```
rviscomi, line 3: clickable
```suggestion
### [pages_with_longest_alts.sql](./pages_with_longest_alts.sql)
```

rviscomi (approved):

PR #1594, opened 2020-11-29
Move Methodology code block to Rainbow styles

bazzadp:
to avoid duplication of code.

Follow on from #1574

### DIFF
--- a/src/static/css/page.css
+++ b/src/static/css/page.css
@@ -690,23 +690,3 @@ pre {
 .code-block code {
   width: 100%;
 }
-
-.code-comment {
-  color: #006400;
-}
-
-.code-keyword {
-  color: #4a3244;
-}
-
-.code-function {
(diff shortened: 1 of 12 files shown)


rviscomi (approved):
What's the process for generating Rainbow-specific markup? Should it be documented in a wiki?

bazzadp:
> What's the process for generating Rainbow-specific markup? Should it be documented in a wiki?

Was the last process? 😉

Was thinking about writing a specific function for Methodology page, but after I discovered how easy it was I figured that was overkill. What about commenting the code like I just added in the last commit?: https://github.com/HTTPArchive/almanac.httparchive.org/pull/1594/commits/3fc4d42f2aaa9e656f204aa3c47d423c98a46b09

rviscomi:
SGTM. The wasn't a process last time, it was manual!

PR #1599, opened 2020-11-30
2020 HTTP2 chapter

dotjs:
Closes #921

PR #1602, opened 2020-11-30
Add no-wrap utility class for better table layouts

bazzadp:
Requested by HTTP/2 chapter in https://github.com/HTTPArchive/almanac.httparchive.org/pull/1599/files#r532518595 and probably useful for other chapters too.

### DIFF
--- a/src/static/css/page.css
+++ b/src/static/css/page.css
@@ -31,8 +31,16 @@
   margin: 0 -1rem;
 }
 
+.width-20 {
+  width: 20%;
+}
+
 .width-45 {
   width: 45%;
+}
+
+.no-wrap {
+  white-space: nowrap;
 }
 
 .index {
--- a/src/static/css/page.css
+++ b/src/static/css/page.css
@@ -31,10 +31,6 @@
   margin: 0 -1rem;
 }
 
-.width-20 {
-  width: 20%;
-}
-
 .width-45 {
   width: 45%;
 }


src/static/css/page.css
```diff
@@ -31,10 +31,18 @@
   margin: 0 -1rem;
 }
 
+.width-20 {
+  width: 20%;
+}
```
rviscomi, line 36: Is there a mobile media query that overrides table widths to 100%? I can't imagine 20% looking legible on a mobile screen.

bazzadp: This is for an individual column not the whole table. If you have a very wide column it greedily forces the first column to be too small.

So like this:

image

Instead of this:

image

But to be honest I'm going to use the `no-wrap` utility class added in the PR anyway so can remove the `width-20` if you want. Was more adding it in case I needed it in future as originally thought about that. Or I could change it to min-width?

Anyway this is what the table looks like with `no-wrap` on that first column:

image

rviscomi: +1 `no-wrap` seems to do exactly what we want in this case

bazzadp: OK I'll just remove the `width-20` one then as not needed.

bazzadp: And it's gone!!

rviscomi (approved):

PR #1606, opened 2020-12-01
Mobile Web: Add two new metrics

obto:
Progress on #910 
Adding two new metrics that were requested. The results are already in the spreadsheet

### DIFF
--- a/sql/2020/12_Mobile_Web/tap_targets.sql
+++ b/sql/2020/12_Mobile_Web/tap_targets.sql
@@ -1,12 +1,12 @@
 #standardSQL
 # % mobile sites with correctly sized tap targets (note: the score is not binary)
 SELECT
-  COUNTIF(color_contrast_score IS NOT NULL) AS total_applicable,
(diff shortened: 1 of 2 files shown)


sql/2020/12_Mobile_Web/tap_targets.sql
```diff
@@ -0,0 +1,12 @@
+#standardSQL
+# % mobile sites with correctly sized tap targets (note: the score is not binary)
+SELECT
+  COUNTIF(color_contrast_score IS NOT NULL) AS total_applicable,
```
rviscomi, line 4:
```suggestion
  COUNTIF(tap_targets_score IS NOT NULL) AS total_applicable,
```

rviscomi (changes requested):

sql/2020/12_Mobile_Web/tap_targets.sql
obto: Fixed haha

rviscomi (approved):

Issue #900, opened 2020-06-27
Media 2020


obto:
# Part I Chapter 5: Media

## Content team
| Authors | Reviewers | Analysts | Draft | Queries | Results |
| ------- | --------- | -------- | ----- | ------- | ------- |
| @tpiros @bseymour | @nhoizey @colinbendell @dougsillars | @smatei | [Doc][~draft-doc] | [*.sql][~sql-dir] | [Sheet][~results-sheet] |

**Content team lead:** @tpiros

Welcome chapter contributors! You'll be using this issue throughout the [chapter lifecycle](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle) to coordinate on the content planning, analysis, and writing stages.

The **content team** is made up of the following contributors:

- [Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)
- [Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)
- [Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)

**New contributors:** If you're interested in joining the content team for this chapter, just leave a comment below and the content team lead will loop you in.

_Note: To ensure that you get notifications when tagged, you must be "[watching](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#notifications)" this repository._

### [**0. Form the content team**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#0-create-content-team)
+ [x] **Jul 6th**: Project owners have selected an author to be the content team lead
+ [x] **Jul 13th**: The content team has at least one author, reviewer, and analyst (minimally viable team formed)

### [**1. Plan content**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
+ [x] **Jul 20th:** The content team has completed the chapter outline in the [draft doc][~draft-doc]
+ [x] **Jul 27th**: Analysts have triaged the feasibility of all proposed metrics

### [**2. Gather data**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
[...]

PR #1612, opened 2020-12-02
Media 2020 queries

smatei:
Minor changes for existing sqls and extra sqls that were not covered by the first one. #900

### DIFF
--- a/sql/2020/05_Media/video_source_types.sql
+++ b/sql/2020/05_Media/video_source_types.sql
@@ -3,7 +3,7 @@
 
 SELECT
   client,
-  lower(video_type) as video_type,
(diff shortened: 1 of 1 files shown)


sql/2020/05_Media/video_source_types.sql
```diff
@@ -3,13 +3,13 @@
 
 SELECT
   client,
-  video_type,
+  lower(video_type) as video_type,
```
rviscomi, line 6: CAPS for all keywords and functions
```suggestion
  LOWER(video_type) AS video_type,
```

rviscomi (changes requested):

Issue #1432, opened 2020-11-06
Edit 2020 Chapters


bazzadp:
Part I. Page Content
 - [x]  - [Chapter 1: CSS](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content/en/2020/css.md) - @bazzadp #1846
 - [x]  - [Chapter 2: JavaScript](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content/en/2020/javascript.md) - @rviscomi #1737
 - [x]  - [Chapter 3: Markup](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content/en/2020/markup.md) - @rviscomi #1435 and #1604 
 - [x]  - [Chapter 4: Fonts](https://github.com/HTTPArchive/almanac.httparchive.org/tree/mainsrc/content/en/2020/fonts.md) - @bazzadp #1816 and #1868
 - [x]  - [Chapter 5: Media](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content/en/2020/media.md) - @bazzadp #1882
 - [x]  - [Chapter 6: Third Parties](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content/en/2020/third-parties.md) - @bazzadp - #1877 

Part II. User Experience
 - [x]  - [Chapter 7: SEO](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content/en/2020/seo.md) - @rviscomi 
 - [x]  - [Chapter 8: Accessibility](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content/en/2020/accessibility.md) - @bazzadp - #1752 
 - [x]  - [Chapter 9: Performance](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content/en/2020/performance) - @bazzadp - #1622 
 - [x]  - [Chapter 10: Privacy](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content/en/2020/privacy.md) - @bazzadp - #1672 and #1935
 - [x]  - [Chapter 11: Security](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content/en/2020/security.md) - @bazzadp - #1927
[...]

PR #1614, opened 2020-12-02
Markup 2020 TODOs

j9t:
Progress on #899 #1432


src/content/en/2020/markup.md
```diff
@@ -121,8 +120,6 @@ Here are the 10 most popular (normalized) languages in our sample. At first we c
 <figcaption>{{ figure_link(caption="Top 10 <code>lang</code> attribute values.", sheets_gid="2047285366", sql_file="pages_almanac_by_device_and_html_lang.sql") }}</figcaption>
 </figure>
 
-{# TODO(authors): Add an interpretation of the lang results. #}
```
j9t, line 124: Explanation for removal: I’d argue there’s little to interpret, or to defer to methodology as anything we see here may have been introduced on that end.

src/content/en/2020/markup.md
```diff
@@ -352,13 +345,11 @@ Standard elements are those that are or were part of the HTML specification. Whi
 <figcaption>{{ figure_link(caption="Low probabilities of finding a given element in pages of the sample.", sheets_gid="184700688", sql_file="pages_element_count_by_device_and_element_type_present.sql") }}</figcaption>
 </figure>
 
-{# TODO(authors): Interpet results. #}
```
j9t, line 355: Explanation: Suggesting to skip this, or ask whether @catalinred or @iandevlin like to draft something.

src/content/en/2020/markup.md
```diff
@@ -352,13 +345,11 @@ Standard elements are those that are or were part of the HTML specification. Whi
...
-
 ### Custom elements
 
 The 2019 edition of the Web Almanac handled [custom elements](../2019/markup#custom-elements) by discussing several non-standard elements. This year, we found it valuable to have a closer look at custom elements. How did we determine these? Roughly by looking at [their definition](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-core-concepts), notably their use of a hyphen. Let's focus on the top elements, in this case elements used on ≥1% of all URLs in the sample:
 
-{# TODO(authors, analysts): Clarify occurrences and percentages _of what_. Pages? Elements? And for desktop or mobile? #}
+{# TODO(authors, analysts): Clarify occurrences and percentages _of what_. Pages? Elements? #}
```
j9t, line 352: Background (for removing part of the TODO): We described in the intro how we usually look at the mobile set. (Open to calling this out again if we feel it needs a reminder in between.)

src/content/en/2020/markup.md
```diff
@@ -80,8 +80,7 @@ A page's document size refers to the amount of HTML bytes transferred over the n
 * The largest document by far weighs 64.16 _MB_, almost deserving its own analysis and chapter in the Web Almanac.
 
 {# TODO(analysts): Should 25,237 bytes be divided by 1000 or 1024 to convert to KB? 1000 seems to be used here but most chapters use 1024. Are the stats above also off? #}
```
j9t, line 82: @Tiggerito?

src/content/en/2020/markup.md
```diff
@@ -183,13 +178,11 @@ This seems to confirm the theory that the total number of `noscript` elements in
 
 What `type` attribute values are used with `script` elements?
 
-{# TODO(authors, analysts): Should this be a figure? #}
```
j9t, line 186: Background: Let’s work with a list here, I believe it is sufficient.

src/content/en/2020/markup.md
Tiggerito: I agree, 1024 should be used. I'm not confident on the above stats. I'd like to use an example we can prove. I think the problem was the bytesHtmlDoc which I think it more true to what we would call the document got cut off at 16,777,215. 

I used Screaming Frog to crawl all the ones in the list, and double checked the big ones in Chrome. Drum roll...

https://www.linkshops.com/ contains a js bundle that is 28.1MB.

https://www.boonterm.com/web/index1.php has 22.8MB of html (embedded images) and references a 34.7MB mp4 file.

https://www.aci.edu.sg/ has 33.7MB of html and took me over a minute to load. Also embedded images.

src/content/en/2020/markup.md
rviscomi: What conclusions would you like to see readers draw from this figure? If there's nothing to say about it, do we need it at all?

src/content/en/2020/markup.md
rviscomi: Ditto

rviscomi (changes requested):

src/content/en/2020/markup.md
j9t: This may raise a very good point here! If the data set suggests that this data may not be representative for the wider Web then we should probably take this out, because what is there to see?

src/content/en/2020/markup.md
j9t: Again, I’d look at @catalinred or @iandevlin whether they like to expand on this.

Other than that, as opposed to the lang section here I don’t agree. Unless you’re coming from an Almanac convention that requires every data point to be interpreted, I wouldn’t be convinced that data can’t stand by itself. I’d suggest that not only are our readers capable of drawing conclusions, but that this can even be refreshing from an editorial perspective, if not to suggest—“wait a minute; what’s this, what does this mean?”

If I misunderstand you, please let me know, otherwise I’d appreciate a bit of leeway here for us to decide on what also _not_ to interpret.

src/content/en/2020/markup.md
bazzadp: The dat set is crawled from a US crawler with `en-US` set as the preferred language. Not sure how common it is to redirect to a home page based on that locale?

Additionally, the set of URLs is based on the CrUX data based off of Chrome data - is that swayed towards western users (therefore explaining the low numbers of Asia sites other than Japanese)?

Therefore I think it raises a very good question as to whether we can rely on this data? At the very least we should add a caveat explaining these influences.

src/content/en/2020/markup.md
bazzadp: My opinion is we should be selective to what we put in the chapter. We have a lot of data - some of it will be interesting and some not. I think we need to pick and choose what to include to keep the chapter interesting.

The point of the almanac is "The Web Almanac is a comprehensive report on the state of the web, backed by real data and trusted web experts" rather than "a list of stats, with experts interpreting them".

Saying that, I think these particular stats ARE interesting. Why are they not used? Are they old elements which are not useful? Have they been replaced by better alternatives? Or are they new elements that haven't taken off yet?

Looking at MDN two if them are obsoleted ([`dir`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dir) and [`basefont`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/basefont) and one ([`rb`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/rp)) is Ruby specific - to me there's a question if that should really be a standard HTML element? Ultimately, I'd never heard of these particular elements and had to look them up to figure this information out, so think a summary of this explaining this would be useful here to save other readers doing the same.

However, I do think this is the authors work, so if you still feel this is not necessary explanation and the stats stand on their own, then I can accept this. But think it's right for @rviscomi to at least ask these questions in case you hadn't considered that interpretation or were making assumptions as to what readers might know because you are HTML experts (and other readers might not be).

src/content/en/2020/markup.md
```diff
@@ -76,17 +76,15 @@ A page's document size refers to the amount of HTML bytes transferred over the n
 
 {# TODO(authors, analysts): Revisit the "largest document" stat and interpretation. #}
 * 1,110 documents are empty (0 bytes).
-* The average document size is 50.35 KB ([in most cases compressed](https://w3techs.com/technologies/details/ce-gzipcompression)).
-* The largest document by far weighs 64.16 _MB_, almost deserving its own analysis and chapter in the Web Almanac.
+* The average document size is 49.17 KB ([in most cases compressed](https://w3techs.com/technologies/details/ce-gzipcompression)).
+* The largest document by far weighs 62.65 _MB_, almost deserving its own analysis and chapter in the Web Almanac.
 
-{# TODO(analysts): Should 25,237 bytes be divided by 1000 or 1024 to convert to KB? 1000 seems to be used here but most chapters use 1024. Are the stats above also off? #}
```
j9t, line 82: Note: I went through this section and re-calculated the values.

@Tiggerito, does this look right?

What I didn’t dive into now are the big documents. The section above seems to work, not sure what to add and what to do there that would benefit the rest of the section.

src/content/en/2020/markup.md
catalinred: I remember the point of this section was to help the readers better understand the data set they're looking at: e.g. where do most of the pages come from and what is the main language used/detected on them.

In my opinion, I think a pie chart with `en`, `en-us`, `ja`, `es` etc would do it here. I'd also add in the chart the `22.36%`  of all documents that specify no `lang` attribute.

bazzadp: I think that's an understandable rational for inclusion, but don't think that's clear from the text and so should make it clear. How about adding some thing like "It should be noted that the HTTP Archive crawls from US data centres with English language settings, so looking at the language pages are written in will be skewed towards English. Nevertheless we present the `lang` attributes seen to give some context to the sites analysed." Or something like that.

I'd discourage pie charts (very difficult to measure the size of relative slices) and think a standard bar graph would better here:

image

catalinred: I have nothing against using a bar chart here, especially that both desktop and mobile stats can be shown in this way.

rviscomi: @bazzadp could you add that as a suggestion for @j9t to review/approve?

bazzadp: Does look like I can do this but I wouild suggest something like this:

```

Here are the 10 most popular (normalized on case) languages in our sample. It is important to note that the HTTP Archive crawls from US data centres with English language settings, so looking at the language pages are written in will be skewed towards English. Nevertheless we present the lang attributes seen to give some context to the sites analysed.

{{ figure_markup(
  image="top-html-lang.png",
  alt="The top HTML lang attritbues.",
  caption="The top HTML `lang` attritutes.",
  description="Bar chart showing the top 10 `lang` attributes using in our crawl with 22.82% of desktop and 22.36% of mobile not setting this, `en` being used on 20.09% and 18.08% respectively, `ja` on 15.17% and 13.27%, `es` on 4.86% and 4.09% , `pt-br` on 2.65% and 2.84%, `ru` on 2.21% 2.53%, `en-gb` on 2.35% and 2.19%, `de` on 1.50% and 1.92%, and finally `fr` being used on 1.55% and 1.43% respectively",
  sheets_gid="2047285366",
  chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vQPKzFb574UnGTcfw5mcD1qR7RYHyGjQTc2hiMuYix0QoTH1DPe54Q2JucXL8bfZ6kjRoAfhk3ckudc/pubchart?oid=1873310240&format=interactive",
  width=600,
  height=371,
  sql_file="pages_almanac_by_device_and_html_lang.sql"
  )
}}
```

j9t: Updated the page (with just minor edits).

src/content/en/2020/markup.md
rviscomi: Suggestion to skip SGTM. Could you update the md to remove the content?

src/content/en/2020/markup.md
rviscomi: Suggestion to remove SGTM.

src/content/en/2020/markup.md
rviscomi: LGTM. Ping @Tiggerito for a quick check if possible.

rviscomi (approved):
@j9t this is looking great and ready for the `unedited: true` label to be removed. Can you do the honors?

src/content/en/2020/markup.md
j9t: Remove? Or keep the update? (We had updated this section.)

src/content/en/2020/markup.md
j9t: I actually felt like I want to add info here. Added a paragraph. What do you think?

j9t:
Included another update.

Checked the other TODOs but these may be okay right now. @catalinred, @iandevlin, @tiggerito, do you have a chance to review those maybe later?

Removed the “unedited” flag with this PR.

Any other thoughts, @rviscomi, @bazzadp?

src/content/en/2020/markup.md
rviscomi: I was referring to the removal of the table. Either way, LGTM.

src/content/en/2020/markup.md
```diff
@@ -351,13 +338,13 @@ Standard elements are those that are or were part of the HTML specification. Whi
 <figcaption>{{ figure_link(caption="Low probabilities of finding a given element in pages of the sample.", sheets_gid="184700688", sql_file="pages_element_count_by_device_and_element_type_present.sql") }}</figcaption>
 </figure>
 
-{# TODO(authors): Interpet results. #}
+We’re including these elements to give an idea what elements may have gone out of favor. But while `dir` and `basefont` were last specified in XHTML 1.0 (2000), the rare use of `rp`, which has been mentioned [as early as 1998](https://www.w3.org/TR/1998/WD-ruby-19981221/#a2-4) but which is also [still part of HTML](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-rp-element), may just suggest that Ruby markup is not very popular.
```
rviscomi, line 341: 🙊 

```suggestion
We're including these elements to give an idea what elements may have gone out of favor. But while `dir` and `basefont` were last specified in XHTML 1.0 (2000), the rare use of `rp`, which has been mentioned [as early as 1998](https://www.w3.org/TR/1998/WD-ruby-19981221/#a2-4) but which is also [still part of HTML](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-rp-element), may just suggest that Ruby markup is not very popular.
```

Tiggerito (approved):
I still think it should be 61.19MB, but what's a few bytes between friends.

catalinred (approved):

j9t:
> I still think it should be 61.19MB, but what's a few bytes between friends.

😅 Maybe I missed that. Updated.

rviscomi (approved):
I think this is good to go! 🚀 
Thanks @j9t and everyone for your help!

PR #1621, opened 2020-12-03
Security SQL 2020: fixes

tomvangoethem:
The `csp_script_source_list_keywords.sql` query might produce some incorrect results if the keywords would appear in other headers as well. The alternative would be to use the JSON representation from the `httparchive.almanac.requests` table, which would require more bytes.

Progress on #906


tomvangoethem:
Alternative query for `csp_script_source_list_keywords.sql` using the JSON representation of the headers:

```
#standardSQL
# CSP: usage of default/script-src, and within the directive usage of strict-dynamic, nonce values, unsafe-inline and unsafe-eval
CREATE TEMPORARY FUNCTION getHeader(headers STRING, headername STRING)
  RETURNS STRING DETERMINISTIC
  LANGUAGE js AS '''
  const parsed_headers = JSON.parse(headers);
  const matching_headers = parsed_headers.filter(h => h.name.toLowerCase() == headername.toLowerCase());
  if (matching_headers.length > 0) {
    return matching_headers[0].value;
  }
  return null;
''';
SELECT
  client,
  total_pages,
  freq_csp,
  freq_default_script_src,
  SAFE_DIVIDE(freq_default_script_src, freq_csp) AS pct_default_script_src_over_csp,
  freq_strict_dynamic,
  SAFE_DIVIDE(freq_strict_dynamic, freq_csp) AS pct_strict_dynamic_over_csp,
  SAFE_DIVIDE(freq_strict_dynamic, freq_default_script_src) AS pct_strict_dynamic_over_csp_with_src,
  freq_nonce,
  SAFE_DIVIDE(freq_nonce, freq_csp) AS pct_nonce_over_csp,
  SAFE_DIVIDE(freq_nonce, freq_default_script_src) AS pct_nonce_over_csp_with_src,
  freq_unsafe_inline,
  SAFE_DIVIDE(freq_unsafe_inline, freq_csp) AS pct_unsafe_inline_over_csp,
  SAFE_DIVIDE(freq_unsafe_inline, freq_default_script_src) AS pct_unsafe_inline_over_csp_with_src,
  freq_unsafe_eval,
  SAFE_DIVIDE(freq_unsafe_eval, freq_csp) AS pct_unsafe_eval_over_csp,
  SAFE_DIVIDE(freq_unsafe_eval, freq_default_script_src) AS pct_unsafe_eval_over_csp_with_src,
FROM (
  SELECT
    client,
    COUNT(0) AS total_pages,
    COUNTIF(csp_header IS NOT NULL) AS freq_csp,
    COUNTIF(REGEXP_CONTAINS(csp_header, '(?i)(default|script)-src')) AS freq_default_script_src,
    COUNTIF(REGEXP_CONTAINS(csp_header, '(?i)strict-dynamic')) AS freq_strict_dynamic,
    COUNTIF(REGEXP_CONTAINS(csp_header, '(?i)nonce-')) AS freq_nonce,
    COUNTIF(REGEXP_CONTAINS(csp_header, '(?i)unsafe-inline')) AS freq_unsafe_inline,
    COUNTIF(REGEXP_CONTAINS(csp_header, '(?i)unsafe-eval')) AS freq_unsafe_eval
  FROM (
    SELECT
      client,
      getHeader(JSON_EXTRACT(payload, '$.response.headers'), 'Content-Security-Policy') AS csp_header
    FROM
    WHERE
      date = "2020-08-01" AND
      firstHtml
  )
  GROUP BY
    client
)
ORDER BY
  client
```

bazzadp:
So which query do you want to use @tomayac ?

tomvangoethem:
Personally I would opt for the one in the comment based on the `requests` table, as it will be correct. I put the other one in the PR as we've used the `summary_requests` table more frequently in the later queries (to reduce queried bytes).

tomayac:
> So which query do you want to use @tomayac ?

I think you tagged me in mistakenly.

bazzadp:
> > So which query do you want to use @tomayac ?
> 
> I think you tagged me in mistakenly.

Oops sorry @tomayac !

bazzadp:
> Personally I would opt for the one in the comment based on the `requests` table, as it will be correct. I put the other one in the PR as we've used the `summary_requests` table more frequently in the later queries (to reduce queried bytes).

OK can you update PR to that then?

tomvangoethem:
> OK can you update PR to that then?

Done!

rviscomi (approved):

PR #1625, opened 2020-12-04
SEO 2020 edits

rviscomi:
Progress on #908 #1432 

Editors' notes:

- global replace of smart quotes to straight quotes
- `code` format for things like `robots.txt` and HTTP headers
- Oxford comma
- replace "homepage" to "home page"
- right alignment of numeric data in tables
- replace passive voice with active voice
- various TODOs and fixes

PR #1628, opened 2020-12-04
Featured quote and 3 stats added

aleyda:


rviscomi:
Thanks @aleyda. Let's combine this with the open PR to edit the chapter: #1625.

I also wanted to give you some context how the stats will appear:

image

For example, on the home page, one of the Markup chapter's featured stats is `85.73% Percentage of pages that use the "living" HTML doctype`. The number and the label are treated separately in the [metadata](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/content/en/2020/markup.md#:~:text=featured_stat_1) as `featured_stat_1` and `featured_stat_label_1`.

Could you take another turn at formatting the stats so that the number is separated from the label? That may also mean choosing between showing the mobile or desktop number. When you're done, please apply these changes to the PR as suggested edits:

image

aleyda:
Thanks for the explanation @rviscomi ! I did the changes [as requested here. ](https://github.com/HTTPArchive/almanac.httparchive.org/pull/1625#discussion_r536344658). I hope they're better now :) Thanks

PR #1634, opened 2020-12-05
no-wrap class added in 2019.css and used at featured chapter stats in index.html

bharatagsrwal:
no-wrap utility class added to `2019.css`

### DIFF
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -1470,7 +1470,7 @@ p.copyright a {
   }
 }
 
-.no-wrap{
+.no-wrap {
   white-space: nowrap;
 }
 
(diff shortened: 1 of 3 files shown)


bazzadp:
Can we remove it from page.css? No point in having it in both and 2019.css is loaded by everything.

bharatagsrwal:
> page.css?

I think we can remove, but first we have to check the dependency of `no-wrap` from `page.css`

bharatagsrwal:
> Can we remove it from page.css? No point in having it in both and 2019.css is loaded by everything.

We can remove it, as I can see `2019.css` is used on every page, and we have `no-wrap` class in `2019.css` as well.

rviscomi (approved):

PR #1642, opened 2020-12-05
Better colour contrast for social media icons

### LINKED ISSUE #1636: "Twitter" and "Github" logo should be little bright on the footer
**URL**
1. Go to almanac.httparchive.org
2. Scroll down to bottom

**Describe the issue**
"Twitter" and "GitHub" logo should be liitle bright on the footer

**Expected behavior**

"Twitter" and "GitHub" logo should be little bright on the footer. logo color and footer background color almost same.
**Screenshots**
image

bazzadp:
Fixes #1636

### DIFF
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -898,7 +898,7 @@ p.copyright a {
 }
 
 .social-media a {
-  color: #8c99a3;
+  color: currentColor;
 }
 
 @media (max-width: 900px) and (min-width: 601px) {


src/static/css/2019.css
```diff
@@ -898,7 +898,7 @@ p.copyright a {
 }
 
 .social-media a {
-  color: #8c99a3;
+  color: currentColor;
```
rviscomi, line 901: Would removing this style entirely have the same effect?

bazzadp: No, tried that as then it gets the link colour (purple) rather than the text colour (off white).

This is what ha-logo has too (presumably for the same reason):

https://github.com/HTTPArchive/almanac.httparchive.org/blob/2ea040c8dbea7d01cee640be13fe21c98e3a0011/src/static/css/2019.css#L828-L834

rviscomi (approved):

PR #1644, opened 2020-12-06
Security SQL 2020: add 2019 results for security header adoption

tomvangoethem:
Progress on #906


tomvangoethem:
Note: query would cost 12TB 🤯

Issue #1630, opened 2020-12-04
Launch the 2020 Web Almanac 🚀


rviscomi:
#### Content
- [x] Home page content updated - PR #1651 
- [x] Foreword added #1406
- [x] All viable chapters added #989 
- [x] Update contributors #1582 #1654 

#### QA
- [x] Staging version deployed: https://20201209t155443-dot-webalmanac.uk.r.appspot.com
- [x] Staging site manually tested for basic functionality
- [x] Automated tests of the staging site pass (LH audits)
- [x] Spell check run on QA site and final spelling mistakes fixed
- [x] Broken links crawl check completed on QA site
- [x] [Twitter](https://cards-dev.twitter.com/validator) and structured data test
- [x] Google Search Console reviewed for errors
- [x] Google Analytics checked to ensure collecting data and errors (e.g. 404s) checked.

#### Deployment
- [x] Full site release to version 2.0.0 at 10:00am UTC-5 (NYC time) on Wednesday, December 9
- [x] Sitemap submission to Google Search Console to pick up launched chapters quicker
- [x] Final QA
- [x] Formal [public announcement](https://twitter.com/rick_viscomi/status/1336718249388625921) by @rviscomi on social channels at 12:00pm UTC-5 (NYC time)
- [x] Monitor [analytics](https://datastudio.google.com/reporting/1c89f8c3-7e11-4376-8815-39966e13180e) and server logs

bazzadp:
Submitted sitemap to Google so closing this.

PR #1651, opened 2020-12-07
Update the post-launch 2020 home page

rviscomi:
Progress on #1630 

This PR does not update non-English languages (help with that would be appreciated). Note that "20 chapters" is baked into the year-agnostic base template, which means that the `2020/index.html` template needs to overwrite the `mission` block.


bazzadp:
>  Note that "20 chapters" is baked into the year-agnostic base template, which means that the `2020/index.html` template needs to overwrite the `mission` block.

Well that's a mistake on my part! We shouldn't have 2019 data in the base templates.

Saying that, if you're going to reuse the exact same wording other than number of chapters then let's generalise this! This is done in https://github.com/HTTPArchive/almanac.httparchive.org/pull/1651/commits/ebe977dc37b70d830115df29e5ded5a622fc7904

Also solves the translation issue.

bazzadp (approved):

Issue #878, opened 2020-06-17
Jamstack 2020


rviscomi:
# Part III Chapter 17: Jamstack

## Content team
| Authors | Reviewers | Analysts | Draft | Queries | Results |
| ------- | --------- | -------- | ----- | ------- | ------- |
| @AhmadAwais  |  @phacks  @maedahbatool| @denar90 @remotesynth | [Doc][~draft-doc] | [*.sql][~sql-dir] | [Sheet][~results-sheet] |

**Content team lead:** @AhmadAwais 

**Published at:** https://almanac.httparchive.org/en/2020/jamstack

Welcome chapter contributors! You'll be using this issue throughout the [chapter lifecycle](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle) to coordinate on the content planning, analysis, and writing stages.

The **content team** is made up of the following contributors:

- [Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)
- [Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)
- [Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)

**New contributors:** If you're interested in joining the content team for this chapter, just leave a comment below, and the content team lead will loop you in.

_Note: To ensure that you get notifications when tagged, you must be "[watching](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#notifications)" this repository._

### [**0. Form the content team**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#0-create-content-team)
+ [x] **Jul 6th**: Project owners have selected an author to be the content team lead
+ [x] **Jul 13th**: The content team has at least one author, reviewer, and analyst (a minimally viable team formed)

### [**1. Plan content**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
+ [x] **Jul 20th:** The content team has completed the chapter outline in the [draft doc][~draft-doc]
+ [x] **Jul 27th**: Analysts have triaged the feasibility of all proposed metrics

### [**2. Gather data**](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Chapter-Lifecycle#1-plan-content)
[...]

PR #1665, opened 2020-12-08
Copyedit of Jamstack chapter

bazzadp:
Makes progress on #1432 #878 

@ahmadawais I've copy edited the Jamstack chapter. One key thing to note is there looks to have been a flaw in the adoption analysis as the percentage of sites in the crawl was taken to be the percentage of Jamstack. So Next.JS was 0.49% of our 6.5 million desktop sites, but this is actually 72.5% of Jamstack sites. So Next.JS's usage is even higher than originally calculated. I've therefore made quite a few edits to this section.

The rest of the edits were fairly standard, though I did reorder some peices.

I've staged a link here: https://20201208t012527-dot-webalmanac.uk.r.appspot.com/en/2020/jamstack

If you're happy with this, then we're all done and good to launch without the "Unedited" label on Wednesday.


ahmadawais (approved):
@bazzadp this is excellent. Looks good to me! ✔︎

I just added an extra line to split a long para into two at the very end. Everything else is good to go.

PR #1671, opened 2020-12-08
Stop doing author check for single chapter generation

bazzadp:
FYI @bharatagsrwal we also generate single chapters sometimes but this is listing basically all contributors not in that as being missing.

Moving into if statement to prevent this.


bharatagsrwal:
Cool @bazzadp, I got it. 
It also has some more problems I am fixing those like those who are missing in the `config/${year}.json` file, is a somewhere little buggy, so I am fixing those and creating a pull request as soon as possible

PR #1678, opened 2020-12-08
Enable 2020 ebook

bazzadp:
We may not have a complete Almanac for launch but still think we should generate the ebook. And will be interesting to see how many pages it will be this year!

358 with chapters merged so far (compared to 405 last year).


bazzadp:
This is a lot bigger than it looks btw. Just had to move the new `swatch` macro form `base_chapter.html` to `base.html`

And similarly also had to move the `unedited` text block out of base_chapter into base for each languages (last year we launched the ebook after everything was edited so never noticed we needed this in base).

No real changes other than that and configuring it.

rviscomi:
Because the website is evergreen but the PDF is not, my inclination would be to wait to publish it until we either have all of the content or everything is edited. How easy would it be to add some kind of version/last-modified to the ebook to make clear how old someone's copy is?

rviscomi (approved):

rviscomi:
It just occurred to me you said 358 pages -- with 9 chapters still to go?! 🤯

bazzadp:
> It just occurred to me you said 358 pages -- with 9 chapters still to go?! 🤯

That's part of the reason I wan't to run it to see! 😁 With chapters like CSS, and Security still be to included (but especially CSS), it's gonna be a monster!

rviscomi:
Should `web_almanac_2020_en.pdf` have the alternating left/right margins? That makes sense for a printed book to account for the binding but maybe less for a digital copy.

PR #1684, opened 2020-12-09
Mobile web chapter: Figures added to markdown

obto:
Figures added to the markdown document

#910


obto:
Have one last figure to add, just requested by Shubhie

obto:
Done

spanicker (approved):
hmm I think you still hyphen when linking to sections and use the section names.
eg. data-sources vs. datasources
Otherwise LGTM.

obto:
Good catch. Must have missed that change before I merged my work with the earlier PR

rviscomi (approved):

PR #1706, opened 2020-12-09
Support redirecting uppercase chapter urls to lowercase

bazzadp:
While reviewing our 404 report in Google Analytics as part of #1630 I noticed one call in the last week for /en/2020/Jamstack with a capital J. We don't support that and 404.

This PR adds a redirect for chapters to lowercase them. Note it does not work for the static pages (ToC, Methodology...etc.) as wanted to keep this simple. URLs are case sensitive and Flask doesn't offer a simple way to do this redirect so if you want it you have to code it as far as I can see.

Our languages already handle this redirect but not the chapters.

Also adds a couple more typo chapters (`js`, `http3`).

It's a reasonably small change (most of the change was writing new test cases!), but haven't see many of these so far, and it is release day so maybe better waiting. But thought I'd put it out there for discussion.


src/config/spelling_ignore_words.txt
```diff
@@ -49,9 +49,11 @@ bendell
 benseymour
 bgsound
 bidirectionality
+bkardell
```
rviscomi, line 52: Looks like there are some unrelated editing changes in this PR. Does it need a sync with `main` or should they be added to an editing PR?

bazzadp: Weird. Merged in `main` and pushed and seems to have resolved it.

rviscomi (approved):

PR #1707, opened 2020-12-09
Fix various issues

lex111:
Mostly fixing naming issues and typos.

### DIFF
--- a/src/content/en/2020/accessibility.md
+++ b/src/content/en/2020/accessibility.md
@@ -69,7 +69,7 @@ Setting an HTML `lang` attribute allows easy translation of a page and better sc
 
 ### Images and their text alternatives
 
-Images are an essential part of the web experience. They can add an enriched context to the surrounding textual information, and not just for sighted users. In 1995, [HTML 2.0](https://www.w3.org/MarkUp/html-spec/html-spec_5.html#SEC5.10) introduced the `alt` attribute, enabling web authors to provide a text alternative for the visual information communicated in an image. A screen reader can convey its visual meaning aurally by announcing the image’s alternative text. Additionally if images are unable to load, the alternative text for a description will be displayed.
+Images are an essential part of the web experience. They can add an enriched context to the surrounding textual information, and not just for sighted users. In 1995, [HTML2.0](https://www.w3.org/MarkUp/html-spec/html-spec_5.html#SEC5.10) introduced the `alt` attribute, enabling web authors to provide a text alternative for the visual information communicated in an image. A screen reader can convey its visual meaning aurally by announcing the image’s alternative text. Additionally if images are unable to load, the alternative text for a description will be displayed.
 
 The 2020 Lighthouse audit data shows that only 54% of sites pass the [test for images with `alt` text](https://dequeuniversity.com/rules/axe/3.5/image-alt). This test looks for the presence of at least one of the `alt`, `aria-label` and `aria-labelledby` attributes on `img` elements. In most cases using the alt attribute is the best choice. Even though `alt` attributes have been around for 25 years, we also found that 21.24% of desktop images and 21.38% of mobile images are lacking alternative text. This is one of the easiest automated checks to test for using your accessibility tool of choice, and should be low hanging fruit and a relatively straightforward problem to solve.
 
@@ -236,7 +236,7 @@ There are 5 rules of ARIA that we need to understand before  making use of this 
 
 ##### The 5 Rules of ARIA (from [W3C’s Using Aria](https://www.w3.org/TR/using-aria/)):
 
-1. If you can use a native HTML element [HTML 5.1] or attribute with the semantics and behavior you require already built in, instead of repurposing an element and adding an ARIA role, state or property to make it accessible, then do so.
+1. If you can use a native HTML element [HTML5.1] or attribute with the semantics and behavior you require already built in, instead of repurposing an element and adding an ARIA role, state or property to make it accessible, then do so.
 2. Do not change native semantics, unless you really have to.
 3. All interactive ARIA controls must be usable with the keyboard.
 4. Do not use `role="presentation"` or `aria-hidden="true"` on a focusable element.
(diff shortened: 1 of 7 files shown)


src/content/en/2020/accessibility.md
```diff
@@ -236,7 +236,7 @@ There are 5 rules of ARIA that we need to understand before  making use of this
 
 ##### The 5 Rules of ARIA (from [W3C’s Using Aria](https://www.w3.org/TR/using-aria/)):
 
-1. If you can use a native HTML element [HTML 5.1] or attribute with the semantics and behavior you require already built in, instead of repurposing an element and adding an ARIA role, state or property to make it accessible, then do so.
+1. If you can use a native HTML element [HTML5.1] or attribute with the semantics and behavior you require already built in, instead of repurposing an element and adding an ARIA role, state or property to make it accessible, then do so.
```
rviscomi, line 239: Could you revert the spacing changes for "HTML"? It uses a space in the spec: https://www.w3.org/TR/html51/ 
```suggestion
1. If you can use a native HTML element [HTML 5.1] or attribute with the semantics and behavior you require already built in, instead of repurposing an element and adding an ARIA role, state or property to make it accessible, then do so.
```

src/content/en/2020/accessibility.md
```diff
@@ -69,7 +69,7 @@ Setting an HTML `lang` attribute allows easy translation of a page and better sc
 
 ### Images and their text alternatives
 
-Images are an essential part of the web experience. They can add an enriched context to the surrounding textual information, and not just for sighted users. In 1995, [HTML 2.0](https://www.w3.org/MarkUp/html-spec/html-spec_5.html#SEC5.10) introduced the `alt` attribute, enabling web authors to provide a text alternative for the visual information communicated in an image. A screen reader can convey its visual meaning aurally by announcing the image’s alternative text. Additionally if images are unable to load, the alternative text for a description will be displayed.
+Images are an essential part of the web experience. They can add an enriched context to the surrounding textual information, and not just for sighted users. In 1995, [HTML2.0](https://www.w3.org/MarkUp/html-spec/html-spec_5.html#SEC5.10) introduced the `alt` attribute, enabling web authors to provide a text alternative for the visual information communicated in an image. A screen reader can convey its visual meaning aurally by announcing the image’s alternative text. Additionally if images are unable to load, the alternative text for a description will be displayed.
```
rviscomi, line 72:
```suggestion
Images are an essential part of the web experience. They can add an enriched context to the surrounding textual information, and not just for sighted users. In 1995, [HTML 2.0](https://www.w3.org/MarkUp/html-spec/html-spec_5.html#SEC5.10) introduced the `alt` attribute, enabling web authors to provide a text alternative for the visual information communicated in an image. A screen reader can convey its visual meaning aurally by announcing the image’s alternative text. Additionally if images are unable to load, the alternative text for a description will be displayed.
```

src/content/en/2020/accessibility.md
```diff
@@ -246,7 +246,7 @@ There are 5 rules of ARIA that we need to understand before  making use of this
 
 One of the most common ways that ARIA is used is by explicitly defining the role for an element, which communicates its purpose to assistive technology.
 
-HTML 5 introduced many new native elements, all which have [implicit semantics](https://www.w3.org/TR/wai-aria-1.1/#implicit_semantics), including roles. For example the `<nav>` element has an implicit `role="navigation"` and does not need to have this role added explicitly in order to convey its purpose information to assistive technology. Currently 64.54% of desktop pages have at least one instance of an ARIA role attribute. The median site has 2 instances of the `role` attribute.
+HTML5 introduced many new native elements, all which have [implicit semantics](https://www.w3.org/TR/wai-aria-1.1/#implicit_semantics), including roles. For example the `<nav>` element has an implicit `role="navigation"` and does not need to have this role added explicitly in order to convey its purpose information to assistive technology. Currently 64.54% of desktop pages have at least one instance of an ARIA role attribute. The median site has 2 instances of the `role` attribute.
```
rviscomi, line 249:
```suggestion
HTML 5 introduced many new native elements, all which have [implicit semantics](https://www.w3.org/TR/wai-aria-1.1/#implicit_semantics), including roles. For example the `<nav>` element has an implicit `role="navigation"` and does not need to have this role added explicitly in order to convey its purpose information to assistive technology. Currently 64.54% of desktop pages have at least one instance of an ARIA role attribute. The median site has 2 instances of the `role` attribute.
```

src/content/en/2020/accessibility.md
```diff
@@ -262,7 +262,7 @@ In the vast majority of these cases, a better pattern than explicitly defining `
 
 ##### Navigation
 
-We found that 22.06% of desktop pages and 21.76% of mobile pages have at least one element with `role="navigation"`, which is a landmark role. Per the 1st rule of ARIA, rather than adding this role to an element, developers should be leveraging the HTML 5 `<nav>` element which comes with the correct semantics implicitly. It is possible that this role has been added explicitly to the `<nav>` element, which would not be an accessibility issue, though it is redundant.
+We found that 22.06% of desktop pages and 21.76% of mobile pages have at least one element with `role="navigation"`, which is a landmark role. Per the 1st rule of ARIA, rather than adding this role to an element, developers should be leveraging the HTML5 `<nav>` element which comes with the correct semantics implicitly. It is possible that this role has been added explicitly to the `<nav>` element, which would not be an accessibility issue, though it is redundant.
```
rviscomi, line 265:
```suggestion
We found that 22.06% of desktop pages and 21.76% of mobile pages have at least one element with `role="navigation"`, which is a landmark role. Per the 1st rule of ARIA, rather than adding this role to an element, developers should be leveraging the HTML 5 `<nav>` element which comes with the correct semantics implicitly. It is possible that this role has been added explicitly to the `<nav>` element, which would not be an accessibility issue, though it is redundant.
```

rviscomi (changes requested):
Just one nit, but otherwise this looks great, thanks for catching these! Let's try to get this in ASAP because we're about to cut the release candidate.

src/content/en/2020/css.md
```diff
@@ -1234,7 +1234,7 @@ Next, we looked at how complex custom property usage is. One way to assess code
   sql_file="custom_property_depth.sql"
 ) }}
 
-Examining the selectors on which custom properties are declared further confirms that most custom property usage in the wild is fairly basic. Two2 out of 3 custom property declarations are on the root element, indicating that they are used essentially as global constants. It is important to note that many popular polyfills have required them to be global in this vein, so developers using said polyfills may not have had a choice.
+Examining the selectors on which custom properties are declared further confirms that most custom property usage in the wild is fairly basic. 2 out of 3 custom property declarations are on the root element, indicating that they are used essentially as global constants. It is important to note that many popular polyfills have required them to be global in this vein, so developers using said polyfills may not have had a choice.
```
bazzadp, line 1237: Since we're gonna change this anyway, let's follow the editors convention of written when less than 10.

```suggestion
Examining the selectors on which custom properties are declared further confirms that most custom property usage in the wild is fairly basic. Two out of three custom property declarations are on the root element, indicating that they are used essentially as global constants. It is important to note that many popular polyfills have required them to be global in this vein, so developers using said polyfills may not have had a choice.
```

lex111:
@rviscomi oops, thanks, fixed!

bazzadp (approved):

rviscomi (approved):

bazzadp:
Thanks @lex111 - much appreciated!!

PR #1714, opened 2020-12-09
Stop including chapters with no featured quote in home page

bazzadp:
This PR avoids this:

image

### DIFF
--- a/src/tools/generate/generate_featured_chapters.js
+++ b/src/tools/generate/generate_featured_chapters.js
@@ -8,12 +8,16 @@ const generate_chapter_featured_quote = (metadata) => {
 
   let featured_quote_obj = {};
   let featured_stats = [];
-  if (featured_quote) {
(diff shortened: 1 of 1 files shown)


rviscomi (commented):
This just removes the quote part but the features stats are still there? If so LGTM.

bazzadp:
No. That’s what happens currently as per the screenshot.

What this does is return an empty object instead of a stat-only object. That makes it ineligible to be included and so we’ll just pick a random quote from any of the other chapters instead. We have 18 chapters with quotes so more than enough until the two missing ones finish out their meta data.

rviscomi:
Oh ok, this is a rare edge case so I'm ok with hiding the chapter until more metadata is added, but given that there's still useful featured content to share I would have thought it be ok to still show the stats. No strong opinions either way, we still want to get that quote!

bazzadp:
Also incentive for the chapter team to finish it out! 😀

rviscomi (approved):

PR #1720, opened 2020-12-09
fixes typo for in-page link in css chapter

argyleink:
no sure why the diff is so heavy?

it's `[which are also highly used](#custom properties)` to `[which are also highly used](#custom-properties)`, adds a `-`


bazzadp:
> no sure why the diff is so heavy?

GitHub is funny about this sometimes 😉

> it's `[which are also highly used](#custom properties)` to `[which are also highly used](#custom-properties)`, adds a `-`

bazzadp (approved):

PR #1728, opened 2020-12-10
Move figure dropdown to figure caption

### LINKED ISSUE #1716: Figure overflow menu issue in Mobile view
Overflow menu for figures can get overlapped with figure itself. I believe we should be able to handle this in a better way

A11Y - Overflow menu

bazzadp:
Fixes #1716 

@rviscomi I wanna suggest again that we move the figure menu to the figure caption thus solving #1716 .

As well as actually fixing the bug raised, I think this just works better in general as on current drop down button is just lost in the top right:

imageMenu button on it's own

In the suggested version it's not as lost:

Menu button near the figure

This applies to charts too where the current menu placement looks lost:

Graph now with menu button in top right

Versus the suggested:

Graph now with menu in the figure caption

The suggested version also works for Big Figures

Big Figure menu

This also allow consistency where the figure menu is always in the same place.

The open menu looks well too:

Chart open menu

Big figure menu open

The changes are very small so the alternative is to merge this and then look to put it in the top right properly when we have time?


bazzadp:
I also have a version working with it in the top right for charts only (where it might be expected for charts) **as well** as the figure caption (for consistency with tables and big number callouts):

Two figure drop downs

However, I think it's redundant to have both and still doesn't look like it's part of the figure - unless we add a border round the figure?

Alternatively could have top right for charts, and figure caption for tables and callouts? But think that's inconsistent and confusing.

My preference is still to have the figure in the caption which avoids all display issues with tables and big figures, avoids the floating effect with the menu being too far off to the right, and is consistent.

rviscomi:
I can try to articulate why my preference is to have the action menu on the figure itself and not the caption. The actions are being taken on the visualization itself, not the caption: viewing the raw data visualized in the chart, the query that made it, and the static image. Menus like this are also commonly positioned in the top corners so I think that helps convey to users what the icon does. Being at the top also means that the menu opens over the figure itself and doesn't drape over nearby content.

I think you also have good reasons to move the menu button. The "better" placement is a subjective call and while I'm not a designer, sometimes I have a gut feeling for how I think it should look. In this case it feels better to be at the top. To help resolve the overlap problem, one idea is to add more padding to the top or right of the figure. We could also consider giving it an opaque background on desktop and only showing it when the figure is hovered. Not sure about mobile.

bazzadp:
> I can try to articulate why my preference is to have the action menu on the figure itself and not the caption. The actions are being taken on the visualization itself, not the caption: viewing the raw data visualized in the chart, the query that made it, and the static image.

I disagree. If it was an embedded menu and part of something sheets provides I'd agree, but forcing in our contextual menu in on top of someone else's component confuses the ownership of this component. This is a menu related to the figure, not a controller of the chart.

Additionally we already give additional source information in figure captions:

Figcaption source

> Menus like this are also commonly positioned in the top corners so I think that helps convey to users what the icon does.

I somewhat agree for charts. The highcharts javascript library (that HTTP Archive main site uses) has certainly popularised this, and Google Sheets themselves use this.

However it's far from the only pattern.

Splunk has a contextual menu that appears at the bottom:

PowerBI chart PowerBI chart with menu

PowerBI and SAS visual analytics are other tools that don't use this pattern.

> Being at the top also means that the menu opens over the figure itself and doesn't drape over nearby content.

Can't argue with that point. You're 100% right.

> I think you also have good reasons to move the menu button. The "better" placement is a subjective call and while I'm not a designer, sometimes I have a gut feeling for how I think it should look. In this case it feels better to be at the top. To help resolve the overlap problem, one idea is to add more padding to the top or right of the figure. We could also consider giving it an opaque background on desktop and only showing it when the figure is hovered. Not sure about mobile.

I think for charts you are correct - top right would probably be "better" (though personally I don't think the figure caption option is bad by any means!). But the reason I'm against this isn't about the charts - I'd be fine putting it in top right if our figures (and this context menu) were just charts (though still do think it's still confusing our contextual menu with the chart controls). I've even shown that's relatively easy with the above screenshot showing it in the correct place for charts.

I've NEVER seen this top-right pattern for tables, nor images, nor call out quotes (aka our big figures). I don't think it's where people expect it to be and think it's just weird for tables to be quite honest. My argument is for consistency across our figures over sacrificing the other two types for an arguably slightly better experience for charts. However as a compromise I'd be OK to have it in top right on charts as well as in the figure if you want as suggested above?

Additionally doing it correctly for tables is more complex.

So I understand your arguments for the charts, but can you let me know your thoughts about our other figure types (tables, images, big numbers)?

bazzadp:
Closing as dealt with in #1778

PR #1730, opened 2020-12-10
Increase line height for content elements

lex111:
This should improve the text readability (a bit, see https://github.com/HTTPArchive/almanac.httparchive.org/pull/1718#issuecomment-742433677), I don't know how it will affect the ebook version, so please check it out there.

| Before   | After    |
| -------- | -------- |
| image | image |

### DIFF
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -771,7 +771,7 @@ code,
 li {
   font-size: 17px;
   font-size: 1.0625rem;
-  line-height: 1.5em;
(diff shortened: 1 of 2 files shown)


bazzadp (approved):
LGTM!

Also takes the ebook to 507 pages! Know you were gutted we were just under that at 499 pages @rviscomi 😁

I added a change for the ebook for big tables as otherwise the [large table in the CSS chapter](https://almanac.httparchive.org/en/2020/css#fig-27) overruns.

PR #1731, opened 2020-12-10
Fix bug with unsetting code background color in places

bazzadp:
@lex111 missed a slight bug in #1729 

`background-color: none;` is not valid CSS, so this wasn't doing anything. Changes to `unset` instead.


lex111 (approved):
Oh, yes, thanks!

PR #1740, opened 2020-12-10
Adjust position of switcher arrow

lex111:
I just noticed that on mobiles the arrow of switcher overlaps the text when the long option is selected.

Actually, since the arrow is absolutely positioned, we need to increase the padding for the option text to compensate width.

| Before   | After    |
| -------- | -------- |
| image | image |

The arrow is now outside the text area (as it should), :

| Before   | After    |
| -------- | -------- |
|image | image |

### DIFF
--- a/src/config/2020.json
+++ b/src/config/2020.json
@@ -266,7 +266,8 @@
       "name": "Alexey Pyltsyn",
       "teams": [
         "translators",
-        "editors"
+        "editors",
+        "developers"
       ],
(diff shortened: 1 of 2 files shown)


bazzadp (approved):
LGTM!

Wanna add yourself to the `developers` team in 2020.json as part of this?

lex111:
> Wanna add yourself to the developers team in 2020.json as part of this?

Sure! :smile:

PR #1744, opened 2020-12-10
Fix relative links in ebook

bazzadp:
The ebook retains links on our figures. At the minute those links all point to http://127.0.0.1:8080 :-(

This was working, as we change from relative to absolute paths in [generate_ebooks.js](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/src/tools/generate/generate_ebooks.js):

https://github.com/HTTPArchive/almanac.httparchive.org/blob/fd2e15d608fac09a50dcf98064dbfdccbd961f7b/src/tools/generate/generate_ebooks.js#L27-L28

I think it broke when we moved to `figure_markup` macro in #1236 as no longer have path in markdown now, so above JS won't help with those (though it will help with other relative links so still good to keep).

So, for the figures, we now need to override when that macro is executed - which this PR now does.

### DIFF
--- a/src/templates/base/base.html
+++ b/src/templates/base/base.html
@@ -777,6 +777,7 @@
 {# Ebook needs a unique fig id. 2019 is already out there so reluctant to change that, so "if" statemnt it is! #}
 {% if ebook %}
 {% set anchor="fig-%s-%s" % (metadata.chapter_number, id) %}
+{% if link.startswith("/") %}{% set link="https://almanac.httparchive.org%s" % link %}{% endif %}
 {% else %}
 {% set anchor="fig-%s" % (id) %}
 {% endif %}


src/templates/base/base.html
```diff
@@ -777,6 +777,7 @@
 {# Ebook needs a unique fig id. 2019 is already out there so reluctant to change that, so "if" statemnt it is! #}
 {% if ebook %}
 {% set anchor="fig-%s-%s" % (metadata.chapter_number, id) %}
+{% if link.startswith("/") %}{% set link="https://almanac.httparchive.org%s" % link %}{% endif %}
```
rviscomi, line 780: Does this account for relative URLs that start with `./` for example?

bazzadp: We don't currently have any of those. Can see how we would have to be honest as then they would be relative to the chapters not the static image directory.

Guess in theory could link to another chapter from an image but don't see why you'd do that.

I see leave for now as relative paths would be more complex to support. It's not the end of the world if the link is wrong - which is why no one's notice it's been broken for a while! The image still displays, just link broken if you click on it while reading PDF.

bazzadp:
Going to merge this and do a release.

PR #1753, opened 2020-12-12
Track ebook clicks from menu

### LINKED ISSUE #1715: Count eBook downloads in Google Analytics
We have made the Almanac available in eBook format but it would be interesting to know how much it's downloaded and currently we don't have that visibility.

Could we fire an event to Google Analytics when the ebook link is clicked?

Ideally I'd like to leave it as a link rather than change to a button (it is a link!). I presume it's possible to intercept the click, fire the event and then continue on to the link? Or does the navigation mean the event won't be found?

Also the link is now in our Table of Contents drop down link, and also as part of a `<select>` on mobile so we should cover those too.

Any of the @HTTPArchive/developers interested in looking at this one?

bazzadp:
Fixes #1715 

Forgot I'd already done this for the ToC page link in #825 !

This adds clicks for the menu option.
It also switchers to `'transport_type': 'beacon'` which is better for tracking links that cause a page `unload` (like the PDF link will).

### DIFF
--- a/src/templates/base/base.html
+++ b/src/templates/base/base.html
@@ -275,7 +275,7 @@
       <a href="{{ url_for('table_of_contents', year=year, lang=lang) }}#ebook">{{ self.ebook_title() }}</a>
     </li>
     <li class="nav-dropdown-list-chapter ebook">
-      <a class="ebook-menu-download" href="/static/pdfs/web_almanac_{{ year }}_{{ lang }}.pdf" data-event="ebook-click" data-label="toc-menu">{{ self.ebook_download_short() }}</a>
+      <a href="/static/pdfs/web_almanac_{{ year }}_{{ lang }}.pdf" data-event="ebook-click" data-label="toc-menu">{{ self.ebook_download_short() }}</a>
     </li>
     {% endif %}
   </ul>
@@ -624,7 +624,7 @@
         });
 
         document.querySelectorAll('[data-event]').forEach(trackableElement => {
-          trackableElement.addEventListener("click", function (event) {
+          trackableElement.addEventListener('click', function (event) {
             gtag('event', event.target.dataset.event, {
               'event_category': 'clicks',
               'event_label': event.target.dataset.label,
--- a/src/templates/base/base.html
+++ b/src/templates/base/base.html
@@ -490,8 +490,8 @@
             //Reset the selector back in case user uses Back button
             var selectedValue = e.target.value;
 
-            if (selectedValue.startsWith('/static/pdfs')) {
-              gtag('event', 'ebook-click', { 'event_category': 'clicks', 'event_label': 'mobile menu ' + document.location.pathname, 'transport_type': 'beacon', 'value': 1 })
+            if (e.target.dataset.event) {
+              gtag('event', e.target.dataset.event, { 'event_category': 'clicks', 'event_label': 'mobile menu ' + document.location.pathname, 'transport_type': 'beacon', 'value': 1 })
             }
             if (selectedValue && selectedValue !== window.location.pathname) {
               e.target.value = window.location.pathname;
--- a/src/templates/base/base.html
+++ b/src/templates/base/base.html
@@ -118,7 +118,7 @@
   <label for="table-of-contents-switcher-{{switcher_name}}" class="visually-hidden">
     {{ self.table_of_contents_switcher() }}
   </label>
-  <select id="table-of-contents-switcher-{{switcher_name}}">
+  <select id="table-of-contents-switcher-{{switcher_name}}" data-label="toc-menu-mobile">
     {% if request.path.endswith("table-of-contents") %}
       <option selected disabled value="{{ url_for('table_of_contents', year=year, lang=lang) }}">{{ self.table_of_contents_title() }}</option>
       <option disabled value="{{ url_for('table_of_contents', year=year, lang=lang) }}#foreword">{{ self.foreword_title() }}</option>
@@ -179,7 +179,7 @@
     {% endif %}
 
     {% if ebook_size_in_mb and ebook_size_in_mb > 0 %}
-    <option value="/static/pdfs/web_almanac_{{ year }}_{{ lang }}.pdf">
+    <option value="/static/pdfs/web_almanac_{{ year }}_{{ lang }}.pdf" data-event="ebook-click">
       {{ self.ebook_download_short() }}
     </option>
     {% endif %}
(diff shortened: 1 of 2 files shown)


src/templates/base/base.html
```diff
@@ -482,12 +482,17 @@
...
         var languageYearSwitchers = document.querySelectorAll('.language-switcher select, .year-switcher select, .table-of-contents-switcher select');
         for (var i = 0; i < languageYearSwitchers.length; i++) {
           languageYearSwitchers[i].addEventListener('change', function(e) {
+
             //Reset the selector back in case user uses Back button
             var selectedValue = e.target.value;
+
+            if (selectedValue.startsWith('/static/pdfs')) {
+              gtag('event', 'ebook-click', { 'event_category': 'clicks', 'event_label': 'mobile menu ' + document.location.pathname, 'transport_type': 'beacon', 'value': 1 })
+            }
```
rviscomi, line 495: Any concerns about brittleness? This statement sniffs out if the option appears to be the ebook download. Why not have the ebook download option unambiguously say what it is, for example `data-type="ebook"`, or more generally support trackable links using something like `data-event="ebook-click"`?
```suggestion
            if (e.target.dataset.event) {
              gtag('event', e.target.dataset.event, { 'event_category': 'clicks', 'event_label': 'mobile menu ' + document.location.pathname, 'transport_type': 'beacon', 'value': 1 })
            }
```

src/templates/base/base.html
```diff
@@ -275,7 +275,7 @@
       <a href="{{ url_for('table_of_contents', year=year, lang=lang) }}#ebook">{{ self.ebook_title() }}</a>
     </li>
     <li class="nav-dropdown-list-chapter ebook">
-      <a href="/static/pdfs/web_almanac_{{ year }}_{{ lang }}.pdf">{{ self.ebook_download_short() }}</a>
+      <a class="ebook-menu-download" href="/static/pdfs/web_almanac_{{ year }}_{{ lang }}.pdf">{{ self.ebook_download_short() }}</a>
```
rviscomi, line 278:
```suggestion
      <a class="ebook-menu-download" href="/static/pdfs/web_almanac_{{ year }}_{{ lang }}.pdf" data-event="ebook-click">{{ self.ebook_download_short() }}</a>
```

src/templates/base/base.html
```diff
@@ -611,6 +616,16 @@
             }
           }
         });
+
+        var ebookLink = document.querySelector('#ebook-toc-download');
+        ebookLink && ebookLink.addEventListener("click", function (event) {
+          gtag('event', 'ebook-click', { 'event_category': 'clicks', 'event_label': 'toc-page', 'transport_type': 'beacon', 'value': 1 })
+        });
+
+        var ebookLink = document.querySelector('.ebook-menu-download');
```
rviscomi, line 625: Could there be more than one element with this class, and would we need to add the event listener for each?

src/templates/base/base.html
```diff
@@ -611,6 +616,16 @@
             }
           }
         });
+
+        var ebookLink = document.querySelector('#ebook-toc-download');
+        ebookLink && ebookLink.addEventListener("click", function (event) {
+          gtag('event', 'ebook-click', { 'event_category': 'clicks', 'event_label': 'toc-page', 'transport_type': 'beacon', 'value': 1 })
```
rviscomi, line 622: How useful is it for the `event_label` to tell us on what page the event happened? Doesn't GA already know that?

src/templates/base/base.html
bazzadp: Good idea. Based it on the link one too much instead of thinking how to do this properly.

src/templates/base/base.html
bazzadp: Good point! There will be two: header and footer menu. Will fix when next on the computer.

src/templates/base/base.html
bazzadp: Not sure it’s that useful at all tbh. But previous event had `toc-page` so had to replace it. Happy to go with just `toc-menu`.

src/templates/base/base.html
rviscomi: What about something generic to build on the `data-event` attribute? For example:

```js
var trackableElements = Array.from(document.querySelectorAll('[data-event]'));
trackableElements.forEach(function (trackableElement) {
  trackableElement.addEventListener('click', function (event) {
    gtag('event', event.target.data.event, { 'event_category': 'clicks', 'event_label': 'menu ' + document.location.pathname, 'transport_type': 'beacon', 'value': 1 });
  });
});
```

bazzadp: OK all fixed. Needed a slightly different method for the `<select>` version but very similar.

src/templates/base/base.html
```diff
@@ -482,10 +482,19 @@
           element.hidden = false;
         });
 
-        // Language and Year switching
+        // Language, Year and ToC switching
         var languageYearSwitchers = document.querySelectorAll('.language-switcher select, .year-switcher select, .table-of-contents-switcher select');
         for (var i = 0; i < languageYearSwitchers.length; i++) {
           languageYearSwitchers[i].addEventListener('change', function(e) {
+
+            if (e.target.dataset.label) {
```
rviscomi, line 490: Will this condition be true for all options? Should it only be true when the switcher has a label and the option has the `data-event` attribute?

bazzadp: Had actually spotted this myself and fixed, but looks like I forgot to `git push`. Should be fixed now.

src/templates/base/base.html
```diff
@@ -275,7 +275,7 @@
       <a href="{{ url_for('table_of_contents', year=year, lang=lang) }}#ebook">{{ self.ebook_title() }}</a>
     </li>
     <li class="nav-dropdown-list-chapter ebook">
-      <a href="/static/pdfs/web_almanac_{{ year }}_{{ lang }}.pdf">{{ self.ebook_download_short() }}</a>
+      <a class="ebook-menu-download" href="/static/pdfs/web_almanac_{{ year }}_{{ lang }}.pdf" data-event="ebook-click" data-label="toc-menu">{{ self.ebook_download_short() }}</a>
```
rviscomi, line 278: Is this class needed anymore?

src/templates/base/base.html
```diff
@@ -611,6 +622,17 @@
             }
           }
         });
+
+        document.querySelectorAll('[data-event]').forEach(trackableElement => {
+          trackableElement.addEventListener("click", function (event) {
```
rviscomi, line 627: To confirm, this won't double-fire for `<option>`s because those are change events and not clicks?

Also one nit about consistent use of single quotes:
```suggestion
          trackableElement.addEventListener('click', function (event) {
```

rviscomi (approved):
Good to merge when the comments are resolved. Thanks for iterating on this!

src/templates/base/base.html
bazzadp: Nope. It's gone!

src/templates/base/base.html
bazzadp: Fixed.

PR #1622, opened 2020-12-03
Copyedit of 2020 Performance chapter

bazzadp:
Made the following changes:

- Fix incorrect heading levels
- Normalised Smart Quotes as per #1485 and changed "it's" to "it is"
- Corrected references to 2019 chapter
- Removal of frequent use of bold in keeping with Web Almanac style
- Corrected Sheets GIDs
- Added some links
- Change of "the amount" to "the number"
- Misc edits for readability

Still to do:
- [x] - Change black on red to white on red in charts, similar to last years charts
- [x] - Font incorrect in figures 1, 12, and 13 
- [x] - Two charts (Figures 2 and 18) are wrong in sheets embed
- [x] - Add full charts for Geo breakdown as per 2019 and set heights and widths in markdown
- [x] - Re-export fallback images in higher quality (@rviscomi it appears that only those lucky enough to have 4k and 5k screens can actually export high quality images when when using the [capture node screenshots](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Figures-Guide#how-to-take-screenshots-of-google-sheet-graphics)  😞 )
- [x] - Few comments for @thefoxis to consider
- [x] - remove "unedited" flag

Staged link: https://20201204t102628-dot-webalmanac.uk.r.appspot.com/en/2020/performance (updated since initially raising this to incorporate all the changes)

PR #1763, opened 2020-12-14
Fix one last TODO comment in 2020 Performance chapter

bazzadp:
Looks like I missed one in #1622 

@thefoxis you OK with this change?

@borisschapira / @Zuckjet FYI as you're busy translating this chapter.


bazzadp:
GitHub not highlight the changes very nicely!

It just changes this:

> Again, it isn't easy to speculate why that might be {# I think it IS easy to speculate why - precisely for the reasons you give! #}. 

to this:

> Again, we can speculate why that might be, but would need further analysis to really be sure.

Zuckjet:
in the section of "FCP by connection type"， this words:

" so ensuring text is visible while web fonts are downloaded is also a worthwhile strategy"

I think it need to change to like this "Ensure text remains visible during webfont load".

should be downloading process not done. sorry for bad format because of away from my computer

bazzadp:
> in the section of "FCP by connection type"， this words:
> 
> " so ensuring text is visible while web fonts are downloaded is also a worthwhile strategy"
> 
> I think it need to change to like this "Ensure text remains visible during webfont load".
> 
> should be downloading process not done. sorry for bad format because of away from my computer

The current text reads fine to me:

> FCP is also significantly affected by font load time, so ensuring text is visible while web fonts are downloaded is also a worthwhile strategy (especially where on slower connections these resources will be costly to fetch).

Not sure I get your concern?

Zuckjet:
my understand is that we should emphasize the process of downloading. that is to say the difference between "font have downloaded" and "font is downloading".

Maybe my understand is mistake, if so just ignore.

bazzadp:
Getting into the depths of English grammar that are beyond me! 😁 However, to me, this:

> FCP is also significantly affected by font load time, so ensuring text is visible while web fonts are **downloaded** is also a worthwhile strategy (especially where on slower connections these resources will be costly to fetch).

reads the same as this:

> FCP is also significantly affected by font load time, so ensuring text is visible while web fonts are **downloading** is also a worthwhile strategy (especially where on slower connections these resources will be costly to fetch).

I don’t think the first implies past tense (i.e. that they have been downloaded already) and it’s more along the lines of “..while web fonts are downloaded in parallel in the background”.

thefoxis (approved):

thefoxis:
@bazzadp approved 👍 thanks for that!

the two sentences above read identical to me too, I think especially that `while` is used before the subject and the verb 😸

PR #1737, opened 2020-12-10
2020 JavaScript edits

rviscomi:
Progress on #897 #1432

PR #1770, opened 2020-12-15
Fix typo in "the"

westonruter:


rviscomi:
Thanks @westonruter! We have a pending fix for this in #1737. I'll see if we can get those changes merged and deployed ASAP.

PR #1778, opened 2020-12-16
Alternative improvement to figure menus

### LINKED ISSUE #1716: Figure overflow menu issue in Mobile view

bazzadp:
Fixes #1716 

This is an alternative to #1728 keeping the menu in the top right but handling it better.

For tables this is a bit more painful and have to inject it into the DOM at build time as need it in the `<div class="table-wrap">` element so it's not all the way on the right.

Staged link: https://20201216t175100-dot-webalmanac.uk.r.appspot.com

[Example graph](https://20201216t175100-dot-webalmanac.uk.r.appspot.com/en/2020/http2#fig-6): 

Example graph

[Example table](https://20201216t175100-dot-webalmanac.uk.r.appspot.com/en/2020/http2#fig-3):

Example table

[Example Big Figure](https://20201216t175100-dot-webalmanac.uk.r.appspot.com/en/2020/http2#fig-2):

Example Big Figure

### DIFF
--- a/src/static/css/page.css
+++ b/src/static/css/page.css
@@ -491,21 +491,15 @@ figcaption {
 
 .figure-dropdown {
   position: absolute;
-  top: 10px;
-  right: -5px;
+  top: -10px;
+  right: -50px;
   display: flex;
   flex-direction: column;
   align-items: flex-end;
 }
 
-.media-figure-wrapper .figure-dropdown {
-  top: -5px;
-  top: -10px;
-}
-
-.content-figure-wrapper .figure-dropdown {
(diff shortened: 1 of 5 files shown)


bazzadp:
Use this when reviewing as Moved the figures under a new `<div>` which makes it look like there's more change than there actually is in base.html.

Ignore whitespace changes

bazzadp:
Merging as agreed with @rviscomi

PR #1789, opened 2020-12-17
Improved home hero image

rviscomi:
When sharing the home page on Twitter, a cropped version of the main graphic is used:

image

Improve this image in two ways:

- give it a 2:1 aspect ratio to prevent cropping
- add the year in the image

I created the graphic in Google Slides ([source](https://docs.google.com/presentation/d/1wSVyFi8gxkEr85mQkWSn8gbDUkZ0oPkrMPu-fmYHlaY/edit?usp=sharing)) which we can reuse to generate new year-specific images.

image


src/templates/base/index.html
```diff
@@ -1,8 +1,8 @@
 {% extends "%s/%s/base.html" % (lang, year) %}
 
-{% block image_url %}https://almanac.httparchive.org/static/images/home-hero-bg.png{% endblock %}
-{% block image_height %}562{% endblock %}
-{% block image_width %}715{% endblock %}
+{% block image_url %}https://almanac.httparchive.org{{ get_versioned_filename('/static/images/home-hero-' + year + '.png') }}{% endblock %}
```
rviscomi, line 3: @bazzadp I'm not sure if I should be using `get_versioned_filename` here. I don't see pngs in `last_updated.json`, but should there be?

rviscomi, line 3: I also wonder if there should be a fallback to `home-hero-bg.png` if the year is out of bounds. Probably not necessary since that would normally be an error page anyway. However, for example, if we forget to upload the 2021 graphic, maybe it would be nice to use the default year-less graphic instead of having nothing. Or should we add this to the tests somehow?

bazzadp: Personally I don't think it's worth tracking images. Very, very, very rare there's breaking changes in these unlike CSS or JS.

bazzadp: Lighthouse test will show error in console log and therefore catch this so don't think its necessary.

bazzadp (changes requested):
Yeah never been particularly happy with the Twitter image on this.

File sizes big. Can we run them through tinypng.com?

Also I quite like the website screenshots some have used:
https://twitter.com/quinaribeiro/status/1339597356686168067?s=20
What's your thoughts on those?

Do we need home-hero-bg.png anymore?

Also can you update ebook template?

bazzadp:
Here's what TinyPng has to say:

image

src/templates/base/index.html
rviscomi: I was asking for its cache-busting side-effect. Images can be long-lived and it would be nice to ensure that any changes take effect immediately after a release.

rviscomi:
> File sizes big. Can we run them through tinypng.com?

Done. I would have expected the Calibre action to fix those. If we need to manually reduce it in tinypng anyway, do we need Calibre? Or is there a tinypng GH action?

> Do we need home-hero-bg.png anymore?

I'd be hesitant to remove it entirely, for example if social sharers are still referring to it. We may also want to reuse it on other pages at some point (eg accessibility statement, or other year-independent pages).

> Also I quite like the website screenshots some have used:
>https://twitter.com/quinaribeiro/status/1339597356686168067?s=20
>What's your thoughts on those?

I think we can do better with a custom graphic, and wouldn't want to tie it to any particular language.

> Also can you update ebook template?

Done.

src/templates/base/base_ebook.html
```diff
@@ -2,9 +2,10 @@
 
 {% set metadata = {"authors":["rviscomi"]} %}
 
-{% block image_url %}https://almanac.httparchive.org/static/images/home-hero-bg.png{% endblock %}
-{% block image_height %}562{% endblock %}
-{% block image_width %}715{% endblock %}
+{% block page_url %}https://almanac.httparchive.org{{ url_for(request.endpoint, **get_view_args(lang=language.lang_code, year=year)) }}{% endblock %}
```
bazzadp, line 5: Do we need this? Seems like it's the default one (set in base.html) so no need to override here?

```suggestion
```

src/templates/base/index.html
bazzadp: That's not a side-effect - that's the whole reason we have that 😁

Yes we could do this, but also then should do for all the figures across all the chapters and years. And that makes the file rather large and its read at server start up so could impact the server - especially as it grows and grows over the years.

We have a maximum cache tie of 3 hours (I'm against really long lived cache times to be honest precisely for this reason) so not sure it's worth the effort.

src/templates/base/base_ebook.html
rviscomi: Good catch, bad copypasta.

bazzadp:
> Done. I would have expected the Calibre action to fix those. If we need to manually reduce it in tinypng anyway, do we need Calibre? Or is there a tinypng GH action?

Yeah I'm not sure why Tinypng is so much better than Calibre to be honest. As I said before, I see Calibre as at least a simple checkpoint as it's integrated in via the GitHub action so if someone forgets, or we're all in a blind panic as we approach publication and no time to do better, we get at least some compression.

There isn't a TinyPng GitHub Action unfortunately. They do have an API key which is free for first 500 images compressed per month (no charge if no compression). Should either look into that or tweak the Calibre settings to do better.

src/templates/base/index.html
```diff
@@ -1,8 +1,8 @@
...
+{% block image_height %}600{% endblock %}
+{% block image_width %}1200{% endblock %}
```
bazzadp, line 5: Height and Width are different for 2020 - can we make it the same?

```
% identify -format '%w %h %b %i\n' home*.png 
1200 600 148774B home-hero-2019.png
960 540 126188B home-hero-2020.png
1200 600 141273B home-hero-bg.png
820 562 38042B home-hero.png
```

rviscomi: Good catch! Done.

src/templates/base/index.html
rviscomi: Would you suggest hard-coding the URL in that case?

Also one other question, WDYT about saving the year-specific images under their respective static/image/[year] subdirectories?

bazzadp (approved):
LGTM!

Reminder to stick our home pages through the Twitter validator after release to refresh the Twitter cache: https://cards-dev.twitter.com/validator

src/templates/base/index.html
bazzadp: > Would you suggest hard-coding the URL in that case?

Can;t be fully hardcoded as you now have a different one for 2019 and 2020 but could be semi-hardcoded with 

```suggestion
{% block image_url %}https://almanac.httparchive.org/static/images/home-hero-{{ year }}.png{% endblock %}
```

Probably more efficient to do that than your current function.

> Also one other question, WDYT about saving the year-specific images under their respective static/image/[year] subdirectories?

Yeah we should probably do that. I've an outstanding question/TODO in #1558 to clean up chapter images and, depending how we do that, could sort this out at same time. For this PR, I say keep it as is for now.

bazzadp: Shall I commit this suggestion and then merge this @rviscomi ?

PR #1793, opened 2020-12-17
Identify links in languages that differ to the page

### LINKED ISSUE #1781: Design proposal: Identify links in languages that differ to the page
## Target pages

Mostly translations, I guess.

## Objectives

Allow users and navigation tools to anticipate the change in language mid-navigation.
Prevent users from being taken to pages whose language they don’t understand.

([Opquast Checklist BP 126](https://checklists.opquast.com/en/qualiteweb/the-main-language-of-a-links-target-page-is-provided-if-different-from-the-original-page))

## Screenshots Before / After (one visual lead, others are possible)

Capture d’écran 2020-12-16 à 22 03 43

Capture d’écran 2020-12-16 à 22 08 36

borisschapira:
Fixes #1781

If this PR was accepted, all translators would have to be notified of this new capability.

### DIFF
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -1601,4 +1601,3 @@ html:not([lang="en"]) [href^="https://en."]::after {
 html[lang="en"] [hreflang="en"]::after {
   content: '';
 }
-
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -1571,3 +1571,34 @@ pre .support.value {
 pre .regexp {
   color: #009926;
 }
+
+/*
+ * Automatically adds language after an anchor if not same language
+ *
+ * Add for links that are obviously in English
+*/
+html:not([lang="en"]) [href*="/en-US/"]::after,
+html:not([lang="en"]) [href*="/en/"]::after,
+html:not([lang="en"]) [href^="https://en."]::after {
+  content: '(en)';
+  vertical-align: super;
+  font-size: 0.6em;
+}
+
+/*
+ * Add links if an explicit `hreflang` attribute exists
+ */
+[hreflang]::after {
+  content: '(' attr(hreflang) ')';
+  vertical-align: super;
+  font-size: 0.6em;
+}
+
+/*
+ * Remove it for English in English pages
+ * (allows us to add this to base content to make it easier for translators)
+ */
+html[lang="en"] [hreflang="en"]::after {
+  content: '';
+}
+
--- a/src/static/css/2019.css
+++ b/src/static/css/2019.css
@@ -1577,9 +1577,9 @@ pre .regexp {
  *
  * Add for links that are obviously in English
 */
-html:not([lang="en"]) [href*="/en-US/"]::after,
-html:not([lang="en"]) [href*="/en/"]::after,
-html:not([lang="en"]) [href^="https://en."]::after {
+html:not([lang="en"]) main a[href*="/en-US/"]::after,
+html:not([lang="en"]) main a[href*="/en/"]::after,
+html:not([lang="en"]) main a[href^="https://en."]::after {
   content: '(en)';
   vertical-align: super;
   font-size: 0.6em;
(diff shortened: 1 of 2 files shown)


src/static/css/page.css
```diff
@@ -780,3 +780,20 @@ pre {
...
+ * used when different lang).
+ *
+ * The first three lines is a fallback for those
+ * that haven’t implemented `hreflang`.
+ */
+html:not([lang="en"]) [href*="/en-US/"]::after,
+html:not([lang="en"]) [href*="/en/"]::after,
+html:not([lang="en"]) [href^="https://en."]::after,
+[hreflang]::after {
+  content: '(' attr(hreflang) ')';
```
bazzadp, line 796: This won’t work for the fallbacks as they won’t have a `hreflang`.  Which is why I think
You need two sets of rules - with the fallback hardcoded to `en`

borisschapira: Oh damn, I'll modify this tomorrow, you're right.

borisschapira: Done

src/static/css/page.css
```diff
@@ -780,3 +780,28 @@ pre {
...
+ * Fallback for the links that are obviously in English
+ * but that haven’t implemented `hreflang`.
+*/
+html:not([lang="en"]) [href*="/en-US/"]::after,
+html:not([lang="en"]) [href*="/en/"]::after,
+html:not([lang="en"]) [href^="https://en."]::after {
+  content: '(en)';
+  vertical-align: super;
+  font-size: 0.6em;
+}
```
bazzadp, line 807: Let's put in other order so the actual `hreflang` has the precedence.

Also let's add an override for `<html lang="en">`/`hreflang="en"` so we can add to English and tackle this in one go (not saying we will add this, but this at least allows us to have that option).

```suggestion
/*
 * Automatically adds language after an anchor if not same language
 *
 * Add for links that are obviously in English
*/
html:not([lang="en"]) [href*="/en-US/"]::after,
html:not([lang="en"]) [href*="/en/"]::after,
html:not([lang="en"]) [href^="https://en."]::after {
  content: '(en)';
  vertical-align: super;
  font-size: 0.6em;
}

/*
 * Add links if an explicit `hreflang` attribute exists
 */
[hreflang]::after {
  content: '(' attr(hreflang) ')';
  vertical-align: super;
  font-size: 0.6em;
}

/*
 * Remove it for English in English pages
 * (allows us to add this to base content to make it easier for translators)
 */
html[lang="en"] [hreflang="en"]::after {
  content: '';
}
```

bazzadp (approved):

bazzadp (changes requested):
Actually just one more change @borisschapira - could we move this from page.css to 2019.css so the whole site can benefit from this? Most pages do load page.css, but some (e.g. Home page, ToC) don't.

borisschapira:
And all pages load 2019.css?

bazzadp:
Correct. I wanna rename this `almanac.css` btw to make it obvious it's used in 2020 pages - see #1558

src/static/css/2019.css
```diff
@@ -1571,3 +1571,33 @@ pre .support.value {
...
+  content: '(' attr(hreflang) ')';
+  vertical-align: super;
+  font-size: 0.6em;
+}
+
+/*
+ * Remove it for English in English pages
+ * (allows us to add this to base content to make it easier for translators)
+ */
+html[lang="en"] [hreflang="en"]::after {
```
bazzadp, line 1601: Actually let's restrict this to `main` links as our DOM size is quite large and can't see this being used on anything but links. Also restricted to `main` means it won't show in language menu where I think this is obvious and not needed:

```suggestion
html:not([lang="en"]) main a[href*="/en-US/"]::after,
html:not([lang="en"]) main a[href*="/en/"]::after,
html:not([lang="en"]) main a[href^="https://en."]::after {
  content: '(en)';
  vertical-align: super;
  font-size: 0.6em;
}

/*
 * Add links if an explicit `hreflang` attribute exists
 */
main a[hreflang]::after {
  content: '(' attr(hreflang) ')';
  vertical-align: super;
  font-size: 0.6em;
}

/*
 * Remove it for English in English pages
 * (allows us to add this to base content to make it easier for translators)
 */
html[lang="en"] main a[hreflang="en"]::after {
```

bazzadp (changes requested):
Sorry one more change!

bazzadp (approved):

PR #1795, opened 2020-12-18
ES: #498: templates, error.html, improve translation in Spanish

hijuliancode:
Makes progress on #498


src/templates/es/error.html
```diff
@@ -1,24 +1,24 @@
...
-    "Unsupported language requested": "Idioma no Admitido Solicitado",
+    "Unsupported year requested": "Solicitud de año no admitida",
+    "Unsupported chapter requested": "Solicitud de capítulo no admitida",
+    "Unsupported language requested": "Solicitud de idioma no admitida",
     "Not Found": "La URL solicitada no se encontró en el servidor. Si ingresó la URL manualmente, verifique su ortografía e intente nuevamente."
   }
 %}
 {% macro expandedErrorMessage(errorCode, errorDescription, errorMessage) %}
-{{ errorCode }} <i lang="en">{{ errorDescription }}</i>: {{ errorMessage }}
+{{ errorCode }} <i lang="es">{{ errorDescription }}</i>: {{ errorMessage }}
```
bazzadp, line 23: This bit is actually in English (it's the original error code returned by the server).

```suggestion
{{ errorCode }} <i lang="en">{{ errorDescription }}</i>: {{ errorMessage }}
```

bazzadp (approved):
The other language error messages (including English) are in Title case, but you're probably right and they probably shouldn't be going with the rest of the conventions used in the Almanac so happy to accept this with one reversion.

PR #1266, opened 2020-09-01
Stop having comma for two person lists

bazzadp:
Fixes #1247

PR #1799, opened 2020-12-18
Oxford commas for Contributor teams

bazzadp:
Looks like we broke contributors with two teams for ebook in #1266 

Two teams screenshot

Still works for more that two teams:

Multiple teams screenshot

Let's move to Oxford commas for both Contributors and Ebook!

Staged here: https://20201218t132416-dot-webalmanac.uk.r.appspot.com/en/2020/contributors#ibnesayeed
And ebook here: https://20201218t132416-dot-webalmanac.uk.r.appspot.com/static/pdfs/web_almanac_2020_en.pdf

Screenshots:

Two teams screenshot fixed

Multiple teams screenshot fixed

### DIFF
--- a/src/static/css/ebook.css
+++ b/src/static/css/ebook.css
@@ -167,15 +167,17 @@ tbody tr:nth-child(even) {
 .contributor-teams {
   line-height: 1;
   color: #757575;
+  font-size: 8px;
+  font-size: 0.8rem;
 }
 
 .contributor-team,
 .contributor-social,
 .contributor-social a,
 .contributor-social a:visited {
+  color: #757575;
   font-size: 8px;
   font-size: 0.8rem;
-  color: #757575;
 }
 
 .authors .social,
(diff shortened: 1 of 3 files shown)


rviscomi (approved):

ibnesayeed (commented):
I would have liked it to be implemented as a [custom template filter](https://flask.palletsprojects.com/en/1.1.x/templating/#registering-filters), so that the conditional logic goes to pure Python files and can be reused everywhere we need oxford commas.

bazzadp:
Think that's quite a bit bigger change but welcome to submit a PR for this.

bazzadp:
Also in this case we want specific spans around the teams, so may not be as easy.

Will merge this and you can look at the custom filter option afterwards if you want @ibnesayeed

ibnesayeed:
> Also in this case we want specific spans around the teams, so may not be as easy.

If Jinja supported supported list comprehensions, it would have been easy to create a list that has markup wrappers around each element before passing through oxford comma filter, but that's not the case.

Issue #1736, opened 2020-12-10
Include Nuxt.js and Docusaurus in the 2020 Jamstack chapter


rviscomi:
Readers are pointing out that Nuxt.js should have been included in the Jamstack chapter.

Is it true that it belongs in the chapter? If so let's rerun the analysis and update the chapter's interpretation. Let's also look into opening a PR to add it to the SSG category in Wappalyzer.

bazzadp:
Unassigned myself for now to be clear on responsibilities here.

I want the content team, as Jamstack experts (which I am not) to decide what ones are missing (Docusaurus was raised too). Since we need to rerun, let's use this as an opportunity to consider all missing Jamstacks rather than do them one at a time.

After that is decided, let's get the queries changed, the extracts rerun, and the wording changed in a PR.

After all that is done, I'm happy to get involved with reviewing PRs, doing any image exports, and whatever I can do to help get this change live. And of course always available for questions in the meantime.

Will continue to monitor this issue (I monitor all issues so don't need to be assigned for that).

rviscomi:
Sorry! From the Twitter thread it had looked like you were involved in some of the analysis rerunning. Are those queries available anywhere for @denar90?

bazzadp:
Yeah that was the danger of me being helpful 😀

I reran just one of the queries by adding an `OR app = "Nuxt.js"` line:

```sql
#standardSQL
# SSG adoptions and top SSGs YoY
SELECT
  _TABLE_SUFFIX AS client,
  2020 AS year,
	app AS ssg,
  COUNT(0) AS freq,
  total,
  COUNT(0) / total AS pct
FROM
  `httparchive.technologies.2020_08_01_*`
JOIN (
  SELECT
    _TABLE_SUFFIX,
    COUNT(0) AS total
  FROM
    `httparchive.summary_pages.2020_08_01_*`
  GROUP BY
    _TABLE_SUFFIX)
USING
  (_TABLE_SUFFIX)
WHERE
  LOWER(category) = "static site generator" OR
  app = "Next.js" OR
  app = 'Nuxt.js'
GROUP BY
  client,
  total,
  ssg
UNION ALL
SELECT
  _TABLE_SUFFIX AS client,
  2019 AS year,
	app AS ssg,
  COUNT(0) AS freq,
  total,
  COUNT(0) / total AS pct
FROM
  `httparchive.technologies.2019_07_01_*`
JOIN (
  SELECT
[...]

PR #1812, opened 2020-12-19
Jamstack 2020 queries: Add Nuxt.js and Docusaurus

denar90:
Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/1736

Also I've updated spread sheet with latest results. cc @ahmadawais

### DIFF
--- a/sql/2020/17_Jamstack/adoption_of_image_formats_in_ssgs.sql
+++ b/sql/2020/17_Jamstack/adoption_of_image_formats_in_ssgs.sql
@@ -23,8 +23,10 @@ JOIN (
   FROM
     `httparchive.technologies.2020_08_01_*`
   WHERE
-  	LOWER(category) = "static site generator" OR
-  	app = "Next.js")
+    LOWER(category) = "static site generator" OR
+    app = "Next.js" OR
+    app = "Nuxt.js" OR
+    app = "Docusaurus")
 USING
   (client, page)
 GROUP BY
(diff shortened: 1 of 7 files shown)


rviscomi:
Thanks @denar90! I've changed your comment from "Fixes" to "Progress on" so that we can keep the issue open until the markdown changes are made, including any text and figure changes.

rviscomi (approved):

ahmadawais:
@denar90 thabks for all your work. Can you update this to include Sphinx, GitBook, MkDoc, and YUI doc? Take a look at the relevant discussion in the issue you tagged above, doc tools that I listed above are producing SSGs and have over 100 sites so we should use the opportunity to include these alongside Docusaurus. 🙌

denar90:
Thank you 🙏 
Of course we can add Sphinx, GitBook, MkDocs, and YUI. Just nit question, are we sure they are Jamstack related? cc @remotesynth

remotesynth:
I kind of have the same concern as @denar90. We seem to be getting off the main point of this research. I'd rather see Docusaurus removed and not cover documentation generators than to see all these added. I wouldn't call these Jamstack tools since they are a) very specific to documentation and b) purely static site generators. For example, YUIdoc is just an API docs generator - this is not for building Jamstack applications.

bazzadp:
To be honest, based on my analysis in https://github.com/HTTPArchive/almanac.httparchive.org/issues/1736#issuecomment-74313597 I’m not sure edits work arguing over. Not of the documentation tools will even register compared to other Jamstack usage. So either leave them all out or include them but all but don’t think it’ll make much difference either way to be honest.

denar90:
I tend to agree with @remotesynth 
Are we ok removing doc tools ? Also cc @rviscomi @ahmadawais

Issue #1571, opened 2020-11-23
Publish 2020 Web Almanac "stories"


rviscomi:
This issue will track the plan to publish one or more Web Almanac [stories](https://amp.dev/about/stories/). The purpose of the stories is to create an engaging way to showcase the stats from each chapter and market the report in web searches so that we drive organic growth in addition to all of our other social media marketing efforts. The current thinking is to create one story for each of the four distinct groups of chapters, grouped by Part I through Part IV as seen in the [ToC](https://almanac.httparchive.org/en/2020/table-of-contents). We'll highlight one or two of the most interesting stats from each chapter in each "page" of the story. Basically, this is secondary marketing material to attract more readers. We're not changing anything about the way the content itself is published.

Here are a few concrete tasks to publish the stories:

- Compile a list of the most interesting stats from all 22 chapters. For chapters whose markdown drafts are already created, we can reuse their featured stats metadata.
- I will coordinate with an external contractor to develop the stories using these stats and our site design/illustrations as inspiration.
_(Side note: Google is an HTTP Archive sponsor and will be covering the contractor's costs directly. For context, we also contracted out the original Web Almanac design and illustrations.)_
- When the stories are delivered to us, we'll need to serve them from the website. They will be developed using the WordPress editor so we'll need to make sure that all of the static resources are addressed and served according to our own directory structure and not the WordPress defaults.
- Implement SEO best practices to make sure the stories are indexable and discoverable in search tools. Here's a helpful [SEO guide to stories](https://developers.google.com/search/docs/guides/web-stories-creation-best-practices).

This is something that should go live on or shortly after the big launch on December 9.

Once above is released, we should create a story for each section:

- [x] - Story 1 - Part I. Page Content - #1800
- [x] - Story 2 - Part II. User Experience - #1864 
- [x] - Story 3 - Part III. Content Publishing - #1880
- [x] - Story 4 - Part IV. Content Distribution - #1898 
- [x] - Updated stories created with 2019 stats as chapters were not ready - #1900

bazzadp:
All done so closing. Thanks for you work on this @swissspidy !

Any ideas how long these take to appear in SERPs? Been googling for "Web Almanac" occasionally but nothing yet, though it does show in AMP section of Google Search Console.

swissspidy:
> Any ideas how long these take to appear in SERPs? Been googling for "Web Almanac" occasionally but nothing yet, though it does show in AMP section of Google Search Console.

[...]

PR #1815, opened 2020-12-20
Add stories to sitemap

rviscomi:
Add stories to sitemap #1571 

This is extremely hacky because I'm not familiar with the sitemap generation code. I'm hard-coding the path and last modified date into the generator. Eventually we should change this to crawl the directory and check for stories, but this was a quick fix to ensure that the published story can be indexed and start appearing organically.

Here's the new output in `sitemap.xml`:

```xml
    <url>
        <loc>https://almanac.httparchive.org/en/2020/stories/page-content.html</loc>
        <lastmod>2020-12-19</lastmod>
    </url>
```


src/tools/generate/generate_sitemap.js
```diff
@@ -57,7 +57,9 @@ const get_static_pages = async (sitemap_languages) => {
     .reduce((x, y) => [...x, ...y], []);
 
   // Get the sitemap entries for those pages
-  let urls = [];
+  let urls = [
+    {url: 'en/2020/stories/page-content.html', lastmod: '2020-12-19'},
+  ];
```
bazzadp, line 62: Revert this back:

```suggestion
  let urls = [];
```

And instead add this to line 7:

```js
const static_pages = [
  'index.html',
  'table_of_contents.html',
  'methodology.html',
  'contributors.html',
  'accessibility_statement.html',
  'stories/page-content.html'
];
```

bazzadp (changes requested):
Since you move this you need to update server/routes.py line 93:

```python
def stories(lang, year, story):
    delattr(request, 'csp_nonce')
    return render_template('%s/%s/stories/%s.html' % (lang, year, story))
```

Though for other templates we've always used underscores rather than dashes but seems to work fine with dashes.

rviscomi:
> Since you move this you need to update server/routes.py line 93:

Good catch! Thanks.

> Though for other templates we've always used underscores rather than dashes but seems to work fine with dashes.

The sitemap generator assumes every file with underscores is served with dashes instead, so the sitemap will point to a 404 URL if we used underscores.

bazzadp:
> The sitemap generator assumes every file with underscores is served with dashes instead, so the sitemap will point to a 404 URL if we used underscores.

Yeah in routes.js we map these, for example:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/3ef5ece5f74f9e9ea84d02961e437fea77fbe3d4/src/server/routes.py#L30-L33

So if we wanted to do the same here we'd need to do this:

```python
def stories(lang, year, story):
    delattr(request, 'csp_nonce')
    return render_template('%s/%s/stories/%s.html' % (lang, year, story.replace('-','_')))
```

Though, as I say, not sure we we did underscores for template names but dashes for routes (before my time!) and it seems to work as is, so up to you whether you want this consistency with other templates or fine to leave as is.

bazzadp:
The URL is referenced in 3 other places in the HTML so they should be changed too:

```html
  <meta property="og:url" content="https://almanac.httparchive.org/en/2020/stories/web-almanac-2020-page-content/">
```

```html
  <script
    type="application/ld+json">{"@context":"http://schema.org","publisher":{"@type":"Organization","name":"The 2020 Web Almanac","logo":{"@type":"ImageObject","url":"/static/stories/web-almanac-2020-page-content/WA_Logo_2-96x96.png"}},"@type":"Article","mainEntityOfPage":"https://almanac.httparchive.org/en/2020/stories/web-almanac-2020-page-content","headline":"Web Almanac Part I: Page Content","datePublished":"2020-12-14T10:54:05-08:00","dateModified":"2020-12-18T13:23:06-08:00","author":{"@type":"Person","name":"marten"},"image":"/static/stories/web-almanac-2020-page-content/WA_Cover-Image_Part-I-640x853.jpg"}</script>
  <link rel="canonical" href="https://almanac.httparchive.org/en/2020/stories/web-almanac-2020-page-content/">
```

rviscomi:
Oh good catch on the HTML references. Those should be fixed now.

bazzadp (approved):
I fixed the errors and also unminified the code to make it somewhat readable.

Also confirmed the page [passes AMP validation](https://validator.ampproject.org/)

Still lots we can clean up here, but think this is enough to launch it. Though will open another PR to minimise the images

rviscomi:
Search Console is still reporting that the story isn't indexed and not included in any sitemaps:

image

I checked https://almanac.httparchive.org/sitemap.xml and confirmed that the story isn't included. It's also not included in a local build from main.

@bazzadp any idea if there was a regression?

bazzadp:
Fixed in #1860 and released and [can see it in the sitemap now](https://almanac.httparchive.org/sitemap.xml).

Updated sitemap (with 2 more URLs - the stories one and HTTP/2 in Spanish) successfully submitted to Google Search Console so hopefully see it indexed soon.

rviscomi:
Thanks for fixing that so quickly!

bazzadp:
Always more motivated to fix screw ups I caused 😁

bazzadp:
Just checked and can see it in GSC already under a new AMP section:

Google Search Console screenshot

also see it in search results:

Web Almanac Stories search term

Though don’t see it appear organically as a story yet for other search terms. Maybe need more time or more than one story for that to happen.

PR #1817, opened 2020-12-21
Add "edited by" bylines

bazzadp:
The @HTTPArchive/editors put an enormous amount of effort into reviewing chapters and are currently not credited for this work other than being listed in the Editors team in our Contributors page. Being heavily involved in the Editing I believe we should credit each chapter editor in the chapter itself given the contribution they make which is often on par with (or even more than!) the other contributors which are already listed in the byline.

Therefore this PR adds an "Edited by" byline, similar to how the other contributors to a chapter are credited:

Chapter bylines

I realise this makes the byline quite large (especially for translations which also have a "translated by" byline), but I think it is no bad thing to show the effort we put into each chapter to ensure the quality.

I would propose that, like the translations, only the primary editors gets the credit and not everyone who makes small edits after (they can be added to the Editors team, but not be credited with a chapter). Of course if two or more people work together to edit a chapter then we can list more than one editor here.

Currently this PR only puts the framework in place along with one example editor, to make it easy to review, but will add the rest if approved (based on #218 and #1432).


bazzadp:
Added other chapters. Merging this as heard no objections and doing a release.

PR #1829, opened 2020-12-22
Correct Twitter Meta data for Story

bazzadp:
Currently not working as image is relative and not absolute and also missing other data.


bazzadp:
BTW it's still not a great image as Twitter prefers landscape and the suggested one it portrait so it's cropped:

image

@rviscomi can we get a better one from designer? Probably preferable to use a different one than the main Alamanc one you created a few days ago, but that's another option.

rviscomi:
I don't expect these stories to be shared very much at all, their purpose is to drive organic traffic from search/discovery. Still, I agree we could use better graphics if needed. WDYT of this?

Web Almanac home hero

bazzadp:
LGTM! Want to add to this branch?

rviscomi:
Added

rviscomi (approved):
Much nicer with proper code formatting too!

bazzadp:
I'm waiting to see what the other parts look like. Wouldn't take much to template this into a Jinja2 template if there's much reuse between all four. That would also allow translations pretty easily as very little text in these – though given your comment about that might not be worth it.

Suspect however that they will all be subtly different with random classes and divs, even if they are basically the same 😞

PR #1833, opened 2020-12-22
Correct ordering of pre-release script

bazzadp:
Currently we generate hashes of our content (including ebooks), then generate the new ebooks. That's the wrong way around!

Luckily as we generate the ebooks each time it hasn't matter so much, but still - it's wrong so let's fix it.


rviscomi:
Ohhh good catch!

PR #1852, opened 2020-12-28
Make clear that Lighthouse stats are from Mobile only

### LINKED ISSUE #1835: Clarify in all chapters that LH scores are mobile only
**Describe the issue**
We only run LightHouse in mobile mode and all LightHouse scores are specific to mobile but this is not explicitly called out in all charts and chapters. 
Example - LightHouse scores for SEO and Accessibility category in CMS chapter don't mention this - https://almanac.httparchive.org/en/2020/cms#lighthouse-scores

This may not be clear for anybody reading web almanac and we should clarify this in chapters and call out 'mobile' explicitly on any charts referencing LightHouse scores.

bazzadp:
Fixes #1835 - mostly by updating graphs and their images to have mobile in the subtitle (though for SEO chapter I was more explicit as no graphs)

CMS updated graph


rockeynebhwani (approved):
Looks good. Just noticed that image sizes for cache-ttl-lighthouse-score.png and
cache-wasted-bytes-lighthouse.png has reduced. I am assuming this was intentional and to align image sizes with other image sizes.

bazzadp:
Yeah I compressed them with TinyPng before uploading. The original author may not have done that. Will run a sweep as part of editing or later.

PR #1858, opened 2020-12-30
Markup 2020 additional edits

bazzadp:
While completing the [Dutch translation of this chapter](https://almanac.httparchive.org/nl/2020/markup) we noticed there were a few TODOs that were not closed out one way or another (as "fixed" or "won't do").

I've reviewed them all and think the biggest issue is the use of "Occurrences" in tables to sometimes mean pages and sometimes mean elements, which is confusing and therefore leaves open misinterpretation.

So I made some fixes to address the outstanding TODOs, and also changed a few other phrasings I still found difficult to understand on a read through (though it's a million times better than my German or Romanian would be, so please don't take that as an insult!).

I realise this chapter was lucky enough to be the first to be edited, so it's had a lot of scrutiny already, and another round of edits might be the last thing you want to see, but think it's best to close out these remaining issues as do think this is a valuable resource that people will depend on until the 2021 edition comes along - and even thereafter. So if you could just spend one more time ensuring you're comfortable with these edits that would be greatly appreciated!

Progress on #899 #1432


src/content/en/2020/markup.md
```diff
@@ -29,7 +29,7 @@ The web is built on HTML. Without HTML there are no web pages, no web sites, no
...
 
-Last year's analysis was based on 5.8 million pages, of which 4.4 million were tested on desktop and 5.3 million on mobile. This year we analyzed 7.5 million pages, of which 5.6 million were tested on desktop and 6.3 million on mobile, using the [latest data](./methodology#websites) on the websites users are visiting in 2020. We do make some comparisons to last year, but just as we've tried to analyze additional metrics for new insights, we've also tried to impart our own personalities and perspectives throughout the chapter.
+Last year's analysis was based on 5.8 million pages, of which 4.4 million were tested on desktop and 5.3 million on mobile. This year we analyzed 7.5 million pages, of which 5.6 million were tested on desktop and 6.3 million on mobile, using the [latest data](./methodology#websites) on the websites users are visiting in 2020. We do make some comparisons to last year but have also analyzed additional metrics for new insights. We've also tried to impart our own personalities and perspectives throughout the chapter.
```
j9t, line 32: Do you feel strongly about this one? I feel it somehow drags out the paragraph (for me the original reads faster), and I think we’d need to tweak further to avoid the double “also.”

src/content/en/2020/markup.md
```diff
@@ -163,9 +157,9 @@ Around 89% of pages contain at least one HTML comment, while about 46% of them c
 <![endif]-->
 ```
 
-The above is a non-standard HTML conditional comment. While those have proven to be helpful in the past in order to tackle browser differences, they are history for some time as Microsoft [dropped conditional comments](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/compatibility/hh801214(v=vs.85)) in Internet Explorer 10.
+The above is a non-standard HTML conditional comment. While those have proven to be helpful in the past in order to tackle browser differences, they are historical for some time as Microsoft [dropped conditional comments](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/compatibility/hh801214(v=vs.85)) in Internet Explorer 10.
```
j9t, line 160: Just to flag this—I’m still not a native speaker but this reads odd. Is this really how one would say this? If you’re sure and this is a solid improvement, feel free to keep it—otherwise I suggest to leave this as is.

src/content/en/2020/markup.md
```diff
@@ -163,9 +157,9 @@ Around 89% of pages contain at least one HTML comment, while about 46% of them c
...
 
-Still, on the above percentile extremes, we found that web pages are using about 6 conditional comments in the 90th percentile, and 1 comment while in the 10th percentile.  Most of the pages include them for helpers such as html5shiv, selectivizr, and respond.js. While being decentish and still active pages, our conclusion is that many of them were using obsolete CMS themes.
+Still, on the above percentile extremes, we found that web pages are using about 6 conditional comments in the 90th percentile, and 1 conditional comment while in the 10th percentile.  Most of the pages include them for helpers such as [html5shiv](https://github.com/aFarkas/html5shiv), [selectivizr](http://selectivizr.com/), and [respond.js](https://github.com/scottjehl/Respond). While being decentish and still active pages, our conclusion is that many of them were using obsolete CMS themes.
```
j9t, line 162: (Nice edit, thanks for adding all the detail.)

src/content/en/2020/markup.md
```diff
@@ -338,7 +321,7 @@ Nothing changed in the Top 7, but the `option` element went a little out of favo
 
 #### `details` and `summary`
 
-Another thing we were curious about was the use of the [`details` and `summary` elements](https://html.spec.whatwg.org/multipage/rendering.html#the-details-and-summary-elements), especially since 2020 brought [broad support](https://caniuse.com/details). Are they being used? Are they attractive for—even popular—among authors? As it turns out, only 0.39% of all tested pages are using them, although it's hard to gauge whether they were all used the correct way in exactly the situations when you need them, "popular" is the wrong word.
+Another thing we were curious about was the use of the [`details` and `summary` elements](https://html.spec.whatwg.org/multipage/rendering.html#the-details-and-summary-elements), especially since 2020 brought [broad support](https://caniuse.com/details). Are they being used? Are they attractive for—even popular—among authors? As it turns out, only 0.39% of all tested pages are using them—although it's hard to gauge whether they were all used the correct way in exactly the situations when you need them.
```
j9t, line 324: No, that doesn’t work yet—with this edit, “although” doesn’t fit anymore. One could remove that last part (after the em dash), but then it takes away our interpretation. I suggest to keep this as it was.

A side note, “using these elements” could be a useful edit as “using them” starts to get unclear. If you could edit that then I wouldn’t need to track it (I had reviewed the chapter independently and am preparing another—small—PR).

src/content/en/2020/markup.md
```diff
@@ -380,12 +363,12 @@ Accordingly, we looked at the number of `details` and `summary` elements and it
...
     </tbody>
   </table>
-<figcaption>{{ figure_link(caption="Adoption of the <code>details</code> and <code>summary</code> elements.", sheets_gid="1406534257", sql_file="pages_element_count_by_device.sql") }}</figcaption>
+  <figcaption>{{ figure_link(caption="Adoption of the <code>details</code> and <code>summary</code> elements.", sheets_gid="1406534257", sql_file="pages_element_count_by_device.sql") }}</figcaption>
 </figure>
 
 ### Probability of element use
 
-Taking another look at element popularity, how likely is it to find a certain element in the DOM of a page? Surely, `html`, `head`, `body` are present on every page (even though [their tags are all optional](https://meiert.com/en/blog/optional-html/)), making them common elements, but what other elements are to be found?
+Taking another look at element popularity, how likely is it to find a certain element in the DOM of a page? Sure, `html`, `head`, `body` are present on every HTML page (even though [these tags are all optional](https://meiert.com/en/blog/optional-html/), browsers will automatically add them), making them common elements, but what other elements are to be commonly found?
```
j9t, line 371: (Works for me but now I wonder if the sentence gets a bit unwieldy. Even for my taste 😅)

src/content/en/2020/markup.md
```diff
@@ -441,7 +424,7 @@ Taking another look at element popularity, how likely is it to find a certain el
   <figcaption>{{ figure_link(caption="High probabilities of finding a given element in pages of the Web Almanac 2020 sample.", sheets_gid="184700688", sql_file="pages_element_count_by_device_and_element_type_present.sql") }}</figcaption>
 </figure>
 
-Standard elements are those that are or were part of the HTML specification. Which ones are you really rarely to find? In our sample, that would bring up the following:
+Standard elements are those that are or were part of the HTML specification, but which ones are rare to find? In our sample, that would bring up the following:
```
j9t, line 427: I’m not sure this makes this clearer or gets it to read better. Do you feel strongly about this (why)? Suggest to leave this as it was.

src/content/en/2020/markup.md
```diff
@@ -469,21 +452,20 @@ Standard elements are those that are or were part of the HTML specification. Whi
...
-{# TODO(authors, analysts): Clarify occurrences and percentages _of what_. Pages? Elements? #}
-
 <figure>
   <table>
     <thead>
       <tr>
         <th>Element</th>
-        <th>Occurrences</th>
-        <th>Percentage</th>
+        <th>Pages</th>
```
j9t, line 467: Off the top of my head, I don’t think this is what the data are about. The occurrences here are *not* pages—they are the number of elements.

We need to be really careful with these labels—a small change in wording can change the whole meaning, as in this case. I strongly advise against just changing them in passing, and I’m also nervous that we (the authors and analysts) get pulled in again for which is going to be a lot more work than it looks like, for given the nature of these edits, we need to review the data sets and should really double-check what is being changed.

So while I get the intent here to make things more clear, ironically it comes with a significant risk of actually making things less clear, or even wrong. I wouldn’t touch the labels anymore.

src/content/en/2020/markup.md
```diff
@@ -562,15 +544,13 @@ The 2019 edition of the Web Almanac handled [custom elements](../2019/markup#cus
...
 
-{# TODO(authors, analysts): What do "cases" mean here: pages/elements? And for desktop or mobile? #}
-
-Other groups that stand out include [AMP markup](https://amp.dev/) with `amp-` elements like `amp-img` (11,700 cases), `amp-analytics` (10,256) and `amp-auto-ads` (7,621), as well as [Angular](https://angular.io/) `app-` elements like `app-root` (16,314), `app-footer` (6,745), and `app-header` (5,274).
+Other groups that stand out include [AMP markup](https://amp.dev/) with `amp-` elements like `amp-img` (11,700 pages), `amp-analytics` (10,256) and `amp-auto-ads` (7,621), as well as [Angular](https://angular.io/) `app-` elements like `app-root` (16,314), `app-footer` (6,745), and `app-header` (5,274).
 
 ### Obsolete elements
 
-There are more questions to ask about the use of HTML, and one may relate to obsolete elements, which are elements like `applet`, `bgsound`, `blink`, `center`, `font`, `frame`, `isindex`, `marquee`, or `spacer`.
+There are more questions to ask about the use of HTML including the use of obsolete elements, which are elements like `applet`, `bgsound`, `blink`, `center`, `font`, `frame`, `isindex`, `marquee`, or `spacer`.
```
j9t, line 553: I suggest to add a comma after “HTML” now, and to consider putting the last sentence fragments in parentheses, like “…obsolete elements (elements like …)”.

src/content/en/2020/markup.md
```diff
@@ -579,7 +559,7 @@ In our mobile dataset of 6.3 million pages, around 0.9 million pages (14.01%) co
     <thead>
       <tr>
         <th>Element</th>
-        <th>Occurances</th>
+        <th>Pages</th>
```
j9t, line 562: See above (#467)—here, too, this would be wrong. This information does *not* relate to pages.

We should fix “Occurances” though—strangely, however, I can’t spot this misspelled variant in the live version of the chapter.

src/content/en/2020/markup.md
```diff
@@ -784,7 +764,7 @@ This section focuses on how attributes are used in documents and explores patter
 
 ### Top attributes
 
-Similar to the section on the most [popular elements](#top-elements), this section delves into the most popular attributes on the web. Given how important the `href` attribute is for the web itself, or the `alt` attribute in order to make information [accessible](./accessibility), would these be most popular attributes?
+Similar to the section on the most [popular elements](#top-elements), this section delves into the most popular attributes on the web. Given how important the `href` attribute is for the web itself, or the `alt` attribute in order to make information [accessible](./#images-and-their-text-alternatives), would these be most popular attributes?
```
j9t, line 767: I think this was meant to go to the Accessibility chapter—can you update?

src/content/en/2020/markup.md
```diff
@@ -1002,7 +981,7 @@ We had a look at the data and in order to better understand the results, we norm
     <thead>
       <tr>
         <th>Content attribute value</th>
-        <th>Occurrences</th>
+        <th>Pages</th>
```
j9t, line 984: Same general problem as mentioned before and no, “pages” would be inaccurate.

src/content/en/2020/markup.md
```diff
@@ -1059,7 +1038,7 @@ When we built our tests we didn't check for the presence of images, but only loo
     <thead>
       <tr>
         <th>Favicon format</th>
-        <th>Occurrences</th>
+        <th>Pages</th>
```
j9t, line 1041: Same general problem as mentioned before and no, “pages” would be inaccurate.

src/content/en/2020/markup.md
```diff
@@ -1132,14 +1111,13 @@ There has been a lot of [discussion](https://adrianroselli.com/2016/01/links-but
...
 
-{# TODO(analysts): Where do these "occurrences" come from? Ideally we have a single sheet to link to with the results used by this table. #}
 <figure>
   <table>
     <thead>
       <tr>
         <th>Button types</th>
-        <th>Occurences</th>
-        <th>Percentage</th>
+        <th>Pages</th>
```
j9t, line 1119: Same general problem as mentioned before and no, “pages” would be inaccurate.

src/content/en/2020/markup.md
```diff
@@ -57,7 +57,7 @@ In this section, we're covering the higher-level aspects of HTML like document t
       <tr>
         <th>Doctype</th>
         <th>Pages</th>
-        <th>Percentage</th>
+        <th>Pages (%)</th>
```
j9t, line 60: I’ve commented on the general issue of changing the headers/labels in #467; this _seems_ fine, however.

j9t (changes requested):
Thanks for doing this.

We _cannot_ put this live yet as it introduces errors with the table header changes. Not only would this actually falsify the information (notably occurrences ≠ pages), it also comes with a cost of at least two more people needing to review both the data and the changes (precisely because there is a big need to be precise, and a good risk—just being confirmed—that a header/label change is inaccurate). I’d stay away from this—or at least reach out ahead of time to work with two people from the team to see who has time and then to prepare these edits together.

src/content/en/2020/markup.md
bazzadp: I could live with the original. I just don't see the connection between you trying to analysis more metrics and trying to impart your personalities. to be there's a stronger connection between the first two phrases and the third is separate. Unless you're trying to imply last year was boring cause they didn't impart enough of their personalities?

But, as I say, it's a small suggestion and I can live with the original.

```suggestion
Last year's analysis was based on 5.8 million pages, of which 4.4 million were tested on desktop and 5.3 million on mobile. This year we analyzed 7.5 million pages, of which 5.6 million were tested on desktop and 6.3 million on mobile, using the [latest data](./methodology#websites) on the websites users are visiting in 2020. We do make some comparisons to last year, but just as we've tried to analyze additional metrics for new insights, we've also tried to impart our own personalities and perspectives throughout the chapter.
```

src/content/en/2020/markup.md
bazzadp: I struggled with this one myself. Neither reads great IMHO. What about this?:

```suggestion
The above is a non-standard HTML conditional comment. While those have proven to be helpful in the past in order to tackle browser differences, they have been consigned to history for some time as Microsoft [dropped conditional comments](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/compatibility/hh801214(v=vs.85)) in Internet Explorer 10.
```

src/content/en/2020/markup.md
bazzadp: I have to read that first sentence several times to understand it - particularly with the last `"popular is the wrong word." bit.

What about dropping that but reverting the em-dash to a comma?

```suggestion
Another thing we were curious about was the use of the [`details` and `summary` elements](https://html.spec.whatwg.org/multipage/rendering.html#the-details-and-summary-elements), especially since 2020 brought [broad support](https://caniuse.com/details). Are they being used? Are they attractive for—even popular—among authors? As it turns out, only 0.39% of all tested pages are using them, although it's hard to gauge whether they were all used the correct way in exactly the situations when you need them.
```

src/content/en/2020/markup.md
bazzadp: For me "surely" doesn't work here and "sure" is I think what you meant.

I also feel it's also important to realise we are analysing post-processed HTML (as we are using a JavaScript custom metic that works on the DOM, rather than looking at the HMTL returned). So even though they are optional, they will have been added for the purposes  of our analysis. I feel that's an important point to get across.

src/content/en/2020/markup.md
bazzadp: "Which ones are you really rarely to find" just does not read well as a native speaker. Maybe it's the alliteration of the r's.

```suggestion
Standard elements are those that are or were part of the HTML specification, but which ones are you rarely to find? In our sample, that would bring up the following:
```

src/content/en/2020/markup.md
bazzadp: That's the biggest thing I think we need to clarify here! The chapter uses "Occurrences" in one place to mean pages (i.e. multiple occurrences on a page still only count as 1), and in other  places to actually mean "Occurrences" (i.e. multiple occurrences on a page count each time).

Looking at both the data and the query I've attempted to clarify which really are occurrences and which are pages. It's entirely possible I got it wrong, so yes it would be good for @Tiggerito to review, but there is little doubt in my mind that the current chapter is presenting incorrect data.

For example this one has the following in the chapter:

image

Which relates to this data [from Sheets](https://docs.google.com/spreadsheets/d/1Ta7amoUeaL4pILhWzH-SCzMX9PsZeb1x_mwrX2C4eY8/edit#gid=184700688) 

image

Therefore it seems clear that "occurrences" here means pages? But that is not clear to the reader.
As a third check[ the query](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2020/03_Markup/pages_element_count_by_device_and_custom_dash_elements.sql) also says this is pages.

Later on you have this:

image

Now we don't have 10million pages so we know "Occurrences" here does not mean pages.

I feel this double usage of the same word to mean two things is NOT clear and you are presenting incorrect data to the readers. And the fact you and I aren't even sure which it is nails that point for me!

So I feel this is something we MUST correct. As I say I've spent considerable time looking at the data and the queries and _think_ I've got them right, but could well be wrong!

src/content/en/2020/markup.md
bazzadp: SGTM!

```suggestion
There are more questions to ask about the use of HTML, including the use of obsolete elements (which are elements like `applet`, `bgsound`, `blink`, `center`, `font`, `frame`, `isindex`, `marquee`, or `spacer`).
```

src/content/en/2020/markup.md
bazzadp: @Tiggerito can you confirm?

Looking at [this part of the query](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2020/03_Markup/pages_element_count_by_device_and_obsolete_elements.sql):

```sql
  COUNT(0) AS freq,
  AS_PERCENT(COUNT(0), total) AS pct_pages_with_obsolete_elements,
  AS_PERCENT(COUNT(0), SUM(COUNT(0)) OVER (PARTITION BY _TABLE_SUFFIX)) AS ratio_compared_to_all_obsolete_elements
```

If we're dividing by total (which is based on pages) as in the second line then surely `COUNT(0)` must be pages too?

bazzadp: Oh and btw looks like I introduced the misspelling of Occurances in #1857 when converting from Markdown tables to HTML (to allow correct alignment of numeric values). I deliberately did that separately to this PR to keep this PR on the edits.

src/content/en/2020/markup.md
bazzadp: Oops! Good spot!

```suggestion
Similar to the section on the most [popular elements](#top-elements), this section delves into the most popular attributes on the web. Given how important the `href` attribute is for the web itself, or the `alt` attribute in order to make information [accessible](./accessibility#images-and-their-text-alternatives), would these be most popular attributes?
```

src/content/en/2020/markup.md
bazzadp: Actually this one I may be wrong. But then I feel the Pages(%) heading is inaccurate?

image

Though to be honest for this one, as there's usually only one per page it's probably the same for both!

@Tiggerito can you confirm?

src/content/en/2020/markup.md
bazzadp: Think you might be right but then Pages (%) is also inaccurate?

image

src/content/en/2020/markup.md
bazzadp: Ah this one I'm definitely wrong! 

```suggestion
        <th>Occurrences</th>
```

src/content/en/2020/markup.md
bazzadp: Yeah tried to standardise on this when I could since it was used in some places but not others.

src/content/en/2020/markup.md
Tiggerito: One per page and the report includes if it is missing (blank). So I say it's pages and % pages.

src/content/en/2020/markup.md
Tiggerito: Same here, one result per page, so pages and % pages is accurate. I think in this case it will pick the first matching tag on the page.

src/content/en/2020/markup.md
Tiggerito: I agree. I discovered that html, head, body are there even if not in the raw html. So showing their percent is inaccurate. They are probably not 100% due to other errors.

bazzadp: Could be some pages are txt or XML or other things other than HTML. So they won’t have html, head and body even post-processing.

src/content/en/2020/markup.md
Tiggerito: The first one is from the script:

pages_element_count_by_device_and_element_type_present.sql

The word present is meant to indicate I'm just counting if there are any. Percent of pages they are used on.

The second one comes from:

pages_element_count_by_device_and_element_type_frequency.sql

Where the word frequency is meant to indicate a full count. So that would be occurrences.

bazzadp: OK so just to be 100% clear, you agree changing this line (which is about the first example) from “Occurrences” to “Pages”, like I’ve done in this PR, is correct?

The second one (headings table) is further down the page and I just included in this comment to make my point that we’re currently using the same word to mean different things - which is the primary thing I’m trying to correct in this PR (Qing with a few other smaller things). I agree this second “headings” table should stay as “Occurrences” as clearly it’s not pages since there more than the 7.5 million pages we check across mobile and desktop 🙂

src/content/en/2020/markup.md
Tiggerito: Yes, I believe this one is pages based. It's only looking to see if an obsolete element exists on a page, nut how many times it is there.

src/content/en/2020/markup.md
Tiggerito: Correct. I think the first example should say Pages as you did.

j9t: Hey—first, I think this is _awesome_ to look into improving the chapter. What I’d like to raise is, is what we’re improving here worth the effort (involving time of currently 3 people) and the risk (of something slipping through)?

In my mind this is a clear “no”—we have no indication that something is wrong, AFAIK no one complained about the wording, and generally it seems this is nice, yes, but also rather minor.

I respect other views here and I’m also willing to support this, but I am doubtful and concerned.

A side note, I want to underline the suggestion to give people a heads-up where possible, and I also suggest to break up PRs where possible. The header/label topic would feel much better to me had it been separated from the other changes—now those changes will need to wait, and so does another PR I was prepared to get out (I might get it ready regardless now, accepting a light risk of conflicts as this also needs to get out).

src/content/en/2020/markup.md
j9t: LGTM.

src/content/en/2020/markup.md
j9t: Actually, I think the em dash works—in the old setting. What do you think of the old version but with an em dash instead of a comma? (In my view it helps.)

src/content/en/2020/markup.md
j9t: I think the “sure” change is fine, but the additional sentence fragment makes the sentence quite long and, unwieldy. I think the old version works but if we want to add more to the sentence we may need to break it up. (I can make a suggestion later if need be.)

src/content/en/2020/markup.md
j9t: How about keeping this but rephrasing the second sentence then? “Which ones are rare to find?” That may address both our points?

j9t (changes requested):

src/content/en/2020/markup.md
bazzadp:
```suggestion
Another thing we were curious about was the use of the [`details` and `summary` elements](https://html.spec.whatwg.org/multipage/rendering.html#the-details-and-summary-elements), especially since 2020 brought [broad support](https://caniuse.com/details). Are they being used? Are they attractive for—even popular—among authors? As it turns out, only 0.39% of all tested pages are using them—although it's hard to gauge whether they were all used the correct way in exactly the situations when you need them—"popular" is the wrong word.
```

src/content/en/2020/markup.md
bazzadp:
```suggestion
Standard elements are those that are or were part of the HTML specification. Which ones are rare to find? In our sample, that would bring up the following:
```

src/content/en/2020/markup.md
bazzadp: As I stated when opening this PR, the motivation for this was not to endlessly tweak or criticise this chapter. During the first translation it was noticed we still had open TODOs in this chapter. This left me with one of four choices"
1. Ignore them and leave open TODOs in each translation - as they are in the Dutch version. If we ever addressed these in the future then we'd have to revisit every translation. I don't think this was a great option to be honest.
2. Assume all the TODOs were in fact done and close them. The chapter is what it is, and we were obviously happy to publish so the TODOs were not that important. That was an option but without looking at the remaining TODOs - which were obviously not addressed sufficiently to warrant removing them.
3. Ask Authors to investigate them. As far as I'm concerned this had been asked once before.
4. Spend considerable time looking through each TODO, closing out the ones that either had been addressed or were not a big deal (e.g. the document size one), and make suggestions to fix the other ones so they could be closed out. This is the choice I made despite it being the most effort on my part. Perhaps in hindsight this was the wrong choice.

> In my mind this is a clear “no”—we have no indication that something is wrong

We have a clear indication something is wrong! Occurrences in one place means number of elements, and in other means number of pages. You then argued that "occurrences here are not pages—they are the number of elements." but both my and @Tiggerito 's analysis shows that is incorrect — it means pages. I'd prefer each chapter to not knowingly continue to publish incorrect information and have put considerable the effort in to correct it, rather than just raising an issue and asking others to do the work.

> A side note, I want to underline the suggestion to give people a heads-up where possible, and I also suggest to break up PRs where possible. The header/label topic would feel much better to me had it been separated from the other changes—now those changes will need to wait, and so does another PR I was prepared to get out (I might get it ready regardless now, accepting a light risk of conflicts as this also needs to get out).

Yes perhaps this was a mistake of mine. I did break up the table reformats (#1587), as didn't feel those required approval, and to keep this PR small. I thought the wording changes were minor enough to include in this and have them reviewed at same time rather than make you review two PRs. I guess I was wrong about this.

I also like the improvements in your PR, despite there being no heads up and no complaints making them necessary, so happy to take them. I wish we'd closed out the one remaining issue here and merged this before you'd opened that though.

src/content/en/2020/markup.md
bazzadp: I'm tired of arguing. Let's remove the offending clause

```suggestion
Taking another look at element popularity, how likely is it to find a certain element in the DOM of a page? Sure, `html`, `head`, `body` are present on every HTML page (even though [these tags are all optional](https://meiert.com/en/blog/optional-html/)), making them common elements, but what other elements are to be commonly found?
```

bazzadp:
@j9t you happy to merge this? think I addressed all the concerns.

iandevlin (approved):

j9t (approved):
Thanks @bazzadp, also @Tiggerito, @iandevlin.

PR #1866, opened 2021-01-05
Allow HTML entities in author bios

### LINKED ISSUE #1865: colon in the bio blurb results in empty bio
A colon in an author's bio trips up processing.  For example,

```
authorsname_bio: Some: bio
```

Will result in `authorname`'s bio being empty.

bazzadp:
Fixes #1865


dtikhonov (approved):
I tested this and it works.

PR #1880, opened 2021-01-11
Add third web story for the 2020 Web Almanac

swissspidy:
Third time is the charm!

Hope there are less issues this time


swissspidy:
@bazzadp nice image optimization skills! I optimized them locally before pushing, but CI still found massive room for improvement. And here you are, being able to optimize images once again! 💪

bazzadp:
Love TinyPNG.com! Wish our action was as efficient (I have an open issue with them to ask why not).

For the larger images I did half the size in photoshop (6000px -> 3000px) as 13MB was a bit large to commit and TinyPNG.com refuses them when that size. After reduction they were 4MB and then TinyPNG made them 1MB. Much better!

bazzadp (approved):
Notice you removed the pixel in https://github.com/HTTPArchive/almanac.httparchive.org/pull/1880/commits/5520f855ccbd0b0022516741517ede2debb94645 as our linting doesn't allow inline styles.

I handled that differently in the first story, by adding a new class:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/7b0e03616d2975cbf0c7044c189613906e0029d0/src/templates/en/2020/stories/page_content.html#L601

And then using that:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/7b0e03616d2975cbf0c7044c189613906e0029d0/src/templates/en/2020/stories/page_content.html#L2363

Not sure if you want to do the same here? Or happy to just remove the style completely?

swissspidy:
> Not sure if you want to do the same here? Or happy to just remove the style completely?

I already did that in my first commit, but I forgot to remove the inline style as well while doing so.

https://github.com/HTTPArchive/almanac.httparchive.org/blob/c23129273197bf4867bc762ddbc58de8e759f59f/src/templates/en/2020/stories/content_publishing.html#L538

https://github.com/HTTPArchive/almanac.httparchive.org/blob/c23129273197bf4867bc762ddbc58de8e759f59f/src/templates/en/2020/stories/content_publishing.html#L1304

bazzadp:
OK good to merge then?

swissspidy:
No objections from my side :-)

PR #1800, opened 2020-12-18
Web Almanac 2020: Add web story for first chapter

swissspidy:
I used a small scraper script to download the story created in WordPress and adapt it for this site's structure. The markup is unfortunately not easy to read because it's all auto-generated from that story editor tool.

Not sure if it works though, as I couldn't test it. When I got the local site running and tried to access `http://127.0.0.1:8080/en/2020/stories/web-almanac-2020-page-content`, I get an `404 Not Found: Unsupported chapter requested` error.

See #1571

PR #1900, opened 2021-01-18
Update stories 2019 content

bazzadp:
The Web Stories added in #1800 and #1880 had 2019 data for two of the chapters as the 2020 chapters were not written yet.

They have since been written so this updates those stats with 2020 content. Not all the stats were updated in 2020 so sometimes had to choose different stats.

I've basically hacked the HMTL and CSS to get it to work so @swissspidy let us know if there's a better way to do this!

Diffs (old on the left, new on the right):


swissspidy:
Nice work!

> I've basically hacked the HMTL and CSS to get it to work so @swissspidy let us know if there's a better way to do this!

The alternative to this is updating the data in the editor and then preparing the stories again for inclusion here. But that would probably take more time than this and create worse diffs (due to auto-generated class names and things like that)

src/templates/en/2020/stories/page_content.html
```diff
@@ -1897,11 +1897,6 @@
                   </div>
                 </div>
               </div>
-              <div data-amp-original-style="position:absolute;pointer-events:none;left:31.79612%;top:95.79288%;width:70.87379%;height:2.26537%;opacity:1" class="amp-wp-ccac7d2">
-                <div id="el-c9dc151c-4749-4c14-b1c8-9a9fced0d640" data-amp-original-style="pointer-events:initial;width:100%;height:100%;display:block;position:absolute;top:0;left:0" class="amp-wp-a1b70b8">
-                  <p class="fill amp-wp-24c460b" data-amp-original-style="white-space:pre-wrap;overflow-wrap:break-word;word-break:break-word;margin:-0.6164383561643838% 0;font-family:'Lato',sans-serif;font-size:0.194175em;line-height:1.5;text-align:initial;padding:0% 0%;color:#000000"><span data-amp-original-style="color: #f7f779" class="amp-wp-66f5b53">*This chapter currently refers to data from 2019.</span></p>
-                </div>
-              </div>
```
swissspidy, line 1904: Nit: could remove the CSS referring to `amp-wp-ccac7d2`, `amp-wp-a1b70b8`, `amp-wp-24c460b`, and `amp-wp-66f5b53`

swissspidy: Ditto for the other story with the one removed paragraph

bazzadp: Yes for `amp-wp-ccac7d2` and `amp-wp-24c460b`

No for `amp-wp-a1b70b8` and `amp-wp-66f5b53` as still being used.

bazzadp: And did the other one now too.

Issue #1750, opened 2020-12-11
Translate content to Dutch


bazzadp:
These are the core templates - without which we cannot release any translated chapters. They are in the [language specific templates directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/templates):
- [ ] base.html - 
- [ ] base_chapter.html - 
- [ ] error.html - 
- [ ] 2019/base.html - 
- [ ] 2019/contributors.html - 
- [ ] 2019/index.html - 
- [ ] 2019/table_of_contents.html -
- [ ] 2020/base.html - 
- [ ] 2020/contributors.html - 
- [ ] 2020/index.html 2019 - 
- [ ] 2020/table_of_contents.html 

These are the chapters to be translated, in rough order of popularity. They exist in the [content directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content):

**2020**
- [ ] performance - 
- [ ] cms - 
- [ ] capabilities -

**2019**
- [ ] javascript - 
- [ ] css - 
- [ ] seo - 
- [ ] performance -
- [ ] http2 - 
- [ ] markup - 
- [ ] third-parties - 
- [ ] fonts - 
- [ ] accessibility - 
- [ ] pwa - 
- [ ] cms - 
- [ ] security - 
- [ ] media - 
- [ ] mobile-web - 
- [ ] cdn - 
- [ ] ecommerce - 
- [ ] page-weight - 
- [ ] caching - 
- [ ] compression - 
- [ ] resource-hints - 

Additionally the following pages need translated too in the [language specific templates directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/templates):
- [ ] 2019/methodology.html -
- [ ] 2020/methodology.html - 
- [ ] accessibility_statement.html - 
- [ ] base_ebook.htm - 

There is no need to translate the chapters HTML pages are they are generated off the markdown combined with the above templates.

Please include "Makes progress on " in all pull requests so a link is created from the PR to this issue.

---

Common notes for writing consistency are here: https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Translators'-Guide. Feel free to edit that and/or add Dutch-specific extras by editing this comment.

Dutch specific extra advice:

noah-vdv:
So the html files don't need to be translated right? Just the files under 2020, 2019 and additional files?

bazzadp:
The core html files DO need to be translated.

[...]

PR #1908, opened 2021-01-21
Remove TODO's from 2020 SEO and Resource Hints chapters

noah-vdv:
Removing some `TODO`'s from chapters noticed in #1750


bazzadp:
Could I ask one more favour @noah-vdv ? Could you remove line 243 and 244 from the Italian version of resource hints as part of this so line numbers stay in sync? Not sure how I didn't notice they'd added this extra line but suspect it was cause the total line numbers added up.

bazzadp (approved):
Thanks mate!

PR #1918, opened 2021-01-24
Move 2019.css to almanac.css

bazzadp:
Makes progress on #1558 

2019.css is. not just used for 2019 chapters and therefore it's name is confusing. This renames it to almanac.css and also adds a redirect in case anyone or anything tries to reference the old name.


bazzadp:
Linting errors are due to [a bug in GitHub super linter](https://github.com/github/super-linter/issues/1078)

PR #1942, opened 2021-01-30
Add typographic punctuation

### LINKED ISSUE #1485: Use of typographic quotes or "smart quotes" in the Web Almanac
Part of our [editing process](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide#suggested-process) is to convert typographic (or so-called "smart" or "curly") quotes and apostrophe's to straight quotes (“ -> ", ” -> ", ‘ -> ', ’ -> ').

This has caught some authors by surprise so I want to make authors more aware of this, explain the reasoning, and also possibly open this up for discussion if we, as a project, want to change this. However I should caveat that last part, by saying that I do not believe now is the right time to make this change - we are all very busy trying to reach the finish line with less than a month to complete our chapters for the launch so do not believe we should be making this change before then for the reasons I'm about to give.

There are good reasons to use smart quotes (primarily, they look nicer), but there are also many good reasons not to use smart quotes. The Web Almanac uses the Google Developer Style Guide for the Web Almanac and [it recommends straight quote marks](https://developers.google.com/style/quotation-marks#straight-and-curly-quotation-marks) and gives reasons for this:

> __Straight and curly quotation marks__
>
> Most typefaces support two forms of quotation marks and apostrophes: straight marks and curly, or typographic, marks. Some tools, like Google Docs, automatically convert straight quotation marks and apostrophes to the curly versions as you type. However, our guidance is to always use straight quotation marks and straight apostrophes, for these reasons:
> 
> - It's easy to get the direction of curly quotation marks (especially apostrophes) wrong. Using straight marks avoids this problem.
> - Code requires straight marks.
> - Many writers use coding tools (for example, Subversion) for working directly with markup. Because these tools do not produce curly marks by default, the writer must manually insert the marks, which is error prone.

I'm not sure if you've even noticed this, but if look at many of Google's technical posts (e.g. [those on web.dev](https://web.dev/speed-tooling-evolutions-cds-2019/#field-data-(crux)-thresholds-adjusted-in-pagespeed-insights)) they use straight quotes. Similarly I'm not sure how many of you noticed the convention we use, despite having (presumably) read the 2019 edition.

Some more reasons to use this to supplement those:
- GitHub (where most of this project is managed) doesn't make the differences the most obvious
- It's MUCH easier to edit for consistency – just do 4 search and replaces (“ -> ", ” -> ", ‘ -> ', ’ -> ') to convert to straight quotes, but much, much harder to go the other way around. Especially when we do include code in our markdown that cannot handle quotes.
- Using straight quotes may not look as nice, but is not wrong. However, using the wrong curly quote (e.g. opening quotes, when you meant to use closing quotes) is wrong. Similarly switching between smart and plain quotes in within a chapter, or even within the Almanac is inconsistent and so also the wrong thing to do.
- Not all authors may use (or even be aware!) of these smart quotes which makes the editing process much harder. We want authors to concentrate on their writing and not items like this. At the same time, putting this burden on editors (who often get squeezed right at the end as timelines slip) is a large effort for them.
- The chapter must be written in Markdown which is much better handled in a code editor than a word processor. Typing these quotes in GitHub and code editors (which do not auto-convert) is more difficult and time consuming.
- While we advise to start writing in Google Docs (which automatically converts to the appropriate smart quote), not everyone does, and even when they do, substantial changes and edits happen in GitHub through pull requests and the like.
- Many, many, many people touch each chapter. From Authors, Reviewers, Editors, Translators (in many languages) to those random typo corrections we receive (and encourage).
- If we want to change this, we would need to revisit all the 2019 chapters to maintain consistency. A quick search shows 94 lines with double quotes, and 440 lines with single quotes in the English chapters alone, but due to the way we include paragraphs on a single line there will likely be multiple instances per line (at least opening and closing quotes for the double quotes anyway!). That's a huge effort.
- The final point may be controversial to the typography people in our team, but IMHO as a native English speaker, this really isn't that important in English (we'll get to other languages in a bit). 

Those are all good reasons for why we do what we do.

However, to present the arguments for the other side:
- It just looks nicer for those that do notice!
- We are primarily a written publication, rather than a coding website. We should strive for the best writing – including typography.
- We are an annual publication, so there are a limited chapters to manage here.
- We strive for excellence in just about every other way, from the experts we use, the data we scrutinise, the editing we do, the fast/accessible/quality website we build. So why not use the best quotes?
- We have typography experts involved who clearly care about this!
- We have allowed smart quotes in the translations. This is for several reasons: it's my opinion these are more important in other languages which make much more use of apostrophe's, accents and other typography, we are under much less time pressure for these, and a lot less hands touch these chapters (typical translation teams are 1,2 or even 3 people) who can all agree and monitor their own use of this.

I hope that explains a bit more background about this situation, it's pros and cons, and why we do what we do. I'm not 100% sure we've made the right decision here, and can see the arguments for both sides. However, I am also aware of the considerable effort to support this properly. I don't like to use my maintainer card often, but will raise it on this occasion, because we maintainers may feel the consequences of the issues expressed above well beyond the short-term pain of getting the English chapters live each year.

I'd be interested to hear other's opinions on this subject.

bazzadp:
Fixes #1485 

This converts simple quotes (" and ') to typographic quotes (“,”, and ‘’), sometimes know as curly quotes or smart quotes.

It does this using the [smartypants](https://www.npmjs.com/package/smartypants) library. I tried a few of these, and this one was the only one I could get working consistently as it ignores code formatting and comments.

Still needs a little fudging for our Jinja2 macros (figure markup) but actually quite clean.

I also fixed a lot of captions and descriptions to allow them to be typographic quotes too.

Currently it only handles English punctuation so translations for languages with different punctuation marks still need to convert.

### DIFF
--- a/src/content/en/2019/caching.md
+++ b/src/content/en/2019/caching.md
@@ -707,7 +707,7 @@ Lighthouse computes a score for each audit, ranging from 0% to 100%, and those s
 
 {{ figure_markup(
   image="fig21.png",
-  caption="Distribution of Lighthouse scores for the \"Uses Long Cache TTL\" audit for mobile web pages.",
+  caption='Distribution of Lighthouse scores for the "Uses Long Cache TTL" audit for mobile web pages.',
   description="A stacked bar chart 38.2% of websites get a score of < 10%, 29.0% of websites get a score between 10% and 39%, 18.7% of websites get a score of 40%-79%, 10.7% of websites get a score of 80% - 99%, and 3.4% of websites get a score of 100%.",
   chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vT3GWCs19Wq0mu0zgIlKRc8zcXgmVEk2xFHuzZACiWVtqOv8FO5gfHwBxa0mhU6O9TBY8ODdN4Zjd_O/pubchart?oid=827424070&format=interactive"
   )
--- a/src/content/en/2019/compression.md
+++ b/src/content/en/2019/compression.md
@@ -336,7 +336,7 @@ Because the [HTTP Archive runs Lighthouse audits](./methodology#lighthouse) for 
 
 {{ figure_markup(
   image="fig11.png",
-  caption="Lighthouse \"enable text compression\" audit scores.",
+  caption='Lighthouse "enable text compression" audit scores.',
   description="Stacked bar chart showing 7.6% are costing less than 10%, 13.2% of sites are scoring between 10-39%, 13.7% of sites scoring between 40-79%, 2.9% of sites scoring between 80-99%, and 62.5% of sites have a pass with over 100% of text assets being compressed.",
   chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vQNIyMEGYE_1W0OdFYLIKsxg6M3o_ZsTTuaX73Zzv6Alw4x4D6oH0jdg9BSgw-jy4E-MmX_Qaf-B98W/pubchart?oid=2048155673&format=interactive",
   width=760,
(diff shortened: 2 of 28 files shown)


bazzadp:
OK not surprising that template changes are too big with this, so going to ignore that failed test as that's the reason it failed.

PR #2031, opened 2021-02-24
Performance improvements to address layout shifts

bazzadp:
A bit of a refactor of the code to address some Core Web Vitals issues flagged in Google Search Console:

Core Web Vitals Screenshot from Google Search Console

It's complaining about Cumulative Layout Shift which is caused by a number of issues:

**Header shifts**
- The header loads without JavaScript initially and the fallback link buttons were a slightly different size than the actual buttons. So changed padding to bring them into line.
- Inlined a bit of JS after header loads to activate the buttons to further reduce this shift.

**Fonts**
- Fonts loading are causing a shift so added preload hints for the critical ones and [switch from `swap` to `block`](https://twitter.com/tunetheweb/status/1364278446311043073). With preload, load times don't really increase and there is less of a jarring shift as fonts fall back into place.
- Preload CSS before fonts as otherwise preloaded fonts incorrectly get highest priority [due to a bug in Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=788757).
- Add prefers-reduced-data checks to avoid loading fonts in this case (also to preload statements)
- Remove use of local fonts, [inline with Google Fonts change](https://twitter.com/googlefonts/status/1323363090969333762). This may seem counter intuitive but they hid the layout shifts on my machine or might have spotted them sooner. Plus with preload and prefers-reduced-data I think it's OK.

**JavaScript**
- The inlined JS was getting far to unwieldy so moved most of it to `chapter.js` and renamed it `almanac.js`. This should now be cached and compiled once, as users browser round the site rather than having to be recompiled on each page. It also consolidates our JS code in one place so easier to spot issues or optimisation opportunities
- Stop upgrading interactive figures if there are no interactive figures
- Moved Google Analytics script to the end of document
- Fixed an issue with the core web vitals script which meant it was not tracking these metrics in Google Analytics properly.

**CSS**
- Remove print stylesheet as it was tiny anyway, and just add to almanac.css, to avoid one more download


bazzadp:
Differences:

[2020 CSS chapter](https://www.webpagetest.org/video/compare.php?tests=210224_DiFZ_f04fbe8bac5fc00c3874826dfc2382bd%2C210223_DiMY_ba7386920761208d513af3c3d25067a5&highlightCLS=1&thumbSize=200&ival=500&end=visual) - had a big, almost full screen CLS and then later one, but that is much reduced:

CSS chapter webpagetest film strip comparison

CSS chapter webpagetest loading chart

[Home Page](https://www.webpagetest.org/video/compare.php?tests=210224_DiT3_1ff569b8d0b5073535628a7fd79ba8d2%2C210223_DiZ2_1abdd576362bb6cb7e5a315457c2e344&highlightCLS=1&thumbSize=200&ival=100&end=visual) had a big, almost full screen CLS and that's been eliminated completely and it now loads twice as fast:

Home page web page test film strip comparison

Home page loading graph

Issue #1044, opened 2020-07-16
Translate content into Turkish


bazzadp:
These are the core templates - without which we cannot release any translated chapters. They are in the [language specific templates directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/templates):
- [ ] base.html - 
- [ ] base_chapter.html - 
- [ ] error.html - 

**2020**
- [ ] base.html - 
- [ ] contributors.html - 
- [ ] index.html  - 
- [ ] table_of_contents.html - 

**2019**
- [ ] base.html - 
- [ ] contributors.html - 
- [ ] index.html  - 
- [ ] table_of_contents.html - 

These are the chapters to be translated, in rough order of popularity. They exist in the [content directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content):

**2020**
- [ ] css - 
- [ ] javascript - 
- [ ] seo - 
- [ ] markup - 
- [ ] accessibility -
- [ ] cms - 
- [ ] performance - 
- [ ] security - 
- [ ] jamstack - 
- [ ] fonts - 
- [ ] mobile-web - 
- [ ] pwa - 
- [ ] http2 - 
- [ ] capabilities - 
- [ ] third-parties - 
- [ ] page-weight - 
- [ ] caching - 
- [ ] resource-hints - 
- [ ] privacy - 
- [ ] compression - 
- [ ] ecommerce - 
- [ ] media - 

**2019**
- [ ] javascript - 
- [ ] css - 
- [ ] seo - 
- [ ] performance - 
- [ ] http2 - 
- [ ] markup - 
- [ ] third-parties -
- [ ] fonts - 
- [ ] accessibility -
- [ ] pwa - 
- [ ] cms - 
- [ ] security - 
- [ ] media - 
- [ ] mobile-web - 
- [ ] cdn - 
- [ ] ecommerce - 
- [ ] page-weight - 
- [ ] caching - 
- [ ] compression - 
- [ ] resource-hints - 

Additionally the following pages need translated too in the [language specific templates directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/templates), but less important than core templates and chapters:
- [ ] accessibility_statement.html - 
- [ ] base_ebook.html- 
- [ ] 2019/methodology.html - 
- [ ] 2020/methodology.html - 

There is no need to translate the chapters HTML pages are they are generated off the markdown combined with the above templates.

[...]

PR #2104, opened 2021-03-29
Create Template Files for Turkish Translation

fatmabadri:
Makes progress on #1044

PR #2108, opened 2021-04-01
Turkish Translation-Template Files of 2019 and 2020 

fatmabadri:
Makes progress on #1044


fatmabadri:
Hi @bazzadp, 
The templates files of 2019/base.html, 2019/index.html, 2020/base.html  and 2020/index.html were translated. I have worked in the same branch which is the same branch of first PR. I also see same commits within this PR. I am not sure if there might be any issue.

bazzadp:
I've moved these files into #2104's branch as this branch appears to be a bit behind that.

PR #2127, opened 2021-04-08
Fix kbd styling

bazzadp:
Spotted this:

kbd element with bold style

Changing it to this inline with the rest of our code style:

kbd element with code style


bazzadp:
Oops - branched from production, instead of main. Will redo.

PR #2180, opened 2021-04-30
Update chapter tracker to 2021

### LINKED ISSUE #2171: Update the chapter progress tracker for 2021
Similar to #989, let's update the progress tracker GH action to parse the 2021 chapter issues and generate a live view of all chapters' milestones and contributors.

cc @ibnesayeed

bazzadp:
Fixes #2171

### DIFF
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -177,7 +177,7 @@ jobs:
               chapters.push({
                 id: issue.number,
                 url: issue.html_url,
-                title: heading.title || issue.title.replace(/\b20[0-9][0-9]\b/,'').trim(),
+                title: heading.title || issue.title.replace(/\b20[0-9][0-9]\b/,'').replace(/  /,' ').trim(),
                 part: heading.part,
                 number: heading.number,
                 links: parseLinks(issue.body),
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -149,7 +149,7 @@ jobs:
             }
 
             const formatTeam = team => {
-              return Object.entries(team).map(([k, v]) => `<span title="${k}: ${v.length ? v.join(', ') : 'TBD'}">${v.length}</span>`).join('+').replace(/\+([^+]*)$/, ' ≥ $1')
+              return Object.entries(team).map(([k, v]) => `<span title="${k}: ${v.length ? v.join(', ') : 'TBD'}">${v.length}</span>`).join('|').replace(/\|([^|]*)$/, ' ($1)')
             }
 
             const formatRow = chapter => {
@@ -177,7 +177,7 @@ jobs:
               chapters.push({
                 id: issue.number,
                 url: issue.html_url,
-                title: heading.title || issue.title,
+                title: heading.title || issue.title.replace(/ 20[0-9][0-9]/,''),
                 part: heading.part,
                 number: heading.number,
                 links: parseLinks(issue.body),
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -149,7 +149,7 @@ jobs:
             }
 
             const formatTeam = team => {
-              return Object.entries(team).map(([k, v]) => `<span title="${k}: ${v.length ? v.join(', ') : 'TBD'}">${v.length}</span>`).join('|').replace(/\|([^|]*)$/, ' ($1)')
+              return Object.entries(team).map(([k, v]) => `<span title="${k}: ${v.length ? v.join(', ') : 'TBD'}">${v.length}</span>`).join(',').replace(/\,([^,]*)$/, ' ($1)')
             }
 
             const formatRow = chapter => {
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -149,7 +149,7 @@ jobs:
             }
 
             const formatTeam = team => {
-              return Object.entries(team).map(([k, v]) => `<span title="${k}: ${v.length ? v.join(', ') : 'TBD'}">${v.length}</span>`).join(',').replace(/\,([^,]*)$/, ' ($1)')
+              return Object.entries(team).map(([k, v]) => `<span title="${k}: ${v.length ? v.join(', ') : 'TBD'}">${v.length}</span>`).join('+').replace(/\+([^+]*)$/, ' ($1)')
             }
 
             const formatRow = chapter => {
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -149,7 +149,7 @@ jobs:
             }
 
             const formatTeam = team => {
-              return Object.entries(team).map(([k, v]) => `<span title="${k}: ${v.length ? v.join(', ') : 'TBD'}">${v.length}</span>`).join('+').replace(/\+([^+]*)$/, ' ($1)').replace(/\+/,' ')
+              return Object.entries(team).map(([k, v]) => `<span title="${k}: ${v.length ? v.join(', ') : 'TBD'}">${v.length}</span>`).join('+').replace(/\+([^+]*)$/, ' ($1)').replaceAll(/\+/,' ')
             }
 
             const formatRow = chapter => {
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -149,7 +149,7 @@ jobs:
             }
 
             const formatTeam = team => {
-              return Object.entries(team).map(([k, v]) => `<span title="${k}: ${v.length ? v.join(', ') : 'TBD'}">${v.length}</span>`).join(' ').replace(/\+([^+]*)$/, ' ≥ $1')
+              return Object.entries(team).map(([k, v]) => `<span title="${k}: ${v.length ? v.join(', ') : 'TBD'}">${v.length}</span>`).join(' ').replace(/\ ([^ ]*)$/, ' ≥ $1')
             }
 
             const formatRow = chapter => {
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -129,7 +129,7 @@ jobs:
             const parseTeam = text => {
               const team = {}
               text.split('\n').forEach(line => {
-                if (/\|\s*\[Sheet\]\[~results-sheet\]\s*\|/.test(line)) {
+                if (/\|\s*\[Sheet\]\[~google-sheets\]\s*\|/.test(line)) {
                   const [, authors, reviewers, analysts] = line.split('|')
                   team.Authors = parseUserNames(authors)
                   team.Reviewers = parseUserNames(reviewers)
--- a/.github/workflows/progress-tracker.yml
+++ b/.github/workflows/progress-tracker.yml
@@ -132,7 +132,7 @@ jobs:
             const parseTeam = text => {
               const team = {}
               text.split('\n').forEach(line => {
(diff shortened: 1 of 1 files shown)


rviscomi:
Great to see this taking shape again! The progress tracker is extremely useful to keep tabs on how everything is moving. Some assorted feedback for your consideration:

- The milestone numbers in the table go 1-6 but 0-5 in the Milestones section. Suggest using 0-5 in the table.
- I've found the 0+0+1≥1 syntax confusing and I'd suggest simplifying it to "X Y Z" format (space delimiter)
- We should also track editor assignments
- The "2021" in each chapter name is redundant for this tracker's purposes. Could we parse that out, for example just "JavaScript"?

ibnesayeed:
Thanks for taking care of it @bazzadp. The other day I created the draft of the tracker issue with the updates set of tasks, but got busy in other work, and postponed it for the weekend. :)

Now, I will review code changes!

ibnesayeed:
I think, may be in a separate PR, we should consider moving this code to [a separate JS script file](https://github.com/actions/github-script#run-a-separate-file) for better linting and IDE assistance.

bazzadp:
OK need to give it up for the evening as family duties are calling!

But got it working. Looks like only 3 chapters have no assignees so far @rviscomi 

> The milestone numbers in the table go 1-6 but 0-5 in the Milestones section. Suggest using 0-5 in the table.
> We should also track editor assignments
Fixed!

> I've found the 0+0+1≥1 syntax confusing and I'd suggest simplifying it to "X Y Z" format (space delimiter)
> The "2021" in each chapter name is redundant for this tracker's purposes. Could we parse that out, for example just "JavaScript"?
Not fixed. Do you want to take these @ibnesayeed ?

I'm turning off the computer for tonight so all yours to review and improve. Feel free to update this branch as you like.

> I think, may be in a separate PR, we should consider moving this code to [a separate JS script file](https://github.com/actions/github-script#run-a-separate-file) for better linting and IDE assistance.

That a comment on all my needless commits ;-)

ibnesayeed:
> I've found the 0+0+1≥1 syntax confusing and I'd suggest simplifying it to "X Y Z" format (space delimiter) -- @rviscomi

I think, that syntax was well-debated and a more accurate representation of the situation. Space separated numbers would work, but the total, that can be less than or equal to the sum (due to multiple roles of the same person in a single chapter) would be awkward to represent without being too verbose.

ibnesayeed:
> > The "2021" in each chapter name is redundant for this tracker's purposes. Could we parse that out, for example just "JavaScript"? -- @rviscomi 
>
> Not fixed. Do you want to take these @ibnesayeed ? -- @bazzadp 

I can take care of it, but do we really want to fix it via code? Last year, chapter issue titles had the year suffix, but chapter titles themselves (first header line of the body) did not have it, which is where the value is picked up from.

bazzadp:
> > I've found the 0+0+1≥1 syntax confusing and I'd suggest simplifying it to "X Y Z" format (space delimiter) -- @rviscomi
> 
> I think, this syntax was well-debated and is a more accurate representation of the situation. Space separated numbers would work, but the total, that can be less than or equal to the sum (due to multiple roles of the same person in a single chapter) would be awkward to represent without being too verbose.

I know what you're trying to do here with this format (the sum of all the team members may be greater than the total total as people do more than one role) however I agree with Rick that that is a little *too* clever and confusing for people who are wondering what this weird formula is all about!

I've replaced it with `0 0 2 0 0 (2)` format where the last number in parenthesis is the team.

Note I had to leave it as `+` initially and then do a `.replace(/\+/g, '')` as otherwise it broken things when tryiong to format the team numbers with parenthesis.

> I can take care of it, but do we really want to fix it via code? Last year, chapter issue titles had the year suffix, but chapter titles themselves (first header line of the body) did not have it, which is where the value is picked up from.

We want the GitHub issue title to include the year to allow easier searching, but I agree it doesn't need to be first line of the issue body.

However, at present the first line of the header is not in the format the action expects (`Part *X* Chapter *Y*: *Chapter Name*`) and so it ignores that and falls back to the GitHub issue title which, as stated above, should have the year.

Eventually we should change to the first line - not least because it would allow adding the chapter number to the issue tracker and allow ordering. However there is likely to be some flux in the chapters as some unfortunately drop out and new ones come in so don't think we should waste our time trying to keep these up to date until they settle down when we have each chapter staffed and confirmed.

So for now the GitHub title is what will be used in the issue tracker and, as mentioned above, that should include the year. So I've added a `.replace(/ 20[0-9][0-9]/, '')` to the issue title so the year is removed.

So I think we're good to merge this now so will upgrade it from draft status if you could code review the actual changes now @ibnesayeed ?

bazzadp:
P.S. I LOVE the more info on the titles available as tool tips - so handy!

.github/workflows/progress-tracker.yml
```diff
@@ -173,7 +177,7 @@ jobs:
               chapters.push({
                 id: issue.number,
                 url: issue.html_url,
-                title: heading.title || issue.title,
+                title: heading.title || issue.title.replace(/ 20[0-9][0-9]/,''),
```
ibnesayeed, line 180: Sticking the pattern to the end to make sure no other 4-digit numbers are replaced (unlikely false negatives).

```suggestion
                title: heading.title || issue.title.replace(/\s+\d{4}$/,''),
```

ibnesayeed:
> I've replaced it with `0 0 2 0 0 (2)` format where the last number in parenthesis is the team.

I still think that space is not a good separator here, because the table cell is not formatted as inline code block, so the space is not using fixed width, as a result, it would be visually difficult to read individual numbers when some of them involve more than one digits. Besides, I do not think spaces will be giving any better clue over plus signs for those unfamiliar with what these numbers represent. That said, I am okay either way, as long as it helps those who manage. :)

.github/workflows/progress-tracker.yml
```diff
@@ -203,8 +207,8 @@ jobs:
             })
 
             const report = [
-              `Chapters | <span title="${Object.entries(icons).map(([k, v]) => `${v} (${k})`).join(', ')}">Links</span> | <span title="${Object.keys(teamMembers).join('+').replace(/\+([^+]*)$/, '≥$1')}">Team</span> | ${trackedTasks.map((v, i) => `<span title="${v}">${i + 1}<span>`).join(' | ')}`,
-              `:--------|:------|:----:${'|:-:'.repeat(totalTrackedTasks)}`
+              `Chapters | <span title="${Object.entries(icons).map(([k, v]) => `${v} (${k})`).join(', ')}">Links</span> | <span title="${Object.keys(teamMembers).join('+').replace(/\+([^+]*)$/, '≥$1')}">Team</span> | ${trackedTasks.map((v, i) => `<span title="${v}">${i}<span>`).join(' | ')}`,
```
ibnesayeed, line 210: Whatever formatting for the number of team members we finalize, the following column header should correspond to it:

```
<span title="${Object.keys(teamMembers).join('+').replace(/\+([^+]*)$/, '≥$1')}">Team</span> 
```

.github/workflows/progress-tracker.yml
bazzadp: What about this instead?:

```suggestion
                title: heading.title || issue.title.replace(/\b20[0-9][0-9]\b/,''),
```

ibnesayeed: I think we should be very specific about the trailing year only. If we want to replace year anywhere in the title then there can be better ways. The change you proposed has a couple of issues: 1) if there are more than one occurrences of years in the title, it will replace the first one, and 2) a word boundary does not include the space character itself, so the replacement may leave spurious white space (not a big deal though).

```
> 'Review of 2020 HTTP/2 Chapter 2021'.replace(/\b20[0-9][0-9]\b/, '')
'Review of  HTTP/2 Chapter 2021'
```

```
> 'Review of 2020 HTTP/2 2021'.replace(/\s+20\d{2}$/, '')
'Review of 2020 HTTP/2'
```

bazzadp: Switched to the `\b` format and `.replace(/  /,' ').trim()` so it covers your scenarios and `2021 JavaScript` or `JavaScript 2021`.

I think we're really getting into the weeds here if we do any more here though! While it's good to future proof and make it reasonably robust, it's only for this chapter tracker which is an internal tool for us, we already have the format for this year's chapters and had to make changes to support 2021 so likely will have to make more changes next year.

I think the current code sufficiently covers us without making it needless overly complex.

.github/workflows/progress-tracker.yml
bazzadp: Updated.

bazzadp:
> > I've replaced it with `0 0 2 0 0 (2)` format where the last number in parenthesis is the team.
> 
> I still think that space is not a good separator here, because the table cell is not formatted as inline code block, so the space is not using fixed width, as a result, it would be visually difficult to read individual numbers when some of them involve more than one digits. Besides, I do not think spaces will be giving any better clue over plus signs for those unfamiliar with what these numbers represent. That said, I am okay either way, as long as it helps those who manage. :)

What's your preference now you've seen both @rviscomi ? I'm easy with either `+` or space (I think the `+` has the same issues you raised for space there tbh @ibnesayeed !) but do think making it look less like a formula by replacing the `>=` with parenthesis is an improvement.

But I also think we're bikeshedding here, and think we should make a decision here, merge and move on. @rviscomi as you'll be the one using this most to PM the project will leave you to have the final decision here.

.github/workflows/progress-tracker.yml
ibnesayeed: I do not like unnecessary chaining of replacements followed by a call to trim. If the intent is to remove a trailing 4-digit number, then using `.replace(/\s+\d{4}$/,'')` is all we need. If we want to limit it to 4-digit numbers that start with `20` then `.replace(/\s+20\d{2}$/, '')` will do the trick. If we want things to be even more tight and replace only the current year then we can use `new Date().getFullYear()` as the find string.

bazzadp: None of those cover the use case of `2022 JavaScript Chapter`, which is not unlikely to be used in the future.

Also, given the frequent use of chains elsewhere in this script, often it really long, complex lines of code, it seems a little odd to complain about this relatively simple one.

ibnesayeed: I did not realize that we are also anticipating titles with leading year, not just trailing ones. :)

We can go for something like `.replace(new Date().getFullYear(), '').trim()`

```
> t1 = '2021 JavaScript Chapter'
> t2 = 'JavaScript Chapter 2021'
> t1.replace(new Date().getFullYear(), '').trim()
"JavaScript Chapter"
> t2.replace(new Date().getFullYear(), '').trim()
"JavaScript Chapter"
```

bazzadp: Does not handle when chapters slip into the following year (like two of them did this year), nor dates in the middle of the title which will have double spaces now - which you, yourself raised!

ibnesayeed: I considered those two situations, but found them to be unlikely because 2) by the time things will slip into the next year, we will already be settled with the section and chapter numbers and the issue title will not be used at that time, and 2) I do not anticipate current year to be anything but as a suffix or prefix to the title.

rviscomi:
> What's your preference now you've seen both @rviscomi ?

Spaces and parentheses LGTM

bazzadp:
Ok gonna merge as is. Think we could debate the year thing until this time next year and neither of us would still be right 😁

@rviscomi once the chapter settle we should edit the first line in each chapter’s issue to the appropriate formatting (including chapter number) so the chapters are ordered correctly in the tracker. Until then they’ll be ordered by issue number.

PR #2181, opened 2021-05-01
Translation of the 2020 Caching chapter to Dutch

noah-vdv:
#1750 
One of the longer chapters


bazzadp (approved):
LGTM! Thanks for fixing the other chapters too!

bazzadp:
> One of the longer chapters

Unfortunately the final two are the longest! But end is in sight...

Issue #1617, opened 2020-12-03
Fixed position chapter CTAs


rviscomi:
At the bottom of each chapter we have three calls to action (CTAs) that we suggest to readers if they'd like to continue learning about the topic:

- discuss it in the comments
- view the raw results spreadsheet
- view the underlying queries

![image](https://user-images.githubusercontent.com/1120896/100968882-2ca1b780-3500-11eb-956f-90b86d90eb77.png)

I'd like to see a design that affixes these buttons to the bottom right corner of the screen so that readers can more easily discover them. I could be dissuaded from showing the results/queries at all times, but I think showing the comment button would be great to get more engagement amongst readers and authors.

I would also propose two other CTAs:

- edit this page
- translate this page

If a reader spots a typo or bad data, they should be one click away from having a GitHub issue pre-populated with the boilerplate info. I've created [issue templates](https://github.com/HTTPArchive/almanac.httparchive.org/issues/new?assignees=&labels=bug%2C+writing&template=2020-content-issue.md&title=Issue+with+the+2020+%5BCHAPTER%5D+chapter) to help make this even more streamlined, and we can even dynamically assign issues to the relevant authors/editors/translators by appending more metadata to that URL.

If a reader is viewing a non-English version of the site but a chapter is only available in English, there should also be a CTA for that person to volunteer to translate it. It would be nice to be one click away from editing the chapter's markdown file, but the button should probably take readers to #923 or someplace where it's easy to start contributing.

For design ideas see [this article](https://web.dev/chrome-ux-report-api/) (mobile) on web.dev:

![image](https://user-images.githubusercontent.com/1120896/100968733-e51b2b80-34ff-11eb-8ef9-3b179ed48e7b.png)

Desktop:

![image](https://user-images.githubusercontent.com/1120896/100968811-0ed45280-3500-11eb-8d5c-159027697fd1.png)

Backlog item (not launch blocking)

AbbyTsai:
Here’s my thoughts in these CTAs and would love to hear yours.

Purpose: Call to audience’s actions.

Proposed actions:
1.Explore raw results and query: position in the end of article. 
   CTA1.Review result,  CTA2.Review queries.  
2.Improve chapters: fixed position at bottom in the right corner.
   CTA3.Comments (via httparchive forum),  CTA4.Edit (via github). 
3.Increase language: fixed position at bottom in the right corner.
  CTA5.Join translator (via github).
   
After: (desktop)
[...]

PR #2196, opened 2021-05-08
Fixed position chapter CTAs

AbbyTsai:
progress on Fixed position chapter CTAs #1617


src/static/css/almanac.css
```diff
@@ -387,6 +386,38 @@ h2.header {
...
+  right: 0.125em;
+}
+
+.alt.btn.cta:nth-child(4) {
+  bottom: 0.6em;
+  right: 0.125em;
+}
+
+.alt.btn.cta > span:nth-child(2) {
+  margin-left: 0.35em;
```
bazzadp, line 420: This all seems a little verbose.

Can't we just use `.cta` for most of these, and maybe give real class names to the three ctas?

Also not sure about the name `cta`. Bit of a generic name. I wonder should we call it `chapter-cta`?

```suggestion
.btn.chapter-cta > svg {
  margin-right: 0px;
}

.chapter-cta {
  position: fixed;
  padding: 10px 15px;
  opacity: 0.9;
  text-decoration: none;
  z-index: 1;
}

.comment-cta {
  bottom: 8.5em;
  right: 0.125em;
}

.suggest-edit-cta {
  bottom: 4.5em;
  right: 0.125em;
}

.help-translate-cta {
  bottom: 0.6em;
  right: 0.125em;
}

.chapter-cta.> span:nth-child(2) {
  margin-left: 0.35em;
```

Still mulling over what we should do with that last one...

src/static/css/almanac.css
```diff
@@ -810,7 +842,8 @@ p.copyright a {
 
   header nav:first-of-type,
   footer nav:first-of-type,
-  header .cta {
+  header .cta,
+  .alt.btn.cta > span {
```
bazzadp, line 846:
```suggestion
  .chapter-cta > span {
```

src/static/css/almanac.css
```diff
@@ -962,7 +995,8 @@ p.copyright a {
...
```
bazzadp, line 999:
```suggestion
  .chapter-cta > span {
```

src/templates/base/base_chapter.html
```diff
@@ -65,14 +65,26 @@ <h2 id="explore-results">
       {{ self.explore_the_results() }}
     </a>
   </h2>
-  <a class="alt btn" href="https://discuss.httparchive.org/t/{{ metadata.get('discuss') }}">
-    <svg width="18" height="18" role="img">
+  <a class="alt btn cta" href="https://discuss.httparchive.org/t/{{ metadata.get('discuss') }}">
```
bazzadp, line 68:
```suggestion
  <a class="alt btn chapter-cta comment-cta" href="https://discuss.httparchive.org/t/{{ metadata.get('discuss') }}">
```

src/templates/base/base_chapter.html
```diff
@@ -65,14 +65,26 @@ <h2 id="explore-results">
...
     </svg>
     <span id="num_comments">{{ self.comments_fallback() }}</span> <span data-translation id="comment-singular">{{ self.comment() }}</span>
     <span data-translation id="comment-plural">{{ self.comments() }}</span>
   </a>
+  <a class="alt btn cta" href="https://github.com/HTTPArchive/almanac.httparchive.org/issues/new?assignees=&labels=bug%2C+writing&template=2020-content-issue.md&title=Issue+with+the+2020+%5BCHAPTER%5D+chapter">
+    <svg width="22" height="22" role="img">
+      <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#github"></use>
+    </svg>
+    <span>Suggest Edit</span>
+  </a>
```
bazzadp, line 81: An svg with `role="image"` must have an accessible name for screen readers. This is what is causing the Lighthouse failures as when you hide the text on mobile and only show the icons, there is no accessible name.

You should add a `<title>` element (which handily will also show as a hover tooltip).

Let's also put the full text in the `<title>` and make the actual text `aria-hidden`, so we get the full text even on mobile view (when text is hidden) but avoids repeating the text on desktop view (which text is not hidden):

```suggestion
  <a class="alt btn chapter-cta suggest-edit-cta" href="https://github.com/HTTPArchive/almanac.httparchive.org/issues/new?assignees=&labels=bug%2C+writing&template=2020-content-issue.md&title=Issue+with+the+2020+%5BCHAPTER%5D+chapter">
    <svg width="22" height="22" role="img">
      <title>GitHub: Suggest Edit</title>
      <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#github"></use>
    </svg>
    <span aria-hidden="true">Suggest Edit</span>
  </a>
```

We should also move this text to each languages templates. But let's look at that after we've addressed all these issues.

src/templates/base/base_chapter.html
```diff
@@ -65,14 +65,26 @@ <h2 id="explore-results">
...
+  <a class="alt btn cta" href="https://github.com/HTTPArchive/almanac.httparchive.org/issues/923">
+    <svg width="22" height="22" role="img">
+      <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#globe"></use>
+    </svg>
+    <span>Help Translate</span>
+  </a>
```
bazzadp, line 87: Same here:

```suggestion
  <a class="alt btn chapter-cta help-translate-cta" href="https://github.com/HTTPArchive/almanac.httparchive.org/issues/923">
    <svg width="22" height="22" role="img">
      <title>GitHub: Help Translate</title>
      <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#globe"></use>
    </svg>
    <span aria-hidden="true">Help Translate</span>
  </a>
```

src/static/css/almanac.css
```diff
@@ -484,6 +515,7 @@ footer.alt-bg a:focus {
 footer {
   padding-top: 40px;
   padding-bottom: 40px;
+  z-index: 2;
```
bazzadp, line 518: I don't think this is needed and think it's a little weird it goes behind the footer.

bazzadp (commented):
This is excellent work @AbbyTsai - works really well!

I've made some suggestions below.

src/static/css/almanac.css
```diff
@@ -387,6 +386,38 @@ h2.header {
   font-size: 0.875rem;
   text-decoration: none;
   text-align: center;
+  box-shadow: 2px 2px 3px 3px lightgray;
```
bazzadp, line 389: Can we add this box-shadow just to the `chapter-cta` class rather than all btns? Looks a bit weird for the buttons on the home screen.

AbbyTsai:
Good point's mentioned. the files have been updated.
thought if we focus on CTA buttons in the full page, then as readers of footer, some of drop-button, social-media icon might be overlapped. shall we consider them behind the footer or any thoughts? 

for example:
image

AbbyTsai:
Thanks. Barry. happy to see the test version and feel free to add anything below together.

Here's some consolidated points would love to have all your thoughts. @rviscomi  @HTTPArchive/designers  

5 CTAs (2 new, 3 alive): call to action (CTA) and hopefully be visible by readers.

CTA1:Comments (alive)     CTA2:suggest edit (new)         CTA3:help translate (new)
CTA4:Review result (alive)  CTA5:Review queries. (alive) 

Purpose: increase readers' engagement. 

- Proposed option A:
  pull CTA1-comment stick on the right bottom by scroll.
  4 CTAs retain in the existed explore result section. 

- Proposed option B:
  pull 3 CTAs1,2,3-comment/edit/help stick on the right bottom by scroll.
  2 CTAs retain in the existed explore result section.

src/static/css/almanac.css
```diff
@@ -389,6 +389,35 @@ h2.header {
...
 }
 
+.btn.chapter-cta > svg {
+  margin: 0px 5px;
+}
+
+.chapter-cta {
+  position: fixed;
+  border: none;
+  box-shadow: 2px 2px 3px 3px lightgray;
```
rviscomi, line 399: Should we apply the `.floating-card` class and reuse its box-shadow styles for consistency?

src/static/css/almanac.css
```diff
@@ -389,6 +389,35 @@ h2.header {
...
+  padding: 10px 15px;
+  opacity: 0.9;
+  text-decoration: none;
+  z-index: 1;
+}
+
+.comment-cta {
```
rviscomi, line 406: Rather than positioning each CTA individually, I'd recommend wrapping them all in a `<div>` and positioning that div on the bottom right. Then the CTAs within the div could be positioned relative to each other according to flexbox.

Something like this (not tested):

```css
#cta-container {
  position: fixed;
  bottom: 0.6em;
  right: 0.125em;
  display: flex;
  flex-direction: column;
}

.chapter-cta {
  margin-top: 1em;
}
```

AbbyTsai:
thanks all the warm, helpful recommendations. 
The update 3CTAs, sticking on the right bottom by scroll and stop at end of article (before chapter-navigation/ footer), look forward to further thoughts/comments.
Also, not quiet sure what and how to fix the internal server error of PR?

bazzadp:
Looks like GitHub is experiencing issues: https://www.githubstatus.com/incidents/zbpwygxwb3gw

Will rerun them again later.

AbbyTsai:
Thanks to the information and hope to see a lovely v here.

src/templates/base/base_chapter.html
```diff
@@ -65,14 +65,7 @@ <h2 id="explore-results">
...
     </a>
   </h2>
-  <a class="alt btn" href="https://discuss.httparchive.org/t/{{ metadata.get('discuss') }}">
-    <svg width="18" height="18" role="img">
-      <title>{{ self.discuss_this_chapter() }}</title>
-      <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#comment"></use>
-    </svg>
-    <span id="num_comments">{{ self.comments_fallback() }}</span> <span data-translation id="comment-singular">{{ self.comment() }}</span>
-    <span data-translation id="comment-plural">{{ self.comments() }}</span>
-  </a>
```
bazzadp, line 75: I still wonder if we should leave this in place here (and add the other new ones as well) and then have the floating CTAs stop before this (or could have it vanish or merge into this but that'll be more tricky).

I've a demo of what I'm talking about here: https://20210517t231435-dot-webalmanac.uk.r.appspot.com/en/2020/fonts

I've also moved the CTAs to be horizontal as feel they cover up less space in the chapter this way and so are less irritating.

What do you think? @rviscomi any comments since you originally asked for this?

I can commit my changes to this branch if we agree this is the way forward as probably easier than explaining all the small changes I did to make this work.

src/templates/base/base_chapter.html
```diff
@@ -308,19 +301,42 @@ <h1 class="title title-lg">
...
+                {{ render_authors() }}
+            </section>
         </article>
-        <div class="chapter_links">
-          {{ render_actions() }}
-        </div>
-        <section class="authors">
-            {{ render_authors() }}
-        </section>
-
```
bazzadp, line 320: I've reverted this. change and moved both the `chapter_links` and the `authors` back out of the article again in my version to allow the CTAs to stop at the right place.

I also had to add a spacer div for where the floating CTAs slot into.

```html
            </div>
            {{ render_byline() }}
            {{ self.main_content() }}
            <div class="spacer"></div>
        </article>
        <div class="chapter_links">
          {{ render_actions() }}
        </div>
        <section class="authors">
            {{ render_authors() }}
        </section>
```

And then added this class to almanac.css:

```css
.spacer {
  min-height: 60px;
}
```

bazzadp (commented):
Love the `position: sticky` changes but spotted a few issues and think we can make it even smoother. Let me know what you think.

AbbyTsai:
Yep, cool idea. These horizonal CTAs play the role gently, but a little thought comes to the mind, would we seemingly put the same buttons both at the end of article and explore result in screen?

bazzadp:
> Yep, cool idea. These horizonal CTAs play the role gently, but a little thought comes to the mind, would we seemingly put the same buttons both at the end of article and explore result in screen?

Yeah I'm not loving that to be honest:

Duplication of CTAs at bottom of article

But I do like that they stop scrolling down at a point, rather than covering the other buttons and/or footer. I think it's cleaner than what we originally had.

As I say, the options are:
1. Leave as is - may it's not that bad?
2. Make the CTA buttons vanish after we reach the bottom of the article. I can't see how to do this with pure CSS so would likely need some IntersectionObserver JavaScript and not sure if that's overkill for this?
3. Make the CTAs rest in the "Explore your Results" section so there is no duplication. I think that could be tricky to pull off and also the tab order will be incorrect as the CTAs are at the top of the article really. Also is a good progressive enhancement in case the position sticky doesn't work (e.g. reader mode).
4. Remove the CTAs from the "Explore your Results" section so there is no duplication like you already had. I do like there there though as they are a way to "Explore the Results" and think it's a bit weird having some there and some not because they are in the scroll. Also is a good progressive enhancement in case the position sticky doesn't work (e.g. reader mode).

@HTTPArchive/designers or @HTTPArchive/developers any thoughts here? Or anyone with between CSS/JS knowledge than me that could make this work better? You can see the current implementation on [this demo deploy](https://20210517t231435-dot-webalmanac.uk.r.appspot.com/en/2020/fonts).

bazzadp:
Having all the options under the "Explore your Options" section also gives options to reduce the floating CTAs if we want in certain circumstances:

- Do we only show the discuss one on mobile as limited screen space? And have all 5 floating CTAs on desktop, but only the 3 on tablet?
- Do we allow a Close button at some point in future to remove them completely for those that find them annoying?
- Do we allow a preference setting at some point in future to remove them completely and permanently for those that find them annoying?

It's always nice to have options 😁

shantsis:
It's quite a lot of actions shown by default - what if they're collapsed first?
Screen Shot 2021-05-20 at 8 36 22 PM

AbbyTsai:
@rviscomi @bazzadp @shantsis 
Here's something would like to hear more thoughts together, as those 3 CTAs aim to touch readers' fingers in.

1.Where's CTAs position? 
   On the top of right side. 

2.Where's CTAs scrolling area? 
   Inside article, from introduction to conclusion-explore result.

3.What's CTAs collapsed look? 
   3 svgs, the desktop's as same as mobile. (Thanks shantsis' good idea and illustration.)

4.What's in explore result? 
  3 CTAs+2 result buttons with text description and show 2 rows in desktop, 5 rows in mobile.

bazzadp:
I wonder if we're trying to over use this. Do we really need "Help Translate" available at all times? And it's fairly common to have "Suggest Edit" just at the end rather than available at all times.

Yes we could hide in a menu, but a menu means we're hiding the comments again - and primary aim of this was to increase that engagement I think. We also wanted to increase interaction with the stats but the availability of those by each figure does that job IMHO.

So I'm wondering if we should leave this as just the comments? That would keep it's size to a minimum and also avoid any concerns about it covering content or repeating the "Explorer the results" section:

Test deploy with just that: https://20210522t125050-dot-webalmanac.uk.r.appspot.com


Mobile:

Mobile CTA over intro Mobile CTA over Explore the Results


Desktop:

Desktop CTA

Desktop CTA over Explore the results

bazzadp:
Had to do some other changes so committed my additions to this branch so they wouldn't be lost. Hope you don't mind @AbbyTsai !

AbbyTsai:
Really like the lately test version where most focus on the comment CTA via discuss. Would appreciate if it's committed as well, as a little bit confused the changes. 

Think that if we still wanna to raise the helper of chapter translate or make it easier to open issue via GitHub, then we might use icon of comment instead of menu and then 3 dedicated icons display by click. Would that collapsed icon attract readers' eye as well? OR we may focus on one comment CTA only. It seemingly depends on our ambition. ^^

rviscomi:
I really like the suggestion by @shantsis in https://github.com/HTTPArchive/almanac.httparchive.org/pull/2196#issuecomment-845571537 to have open/closed states. That allows us to add more actions without overcrowding in the default (closed) state. I can also imagine using transitions to animate between states.

So I'd propose adding all of the buttons in "Explore the results" to the fixed position CTA's "open" state.

This is where the sticky CTA ends up for me at the bottom of the chapter:

image

I'd suggest changing the resting spot to be before the "Explore the results" section and after any "Conclusion" text content to avoid the overlap.

One other thing I noticed is that the "Suggest Edit" button opens a new GitHub issue with the title `Issue with the 2020 [CHAPTER] chapter`. Could we replace `[CHAPTER]` with the actual chapter name? It should also make the year variable.

And one last piece of feedback. Regarding the comments CTA HTML:

```html
<a class="alt btn chapter-cta comment-cta" href="https://discuss.httparchive.org/t/1757">
                <svg width="22" height="22" role="img">
                  <title>Discuss this chapter</title>
                  <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#comment"></use>
                </svg>
                <span class="num_comments" aria-hidden="true">8</span> <span data-translation="" class="comment-singular not_mobile">comment</span>
                <span class="comment-plural not_mobile" aria-hidden="true">comments</span>
              </a>
```

When there are 8 comments as there are on the [2019 CSS](https://almanac.httparchive.org/en/2019/css) chapter, shouldn't the `comment` span have the `aria-hidden="true"` attribute?

bazzadp:
I've fixed the hover overlap as suggested, and corrected my bad with the aria attributes. I also moved the CSS from almanac.css (loaded by the whole site) to page.css (loaded by chapters only)

Latest test version here: https://20210526t153908-dot-webalmanac.uk.r.appspot.com/en/2020/fonts

I'm still in two minds whether a menu is the best to be honest. While I agree it gives us more options, I quite like having the main option (Discuss) visible as think that increases the chances of people actually engaging with the discussion. My worry is that with it behind a menu more people will ignore it. Anyone any opinions here?

shantsis:
If we don't go with the menu idea I think it's best to only have 1 CTA (the comments or whatever we're promoting). Having 3 floating around the page is too much and I don't think most of them would be frequently used.

AbbyTsai:
Here might be an one of choice for our consideration. ^^
CTA-close view:
image

CTA-open view:
image

bazzadp:
Looks good, though bit of a gap between "Conclusion" and "Explore the results" so should try to reduce that.

Trying to decide between horizontal or vertical. Horizontal probably is better as can keep open and still read the article.

I wonder should the number of comments be shown even in closed state to address my concern about engagement being lower if requires opening?

AbbyTsai:
Thanks to Barry's nice insight.

**Topic in readers engagement:**
Yep, magical number may inspire our curious mind to see how many comments are.
Horizontal look is lovely and the most article isn't covered by it. but then a guess is talking: 
does it might be ignored by reader as well? or doesn't?

**Topic in implement:**
I'm trying to implement a idea (href#) into sticky button from [GUI-CHALLENGE of web.dev](https://web.dev/blog/) 
(thanks author @argyleink ), but fail without luck. 
The screen dances after the button is clicked to either open or close. [Here's a example.](https://codepen.io/AbbyTsai/pen/poeaJMa) 
Does someone have eye to solve it?  Thanks.

bazzadp:
> The screen dances after the button is clicked to either open or close. [Here's a example.](https://codepen.io/AbbyTsai/pen/poeaJMa)
> Does someone have eye to solve it? Thanks.

I think it's because you've used a link and it keeps going to that. The toggle should really be implemented with JavaScript anyway so we can set the appropriate accessibility attributes on it. Taking your code pen and changing it to do it with JavaScript shows no jumps: https://codepen.io/tunetheweb/pen/jOBxRZv?editors=1111

AbbyTsai:
Another sweet css:hover might has a similar effect without the dance. ([demo](https://codepen.io/AbbyTsai/pen/poeaJMa))
The thing we'd consider is the influence between mouse-over and mouse-click in desktop, but mobile.
Comparison in reader's action, then mouse-over might be an easy one to trigger, or probably mouse-click?  
Also curious how the reason you guys determine the way in development, generally, as there might be various methods (css, js, .. library etc) ?

tunetheweb:
> Another sweet css:hover might has a similar effect without the dance. ([demo](https://codepen.io/AbbyTsai/pen/poeaJMa))

It's still not keyboard accessible. Nor does it add the appropriate aria attributes necessary for screen-readers.

> The thing we'd consider is the influence between mouse-over and mouse-click in desktop, but mobile.
> Comparison in reader's action, then mouse-over might be an easy one to trigger, or probably mouse-click?

Another reason not to try a pure CSS solution.

> Also curious how the reason you guys determine the way in development, generally, as there might be various methods (css, js, .. library etc) ?

Well that's the trick! It's a combination of what's the best solution for the users (this should be the primary concern!), what's the easiest and cleanest to code, and how maintainable it is in the future.

Adding a library is easy, but can add bloat, and add future work to maintain it as it changes. So need to be sure it's worth it.

We've (only half consciously) tried to avoid libraries for the Web Almanac site itself (though do use lots for our build process) as it's a fairly simple site at the end of the day and not needed them.

tunetheweb:
> Another sweet css:hover might has a similar effect without the dance. ([demo](https://codepen.io/AbbyTsai/pen/poeaJMa))

It's still not keyboard accessible. Nor does it add the appropriate aria attributes necessary for screen-readers.

> The thing we'd consider is the influence between mouse-over and mouse-click in desktop, but mobile.
> Comparison in reader's action, then mouse-over might be an easy one to trigger, or probably mouse-click?

Another reason not to try a pure CSS solution.

> Also curious how the reason you guys determine the way in development, generally, as there might be various methods (css, js, .. library etc) ?

Well that's the trick! It's a combination of what's the best solution for the users (this should be the primary concern!), what's the easiest and cleanest to code, and how maintainable it is in the future.

Adding a library is easy, but can add bloat, and add future work to maintain it as it changes. So need to be sure it's worth it.

We've (only half consciously) tried to avoid libraries for the Web Almanac site itself (though do use lots for our build process) as it's a fairly simple site at the end of the day and not needed them.

AbbyTsai:
Thanks Barry for the nice sharing in development principle of user first and warm sample in JS version.

Happy to see these CTAs would help our users find an easy way to get involved by the articles.

Not quiet sure if we try to add a role, aria attributes to its html element on css:mouse over, does then the screen read the missing keyboard? or? 

In summary, love to see our mouse-click button alive if experience show users enjoy it, or it's a better way to keyboard accessible.

tunetheweb:
> Not quiet sure if we try to add a role, aria attributes to its html element on css:mouse over, does then the screen read the missing keyboard? or?

Mouse over doesn’t work for keyboard or mobile users. Also it’s very easy to hover away by accident and then you lose the menu. Best to just do it on click. Then it’s easy to add the necessary ARIA attributes as well.

There’s been a fair bit of research on this (see links in answer to [this question for just some of them](https://ux.stackexchange.com/questions/94838/should-nav-items-be-hover-or-click-to-activate-dropdown)). Bootstrap 4 for example moved from hover menus to click menus.

shantsis:
Can’t a mouse over action also be triggered by on keyboard focus? It may be a bit more work but doesn’t have to be a click.

tunetheweb:
Yes it can but doesn’t address mobile or the ARIA requirements. The reality is we need JavaScript to make accessible menus at present. And, given that this is a progressive enhancement of the “Explore the results” section that’s not a bad thing. So what’s the reason for not using JavaScript?

tunetheweb:
Interesting article on this topic which has just been published: https://www.smashingmagazine.com/2021/06/css-javascript-requirements-accessible-components/

tunetheweb:
This pull request is dragging out, and growing arms and legs as we add more features, complexity and opinions to it. I'm thinking we should merge it as is, and then iterate upon it. What do you think @AbbyTsai / @rviscomi / @shantsis ?

As a reminder the current code in this branch is basically a floating comments widget for mobile and desktop and some extra calls to action in the Explore the Results section. It can be seen here: https://20210614t202814-dot-webalmanac.uk.r.appspot.com/en/2020/fonts.

I think these are good features, that are beneficial and are mergeable as is. We can add the extra ideas (expandable menu with more options) in a new branch and pull request.

One thing I'd like to do is before merging, is internationalise it, but I am happy to do that and commit to this branch if you've not moved it forward offline @AbbyTsai ?

Let me know your thoughts?

shantsis:
Design wise it seems fine for now. I was trying it on my mobile and was struggling quite a lot to activate it - see the video linked. https://drive.google.com/file/d/1Fhh9gR7zi50X3KdlolNYE6KEflFzyShf/view?usp=drivesdk

tunetheweb:
Oh good spot! Moving to bottom we can make it work regardless of whether menu is showing or not.

Try this: https://20210614t205125-dot-webalmanac.uk.r.appspot.com/en/2020/fonts

shantsis:
Much better thanks 😊

rviscomi:
Merge and iterate SGTM!

tunetheweb:
I've added the internalisation pieces so think this is good to merge now.

@AbbyTsai let me know your thoughts before I do that.

tunetheweb:
Oh and latest staged version: https://20210615t003135-dot-webalmanac.uk.r.appspot.com/en/2020/fonts

AbbyTsai:
Great to see it and warm thanks to all support by Barry, Shantsis, Rick.

tunetheweb (approved):

tunetheweb:
And it's live! Thanks @AbbyTsai (and @shantsis / @rviscomi for input).

Will be interesting to see if comments increase.
And also Edits/Translations.
Though it is the quiet time between editions so readership is traditionally down at these times.

PR #2214, opened 2021-05-30
Add chapter number to the chapter footer navigation (HTTPArchive#2209)

LucasBona05:


src/package-lock.json
```diff
@@ -1,1214 +1,8 @@
 {
   "name": "almanac.httparchive.org",
   "version": "0.0.1",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
bazzadp, line 4: Can we revert these changes? We use npm v7 on this project so version 2 of the package-lock.json file. It's that that you are using an older version, but no need to change the package-lock file.

src/server/helpers.py
```diff
@@ -112,7 +112,6 @@ def get_chapter_nextprev(config, chapter_slug):
 
     if not found:
         return (None, None)
-
```
bazzadp, line 115: Can we revert this as nothing really to do with your change?

bazzadp (changes requested):
LGTM!

Can you revert the two unintended changes?

Also add yourself to the `src/config/2021.json` file in the 'developers' team? We try to keep it in alphabetical order by first name so add yourself around line 247 I guess?

rviscomi:
Thanks for fixing this @LucasBona05! 🎉

Issue #1293, opened 2020-09-15
Add additional columns to Almanac Requests table


tunetheweb:
The `httparchive.almanac.requests` table includes a `respHttpVersion` and a `reqHttpVersion` column neither of which contain any meaningful data.

All the HTTP/2 queries last year were based on the `JSON_EXTRACT_SCALAR(payload, "$._protocol")` column which does appear to be mostly accurate. We should add the protocol column to the `httparchive.almanac.requests` table and/or replace the `respHttpVersion` and a `reqHttpVersion` columns with this data to make it easier and cheaper to query HTTP version requests.

**Edit.** Complete list of headers we should add if possible:
- `$._protocol`
- `$._was_pushed`
- `$._tls_version`
- `$._tls_cipher_suite`
- `$._securityDetails.issuer`
- `$._securityDetails.keyExchange`
- `$._securityDetails.cipher`
- `$._securityDetails.protocol` - not sure how this is different than TLS Version to be honest
- `$.request.headers` - in JSON format rather than like `reqOtherHeaders`
- `$.response.headers` - in JSON format rather than like `respOtherHeaders`
- CrUX rank from `summary_pages`

Also fix the `startedDateTime` field as [off by 1000](https://httparchive.slack.com/archives/C016QLMT0BZ/p1604932577129800).

CC: @gregorywolf

rviscomi:
Yeah I don't know why those fields have such weird values. I think making the `protocol` field more easily accessible in the request data sounds like a great idea. @bazzadp @gregorywolf is this something you'd like to see done ASAP in order to update the HTTP/2 chapter queries or is it for 2021+?

bazzadp:
I've checked and they are wrong in the HAR.

If we could update it ASAP, it would be really useful and save @gregorywolf (and therefore your coupons!) a lot. However if that's a lot of effort than can just query this from the `payload` column. It's just of the 17 queries @gregorywolf has so far, 16 are doing a 7.5TB query each for the `payload` data! 😞  So we should maybe thing about joining queries or creating another temp table if we don't want to touch `requests` table now. From a quick look I reckon about half the queries would no longer need to use this expensive `payload` column if we had this.

And if we added these columns while we are at it, then perhaps none would need to look at the `payload` column at all:
- `_was_pushed`
- `response.headers.link` - useful for Resource Hints chapter too
- `response.headers.alt-svc` - could be come important in future as QUIC and HTTP/3 takes hold
[...]

PR #2227, opened 2021-06-05
Privacy 2021 queries

max-ostapenko:


VictorLeP:
Queries that rely on headers depend on a possible format change for headers in https://github.com/HTTPArchive/almanac.httparchive.org/issues/1293.

tunetheweb:
Hi @max-ostapenko , I've just merged a SQL Linter into `main`. When you get a chance could you merge into this branch so your new queries will be linted according to the Web Almanac style?

sql/2021/privacy/purposes_for_iab_tcf_v2.sql
```diff
@@ -0,0 +1,111 @@
...
+RETURNS Array<STRUCT<key String,value String>>
+AS
+((select
+  array(
+    select as struct 
+      trim(split(kv, ':')[SAFE_OFFSET(0)]) as key, 
+      trim(split(kv, ':')[SAFE_OFFSET(1)]) as value
+    from t.kv kv
+  ) 
+from unnest([struct(split(translate(input, '{}"', '')) as kv)]) t
```
tunetheweb, line 17: Looks like the linter doesn't support this syntax. So tell it to ignore it (plus fixing the capitalisation issues at the same time!):

```suggestion
FROM UNNEST([STRUCT(SPLIT(TRANSLATE(input, '{}"', '')) AS kv)]) t -- noqa: PRS
```

Issue raised here: https://github.com/sqlfluff/sqlfluff/issues/1271 and fix has been merged so will be in the next release of the linter, but for now do the above.

sql/2021/privacy/top100_domains_that_set_cookies_via_response_header.sql
```diff
@@ -0,0 +1,59 @@
+#standardSQL
+# Top100 domains that set cookies via response header
+
+CREATE TEMPORARY FUNCTION
+  cookieNames(headers STRING)
+  RETURNS ARRAY<STRING> DETERMINISTIC
+  LANGUAGE js AS '''
```
tunetheweb, line 7:
```suggestion
CREATE TEMPORARY FUNCTION cookieNames(headers STRING) -- noqa: PRS
-- SQL Linter cannot handle DETERMINISTIC keyword so needs noqa ignore command on previous line
  RETURNS ARRAY<STRING> DETERMINISTIC
  LANGUAGE js AS '''
```

tunetheweb (changes requested):
Made some comments on some of the trickier linting issues. The rest should be obvious but gimme a shout if you can't figure them out.

You can also compare to the 2020 queries where they exist as they all lint now.

max-ostapenko:
@tunetheweb is there a way to run some SQL prettifier to match linter requirements?
We get a lot of errors in many files and it's gonna take time.

tunetheweb:
I've a fixer for some of the rules as have SQL Fluff downloaded locally. Just ran that and pushed a commit in https://github.com/HTTPArchive/almanac.httparchive.org/pull/2227/commits/3f659b427cf544679a76615f5c8cc937541c8ca5. Let's see how that goes.

max-ostapenko:
@tunetheweb please, have a look on code scanning check running in this branch. There is a lot of errors and warnings for unchanged files.
I'm not sure JS scan is expected to run on HTML files.

tunetheweb:
That looks very wrong! Is the branch up to date with `main`? That's the first thing I'd try.

max-ostapenko:
> Is the branch up to date with main?

Yes.

tunetheweb:
Reran the CodeQL jobs. Fingers crossed it magically works, cause I can't explain it. But the fact it's failed for the last four commits is not making me hopefully it'll magically fix itself...

tunetheweb:
Doing an experiment in #2333 

I suspect CodeQL has been enhanced to allow JavaScripts checks to be run on HTML for inline `<script>` tags. Which is fine, except that it doesn't like Jinja code as think it's confusing it as it looks a bit like JavaScript so thinks it can treat it as such when it can't. You're the unfortunate person unlucky enough to have submitted a Python change in a bit, (which kicks off both Python and JS CodeQL checks) and so first to experience this issue.

If that turns out to be the case, then I'll change #2333 to exclude the `src/templates` directory from CodeQL scanning and hopefully that will fix it. If so then will merge it to `main` and you can then pull it into this branch.

tunetheweb:
Yup looks like that was the issue - good guess from me!

Fix merged into `main` now, so have updated this branch so hopefully the jobs will pass now.

tunetheweb:
This still good to review? Or was it just for that issue that you tagged me to review it?

If good to review can you untag me again as reviewer, and tag the analyst group instead? I'm in that too but just means someone else might pick it up in the meantime too.

sql/2021/privacy/most_common_cmps_for_iab_tcf_v2.sql
```diff
@@ -0,0 +1,45 @@
...
+
+pages_iab_tcf_v2 AS (
+  SELECT
+    client,
+    JSON_QUERY(metrics, "$.iab_tcf_v2.data") AS metrics
+  FROM
+    pages_privacy
+  WHERE
+    JSON_QUERY(metrics, "$.iab_tcf_v2.data") IS NOT NULL
+)
```
tunetheweb, line 31: I guess you are aiming for consistent use of the `pages_privacy` table, but seems these are fairly simple and we could simplify by merging the first and third table (this feedback applies throughout):

```suggestion
WITH pages_iab_tcf_v2 AS (
  SELECT
    _TABLE_SUFFIX AS client,
    JSON_QUERY(metrics, "$._privacy.iab_tcf_v2.data") AS metrics
  FROM
    `httparchive.pages.2021_07_01_*`
  WHERE
    JSON_QUERY(metrics, "$._privacy.iab_tcf_v2.data") IS NOT NULL
),

total_nb_pages AS (
  SELECT
    _TABLE_SUFFIX AS client,
    COUNT(0) AS total_nb_pages
  FROM
    `httparchive.pages.2021_07_01_*`
  GROUP BY
    1
),
```

sql/2021/privacy/most_common_cmps_for_iab_tcf_v2.sql
```diff
@@ -0,0 +1,45 @@
...
+
+SELECT
+  client,
+  JSON_VALUE(metrics, '$.cmpId') AS cmpId,
+  COUNT(0) AS nb_websites,
+  ROUND(COUNT(0) / MIN(total_nb_pages.total_nb_pages), 2) AS pct_websites
+FROM
+  pages_iab_tcf_v2 JOIN total_nb_pages USING (client)
```
tunetheweb, line 39: Formatting nit:

```suggestion
FROM
  pages_iab_tcf_v2
JOIN
  total_nb_pages
USING (client)
```

sql/2021/privacy/most_common_cmps_for_iab_tcf_v2.sql
```diff
@@ -0,0 +1,45 @@
...
+WHERE
+  JSON_VALUE(metrics, '$.cmpId') IS NOT NULL
+GROUP BY
+  1, 2
+ORDER BY
+  3 DESC
```
tunetheweb, line 45: Can we use column names rather than numbers to make this more obvious:

```suggestion
  client,
  cmpId
ORDER BY
  nb_websites DESC
```

sql/2021/privacy/most_common_featurepolicy_permissionspolicy_directive_values.sql
```diff
@@ -0,0 +1,141 @@
...
+
+WITH page_ranks AS (
+  SELECT
+    client,
+    page,
+    rank
+  FROM
+    `httparchive.almanac.summary_response_bodies`
+  WHERE
+    date = '2021-07-01' AND firstHtml = TRUE
```
tunetheweb, line 12: Formatting nit, list WHERE clauses separately where it makes no difference:

```suggestion
    date = '2021-07-01' AND
    firstHtml = TRUE
```

(applies throughout)

sql/2021/privacy/most_common_featurepolicy_permissionspolicy_directive_values.sql
```diff
@@ -0,0 +1,141 @@
...
+),
+
+response_headers AS (
+  SELECT
+    client,
+    page,
+    LOWER(JSON_VALUE(response_header, '$.name')) AS header_name,
+    LOWER(JSON_VALUE(response_header, '$.value')) AS header_value
+  FROM
+    `httparchive.almanac.summary_response_bodies`,
```
tunetheweb, line 22: Why do we need to use `response_bodies`? Try our best not to use that is at al possible. Can we use `summary_requests` (which is actually requests and responses btw):

```suggestion
    `httparchive.almanac.summary_requests`,
```

sql/2021/privacy/most_common_featurepolicy_permissionspolicy_directive_values.sql
```diff
@@ -0,0 +1,141 @@
...
+    client,
+    rank,
+    COUNT(DISTINCT page) AS total_nb_pages
+  FROM
+    `httparchive.almanac.summary_response_bodies`
+  WHERE
+    date = '2021-07-01' AND firstHtml = TRUE
+  GROUP BY
+    1, 2
+),
```
tunetheweb, line 58: Same comment about using `summary_requests` over `summary_response_bodies`

Also not sure you understand how the ranking works.
The `rank` column will be 1,000 (if it's in the top 1,000 websites), 10,000 (of it's in the next 9,900 sites)...etc.

See this query: https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2021/pwa/sw_adoption_over_time_ranking.sql for an example as to how to query this. Basically we create a `rank_grouping` column and then use that to include all URLs where `rank <= rank_grouping`.

tunetheweb (changes requested):
Not been through all of the queries but spotted a few things that I think apply in general so will stop here for now.

Let me know your thoughts and once those are addressed (across all queries) can then have another review.

sql/2021/privacy/most_common_cmps_for_iab_tcf_v2.sql
VictorLeP: Fixed in 3a1cfe8

sql/2021/privacy/most_common_featurepolicy_permissionspolicy_directive_values.sql
VictorLeP: Fixed in e878bce

sql/2021/privacy/most_common_cmps_for_iab_tcf_v2.sql
VictorLeP: Fixed in `c642950`

sql/2021/privacy/most_common_cmps_for_iab_tcf_v2.sql
VictorLeP: Fixed in `6a94989`

sql/2021/privacy/most_common_featurepolicy_permissionspolicy_directive_values.sql
VictorLeP: Is this supposed to be a drop-in replacement? I don't see a `response_headers` column in `summary_requests`.

tunetheweb: Sorry my mistake! It’s the `requests` table not `summary_requests`

VictorLeP: Does this actually matter? We'll be accessing the same columns anyway, so it should be exactly the same data (size).
Unless it's about the 50 million extra rows in `requests` versus `summary_response_bodies`?
(obviously it's just a search/replace, so not much effort to change, but I'm curious as to the reason why)

sql/2021/privacy/most_common_featurepolicy_permissionspolicy_directive_values.sql
VictorLeP: I think I get the ranking, in that we'll indeed see (10000-1000) sites with `rank = 10000` when we group like in the current query.
I'm thinking it might be interesting though to know the numbers for each 'slice' of ranks, instead of 'duplicating' highly ranked sites? (so having the data for 1-1000, 1001-10000, 10001-100000, ... instead of for 1-1000, 1-10000, 1-100000, ...) That way we distinctly see what's happening for lower-ranked sites.
I'd be interested to hear what the others think about this though.

sql/2021/privacy/most_common_featurepolicy_permissionspolicy_directive_values.sql
tunetheweb: You're not wrong in this case.

But I still think we should change it, for a few reasons:

The `response_bodies` table just scares the heck outta me because it's so massive! It costs $227 to query the almanac version in full, and $162 (per month!) for the monthly version!

It's a red flag to me when I see it used. This year we scanned our 2020 code base for any use of it to see if we could implement those metrics for cheaper in custom metrics. Having it used like this will just add noise to that.

Yes the `almanac` version of that table (`almanac.summary_response_bodies`) is partitioned on certain key columns so you can just be charged for those columns but it's still easy to get the full cost, for example when:
- Changing to `SELECT *` for debugging
- Changing to monthly tables to see if it's changed since the Almanac was published (we make the queries easily available on the Web Almanac so try to assume someone might run them themselves, even if no one does).

Plus it's a simple drop in replacement - no further code changes required! So no real reason NOT to change it IMHO.

Does that make sense?

VictorLeP:
I've checked and I don't think we're using `PARTITION BY` in a "dangerous way" (https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide#a-warning-on-totals).

sql/2021/privacy/most_common_featurepolicy_permissionspolicy_directive_values.sql
tunetheweb: I thought the same initially but changed my mind. I think it's really confusing for a reader to see the "rank" of 10,000 and to have to realise it's actually only the next segment of 9,000 pages, rather than the top 10,000 sites.

And labelling it like this (which is what I initially did), is honest, but still quite confusing:

image

It also makes it less comparable to other stats (e..g those based on Alexa top 1M)

Plus we should be consistent in this usage across the Web Almanac and, at the moment, we're going with the alternative method (cause we think it's less confusing). But happy to hear other's thoughts on this!

tunetheweb:
> I've checked and I don't think we're using `PARTITION BY` in a "dangerous way" (https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide#a-warning-on-totals).

Had a quick look myself there and agree with you. Good stuff.

sql/2021/privacy/most_common_featurepolicy_permissionspolicy_directive_values.sql
VictorLeP: Sure; fixed in ec2635c.

sql/2021/privacy/most_common_cmps_for_iab_tcf_v2.sql
```diff
@@ -0,0 +1,42 @@
...
+  total_nb_pages
+USING (client)
+WHERE
+  JSON_VALUE(metrics, '$.cmpId') IS NOT NULL
+GROUP BY
+  client,
+  cmpId
+ORDER BY
+  client ASC,
+  nb_websites DESC
```
tunetheweb, line 42: This query no longer runs :-(

The problem is the `_privacy` is not real JSON but escaped JSON. So you need to use this:

```sql
    JSON_QUERY(JSON_VALUE(payload, "$._privacy"), "$.iab_tcf_v2.data") AS metrics
```

as opposed to

```sql
   JSON_QUERY(payload, "$._privacy.iab_tcf_v2.data") AS metrics
```

Also your first `WHERE` clause is incorrect.

I also like to include the total in the final query (it's a nice spot check for `OVER` issue).

I actually think this query can be simplified further, so including the above changes, we can simply do something like this with only one CTE:

```suggestion
WITH total_pages AS (
  SELECT
    _TABLE_SUFFIX,
    COUNT(0) AS total
  FROM
    `httparchive.pages.2021_07_01_*`
  GROUP BY
    _TABLE_SUFFIX
)

SELECT
  _TABLE_SUFFIX AS client,
  JSON_QUERY(JSON_VALUE(payload, "$._privacy"), "$.iab_tcf_v2.data.cmpId") AS cmpId,
  COUNT(0) AS number_websites,
  total,
  COUNT(0) / MIN(total) AS pct_websites
FROM
  `httparchive.pages.2021_07_01_*`
JOIN
  total_pages
USING (_TABLE_SUFFIX)
WHERE
  JSON_QUERY(JSON_VALUE(payload, "$._privacy"), "$.iab_tcf_v2.data.cmpId") IS NOT NULL
GROUP BY
  client,
  total,
  cmpId
ORDER BY
  client ASC,
  number_websites DESC
```

WDYT? Can you check the other queries for similar issues and, make similar changes to them? Happy to use multiple CTEs where it simplifies the query, but don't think it does in this case to be honest. Some of your other queries are more complicated so may make more sense for them.

tunetheweb (changes requested):

sql/2021/privacy/most_common_cmps_for_iab_tcf_v2.sql
VictorLeP: Yeah, looks like I left most queries using custom metrics in a horribly broken state :sweat_smile: 
For simple queries, the simplification is fine. However, for a query like https://github.com/HTTPArchive/almanac.httparchive.org/blob/c642950b56caddc3894e5738a841c2559188ca50/sql/2021/privacy/nb_websites_with_features_based_on_string_search.sql, I wonder whether repeatedly extracting `JSON_VALUE(payload, "$._privacy")` (everywhere where it incorrectly says `metrics` now) instead of just once in a CTE could reduce performance?

tunetheweb: Agreed. Use CTEs where appropriate but don’t unnecessarily use them for simple queries.

And btw all our custom metrics used to be written this way so not your fault. It’s only more recently (e.g. PWA custom metric) that we’ve changed how to do it in a way that saves them as proper JSON instead of an escaped string.

VictorLeP: Done in b3cedee and ef44cab.

sql/2021/privacy/trackers_whotracksme.py
```diff
@@ -0,0 +1,44 @@
+"""
+Retrieve and extract trackers as identified by WhoTracks.me.
+
+First, download the latest 'tracker database'
+(https://github.com/ghostery/whotracks.me/tree/master/whotracksme/data/assets):
+`aws s3 cp --no-sign-request s3://data.whotracks.me/trackerdb.sql .`
+
+Then, execute this script.
+Finally, copy the result into the tracker-related SQL queries.
+"""
```
tunetheweb, line 10: Let's get the results of this query added as a date table to almanac schema. We have precedence in this with other chapters (e.g. `h2_prioritization_cdns` is a special table just for the HTTP chapter from an external source).

That will make your other queries in this PR a lot smaller as not repeating the whole dataset each time.

Will also make it easier to manage.

Can you reach out to @paulcalvano to get this set up?
Should also consider this for the haveIbeenpwned data.

BTW are there any licencing issues with using either o these datasets?

sql/2021/privacy/nb_websites_with_device_sensor_blink_usage.sql
```diff
@@ -0,0 +1,21 @@
...
+WHERE
+  yyyymmdd = '20210701' AND
+  (
+    feature LIKE '%DeviceMotion%' OR
+    feature LIKE '%DeviceOrientation%' OR
+    feature LIKE '%DeviceProximity%' OR
+    feature LIKE '%DeviceLight%'
+  )
+ORDER BY
+  2 ASC, 1 ASC
```
tunetheweb, line 21:
```suggestion
ORDER BY
  client,
  feature
```

sql/2021/privacy/nb_websites_with_dnt_or_gpc_blink_usage.sql
```diff
@@ -0,0 +1,21 @@
...
+  num_urls,
+  total_urls,
+  pct_urls AS pct_urls
+FROM
+  `httparchive.blink_features.usage`
+WHERE
+  yyyymmdd = '20210701' AND
+  feature = 'NavigatorDoNotTrack'
+ORDER BY
+  2 ASC, 1 ASC
```
tunetheweb, line 16:
```suggestion
ORDER BY
  feature,
  client
```

sql/2021/privacy/nb_websites_with_featurepolicy_blink_usage.sql
```diff
@@ -0,0 +1,40 @@
...
+  num_urls,
+  total_urls,
+  pct_urls AS pct_urls
+FROM
+  `httparchive.blink_features.usage`
+WHERE
+  yyyymmdd = '20210701' AND
+  feature LIKE '%FeaturePolicy%'
+ORDER BY
+  2 ASC, 1 ASC
```
tunetheweb, line 16:
```suggestion
ORDER BY
  feature,
  client
```

sql/2021/privacy/nb_websites_with_geolocation_blink_usage.sql
```diff
@@ -0,0 +1,29 @@
...
+  num_urls,
+  total_urls,
+  pct_urls AS pct_urls
+FROM
+  `httparchive.blink_features.usage`
+WHERE
+  yyyymmdd = '20210701' AND
+  feature LIKE '%Geolocation%'
+ORDER BY
+  2 ASC, 1 ASC
```
tunetheweb, line 16:
```suggestion
ORDER BY
  feature,
  client
```

sql/2021/privacy/nb_websites_with_geolocation_blink_usage.sql
```diff
@@ -0,0 +1,29 @@
+#standardSQL
+# Pages that use geolocation features (based on Blink features)
+
+SELECT
+  client,
+  feature,
+  num_urls,
+  total_urls,
+  pct_urls AS pct_urls
```
tunetheweb, line 9:
```suggestion
  pct_urls
```

sql/2021/privacy/nb_websites_with_mediadevices_blink_usage.sql
```diff
@@ -0,0 +1,37 @@
...
+  (
+    feature LIKE '%MediaDevices%' OR
+    feature LIKE '%EnumerateDevices%' OR
+    feature LIKE '%GetUserMedia%' OR
+    feature LIKE '%GetDisplayMedia%' OR
+    feature LIKE '%Camera%' OR
+    feature LIKE '%Microphone%'
+  )
+ORDER BY
+  2 ASC, 1 ASC
```
tunetheweb, line 23:
```suggestion
ORDER BY
  feature,
  client
```

sql/2021/privacy/nb_websites_with_privacy_sandbox_blink_usage.sql
```diff
@@ -0,0 +1,28 @@
...
+    feature = 'V8Navigator_RunAdAuction_Method' OR
+    feature = 'ConversionRegistration' OR
+    feature = 'ImpressionRegistration' OR
+    feature = 'ConversionAPIAll' OR
+    feature = 'SamePartyCookieAttribute' OR
+    feature = 'V8Document_HasTrustToken_Method' OR
+    feature = 'HTMLFencedFrameElement'
+  )
+ORDER BY
+  2 ASC, 1 ASC
```
tunetheweb, line 28:
```suggestion
ORDER BY
  feature,
  client
```

sql/2021/privacy/most_common_cmps_for_iab_tcf_v2.sql
```diff
@@ -0,0 +1,34 @@
...
+  totals
+USING (_TABLE_SUFFIX)
+WHERE
+  JSON_VALUE(JSON_VALUE(payload, "$._privacy"), '$.iab_tcf_v2.data.cmpId') IS NOT NULL
+GROUP BY
+  client,
+  total_websites,
+  cmpId
+ORDER BY
+  client ASC,
```
tunetheweb, line 33: As `ASC` is default there's no need to explicitly say this and we generally tend not to bother with it. Though appreciate some might prefer to be explicit.

sql/2021/privacy/most_common_strings_for_iab_usp.sql
```diff
@@ -0,0 +1,44 @@
...
+USING (client)
+WHERE
+  JSON_VALUE(metrics, '$.uspString') IS NOT NULL
+GROUP BY
+  client,
+  uspString,
+  total_websites
+ORDER BY
+  client ASC,
+  nb_websites DESC
```
tunetheweb, line 44: Can simplify this one:

```suggestion
WITH totals AS (
  SELECT
    _TABLE_SUFFIX,
    COUNT(0) AS total_websites
  FROM
    `httparchive.pages.2021_07_01_*`
  GROUP BY
    _TABLE_SUFFIX
)

SELECT
  _TABLE_SUFFIX AS client,
  JSON_QUERY(JSON_VALUE(payload, "$._privacy"), "$.iab_usp.privacy_string.uspString") AS uspString,
  COUNT(0) AS nb_websites,
  total_websites,
  COUNT(0) / ANY_VALUE(total_websites) AS pct_websites
FROM
  `httparchive.pages.2021_07_01_*`
JOIN
  totals
USING (_TABLE_SUFFIX)
WHERE
  JSON_QUERY(JSON_VALUE(payload, "$._privacy"), "$.iab_usp.privacy_string.uspString") IS NOT NULL
GROUP BY
  client,
  uspString,
  total_websites
ORDER BY
  client ASC,
  nb_websites DESC
```

sql/2021/privacy/nb_websites_using_each_cmp.sql
```diff
@@ -0,0 +1,47 @@
...
+    cmp_app,
+    COUNT(DISTINCT url) OVER (PARTITION BY client) AS total_pages,
+    COUNT(DISTINCT url) / COUNT(DISTINCT url) OVER (PARTITION BY client) AS pct_pages_with_cmp
+  FROM
+    apps
+  GROUP BY
+    client,
+    url,
+    cmp_app
+)
```
tunetheweb, line 31: Total pages when grouped bu the same url? Does this make sense?

sql/2021/privacy/nb_websites_using_each_fingerprinting.sql
```diff
@@ -0,0 +1,47 @@
...
+    fingerprinting_app,
+    COUNT(DISTINCT url) OVER (PARTITION BY client) AS total_pages,
+    COUNT(DISTINCT url) / COUNT(DISTINCT url) OVER (PARTITION BY client) AS pct_pages_with_fingerprinting
+  FROM
+    apps
+  GROUP BY
+    client,
+    url,
+    fingerprinting_app
+)
```
tunetheweb, line 31: Ditto

sql/2021/privacy/nb_websites_using_each_geolocation.sql
```diff
@@ -0,0 +1,47 @@
...
+    geolocation_app,
+    COUNT(DISTINCT url) OVER (PARTITION BY client) AS total_pages,
+    COUNT(DISTINCT url) / COUNT(DISTINCT url) OVER (PARTITION BY client) AS pct_pages_with_geolocation
+  FROM
+    apps
+  GROUP BY
+    client,
+    url,
+    geolocation_app
+)
```
tunetheweb, line 31: Same again

sql/2021/privacy/nb_websites_using_each_retargeting.sql
```diff
@@ -0,0 +1,47 @@
...
+    retargeting_app,
+    COUNT(DISTINCT url) OVER (PARTITION BY client) AS total_pages,
+    COUNT(DISTINCT url) / COUNT(DISTINCT url) OVER (PARTITION BY client) AS pct_pages_with_retargeting
+  FROM
+    apps
+  GROUP BY
+    client,
+    url,
+    retargeting_app
+)
```
tunetheweb, line 31: And again

sql/2021/privacy/nb_websites_with_gpc_well_known.sql
```diff
@@ -0,0 +1,22 @@
...
+  client,
+  COUNT(0) AS nb_websites -- crawled sites containing at leat one origin trial
+FROM
+  pages_well_known
+WHERE
+  JSON_VALUE(pages_well_known.metrics, '$."/.well-known/gpc.json".found') = "true"
+GROUP BY
+  client
+ORDER BY
+  client ASC
```
tunetheweb, line 22: Can this be simplified into one query?

Should it join with a totals query (so back up to two tables!) to give a percentage?

sql/2021/privacy/nb_websites_with_origin_trial_from_token.sql
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  // ignore expiry date: evaluation happens retroactively
+  //// let expiryDate = new Date(expiry * 1000);
+  //// expiryElem = expiryDate;
+  //// if (expiryDate < new Date()) {
+  ////   origin_trial_metadata.validityElem = 'Expired';
+  //// return origin_trial_metadata;
+  //// }
+
```
tunetheweb, line 142: Is this needed? If not can we remove?

```suggestion
```

sql/2021/privacy/nb_websites_with_privacy_service.sql
```diff
@@ -0,0 +1,35 @@
...
+    url,
+    LOGICAL_OR(category = "Cookie compliance") AS with_cookie_compliance,
+    LOGICAL_OR(category = "Browser fingerprinting") AS with_browser_fingerprinting,
+    LOGICAL_OR(category = "Retargeting") AS with_retargeting,
+    LOGICAL_OR(category = "Geolocation") AS with_geolocation
+  FROM `httparchive.technologies.2021_07_01_*`
+  GROUP BY
+    client,
+    url
+)
```
tunetheweb, line 17: Does this need a separate CTE?

tunetheweb (changes requested):
Some more feedback for you.

Not far to go now - I promise!

sql/2021/privacy/most_common_cmps_for_iab_tcf_v2.sql
max-ostapenko: I don't see a value changing it to implicit here.

tunetheweb: Main value would be in being consistent across chapters, so there's not an expectation of lack of explicit ASC means. It's not one we've enforced before to be honest, but looking across the current queries I can only find 12 instances of being explicit with ASC so definitely in the minority.

sql/2021/privacy/nb_websites_using_each_cmp.sql
max-ostapenko: While it's a window function there is no calculation error, but it may be simplified.

sql/2021/privacy/nb_websites_using_each_fingerprinting.sql
max-ostapenko: Adjusted the same way as above

sql/2021/privacy/nb_websites_with_gpc_well_known.sql
max-ostapenko: Adjusted to a version with percentages.

sql/2021/privacy/nb_websites_with_privacy_service.sql
max-ostapenko: Adjusted to a single SELECT

sql/2021/privacy/trackers_whotracksme.py
max-ostapenko: HaveIBeenPwned is on CC licence - https://haveibeenpwned.com/API/v3#License
and whotracks me is on MIT - https://github.com/ghostery/whotracks.me/blob/master/LICENSE.md
So I suppose there is no issue with licensing, right?

@paulcalvano how would you prefer us to prepare this data for BigQuery?

sql/2021/privacy/most_common_cmps_for_iab_tcf_v2.sql
```diff
@@ -0,0 +1,34 @@
...
+  FROM
+    `httparchive.pages.2021_07_01_*`
+  GROUP BY
+    _TABLE_SUFFIX
+)
+
+SELECT
+  _TABLE_SUFFIX AS client,
+  JSON_VALUE(JSON_VALUE(payload, "$._privacy"), '$.iab_tcf_v2.data.cmpId') AS cmpId,
+  COUNT(0) AS nb_websites,
```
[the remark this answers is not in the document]
max-ostapenko: @tunetheweb Should the same be applied to `pct_` as `percentage_of_`?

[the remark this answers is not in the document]
tunetheweb: Well the % usually gives that away if you format the columns in sheets.

We normally put it at the end:
- `number_of_trackers`
- `number_of_trackers_pct`

Though to be honest the `number_of_` is a bit superfluous and that could simply be:
- `trackers`
- `trackers_pct`

I guess the `nb_` just threw me for a bit there. But sometimes I’m just a bit slow 😁

sql/2021/privacy/most_common_featurepolicy_permissionspolicy_directive_values.sql
tunetheweb: Discussed with the other maintainers and we're gonna have to insist on the ranking by top XXX, rather than each slice. For all the reasons I gave above. Will add to the Analysts Guide when I get a second so clear in future.

So would you mind correcting the query with this in mind?

sql/2021/privacy/most_common_cmps_for_iab_tcf_v2.sql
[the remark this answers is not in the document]
tunetheweb: See it's updated so resolving this comment.

sql/2021/privacy/trackers_whotracksme.py
tunetheweb: Licencing looks good to me!

If you provide the data in a Google Sheets document I imagine @paulcalvano can get that inserted for you!

rviscomi:
Hi everyone, checking in to see if I can help move things along.

Looks like there are unresolved changes requested by the reviewer. Assigning to @max-ostapenko. Max, feel free to reassign back to @tunetheweb and/or re-request a review when this is ready. Not much more to go!

max-ostapenko:
Thanks, I'll resolve the rest of adjustments today.

sql/2021/privacy/trackers_whotracksme.py
max-ostapenko: @paulcalvano could you upload [this CSV](https://github.com/HTTPArchive/almanac.httparchive.org/blob/privacy-sql-2021/sql/2021/privacy/trackers.csv) to the BigQuery please, so that we can relate to it in the queries [for example](https://github.com/HTTPArchive/almanac.httparchive.org/pull/2227/files#diff-99583fbced3f94f4f9705727dfc2f600d12225cb24cb7c1a2351ad6c44046679R4)?

VictorLeP:
Does the [`urlShort` issue](https://github.com/HTTPArchive/almanac.httparchive.org/pull/2262#discussion_r721779951) apply to us as well? We use in the WhoTracks.me based queries: [1](https://github.com/HTTPArchive/almanac.httparchive.org/blob/privacy-sql-2021/sql/2021/privacy/most_common_tracker_categories.sql), [2](https://github.com/HTTPArchive/almanac.httparchive.org/blob/privacy-sql-2021/sql/2021/privacy/most_common_trackers.sql), [3](https://github.com/HTTPArchive/almanac.httparchive.org/blob/privacy-sql-2021/sql/2021/privacy/number_of_websites_with_nb_trackers.sql)

tunetheweb:
> Does the [`urlShort` issue](https://github.com/HTTPArchive/almanac.httparchive.org/pull/2262#discussion_r721779951) apply to us as well? We use in the WhoTracks.me based queries: [1](https://github.com/HTTPArchive/almanac.httparchive.org/blob/privacy-sql-2021/sql/2021/privacy/most_common_tracker_categories.sql), [2](https://github.com/HTTPArchive/almanac.httparchive.org/blob/privacy-sql-2021/sql/2021/privacy/most_common_trackers.sql), [3](https://github.com/HTTPArchive/almanac.httparchive.org/blob/privacy-sql-2021/sql/2021/privacy/number_of_websites_with_nb_trackers.sql)

Yes I’d say so.

VictorLeP:
> > Does the [`urlShort` issue](https://github.com/HTTPArchive/almanac.httparchive.org/pull/2262#discussion_r721779951) apply to us as well? We use in the WhoTracks.me based queries: [1](https://github.com/HTTPArchive/almanac.httparchive.org/blob/privacy-sql-2021/sql/2021/privacy/most_common_tracker_categories.sql), [2](https://github.com/HTTPArchive/almanac.httparchive.org/blob/privacy-sql-2021/sql/2021/privacy/most_common_trackers.sql), [3](https://github.com/HTTPArchive/almanac.httparchive.org/blob/privacy-sql-2021/sql/2021/privacy/number_of_websites_with_nb_trackers.sql)
> 
> Yes I’d say so.

I'm not sure it applies actually: we use `urlShort` to get the URLs (actually domains) of any resources that were loaded from a page, including resources on other domains. There would therefore not be anything in `summary_pages` of use to us.

tunetheweb:
> > > Does the [`urlShort` issue](https://github.com/HTTPArchive/almanac.httparchive.org/pull/2262#discussion_r721779951) apply to us as well? We use in the WhoTracks.me based queries: [1](https://github.com/HTTPArchive/almanac.httparchive.org/blob/privacy-sql-2021/sql/2021/privacy/most_common_tracker_categories.sql), [2](https://github.com/HTTPArchive/almanac.httparchive.org/blob/privacy-sql-2021/sql/2021/privacy/most_common_trackers.sql), [3](https://github.com/HTTPArchive/almanac.httparchive.org/blob/privacy-sql-2021/sql/2021/privacy/number_of_websites_with_nb_trackers.sql)
> > 
> > 
> > Yes I’d say so.
> 
> I'm not sure it applies actually: we use `urlShort` to get the URLs (actually domains) of any resources that were loaded from a page, including resources on other domains. There would therefore not be anything in `summary_pages` of use to us.

Ah good point. If looking at resources a page loads, then fine to use this. If looking at the top level page it's not fine to use this.

rviscomi:
@max-ostapenko could you give a status update for this PR? How close are we to approval/merging?

rviscomi:
I've just added the table to `httparchive.almanac.whotracksme`. Hopefully that unblocks you all.

paulcalvano:
Sorry for not responding sooner.   I just wrote a simple nodejs script to convert the breaches JSON to a file I could upload into BigQuery.   The breaches table is at `httparchive:almanac.breaches` and it also contains a `date` column so we can append to it next year.

tunetheweb:
Great stuff @paulcalvano !

Could you add the script to this branch in the sql/utils folder?

@max-ostapenko guess we can remove your Python script then?

PR #2252, opened 2021-06-19
Breaking "Service worker usage" into separate paragraphs

demianrenzulli:
In the 2020 PWA chapter, the [Service worker usage](https://almanac.httparchive.org/en/2020/pwa#service-worker-usage) contains two groups of metrics: the first one refers to the number and percentage of URLs controlled by service workers, and the second one to the % of page loads.

While reading the section, I was expecting the chart to appear immediately after the metric, but it's being shown at the end, creating confusion of which of the two groups it represents.

This small change aims to fix that.


tunetheweb:
Ah forgot there were a few translations of this and we have a check to keep them in sync. Could you make same changes there? Appreciate you don’t speak those languages (I assume!) but hopefully with a bit of Google Translate’s help you’ll be able to figure out where to put the same break in those?

demianrenzulli:
Thanks for letting me know, @tunetheweb. I just pushed the changes for JP and NL, which are the other translations.

I didn't know the Almanac was localized in this way. The PWA team we have formed is quite diverse, so we might have folks collaborating to translate it into their own languages.

tunetheweb (approved):

tunetheweb:
Thanks @demianrenzulli ! Will include in next release!

PR #2255, opened 2021-06-22
ES #498: 2020 SEO Spanish Translation

carloscastromx:
Makes progress on #498

This is my first translated chapter, I would appreciate any comments.

I also found some typos in the source content in English and fixed them in this pull request for review.


tunetheweb (changes requested):
Looks pretty good @carloscastromx ! Just a few small things to fix (which should also sort out that failing test) and then I'm happy from a technical point of view – but can't comment on the Spanish I'm afraid! 😁

carloscastromx:
Thanks for the review! I think it's all ok @bazzadp 😄

tunetheweb (approved):
LGTM!

@HTTPArchive/spanish-translators let me know if anyone has time to review this. Otherwise if I hear nothing in next few days then will merge away.

Issue #2158, opened 2021-04-27
Resource Hints 2021


rviscomi:
# Part IV Chapter 21: Resource Hints

![Resource Hints illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/resource-hints/hero_lg.jpg)

If you're interested in contributing to the Resource Hints chapter of the 2021 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| --- | ------- | --------- | -------- | ------- | -- |
| @kevinfarrugia | @kevinfarrugia | @siakaramalegos @Zizzamia @samarpanda | @Nithanaroy | - | @rviscomi |

<details>
<summary><b>Expand for more information about each role</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
- **[Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)** are responsible for researching the stats and trends used throughout the Almanac. Analysts work closely with authors and reviewers during the planning phase to give direction on the types of stats that are possible from the dataset, and during the analyzing/writing phases to ensure that the stats are used correctly.
[...]

PR #2264, opened 2021-07-01
Update README.md

kevinfarrugia:
_This is a draft PR to keep track of the metrics whose queries need to be written for the Resource Hints 2021 chapter._ 

Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/2158 based on the [chapter outline](https://docs.google.com/document/d/1zIKdkS65X9u3rGZOn1kG0rk9YveFb0zhdORBFOoIgQQ/edit)
cc @Nithanaroy  

### Incorrect Usage
- [ ] Preloading multiple font types. 
- [ ] Unsupported Type
- [ ] Incorrect type.
- [ ] Missing crossorigin on files may create a duplicate request for the same file
- [ ] Unused or low-priority files:
- [ ] Are users preloading images which are not in the viewport?

### DIFF
--- a/sql/2021/resource-hints/README.md
+++ b/sql/2021/resource-hints/README.md
@@ -8,3 +8,10 @@
 
   Analysts: if helpful, you can use this README to give additional info about the queries.
 -->
+
+Resources:
+
+- [Planning doc](https://docs.google.com/document/d/1zIKdkS65X9u3rGZOn1kG0rk9YveFb0zhdORBFOoIgQQ/edit?usp=sharing)
+- [2019 chapter](https://almanac.httparchive.org/en/2019/resource-hints)
+- [2018 chapter](https://almanac.httparchive.org/en/2018/resource-hints)
+


sql/2021/resource-hints/preload-redundant-font-types.sql
```diff
@@ -0,0 +1,73 @@
...
+    rd as redundantDownloads,
+    count(0) as freq,
+    AS_PERCENT(COUNT(0), SUM(COUNT(0)) OVER (PARTITION BY client)) AS pct_m405
+FROM (
+    SELECT 
+        _TABLE_SUFFIX AS client,
+        JSON_QUERY(payload, '$._almanac') AS almanac, 
+        getUnnecessaryFontDownloadsCount(JSON_EXTRACT_SCALAR(payload, '$._almanac')) AS rd 
+    FROM
+        `httparchive.sample_data.pages*`
```
Nithanaroy, line 64: Not sure which tables we use here eventually

kevinfarrugia: Yes, this would be changed to once the data is available:
```
```

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
```diff
@@ -0,0 +1,77 @@
...
+    ROUND(SAFE_DIVIDE(freq, total), 4)
+);
+
+# retruns the number of redundant font downloads
+# use https://replit.com/@NitinPasumarthy/almanac-2021-redundant-font-preloads#index.js to run the JS on several test payloads
+CREATE TEMPORARY FUNCTION getUnnecessaryFontDownloadsCount(almanac_string STRING)
+RETURNS INT64 LANGUAGE js AS '''
+try {
+    var almanac = JSON.parse(almanac_string)
+    if (Array.isArray(almanac) || typeof almanac != 'object') return -1;
```
rviscomi, line 14:
```suggestion
    if (Array.isArray(almanac) || typeof almanac != 'object') return null;
```

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
```diff
@@ -0,0 +1,77 @@
...
+    rd as redundantDownloads,
+    count(0) as freq,
+    AS_PERCENT(COUNT(0), SUM(COUNT(0)) OVER (PARTITION BY client)) AS pct
+FROM (
+    SELECT 
+        _TABLE_SUFFIX AS client,
+        JSON_QUERY(payload, '$._almanac') AS almanac, 
+        getUnnecessaryFontDownloadsCount(JSON_EXTRACT_SCALAR(payload, '$._almanac')) AS rd 
+    FROM
+        `httparchive.sample_data.pages*`
```
rviscomi, line 69: July dataset

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
```diff
@@ -0,0 +1,77 @@
...
+SELECT
+    client,
+    rd as redundantDownloads,
+    count(0) as freq,
+    AS_PERCENT(COUNT(0), SUM(COUNT(0)) OVER (PARTITION BY client)) AS pct
+FROM (
+    SELECT 
+        _TABLE_SUFFIX AS client,
+        JSON_QUERY(payload, '$._almanac') AS almanac, 
+        getUnnecessaryFontDownloadsCount(JSON_EXTRACT_SCALAR(payload, '$._almanac')) AS rd
```
rviscomi, line 67: Same comment about `rd`

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
```diff
@@ -0,0 +1,77 @@
...
+    });
+    
+    // only include sites which preload at least one font file
+    if (fontFiles.length === 0) {
+        return -1;
+    }
+    
+    var fontsWithExt = {} // {"woff2": Set( 'font1', 'font2' ), "ttf": Set( 'font2' )}
+    for(var i = 0; i < fontFiles.length; i++) {
+        var [fontName, ext] = fontFiles[i].split(".")
```
rviscomi, line 34: A valid file name could have multiple dots in it. Maybe search for the index of the last dot and use that position to split the string in two?

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
```diff
@@ -0,0 +1,77 @@
+#standardSQL
```
rviscomi, line 1: General comment about the file name

I'd suggest something simpler like `redundant_preload_font_downloads.sql`. The `pages_almanac_by_device_and` part is overly descriptive.

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
```diff
@@ -0,0 +1,77 @@
...
+FROM (
+    SELECT 
+        _TABLE_SUFFIX AS client,
+        JSON_QUERY(payload, '$._almanac') AS almanac, 
+        getUnnecessaryFontDownloadsCount(JSON_EXTRACT_SCALAR(payload, '$._almanac')) AS rd 
+    FROM
+        `httparchive.sample_data.pages*`
+)
+WHERE 
+    rd > -1
```
rviscomi, line 72:
```suggestion
    rd IS NOT NULL
```

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
```diff
@@ -0,0 +1,77 @@
...
+}
+catch {
+    return -1
+}
+''';
+
+SELECT
+    client,
+    rd as redundantDownloads,
+    count(0) as freq,
```
rviscomi, line 61: Nit: uppercase for function names and keywords
```suggestion
    rd AS redundantDownloads,
    COUNT(0) AS freq,
```

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
```diff
@@ -0,0 +1,77 @@
...
+        } else {
+            fontsWithExt[ext] = new Set([fontName])
+        }
+    }
+
+    var redundantDownloads = 0
+    if("woff2" in fontsWithExt) {
+        for(var ext of Object.keys(fontsWithExt) ) {
+            if(ext !== "woff2") {
+                redundantDownloads += [...fontsWithExt["woff2"]].filter(f => fontsWithExt[ext].has(f)).length
```
rviscomi, line 46: Not sure I understand how this works. How can the `woff2` bucket contain anything other than font files with `woff2` extensions?

rviscomi (changes requested):

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
Nithanaroy: Ah good catch. Not sure, why I missed that! 🤦

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
Nithanaroy: Sorry for a counterintuitive implementation. Here all we are doing is 
```js
fontsWithExt['woff2'] - fontsWithExt['woff']
fontsWithExt['woff2'] - fontsWithExt['ttf']
...
```
where `fontsWithExt[extension]` is a **set** of file names with that extension. This way we obtain the total number of font downloads where both woff and woff2 are downloaded which is redundant. The hypothesis is that, browsers that support preload, also support woff2. So, preloading woff in those scenarios is a waste.

rviscomi: Oh ok, so given `foo.woff` and `foo.woff2`, "foo" will appear in both the `woff` and `woff2` buckets and counted as a redundant download. Assuming that the base name of the file is the same.

In last year's Fonts chapter, they looked at the font metadata itself to see what the canonical family name of the font was. That's more complicated but it may yield more reliable results, for example to handle the same font with different base names and formats.

tunetheweb:
Hi @kevinfarrugia , I've just merged a SQL Linter into `main`. When you get a chance could you merge into this branch so your new queries will be linted according to the Web Almanac style?

tunetheweb:
FYI @kevinfarrugia I have just merged `main` into this branch and run the linting fixes.

kevinfarrugia:
Thank you @tunetheweb . Will be jumping back onto this and catching up in the coming weeks as I took a few weeks of vacation. :)

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
Nithanaroy: Oh nice. I did not know about this concept. Do you remember which query has this implementation? I did a quick glance in ./sql/2020/resource-hints folder but nothing stood out matching this description.

Nithanaroy:
@rviscomi what's the practice with resolving comments? Should I resolve or the comment author does?

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
rviscomi: It was the Fonts chapter, so you'd need to look in /sql/2020/fonts. Here's an example of a query that extracted name metadata from the font file itself: https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2020/fonts/04_14.variable_font.sql

rviscomi (commented):
> @rviscomi what's the practice with resolving comments? Should I resolve or the comment author does?

If there are still changes required to resolve the comments, please leave them unresolved.

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
Nithanaroy: Thanks Rick. I'm unable to follow how I can use the info from `JSON_EXTRACT(payload, '$._font_details')` to understand if two fonts have the same base name. Also, we only wish to consider `foo.woff` and `foo.woff2` redundant only if they are in link-preload tags.

```json
{
    "table_sizes": {
        "GSUB": 59,
        "OS/2": 65,
        "cmap": 194,
        "glyf": 1895,
        "head": 49,
        "hhea": 29,
        "hmtx": 63,
        "loca": 60,
        "maxp": 31,
        "name": 296,
        "post": 207
    },
    "names": {
        "1": "iconFont",
        "2": "Regular",
        "3": "iconFont",
        "4": "iconFont",
        "5": "Version 1.0",
        "6": "iconFont",
        "10": "Generated by svg2ttf from Fontello project.",
        "11": "http://fontello.com"
    },
    "OS2": {
        "version": 1,
        "xAvgCharWidth": 540,
        "usWeightClass": 400,
        "usWidthClass": 5,
        "fsType": 0,
        "ySubscriptXSize": 813,
        "ySubscriptYSize": 350,
        "ySubscriptXOffset": 0,
        "ySubscriptYOffset": 70,
        "ySuperscriptXSize": 813,
        "ySuperscriptYSize": 350,
        "ySuperscriptXOffset": 0,
        "ySuperscriptYOffset": 240,
        "yStrikeoutSize": 24,
        "yStrikeoutPosition": 129,
        "sFamilyClass": 0,
        "ulUnicodeRange1": 0,
        "ulUnicodeRange2": 0,
        "ulUnicodeRange3": 0,
        "ulUnicodeRange4": 0,
        "achVendID": "PfEd",
        "fsSelection": 64,
        "usFirstCharIndex": 57345,
        "usLastCharIndex": 57372,
        "sTypoAscender": 500,
        "sTypoDescender": 0,
        "sTypoLineGap": 45,
        "usWinAscent": 501,
        "usWinDescent": 4,
        "ulCodePageRange1": 1,
        "ulCodePageRange2": 0,
        "panose": {
            "bFamilyType": 2,
            "bSerifStyle": 0,
            "bWeight": 5,
            "bProportion": 3,
            "bContrast": 0,
            "bStrokeVariation": 0,
            "bArmStyle": 0,
            "bLetterForm": 0,
            "bMidline": 0,
            "bXHeight": 0
        }
    },
    "post": {
        "formatType": 2,
        "italicAngle": 0,
        "underlinePosition": 5,
        "underlineThickness": 0,
        "isFixedPitch": 0,
        "minMemType42": 0,
        "maxMemType42": 0,
        "minMemType1": 0,
        "maxMemType1": 0
    },
    "counts": {
        "num_cmap_codepoints": 28,
        "num_glyphs": 29
    }
}
```

rviscomi: In this case the name of the font is `iconFont`, so you could compare that against other fonts, as opposed to comparing them purely by filename.

rviscomi:
@kevinfarrugia do you feel comfortable that you'd be able to get these queries written by the milestone deadline (end of the month)? If not, anything we can help with?

Nithanaroy:
@kevinfarrugia need a quick review of db47b9b commit. Want to make sure I understand the problem you want to capture correctly.

Nithanaroy:
@kevinfarrugia d0c29fa returned empty results at least on the sample dataset. Again need your quick review to make sure I understood your intent clearly.

Nithanaroy:
Added the grouped results in https://docs.google.com/spreadsheets/d/1Mw6TjkIClRtlZPHbij5corOZbaSUp-vgTVq3Ig18IwQ/edit#gid=1764929664 using the bc47cb2 commit

sql/2021/resource-hints/pages_almanac_by_device_dns-prefetch_domains.sql
```diff
@@ -0,0 +1,67 @@
...
+            SELECT
+              _TABLE_SUFFIX AS client,
+              getResourceHints(payload, "dns-prefetch") AS hints
+            FROM
+              `httparchive.pages.2021_07_01_*`
+        )
+    )
+    CROSS JOIN UNNEST(value) AS href
+    WHERE
+      ARRAY_LENGTH(value) > 0
```
Nithanaroy, line 55: Do we want to include `hints` that don't have hrefs in the final fractions? If yes, we would need `LEFT OUTER JOIN` without this `WHERE` condition.

```sql
SELECT id, COUNT(0)
FROM (
  WITH abc AS (
  SELECT 1 AS id, [1, 2, 3] AS vals
  UNION ALL
  SELECT 2 AS id, NULL AS vals
  )
  SELECT id, val
  FROM abc
  LEFT JOIN UNNEST(vals) AS val
  -- CROSS JOIN UNNEST(vals) AS val
)
GROUP BY id
```

kevinfarrugia: Thank you Nitin. We only want to include hints that have an `href` value as otherwise they would be invalid. However I am currently excluding relative hrefs (such as `/my-script.js`) and only considering absolute ones, such as `https://example.com/my-script.js`. It might make sense to include both for the `preload` resource hint.

kevinfarrugia: Updated queries to remove `WHERE domain IS NOT NULL`

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
kevinfarrugia: @rviscomi Sorry I missed this comment. The `pages_almanac_by_device`...etc was copied from some of last years chapters (specifically Markup 2020) as I was told that the queries are a good example to copy. Is there a preferred naming convention?

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
kevinfarrugia: @tunetheweb We are taking care of
> if a font is preloaded but that font is never used at all (even in another format). I believe there's a Lighthouse test for this.
in the query [sql/2021/resource-hints/pages_consoleLog_by_device_unused_preload.sql](https://github.com/HTTPArchive/almanac.httparchive.org/pull/2264/files#diff-6afb590c58532e8c89bc9993f28d8c11da5c033b309e64fe23649a46dd9a9125). This covers images, fonts...etc.

We created a separate query for the duplicate fonts because they are not flagged in the console or in Lighthouse and all fonts are downloaded, making it quite obscure to find. 

Demo: https://kevinfarrugia.github.io/resource-hints/preload-multiple-fonts

Regarding iframes, haven't considered them at all tbh.

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
rviscomi: No real convention, just short and simple. More about "what am I measuring" rather than "how am I measuring it". pages/almanac describe the how and device is measured on every metric so redundant.

sql/2021/resource-hints/pages_almanac_by_device_and_redundant_preload_font_downloads.sql
kevinfarrugia: I will check whether we are able to update the query to use the metadata to get a more accurate result and compare it with the sample_data.

kevinfarrugia: I was unable to use the font metadata since we are limited to the data available in the preload link (i.e. the `href` attribute) and will go ahead with the filename assumption for now.

sql/2021/resource-hints/consoleLog_incorrect_type.sql
```diff
@@ -0,0 +1,34 @@
+#standardSQL
+# returns the number of pages which preload a resource of the incorrect script type
+
+# helper to create percent fields
+CREATE TEMP FUNCTION AS_PERCENT (freq FLOAT64, total FLOAT64) RETURNS FLOAT64 AS (
+  ROUND(SAFE_DIVIDE(freq, total), 4)
+);
+
```
rviscomi, line 8: It's preferred to calculate the raw quotient in SQL and format it as a percent in Sheets. This gives you more control over the number of decimal places depending on the values. So I'd suggest removing it from all of these queries.

Also this function appears to be unused here.

sql/2021/resource-hints/consoleLog_incorrect_type.sql
```diff
@@ -0,0 +1,34 @@
...
+SELECT
+  client,
+  ARRAY_LENGTH(value) AS freq,
+  COUNT(0) AS total
```
rviscomi, line 12: Are you looking for the number of pages having this value, or just the number of times it appears in the logs?

If number of pages, you'd need something like `COUNT(DISTINCT url)` where `url` is included in the inner selects below.

I'd suggest renaming this to something like `freq` (frequency) or `pages`, and reserving `total` for the denominator, which you can use to calculate the percentage. So consider something like this:
```suggestion
  COUNT(0) AS freq,
  SUM(COUNT(0)) OVER (PARTITION BY client) AS total,
  COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client) AS pct
```

sql/2021/resource-hints/consoleLog_incorrect_type.sql
```diff
@@ -0,0 +1,34 @@
...
+FROM (
+  SELECT
+    client,
+    REGEXP_EXTRACT_ALL(consoleLog, r'A preload for (.*?) is found, but is not used because the script type does not match.') AS value
```
rviscomi, line 16: What does the `?` do in this regex?

sql/2021/resource-hints/consoleLog_missing_crossorigin.sql
```diff
@@ -0,0 +1,45 @@
...
+    "fonts.googleapis.com",
+    TRIM(TRIM(REGEXP_EXTRACT(href, r'\.[0-9a-z]+(?:[\?#]|$)'), '?'), '#')
+  )
+);
+
+SELECT
+  client,
+  GET_TYPE(TRIM(href, '\'')) AS type,
+  COUNT(0) AS total,
+  AS_PERCENT(COUNT(0), SUM(COUNT(0)) OVER (PARTITION BY client)) AS pct
```
rviscomi, line 22:
```suggestion
  COUNT(0) AS freq,
  SUM(COUNT(0)) OVER (PARTITION BY client) AS total,
  COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client) AS pct
```

sql/2021/resource-hints/consoleLog_missing_crossorigin.sql
```diff
@@ -0,0 +1,45 @@
+#standardSQL
+# returns the file types for preload tags without the required crossorigin attribute
+
+# helper to create percent fields
+CREATE TEMP FUNCTION AS_PERCENT (freq FLOAT64, total FLOAT64) RETURNS FLOAT64 AS (
+  ROUND(SAFE_DIVIDE(freq, total), 4)
+);
+
```
rviscomi, line 8:
```suggestion
```

sql/2021/resource-hints/consoleLog_missing_crossorigin.sql
```diff
@@ -0,0 +1,45 @@
...
+# helper to convert the href into a type
+CREATE TEMP FUNCTION GET_TYPE (href STRING) RETURNS STRING AS (
+  IF(
+    REGEXP_CONTAINS(href, r'fonts\.googleapis\.com'),
+    "fonts.googleapis.com",
+    TRIM(TRIM(REGEXP_EXTRACT(href, r'\.[0-9a-z]+(?:[\?#]|$)'), '?'), '#')
```
rviscomi, line 14: The function sounds like it's returning a "type" but the value seems more like a host name or path name. Do any of the [`NET`](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#net_functions) functions do what you're looking for (HOST, REG_DOMAIN)?

sql/2021/resource-hints/consoleLog_unused_preload.sql
```diff
@@ -0,0 +1,33 @@
...
+
+SELECT
+  client,
+  ARRAY_LENGTH(value) AS freq,
+  COUNT(0) AS total,
+  AS_PERCENT(COUNT(0), SUM(COUNT(0)) OVER (PARTITION BY client)) AS pct
+FROM (
+  SELECT
+    client,
+    REGEXP_EXTRACT_ALL(consoleLog, r'The resource (.*?) was preloaded using link preload but not used within a few seconds from the window\'s load event') AS value
```
rviscomi, line 17: Aside: I think it's really clever that you're analyzing the console logs as a heuristic for resource hint optimization 😃

sql/2021/resource-hints/dns-prefetch_domains.sql
```diff
@@ -0,0 +1,65 @@
...
+CREATE TEMPORARY FUNCTION getHrefs(payload STRING)
+RETURNS ARRAY<STRING>
+LANGUAGE js AS '''
+try {
+    var $ = JSON.parse(payload);
+    return $.map(n => n.href);
+} catch (e) {
+    return [];
+}
+''' ;
```
rviscomi, line 30: The SQL takes the output of the first function and passes it into the second function, so they could be combined.
```suggestion
CREATE TEMPORARY FUNCTION getResourceHintHrefs(payload STRING, hint STRING)
RETURNS ARRAY<STRING>
LANGUAGE js AS '''
try {
  var $ = JSON.parse(payload);
  var almanac = JSON.parse($._almanac);
  return almanac['link-nodes'].nodes.filter(link => link.rel.toLowerCase() == hint).map(link => link.href);
} catch (e) {
  return [];
}
''';
```

sql/2021/resource-hints/dns-prefetch_domains.sql
```diff
@@ -0,0 +1,65 @@
...
+          getHrefs(hints) AS value
+        FROM (
+            SELECT
+              _TABLE_SUFFIX AS client,
+              getResourceHints(payload, "dns-prefetch") AS hints
+            FROM
+              `httparchive.pages.2021_07_01_*`
+        )
+    )
+    CROSS JOIN UNNEST(value) AS href
```
rviscomi, line 53: With the functions combined, this section could be consolidated
```suggestion
    SELECT
      _TABLE_SUFFIX AS client,
      REGEXP_EXTRACT(href, r'\/\/(.+?)$') AS domain
    FROM
      `httparchive.pages.2021_07_01_*`,
      UNNEST(getResourceHintHrefs(payload, "dns-prefetch") AS href
```

sql/2021/resource-hints/dns-prefetch_domains_by_url.sql
```diff
@@ -0,0 +1,79 @@
+#standardSQL
+# Most popular domains users dns-prefetch to
+# capped to one hit per url to avoid having the results skewed by websites which dns-prefetch the same domain many times (!)
+
+# helper to create percent fields
+CREATE TEMP FUNCTION AS_PERCENT (freq FLOAT64, total FLOAT64) RETURNS FLOAT64 AS (
```
rviscomi, line 6: Remove

sql/2021/resource-hints/dns-prefetch_domains_by_url.sql
```diff
@@ -0,0 +1,79 @@
...
+  ROUND(SAFE_DIVIDE(freq, total), 4)
+);
+
+CREATE TEMPORARY FUNCTION getResourceHints(payload STRING, hint STRING)
```
rviscomi, line 10: Combine

sql/2021/resource-hints/dns-prefetch_domains_by_url.sql
```diff
@@ -0,0 +1,79 @@
...
+} catch (e) {
+    return [];
+}
+''' ;
+
+SELECT
+  client,
+  domain,
+  COUNT(0) AS freq,
+  AS_PERCENT(COUNT(0), SUM(COUNT(0)) OVER()) AS pct
```
rviscomi, line 37: Evaluate freq, total, and pct fields

sql/2021/resource-hints/dns-prefetch_domains_by_url.sql
```diff
@@ -0,0 +1,79 @@
...
+FROM (
+    SELECT
+      client,
+      url,
+      domain
+    FROM (
```
rviscomi, line 43: Consolidate SELECTs

sql/2021/resource-hints/hints_adoption_by_rank.sql
```diff
@@ -0,0 +1,61 @@
...
+} catch (e) {
+  return hints.reduce((results, hint) => {
+    results[hint] = false;
+    return results;
+  }, {});
+}
+''' ;
+
+SELECT
+  client,
```
rviscomi, line 24: I'd suggest bringing the rank field up higher in the output, to make it clearer in the results that the results are grouped by rank
```suggestion
  client,
  rank,
```

sql/2021/resource-hints/hints_adoption_by_rank.sql
```diff
@@ -0,0 +1,61 @@
...
+  COUNTIF(hints.prefetch) / COUNT(0) AS pct_prefetch,
+  COUNTIF(hints.preconnect) AS preconnect,
+  COUNTIF(hints.preconnect) / COUNT(0) AS pct_preconnect,
+  COUNTIF(hints.prerender) AS prerender,
+  COUNTIF(hints.prerender) / COUNT(0) AS pct_prerender,
+  COUNTIF(hints.`dns-prefetch`) AS dns_prefetch,
+  COUNTIF(hints.`dns-prefetch`) / COUNT(0) AS pct_dns_prefetch,
+  COUNTIF(hints.modulepreload) AS modulepreload,
+  COUNTIF(hints.modulepreload) / COUNT(0) AS pct_modulepreload,
+  rank
```
rviscomi, line 37:
```suggestion
  COUNTIF(hints.modulepreload) / COUNT(0) AS pct_modulepreload
```

sql/2021/resource-hints/hints_adoption_by_rank.sql
```diff
@@ -0,0 +1,61 @@
...
+JOIN (
+    SELECT
+      _TABLE_SUFFIX AS client,
+      url AS page,
+      rank
+    FROM
+      `httparchive.summary_pages.2021_07_01_*`
+)
+USING
+  (client, page)
```
rviscomi, line 55: By default ranks are exclusive, so for example the top 100,000 rank contains 99,000 URLs because the top 1,000 is its own exclusive segment. So to ensure that the top 100k contains 100k URLs we combine them like this:
```suggestion
  (client, page),
  UNNEST([1000, 10000, 100000, 1000000, 10000000]) AS rank
WHERE
  _rank <= rank
```

sql/2021/resource-hints/hints_adoption_by_rank.sql
```diff
@@ -0,0 +1,61 @@
...
+    url AS page,
+    getResourceHints(payload) AS hints
+  FROM
+    `httparchive.pages.2021_07_01_*`
+)
+JOIN (
+    SELECT
+      _TABLE_SUFFIX AS client,
+      url AS page,
+      rank
```
rviscomi, line 50: Aliasing for simplicity
```suggestion
      rank AS _rank
```

rviscomi (changes requested):
Reviewed the first 9 of 27 queries. Some of these comments apply to patterns used by many queries, so I'll submit my review now as a first pass to unblock you.

sql/2021/resource-hints/consoleLog_incorrect_type.sql
kevinfarrugia: Non greedy, so if there are multiple matches it takes the shortest one.

sql/2021/resource-hints/dns-prefetch_domains.sql
kevinfarrugia: Don't think I could consolidate as much as if I consolidate the three SELECT statements I would be unable to use the condition
```
  WHERE
    ARRAY_LENGTH(value) > 0
```

and need to replace it with 

```
  WHERE
    ARRAY_LENGTH(getResourceHintHrefs(payload, "dns-prefetch")) > 0
```

I can however consolidate two of the statements.

sql/2021/resource-hints/hints_adoption_by_rank.sql
kevinfarrugia: Good to know. Would suggest having this documented somewhere - unless it already is and I missed it.

sql/2021/resource-hints/consoleLog_incorrect_type.sql
tunetheweb: I updated already in https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide#formatting-results but just added a comment about rounding too as see I hadn't explicitly mentioned that.

sql/2021/resource-hints/hints_adoption_by_rank.sql
tunetheweb: Yeah should be in Analysts guide. Will add...

sql/2021/resource-hints/dns-prefetch_domains.sql
rviscomi: You could probably remove that part from the query and it'd still work because the CROSS JOIN UNNEST won't produce any `href` values for pages in which the array is empty.

sql/2021/resource-hints/lazy_iframes.sql
```diff
@@ -0,0 +1,39 @@
...
+    return null
+}
+''' ;
+WITH iframe_stats_tb AS (
+  SELECT
+    _TABLE_SUFFIX AS client,
+    JSON_EXTRACT_SCALAR(payload,
+      '$._almanac') AS almanac,
+    countLazyIframes(JSON_EXTRACT_SCALAR(payload,
+        '$._almanac')) AS res
```
rviscomi, line 24: `almanac` is unused

Also replacing `res` with a more descriptive field name, preferring underscore naming convention for SQL field names

```suggestion
    countLazyIframes(JSON_EXTRACT_SCALAR(payload, '$._almanac')) AS num_lazy_iframes
```

sql/2021/resource-hints/lazy_iframes.sql
```diff
@@ -0,0 +1,39 @@
...
+  FROM
+    `httparchive.pages.2021_07_01_*`
+)
+
+SELECT
+  client,
+  res AS numLazyIframes,
+  COUNT(0) AS numPages
```
rviscomi, line 32: Also include relative stats in terms of % of pages.

```suggestion
  num_lazy_iframes,
  COUNT(0) AS pages,
  SUM(COUNT(0)) OVER (PARTITION BY client) AS total,
  COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client) AS pct
```

sql/2021/resource-hints/lazy_iframes.sql
```diff
@@ -0,0 +1,39 @@
...
+FROM
+  iframe_stats_tb
+WHERE
+  res > 0
```
rviscomi, line 36:
```suggestion
  num_lazy_iframes > 0
```

sql/2021/resource-hints/lazy_iframes.sql
```diff
@@ -0,0 +1,39 @@
...
+GROUP BY
+  client,
+  res
```
rviscomi, line 39:
```suggestion
  num_lazy_iframes
```

sql/2021/resource-hints/lazy_viewport_images.sql
```diff
@@ -0,0 +1,40 @@
+  # standardSQL
+  # Lazily loaded viewport images
```
rviscomi, line 2: Ditto feedback on iframe query

sql/2021/resource-hints/preconnect_hosts.sql
```diff
@@ -0,0 +1,45 @@
...
+}
+''' ;
+
+SELECT
+  client,
+  host,
+  COUNT(0) AS freq,
+  COUNT(0) / SUM(COUNT(0)) OVER () AS pct
+FROM (
+    SELECT
```
rviscomi, line 23: Ditto feedback about combining SELECT statements

sql/2021/resource-hints/preconnect_hosts.sql
```diff
@@ -0,0 +1,45 @@
...
+} catch (e) {  
+  return [];
+}
+''' ;
+
+SELECT
+  client,
+  host,
+  COUNT(0) AS freq,
+  COUNT(0) / SUM(COUNT(0)) OVER () AS pct
```
rviscomi, line 21: Omitting the `PARTITION BY client` will add desktop and mobile results together.

Also a good idea to always include the total in the results when calculating percents.
```suggestion
  SUM(COUNT(0)) OVER (PARTITION BY client) AS total,
  COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client) AS pct
```

sql/2021/resource-hints/preload-fonts.sql
```diff
@@ -0,0 +1,14 @@
+#standardSQL
+# Distribution of Lighthouse scores for the 'Preload Fonts' audit
+
+
+SELECT
+  JSON_EXTRACT_SCALAR(report, "$.audits.preload-fonts.score") AS score,
+  COUNT(0) AS freq,
+  COUNT(0) / SUM(COUNT(0)) OVER () AS pct
```
rviscomi, line 8: Include the total

sql/2021/resource-hints/preload_below_fold_images.sql
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    return unnecessaryImgPreloads;
+}
+catch {
+    return null
+}
+''' ;
+WITH
+ImageStats AS (
```
rviscomi, line 36: Ditto feedback on earlier iframe query

sql/2021/resource-hints/preload_hosts.sql
```diff
@@ -0,0 +1,40 @@
...
+} catch (e) {  
+  return [];
+}
+''' ;
+
+SELECT
+  client,
+  host,
+  COUNT(0) AS freq,
+  COUNT(0) / SUM(COUNT(0)) OVER () AS pct
```
rviscomi, line 21: total and PARTITION

sql/2021/resource-hints/preload_hosts_by_url.sql
```diff
@@ -0,0 +1,58 @@
...
+}
+''' ;
+
+SELECT
+  client,
+  host,
+  COUNT(0) AS freq,
+  COUNT(0) / SUM(COUNT(0)) OVER () AS pct
+FROM (
+    SELECT
```
rviscomi, line 24: total and PARTITION

Also, ditto feedback on simplifying inner SELECTs

sql/2021/resource-hints/uses-rel-preconnect.sql
```diff
@@ -0,0 +1,14 @@
+#standardSQL
+# Distribution of Lighthouse scores for the 'Preconnect to required origins' audit
+
+
+SELECT
+  JSON_EXTRACT_SCALAR(report, "$.audits.uses-rel-preconnect.score") AS score,
+  COUNT(0) AS freq,
+  COUNT(0) / SUM(COUNT(0)) OVER () AS pct
```
rviscomi, line 8: total

sql/2021/resource-hints/uses-rel-preload.sql
```diff
@@ -0,0 +1,14 @@
+#standardSQL
+# Distribution of Lighthouse scores for the 'Preload key requests' audit
+
+
+SELECT
+  JSON_EXTRACT_SCALAR(report, "$.audits.uses-rel-preload.score") AS score,
+  COUNT(0) AS freq,
+  COUNT(0) / SUM(COUNT(0)) OVER () AS pct
```
rviscomi, line 8: total

rviscomi (changes requested):

sql/2021/resource-hints/unused_preloaded_fonts.sql
```diff
@@ -0,0 +1,82 @@
...
+
+      return {
+        ...acc,
+        [filename]: [...(acc[filename] ? acc[filename] : []), ext],
+      };
+    }
+  }, {});
+
+  return Object.values(preloadedFontSet).reduce((acc, n) => {
+    if (n.find((m) => m === ".woff2")) {
```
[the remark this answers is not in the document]
kevinfarrugia: Any browser which supports `<link rel="preload" />` also supports .woff2; so preloading older font types (.woff, .eot, .ttf, .svg) for cross-browser compatibility when you have already preloaded woff2 is wasteful.

Having said that, this could be extended to .woff too.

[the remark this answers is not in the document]
tunetheweb: FYI we preload woff at work as don’t have woff2 versions. Must dig up that email thread as to why we can’t get them or if it’s possible to convert…

Anyway that’s one use case (maybe an edge one?) of why to preload older versions.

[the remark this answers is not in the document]
kevinfarrugia: In that case you'd just want to preload woff though, correct? I.e. you do not preload OTF, TTF, EOT...etc.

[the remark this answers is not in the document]
tunetheweb: Yes that's correct. Sorry didn't look at full context before butting in!

sql/2021/resource-hints/lazy_iframes.sql
Nithanaroy: Learned something new! Thanks Rick :)

sql/2021/resource-hints/unused_preloaded_fonts.sql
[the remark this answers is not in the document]
kevinfarrugia: Updated to include either woff and woff2.

Issue #2153, opened 2021-04-27
PWA 2021


rviscomi:
# Part II Chapter 16: PWA

![PWA illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/pwa/hero_lg.jpg)

If you're interested in contributing to the PWA chapter of the 2021 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| --- | ------- | --------- | -------- | ------- | -- |
| @demianrenzulli | @demianrenzulli | @webmaxru @Schweinepriester @thepassle @hemanth @tropicadri @andreban @jeffposnick  | @tunetheweb @demianrenzulli | - | @obto |

<details>
<summary><b>Expand for more information about each role</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
- **[Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)** are responsible for researching the stats and trends used throughout the Almanac. Analysts work closely with authors and reviewers during the planning phase to give direction on the types of stats that are possible from the dataset, and during the analyzing/writing phases to ensure that the stats are used correctly.
[...]

PR #2272, opened 2021-07-06
PWA 2021 queries

tunetheweb:
Makes progress on #2153 

- [x] - beforeinstallprompt_usage.sql
- [x] - lighthouse_pwa_audits.sql (PWA pages and All pages)
- [x] - lighthouse_pwa_score.sql (PWA pages and All pages)
- [x] - manifests_and_service_workers.sql
- [x] - manifests_not_json_parsable.sql (PWA pages and All pages)
- [x] - manifests_preferring_native_apps.sql (PWA pages and All pages)
- [x] - popular_pwa_libraries.sql
- [x] - popular_pwa_libraries_helper.sql
- [x] - pwa_blink_usage.sql
- [x] - sw_adoption_over_time.sql
- [x] - sw_adoption_over_time_ranking.sql
- [x] - sw_events.sql
- [x] - top_manifest_categories.sql (PWA pages and All pages)
- [x] - top_manifest_display_values.sql (PWA pages and All pages)
- [x] - top_manifest_icon_sizes.sql (PWA pages and All pages)
- [x] - top_manifest_orientations.sql (PWA pages and All pages)
- [x] - top_manifest_properties.sql (PWA pages and All pages)
- [x] - workbox_package_methods.sql
- [x] - workbox_usage.sql
- [x] - Workbox versions
- [x] - Usage of assetlinks (already captured [here](https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/well-known.js#L45), so I assume we can just reuse that one)
- [x] - CrUX currently measures the [acceptance rate of notifications](https://developers.google.com/web/tools/chrome-user-experience-report#notification_permissions). Could we do a query that just aggregates this information across all origins?
- [x] - Update all queries to stop pointing at `sample_data`.
- [x] - Decide how to measure sw usage given conversation at https://github.com/HTTPArchive/almanac.httparchive.org/issues/2153#issuecomment-874570418
- [x] - install_events.sql
- [x] - sw-registration-properties.sql
- [x] - sw_methods.sql
- [x] - sw_objects.sql


sql/2021/pwa/lighthouse_pwa_audits.sql
```diff
@@ -0,0 +1,42 @@
...
+FROM
+  `httparchive.lighthouse.2021_07_01_mobile`,
+  UNNEST(getAudits(report, "pwa")) AS audits
+WHERE
+  LENGTH(report) < 20000000  # necessary to avoid out of memory issues. Excludes 16 very large results
+GROUP BY
+  audits.id
+ORDER BY
+  median_weight DESC,
+  id
```
demianrenzulli, line 42: Compared with 2020, Lighthouse has removed the **load-fast-enough-for-pwa** and **works-offline** ([issue](https://github.com/GoogleChrome/lighthouse/pull/11806)). Those used to be available in last year’s `sample_data_2020` dataset.

For this year I wanted to check the YoY growth of sites that respond offline, but that won't be possible.
In any case, this query LGTM.

tunetheweb: Actually that reminds me I want to rewrite this a bit [after learnings elsewhere](https://github.com/HTTPArchive/bigquery/pull/114).

Currently we take the whole Lighthouse JSON and sent it to the JavaScript UDF for processing. This causes that to fail for large payloads - hence the WHERE clause to exclude this.

But it’s more efficient to JSON_EXTRACT the bit we want (PWA audits) and then send that smaller JSON to the UDF instead. Then don’t need that WHERE clause. Plus it runs a lot faster (in that other query at least so guessing same here). I guess JSON_EXTRACT is more efficient than passing large objects between SQL and JavaScript.

Will leave this comment open as a reminder to myself to refactor this a little before merging.

sql/2021/pwa/lighthouse_pwa_audits_sw.sql
```diff
@@ -0,0 +1,48 @@
...
+  MAX(audits.description) AS description
+FROM
+  `httparchive.lighthouse.2021_07_01_mobile`
+JOIN
+  `httparchive.pages.2021_07_01_mobile`
+USING (url),
+  UNNEST(getAudits(report, "pwa")) AS audits
+WHERE
+  JSON_EXTRACT(payload, '$._pwa') != "[]" AND
+  JSON_EXTRACT(payload, '$._pwa.serviceWorkers') != "[]" AND
```
demianrenzulli, line 41: @tunetheweb: leaving this comment here, so we can discuss if it makes sense to include `$._pwa.serviceWorkerHeuristic` in the final query.

sql/2021/pwa/lighthouse_pwa_score.sql
```diff
@@ -0,0 +1,19 @@
+#standardSQL
+# Percentiles of lighthouse pwa score
+# This metric comes from Lighthouse only and is only available in mobile in HTTP Archive dataset
+SELECT
+  '2021_07_01' AS date,
+  percentile,
+  APPROX_QUANTILES(score, 1000)[OFFSET(percentile * 10)] AS score
```
demianrenzulli, line 7: I ran this query on `httparchive.sample_data.lighthouse_mobile_10k` and obtained the following results:

Row | Date            | Percentile | Score |
1      | 2021_07_01 | 10              | 0.25 |  
2      | 2021_07_01 | 25             | 0.33 |  
3      | 2021_07_01 | 50             | 0.42 |  
4      | 2021_07_01 | 75             | 0.5 |  
5      | 2021_07_01 | 90             | 0.5

Taking a look at the last row, does it mean that 90% of the reports have, **at least** a pwa score of 0.5?
Isn't that a very high value?

tunetheweb: Yup. Consistent with what I got before: https://www.tunetheweb.com/blog/what-do-lighthouse-scores-look-like-across-the-web/

My own blog gets a score of 0.5 despite not having a service worker not a manifest. There’s plenty of audits in this category that non-PWA websites can pass (e.g. uses HTTPS, loads fast enough, sets viewport… etc.).

rviscomi: Isn't it the other way around? If the 90th percentile is 0.5, that means 10% of reports are at least 0.5. In other words, 90% of reports are less than 0.5.

tunetheweb: Sorry yes you’re right, misread @demianrenzulli ’s comment. 90% of pages have a score of 0.5 or less.

Btw I always thought that the percentile was inclusive (i.e. “90% of pages have a score of 0.5 or less.” rather than “90% of pages have a score of less than 0.5.”)? Not that that makes that much of a difference.

demianrenzulli: Thanks for claryifing, folks.

As done in previous editions of the Web Almanac, I think that it would be interesting to present a conclusion that could be more or less actionable and informative, besides the raw data.

In this particular case, for example, would it be fair to assume that the most meaningful conclusion is:  "Only 10% of the reports have a PWA score higher or equals than 0.5?"

The other conclusion doesn't sound very useful: “90% of pages have a score of 0.5 or less.”. One could also also say something like "100% of pages have a score of less or equals than 1" and that would be right.

Leaving this discussion open so we can discuss in more detail when the time comes.

sql/2021/pwa/manifests_and_service_workers.sql
```diff
@@ -0,0 +1,22 @@
...
+  client,
+  SAFE_DIVIDE(SUM(ServiceWorker), SUM(COUNT(0)) OVER (PARTITION BY client)) AS serviceworkers_pct,
+  SAFE_DIVIDE(SUM(manifests), SUM(COUNT(0)) OVER (PARTITION BY client)) AS manifests_pct,
+  SAFE_DIVIDE(COUNTIF(ServiceWorker > 0 OR manifests > 0), SUM(COUNT(0)) OVER (PARTITION BY client)) AS either_pct,
+  SAFE_DIVIDE(COUNTIF(ServiceWorker > 0 AND manifests > 0), SUM(COUNT(0)) OVER (PARTITION BY client)) AS both_pct,
+  SUM(COUNT(0)) OVER (PARTITION BY client) AS total
+FROM (
+  SELECT
+    _TABLE_SUFFIX AS client,
+    IF(JSON_EXTRACT(payload, '$._pwa') != "[]" AND JSON_EXTRACT(payload, '$._pwa.serviceWorkers') != "[]",1,0) AS ServiceWorker,
```
demianrenzulli, line 13: The test run shows a very low number for the present of each and also of both. This seems to be consistent with [2020](https://almanac.httparchive.org/en/2020/pwa#web-app-manifests). Once again it would be interesting to check YoY on growth.
Here we might also want to consider adding the `$._pwa.serviceWorkerHeuristic` in the final query.

sql/2021/pwa/manifests_not_json_parsable.sql
```diff
@@ -0,0 +1,39 @@
+#standardSQL
+# Manifests that are not JSON parsable for all pages
```
demianrenzulli, line 2: What is the goal of this report? Is it to measure the number of "invalid" Manifests, or is there any other reason for it?

sql/2021/pwa/manifests_not_json_parsable_sw.sql
```diff
@@ -0,0 +1,40 @@
+#standardSQL
+# Manifests that are not JSON parsable for service worker pages
```
demianrenzulli, line 2: Same observations as the previous one. Here we could also consider adding the `$._pwa.serviceWorkerHeuristic`.

sql/2021/pwa/lighthouse_pwa_audits_sw.sql
tunetheweb: Yes, as discussed previously, we’ll make the call on using this for all the queries after a July has run and we see the full numbers.

sql/2021/pwa/lighthouse_pwa_score.sql
tunetheweb: You don’t have to use every query in the chapter. Some just won’t have any interesting insights and to be honest I think this is one of those that’s not very interesting. We didn’t use it in last two years despite running it. Should we drop it completely? Maybe. But on other have it’s written so no harm running. Who knows maybe this year it’ll surprise us.

I think the individual audits are more interesting and we did use them last year.

sql/2021/pwa/manifests_not_json_parsable.sql
tunetheweb: Not sure to be honest. It was copied from last two years queries.

But yeah good to see malformed manifests are mostly valid JSON.

Might also be worth looking at 404 ones? I.e. where a manifest is referenced in the HTML but not actually served (or an HTML 404 page served instead of a JSON 200 ).

sql/2021/pwa/manifests_preferring_native_apps_sw.sql
```diff
@@ -0,0 +1,35 @@
+#standardSQL
+# % manifests preferring native apps for service worker pages
```
demianrenzulli, line 2: Use this in the "Installability" section, where we'll also count usage of the `beforeinstallprompt` event.

sql/2021/pwa/popular_pwa_libraries.sql
```diff
@@ -0,0 +1,58 @@
...
+    LOWER(body) NOT LIKE '%analytics-helper.js%') AS none_of_the_above
+FROM
+  (
+    SELECT DISTINCT
+      date,
+      client,
+      page,
+      body
+    FROM
+      `httparchive.almanac.service_workers`
```
demianrenzulli, line 46: Should we replace `httparchive.almanac.service_workers` with the new script?

sql/2021/pwa/popular_pwa_libraries_helper.sql
```diff
@@ -0,0 +1,36 @@
...
+# TODO - CONVERT TO CUSTOM METRICS - IF WE STILL NEED THIS?
+#
+SELECT
+  client,
+  importscript,
+  COUNT(DISTINCT page) AS pages,
+  total,
+  COUNT(DISTINCT page) / total AS pct
+FROM
+  (SELECT DISTINCT * FROM `httparchive.almanac.service_workers`)
```
demianrenzulli, line 14: Same comment to replace `httparchive.almanac.service_workers` with the new script.

sql/2021/pwa/pwa_blink_usage.sql
```diff
@@ -0,0 +1,20 @@
...
+  yyyymmdd = '20210701' AND
+  (
+    feature LIKE '%ServiceWorker%' OR
+    feature LIKE '%BackgroundSync%' OR
+    feature LIKE '%GetInstalledRelatedApps%'
+  )
+ORDER BY
+  client,
+  num_urls,
+  feature,
```
demianrenzulli, line 20: BigQuery console complaints that there's a comma at the end of this query, but I see that it's also present in the [2020 version](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2020/pwa/pwa_blink_usage.sql).

sql/2021/pwa/sw_adoption_over_time.sql
```diff
@@ -0,0 +1,15 @@
...
+SELECT
+  yyyymmdd AS date,
+  client,
+  num_urls AS freq,
+  total_urls AS total,
+  pct_urls AS pct
+FROM
+  `httparchive.blink_features.usage`
+WHERE
+  feature = 'ServiceWorkerControlledPage'
```
demianrenzulli, line 12: This is one of the mos interesting queries. Taking a look at [the chart](https://www.chromestatus.com/metrics/feature/timeline/popularity/990), we might be able to reach to 20% of pages controlled by a service worker by the time the Almanac launches, which is a great stat. A couple of thoughts:

- I recall you've mentioned that this year we would be able to see if the top sites use service workers, in order to support our hypothesis that, even when 1% of URLs have service workers, 20% of traffic is controlled, because top sites implemented it. Is that correct?
- We asked internally but we were not able to explain why from [Nov. 2019 and May 2020](https://www.chromestatus.com/metrics/feature/timeline/popularity/990) there's a pronounced drop in usage, that is restored afterwards. Maybe we can leave a note indicating that we don't have a reason to explain this?

sql/2021/pwa/sw_events.sql
```diff
@@ -0,0 +1,28 @@
...
+  event,
+  COUNT(DISTINCT page) AS freq,
+  total,
+  COUNT(DISTINCT page) / total AS pct
+FROM
+  `httparchive.almanac.service_workers`
+JOIN
+  (SELECT client, COUNT(DISTINCT page) AS total FROM `httparchive.almanac.service_workers` WHERE date = '2021-07-01' GROUP BY client)
+USING (client),
+  UNNEST(ARRAY_CONCAT(
```
demianrenzulli, line 17: This one is unchecked, so I assume I can just update it with the new methods defined in [pwa.js](https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/pwa.js).

sql/2021/pwa/top_manifest_categories.sql
```diff
@@ -0,0 +1,38 @@
+#standardSQL
+# Top manifest categories
```
demianrenzulli, line 2: This is one of those things that sound like great information, for example, to understand which verticals use or benefit more from PWAs. Despite that, I'm still unsure about what's the real usefulness of it (I'll ask).
Regarding the query itself, besides the distribution, it would be useful to know the % of sites that, having a manifest, have this field.

sql/2021/pwa/lighthouse_pwa_audits.sql
tunetheweb: Rewrote these to be more efficient in https://github.com/HTTPArchive/almanac.httparchive.org/pull/2272/commits/9c83af78c0987c6a6a7e105c30faa54ae283ef5c. Resolving.

sql/2021/pwa/popular_pwa_libraries.sql
tunetheweb: Yup. These are unticked still. You were going to have a look at the JS function for the ones not ticked yet?

sql/2021/pwa/popular_pwa_libraries_helper.sql
tunetheweb: Yup. These are unticked still. You were going to have a look at the JS function for the ones not ticked yet?

sql/2021/pwa/pwa_blink_usage.sql
tunetheweb: Removed.

sql/2021/pwa/sw_adoption_over_time.sql
tunetheweb: > I recall you've mentioned that this year we would be able to see if the top sites use service workers, in order to support our hypothesis that, even when 1% of URLs have service workers, 20% of traffic is controlled, because top sites implemented it. Is that correct?

I've added a query for this for top 1,000 sites in https://github.com/HTTPArchive/almanac.httparchive.org/pull/2272/commits/e519676de7e4c076341c953a56dfaab6c4339454. Note this is only available from May this year.

All sites:

date | client | freq | total | pct
-- | -- | -- | -- | --
20210601 | desktop | 74844 | 6435567 | 1.16%
20210601 | mobile | 84998 | 7452047 | 1.14%
20210501 | desktop | 70558 | 6389672 | 1.10%
20210501 | mobile | 79878 | 7364482 | 1.08%

Top 1000 sites (note not actually 1000 since we can't reach some sites):

date | client | freq | total | pct
-- | -- | -- | -- | --
20210601 | desktop | 60 | 803 | 7.47%
20210601 | mobile | 72 | 942 | 7.64%
20210501 | desktop | 60 | 796 | 7.54%
20210501 | mobile | 73 | 938 | 7.78%

An interesting difference I think you'll agree!

> We asked internally but we were not able to explain why from Nov. 2019 and May 2020 there's a pronounced drop in usage, that is restored afterwards. Maybe we can leave a note indicating that we don't have a reason to explain this?

It's interesting that we only see that drop in the Chrome usage, and not in the [HTTP Archive crawl](https://httparchive.org/reports/progressive-web-apps#swControlledPages). This suggests either Chrome stopped logging this correctly, or a big player (e.g. Facebook) stopped using Service Workers for a period. Unfortunately as we don't have access to the underlying data of Chrome's usage we can't take this any further.

Last year we didn't use Chrome's usage graph, and use the crawl graph, and only refered to Chrome's usage: "While that usage may seem low, it is important that we realize that other measurements equate that to 16.6% of the [web traffic](https://www.chromestatus.com/metrics/feature/timeline/popularity/990)–the difference being due to high traffic websites tending to use service workers more."

Probably best to do the same thing year to avoid confusing readers with this anomaly.

> How far in the past can we go with this report? It would be interesting to see it as a timeseries, starting when service workers were launched (around 2015), to our days.

We only have `blink_usage` stats back to January 2017 but even then it's down to a couple of hundred sites so doubt it's useful before that.

tunetheweb: In fact here's a better graph looking at all the ranking categories for June 2021:

Service Worker Controlled Pages by Ranking

And here's a timeline of the mobile results (though less interesting as only have two months worth of data at present):

SW Controlled Pages by Ranking over Time

sql/2021/pwa/top_manifest_categories.sql
tunetheweb: The query already does that. Running it for June for example we have:

image

So 252 of the 2603 desktop sites that have a manifest use the News category.
And 150 of the 2603 desktop sites use the Education category.
As you can see in both cases it's measured against the total of 2603 manifest sites.

Remember that we only got the tail end of the June run so these numbers will be small.

sql/2021/pwa/popular_pwa_libraries.sql
demianrenzulli: Yes, I've noticed this wasn't finished. I'll take a look at those remaining queries.

sql/2021/pwa/sw_adoption_over_time.sql
demianrenzulli: All these comments are great, thanks for working on this @tunetheweb!

I like the chart showing how the SW usage decreases as one broadens the scope of the search beyond the top sites. I think we can use it in the SW section to explain why the number of URLs controlled by SW differs from the actual percentage of page loads.

I think we are good in terms of this section and queries 👍

sql/2021/pwa/top_manifest_categories.sql
demianrenzulli: Perfect, thank you!

demianrenzulli:
@tunetheweb: regarding a comment you left in some of the queries asking if we should detect more than one manifest. Here are situations where you can find multiple manifests for a given domain:

- Publishing standalone webapps via subdomains: https://video.televisa.com/manifest.json is different from https://www.televisa.com/manifest.json.
- Same thing, but via paths: I've noticed that Google Maps Manifest: https://www.google.com/maps/preview/pwa/manifest is different from google.com manifest.
- Publishing localized manifests for different languages. 

From these (1) wouldn't make sense to capture, because the manifests would be published on subdomains (essentially different origins). Regarding (1): it might make more sense, but there are few sides that have nested apps, like in Google Maps example (also this technique is not encouraged). (3) might be interesting, but it's infrequent.

With that said, the current detection seems to be good enough for most cases.

sql/2021/pwa/sw_adoption_over_time.sql
demianrenzulli: @tunetheweb: one more question regarding this: do you have the query to get the list of top 100 / 1000 sites?
I'd like to go over them "manually" to see if there's any interesting pattern to extract.

tunetheweb: Saved here (along with the query needed to get it):
https://docs.google.com/spreadsheets/d/10KT5CxH6VKRktg90MwfzXuSXn67pxUuNiIxR94B5FSg/edit?usp=sharing

demianrenzulli: Thanks for sharing the list @tunetheweb! I went over it, and taking a look at the origins listed there, it makes sense that a 1% of the URLs controlled by service workers, represent almost 20% of the web traffic.

As you know, the limitation with that list is that it leaves out potentially high traffic sites that only register the service worker after the user signs in. This is the case of Facebook, for example.

In that case, the URL won't show up, but the traffic will still be counted for the 20% of traffic controlled by service workers.

In general, this is a very good approximation, though.

sql/2021/pwa/lighthouse_pwa_audits_sw.sql
tunetheweb: July has now finished so let's look at the stats:

Running Lighthouse we see:

id | num_pages | total | total_applicable | pct
-- | -- | -- | -- | --
service-worker | 107,775 | 7,262,217 | 7,165,208 | 1.50%

Running blink usage and our custom metrics we have:

client | sw_blink_usage | serviceworkers | serviceWorkerHeuristic | total | serviceworkers_pct | serviceWorkerHeuristic_pct | sw_blink_usage_pct
-- | -- | -- | -- | -- | -- | -- | --
desktop | 77,020 | 55,723 | 191,745 | 6,286,373 | 0.89% | 3.05% | 1.23%
mobile | 89,510 | 76,011 | 240,942 | 7,491,840 | 1.01% | 3.22% | 1.19%

So it appears Lighthouse over counts compared to blink usage and our custom stat, and the heuristic way over counts. Given that, it seems our original `serviceworkers` is a better metric to use rather than the `serviceWorkerHeuristic` megric even if it does under count blink usage by 0.34% on desktop and 0.18% on mobile.

Lighthouse over counts blink usage by 0.31%, and our `serviceworkers` metric by 0.49%.

Comparing this with [Last years stats](https://docs.google.com/spreadsheets/d/1AOqCkb5EggnE8BngpxvxGj_fCfssT9sZ0ElCQKp4pOw/edit#gid=814184412):

[Lighthouse](https://docs.google.com/spreadsheets/d/1AOqCkb5EggnE8BngpxvxGj_fCfssT9sZ0ElCQKp4pOw/edit#gid=814184412):

id | num_pages | total | total_applicable | pct
-- | -- | -- | -- | --
service-worker | 64,373 | 6,290,131 | 6,252,154 | 1.03%

[Blink usage](https://docs.google.com/spreadsheets/d/1AOqCkb5EggnE8BngpxvxGj_fCfssT9sZ0ElCQKp4pOw/edit#gid=2077755325):

client | feature | num_urls | total_urls | pct
-- | -- | -- | -- | --
desktop | ServiceWorkerControlledPage | 49,305 | 5,593,642 | 0.88%
mobile | ServiceWorkerControlledPage | 55,019 | 6,347,919 | 0.87%

[Almanac Custom Stats](https://docs.google.com/spreadsheets/d/1AOqCkb5EggnE8BngpxvxGj_fCfssT9sZ0ElCQKp4pOw/edit#gid=227354805):

date | client | serviceworkers_pct | total
-- | -- | -- | -- | -- | -- | --
2020-08-01 | desktop | 0.76% | 5,593,642
2020-08-01 | mobile | 0.84% | 6,347,640

We can see our custom stat also undercounted blink usage last year, though by less (0.12% on desktop, and 0.03% on mobile), and both undercounted Lighthouse even more (by 0.16% for blink usage and 0.19% for our stat).

This again points us at using the `serviceworkers` custom metric over `serviceWorkerHeuristic` as, although the divergence from blink usage and Lighthouse has got slightly worse, it's similar to what we used last year and the `serviceWorkerHeuristic` is just over counting way too much.

It's also notable that we have nearly a million extra URLs in this years data, compared to last year. Whether this is new sites, or just more of a long tail (who may or may not use Service Workers more or less) remains to be seen.

All of these stats show that we have different numbers for each, and probably none of these numbers are "correct", and the truth probably lies somewhere in between these all.

However I do think the numbers we have gathered with our `serviceworkers` metric are a significant proportion of the PWA sites out there, and so are likely to be representative of the usage of service workers and manifests in a similar proportion to they were last year, so while we can prefer blink usage data for reporting actual service worker usage (as we did last year), the custom metric data can still be used (as it was last year, though in a slightly different guise) for other data that we do not have other sources for (like manifest usage, popularity of SW events, workbox usage...etc.). Though, like last year, I would advise reporting in terms of percentages rather than absolute number of sites, since our data is not (and cannot be) 100% accurate to the exact usage. This is the preference within the Web Almanac anyway, to allow easier comparison year on year amongst other reasons.

Let me know if you agree and so can resolve this comment (and others like it) is we stick with `serviceworks` rather than switching to `serviceWorkerHeuristic`?

**Queries:**

Lighthouse:

```sql
#standardSQL
# Get summary of all lighthouse scores for a category
# Note scores, weightings, groups and descriptions may be off in mixed months when new versions of Lighthouse roles out

CREATE TEMPORARY FUNCTION getAudits(auditRefs STRING, audits STRING)
RETURNS ARRAY<STRUCT<id STRING, weight INT64, audit_group STRING, title STRING, description STRING, score INT64>> LANGUAGE js AS '''
var auditrefs = JSON.parse(auditRefs);
var audits = JSON.parse(audits);
var results = [];
for (auditref of auditrefs) {
  results.push({
    id: auditref.id,
    weight: auditref.weight,
    audit_group: auditref.group,
    description: audits[auditref.id].description,
    score: audits[auditref.id].score
  });
}
return results;
''';

SELECT
  audits.id AS id,
  COUNTIF(audits.score > 0) AS num_pages,
  COUNT(0) AS total,
  SAFE_DIVIDE(COUNTIF(audits.score > 0), COUNTIF(audits.score IS NOT NULL)) AS pct,
  APPROX_QUANTILES(audits.weight, 100)[OFFSET(50)] AS median_weight,
  MAX(audits.audit_group) AS audit_group,
  MAX(audits.description) AS description
FROM
  `httparchive.lighthouse.2021_07_01_mobile`,
  --`httparchive.sample_data.lighthouse_mobile_10k`,
  UNNEST(getAudits(JSON_EXTRACT(report, '$.categories.pwa.auditRefs'),JSON_EXTRACT(report, '$.audits'))) AS audits
GROUP BY
  audits.id
ORDER BY
  median_weight DESC,
  id
```

Blink usage and custom metrics:
```sql
#standardSQL
# Counting Manifests and Service Workers
SELECT
  client,
  SUM(ServiceWorkerControlledPage) as sw_blink_usage,
  SUM(ServiceWorker) as serviceworkers,
  SUM(serviceWorkerHeuristic) as serviceWorkerHeuristic,
  SAFE_DIVIDE(SUM(ServiceWorker), SUM(COUNT(0)) OVER (PARTITION BY client)) AS serviceworkers_pct,
  SAFE_DIVIDE(SUM(serviceWorkerHeuristic), SUM(COUNT(0)) OVER (PARTITION BY client)) AS serviceWorkerHeuristic_pct,
  SAFE_DIVIDE(SUM(ServiceWorkerControlledPage), SUM(COUNT(0)) OVER (PARTITION BY client)) AS sw_blink_usage_pct
FROM (
  SELECT
    _TABLE_SUFFIX AS client,
    IF(JSON_EXTRACT(payload, '$._pwa') != "[]" AND JSON_EXTRACT(payload, '$._pwa.serviceWorkers') != "[]",1,0) AS ServiceWorker,
    IF(JSON_EXTRACT(payload, '$._pwa') != "[]" AND JSON_EXTRACT(payload, '$._pwa.serviceWorkerHeuristic') = 'true',1,0) AS serviceWorkerHeuristic,
    IF(IFNULL(
    JSON_EXTRACT(payload, '$._blinkFeatureFirstUsed.Features.ServiceWorkerControlledPage'),
    JSON_EXTRACT(payload, '$._blinkFeatureFirstUsed.Features.990'))
  IS NOT NULL, 1, 0) AS ServiceWorkerControlledPage,
  FROM
    --`httparchive.sample_data.pages_*`
  )
GROUP BY
  client
ORDER BY
  client
```

tunetheweb: Actually looking into this some more I'm not convinced.

I've created a scratch table with the following SQL:

```sql
#standardSQL
CREATE TABLE
`httparchive.scratchspace.serviceworkers_2021_07_01_mobile`
AS
SELECT
  url,
  JSON_EXTRACT(report, '$.audits.service-worker.score') as lh_service_worker,
  IF(IFNULL(
    JSON_EXTRACT(payload, '$._blinkFeatureFirstUsed.Features.ServiceWorkerControlledPage'),
    JSON_EXTRACT(payload, '$._blinkFeatureFirstUsed.Features.990'))
  IS NOT NULL, 1, 0) AS ServiceWorkerControlledPage,
  IF(JSON_EXTRACT(payload, '$._pwa') != "[]" AND JSON_EXTRACT(payload, '$._pwa.serviceWorkers') != "[]",1,0) AS ServiceWorker,
  IF(JSON_EXTRACT(payload, '$._pwa') != "[]" AND JSON_EXTRACT(payload, '$._pwa.workboxInfo') != "[]",1,0) AS workboxInfo,
  IF(JSON_EXTRACT(payload, '$._pwa') != "[]" AND JSON_EXTRACT(payload, '$._pwa.swEventListenersInfo') != "[]",1,0) AS swEventListenersInfo,
  IF(JSON_EXTRACT(payload, '$._pwa') != "[]" AND JSON_EXTRACT(payload, '$._pwa.swMethodsInfo') != "[]",1,0) AS swMethodsInfo,
  IF(JSON_EXTRACT(payload, '$._pwa') != "[]" AND JSON_EXTRACT(payload, '$._pwa.serviceWorkerHeuristic') = 'true',1,0) AS serviceWorkerHeuristic,
  JSON_EXTRACT(payload, '$._pwa') as pwaCustomMetric
FROM
    `httparchive.pages.2021_07_01_mobile`
JOIN
    `httparchive.lighthouse.2021_07_01_mobile`
USING
  (url)
WHERE
  (JSON_EXTRACT(payload, '$._pwa') != "[]" AND
    (JSON_EXTRACT(payload, '$._pwa.serviceWorkerHeuristic') = 'true' OR JSON_EXTRACT(payload, '$._pwa.serviceWorkers') != "[]")
  )
 OR JSON_EXTRACT(payload, '$._blinkFeatureFirstUsed.Features.ServiceWorkerControlledPage') IS NOT NULL
 OR JSON_EXTRACT(payload, '$._blinkFeatureFirstUsed.Features.990') IS NOT NULL
 OR IFNULL(JSON_EXTRACT(report, '$.audits.service-worker.score'),'0') = '1'
```

So This saves all the lighthouse service worker score, the ServiceWorkerControlledPage metric, and all our PWA custom metric JSON into a new table called `httparchive.scratchspace.serviceworkers_2021_07_01_mobile`, which allows us to run cheap queries while we investigate this further. For example:

```sql
#standardSQL
SELECT
  SUM(IF(lh_service_worker = '1',1,0)) AS lighthouse,
  SUM(ServiceWorkerControlledPage) AS blink,
  SUM(ServiceWorker) AS ServiceWorker,
  SUM(workboxInfo) AS workboxInfo,
  SUM(swEventListenersInfo) AS swEventListenersInfo,
  SUM(swMethodsInfo) AS swMethodsInfo,
  SUM(serviceWorkerHeuristic) AS serviceWorkerHeuristic
FROM
  `httparchive.scratchspace.serviceworkers_2021_07_01_mobile`
```
Which gives:

lighthouse | blink | ServiceWorker | workboxInfo | swEventListenersInfo | swMethodsInfo | serviceWorkerHeuristic
-- | -- | -- | -- | -- | -- | --
107,775 | 86,175 | 72,366 | 57,524 | 218,791 | 139,458 | 230,067

So it's the `swEventListenersInfo` and `swMethodsInfo` which are driving up the `serviceWorkerHeuristic` count.

If I look at individual pages for ones our heuristic thinks uses a service worker but the rest don't:

```sql
#standardSQL
SELECT
    url,
    JSON_EXTRACT(pwaCustomMetric, '$.serviceWorkers') AS ServiceWorker,
    JSON_EXTRACT(pwaCustomMetric, '$.swEventListenersInfo') AS swEventListenersInfo,
    JSON_EXTRACT(pwaCustomMetric, '$.swMethodsInfo') AS swMethodsInfo
FROM
  `httparchive.scratchspace.serviceworkers_2021_07_01_mobile`
WHERE
  lh_service_worker = '0'
  AND ServiceWorkerControlledPage = 0
  AND ServiceWorker = 0
  AND serviceWorkerHeuristic = 1
```

When I run a few of them through Chrome they do use a service worker.

The `serviceWorker` custom metric looks for an instance of `navigator.serviceWorker.register` and expects the provided file to also be in the URL (which we know doesn't always work for base URL) but misses things like this (as the argument is a variable and not the URL):

```js
const serviceWorkerPath = '/runtime-service-worker.js?v=2';
navigator.serviceWorker.register(serviceWorkerPath, { scope: './' })
```

So it will definitely undercount. I really think we should enhance this custom metric to look for any instance of `navigator.serviceWorker.register`, as well as an instance that matches a url fetched. In fact maybe we should even drop that `serviceWorker` object since it will only catch simple ones?

Lighthouse sometimes fails the audit as they don't register a `start_url` which explains the differences with those. And suggests lighthouse undercounts too.

Which leaves us with blink_features. [This seems to be the code for that](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/modules/service_worker/service_worker_container.cc;l=494;drc=a7dbdae762701a9f07a2d8d0f2335c4e911ceba2) but don't know Chrome code well enough to understand why it might undercount.

Not sure if you have any insights here @demianrenzulli or any contacts within Google that could explain if it's too cautious?

I have however found some examples where the heuristic got it wrong:
- https://lostcoastoutpost.com/ seems to have the serviceWorker commented out
- https://www.tkvertriebs-auto.de/ uses firebase but doesn't register a service worker
- ...etc.
Both of these have a swEventListenersInfo but no swMethodsInfo so maybe we should exclude those from the heuristics going forward.

So I'm not sure if we should use `serviveWorker` or `serviceWorkerHeuristics`? In fact I'm tempted to use the later (or perhaps a combination of `serviceWorker and `swMethodsInfo` not being null?) as think the former is undercounting based on spot checks.

WDYT @demianrenzulli ?

tunetheweb:
@demianrenzulli I have updated all the queries (thanks for the JavaScript help!) and have run them all on the full dataset and put the data here: https://docs.google.com/spreadsheets/d/16AkIdDBBkCR5Kgb7kyfYvnNLQBu23Vsh7MUSFHW9RtA/edit#gid=398503119

I choose to use the `serviceWorkerHeuristic` metric for those queries where we want to look at just PWA pages.

Please review the queries and the data and let me know if you see anything that requires further investigation.

rockeynebhwani:
@tunetheweb - Usage of assetlinks look very high in your results sheet (approx. 89% of all sites have this as per the sheet). This looks very high. Are we sure the queries are working as expected?

tunetheweb:
> @tunetheweb - Usage of assetlinks look very high in your results sheet (approx. 89% of all sites have this as per the sheet). This looks very high. Are we sure the queries are working as expected?

Yup you are correct. Query was hopelessly naive. Fixed in https://github.com/HTTPArchive/almanac.httparchive.org/pull/2272/commits/d703954bdb25cb702ac650073f842be2ef029b09 and data set updated.

demianrenzulli:
Hi @tunetheweb, I've added some additional queries:

- **install_events.sql:** This one should replace **beforeinstallprompt_usage.sql** as it checks for both `beforeinstallprompt` and `appinstalled`. This is the only query that doesn't use `serviceWorkerHeuristics` in the `WHERE` clause to filter results, as both are Window events. I need to add the filter to exclude false positives (e.g. the Youtube library that calls `beforeinstallprompt`).
- **sw-registration-properties.sql:** Usage of properties in the [ServiceWorkerRegistration object](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration).
- **sw_methods.sql:** Usage of service worker methods (currently only `skipWaiting()`).
- **sw_objects.sql**: Usage of common service worker objects and their methods (e.g. `cache`, `clients`, etc).

I've noticed something else in the existing queries:

**sw_events.sql** only checks for `swEventListenersInfo`, but it should combine this with `swPropertiesInfo`, since both are the same thing, but declared differently:

- `self.addEventListener('install', function(event) {...};` vs. `self.oninstall = function(event) { ... };`

tunetheweb:
> Hi @tunetheweb, I've added some additional queries:

Cool. BTW, can you update the list in the top issue when adding new queries in future? It helps us track whether this PR is ready to go. I've added them now and ticked them all apart from `install_event.sql` as that's still being worked on so we can see one query still to be completed.

I've fixed the parsing errors and run all the new queries on the full data set (except for `install_event.sql` which I ran on sample data set only as still WIP) and added to [the results sheet](https://docs.google.com/spreadsheets/d/16AkIdDBBkCR5Kgb7kyfYvnNLQBu23Vsh7MUSFHW9RtA/edit#gid=398503119).

> * **install_events.sql:** This one should replace **beforeinstallprompt_usage.sql** as it checks for both `beforeinstallprompt` and `appinstalled`.

I've deleted **beforeinstallprompt_usage.sql**.

> This is the only query that doesn't use `serviceWorkerHeuristics` in the `WHERE` clause to filter results, as both are Window events. I need to add the filter to exclude false positives (e.g. the Youtube library that calls `beforeinstallprompt`).

I presume you're still working on that?

> I've noticed something else in the existing queries:
> 
> **sw_events.sql** only checks for `swEventListenersInfo`, but it should combine this with `swPropertiesInfo`, since both are the same thing, but declared differently:
> 
> * `self.addEventListener('install', function(event) {...};` vs. `self.oninstall = function(event) { ... };`

OK I've fixed that now. Didn't make much difference to the stats so seems like most people were only using `addEventListener` method (not unexpected to be honest).

So, other than the filtering you want to add on `install_events.sql` I think we're done?

sql/2021/pwa/lighthouse_pwa_audits_sw.sql
tunetheweb: Resolving this as we've decided to use `serviceWorkerHeuristics`.

demianrenzulli:
Thanks for working on these changes @tunetheweb!

> So, other than the filtering you want to add on **install_events.sql** I think we're done?

Exactly, I'll be working on that today, and commit my changes later.

sql/2021/pwa/manifests_not_json_parsable.sql
tunetheweb: Had a look at the 404 suggestion and 99.86% of sites return a 200. So not that interesting, so won't bother including. Will leave you to decide whether you wanna use the existing stat or not. Resolving.

client | status | freq | total | pct
-- | -- | -- | -- | --
desktop |   | 122 | 98,655 | 0.12%
desktop | 304 | 45 | 98,655 | 0.05%
desktop | 200 | 98,488 | 98,655 | 99.83%
mobile | 404 | 1 | 128,541 | 0.00%
mobile | - | 5 | 128,541 | 0.00%
mobile | 200 | 128,364 | 128,541 | 99.86%
mobile |   | 118 | 128,541 | 0.09%
mobile | 304 | 53 | 128,541 | 0.04%

```sql
CREATE TEMP FUNCTION jsonObjectKeys(input STRING)
RETURNS Array<String>
LANGUAGE js AS """
  try {
    return Object.keys(JSON.parse(input));
  } catch {
    return "['Invalid']";
  }
""";

SELECT
  _TABLE_SUFFIX as client,
  status,
  COUNT(DISTINCT url) AS freq,
  SUM(COUNT(DISTINCT url)) OVER (PARTITION BY _TABLE_SUFFIX) AS total,
  COUNT(DISTINCT url) / SUM(COUNT(DISTINCT url)) OVER (PARTITION BY _TABLE_SUFFIX) AS pct
FROM
  --`httparchive.sample_data.pages_*`,
  UNNEST(jsonObjectKeys(JSON_EXTRACT(payload, '$._pwa.manifests'))) AS manifest_url
LEFT OUTER JOIN
  (
    SELECT
      _TABLE_SUFFIX as _TABLE_SUFFIX,
      p.url,
      r.url as manifest_url,
      status
    FROM
      --`httparchive.sample_data.summary_requests_*` r
      `httparchive.summary_requests.2021_07_01_*` r
    JOIN
      --`httparchive.sample_data.summary_pages_*` p
      `httparchive.summary_pages.2021_07_01_*` p
    USING (_TABLE_SUFFIX, pageid)
  )
USING (_TABLE_SUFFIX, url, manifest_url)
WHERE
  JSON_EXTRACT(payload, '$._pwa') != "[]" AND
  JSON_EXTRACT(payload, '$._pwa.manifests') != "[]" AND
  JSON_EXTRACT(payload, '$._pwa.serviceWorkerHeuristic') = "true"
GROUP BY
  client,
  status

```

GJFR:
> > @tunetheweb - Usage of assetlinks look very high in your results sheet (approx. 89% of all sites have this as per the sheet). This looks very high. Are we sure the queries are working as expected?
> 
> Yup you are correct. Query was hopelessly naive. Fixed in [d703954](https://github.com/HTTPArchive/almanac.httparchive.org/commit/d703954bdb25cb702ac650073f842be2ef029b09) and data set updated.

Just a heads-up; this will probably be an undercount due to my naive error handling in the custom metric script. For more info: I've made a PR with a fix [here](https://github.com/HTTPArchive/legacy.httparchive.org/pull/230).

sql/2021/pwa/pwa_notification_acceptance_rates.sql
```diff
@@ -0,0 +1,63 @@
...
+SELECT
+  date,
+  client,
+
+  ROUND(APPROX_QUANTILES(notification_permission_accept, 1000 RESPECT NULLS)[OFFSET(100)], 3) AS notification_permission_accept_10th_percentile,
+  ROUND(APPROX_QUANTILES(notification_permission_accept, 1000 RESPECT NULLS)[OFFSET(250)], 3) AS notification_permission_accept_25th_percentile,
+  ROUND(APPROX_QUANTILES(notification_permission_accept, 1000 RESPECT NULLS)[OFFSET(500)], 3) AS notification_permission_accept_50th_percentile,
+  ROUND(APPROX_QUANTILES(notification_permission_accept, 1000 RESPECT NULLS)[OFFSET(750)], 3) AS notification_permission_accept_75th_percentile,
+  ROUND(APPROX_QUANTILES(notification_permission_accept, 1000 RESPECT NULLS)[OFFSET(900)], 3) AS notification_permission_accept_90th_percentile,
+  ROUND(APPROX_QUANTILES(notification_permission_accept, 1000 RESPECT NULLS)[OFFSET(1000)], 3) AS notification_permission_accept_100th_percentile,
```
rviscomi, line 14: Use a `percentile` field from `UNNEST([10, 25, 50, 75, 90, 100])` instead?

sql/2021/pwa/sw_adoption_over_time.sql
```diff
@@ -0,0 +1,15 @@
+#standardSQL
+# SW adoption over time
+SELECT DISTINCT
+  SUBSTRING(yyyymmdd, 0, 4) || '-' || SUBSTRING(yyyymmdd, 5, 2) || '-' || RIGHT(yyyymmdd, 2) AS date,
```
rviscomi, line 4: Here's a more standard way of formatting the date
```suggestion
  REGEXP_REPLACE(yyyymmdd, r'(\d{4})(\d{2})(\d{2})', r'\1-\2\-3') AS date,
```

sql/2021/pwa/sw_events.sql
```diff
@@ -0,0 +1,66 @@
...
+
+SELECT
+  _TABLE_SUFFIX AS client,
+  event,
+  COUNT(DISTINCT url) AS freq,
+  total,
+  COUNT(DISTINCT url) / total AS pct
+FROM
+  `httparchive.pages.2021_07_01_*`,
+  UNNEST(getSWEvents(parseField(JSON_EXTRACT(payload, '$._pwa.swEventListenersInfo')), parseField(JSON_EXTRACT(payload, '$._pwa.swPropertiesInfo')))) AS event
```
rviscomi, line 38: Any reason why you don't pass the JSON_EXTRACT'd value of `$._pwa`, which should be relatively small, and extract the listeners/properties arrays in the UDF?

sql/2021/pwa/sw_objects.sql
```diff
@@ -0,0 +1,51 @@
...
+''';
+SELECT
+  _TABLE_SUFFIX AS client,
+  sw_object,
+  COUNT(DISTINCT url) AS freq,
+  total,
+  COUNT(DISTINCT url) / total AS pct
+FROM
+  `httparchive.sample_data.pages_*`,
+  --`httparchive.pages.2021_07_01_*`,
```
rviscomi, line 24: Does this need to be rerun?

```suggestion
  `httparchive.pages.2021_07_01_*`,
```

sql/2021/pwa/sw_objects.sql
```diff
@@ -0,0 +1,51 @@
...
+  UNNEST(getSWObjects(JSON_EXTRACT(payload, '$._pwa.swObjectsInfo'))) AS sw_object
+JOIN
+  (
+    SELECT
+      _TABLE_SUFFIX,
+      COUNT(0) AS total
+    FROM
+      `httparchive.sample_data.pages_*`
+    -- `httparchive.pages.2021_07_01_*`
```
rviscomi, line 33:
```suggestion
      `httparchive.pages.2021_07_01_*`
```

sql/2021/pwa/top_manifest_categories.sql
```diff
@@ -0,0 +1,89 @@
...
+  (
+    SELECT
+      _TABLE_SUFFIX,
+      COUNT(0) AS total
+    FROM
+      `httparchive.pages.2021_07_01_*`
+    WHERE
+      JSON_EXTRACT(payload, '$._pwa') != "[]" AND
+      JSON_EXTRACT(payload, '$._pwa.manifests') != "[]" AND
+      JSON_EXTRACT(payload, '$._pwa.serviceWorkerHeuristic') = "true"
```
rviscomi, line 36: There are 12 `JSON_EXTRACT` statements in this query. Any way to simplify that?

sql/2021/pwa/top_manifest_display_values.sql
```diff
@@ -0,0 +1,58 @@
...
+  `httparchive.pages.2021_07_01_*`
+WHERE
+  JSON_EXTRACT(payload, '$._pwa') != "[]" AND
+  JSON_EXTRACT(payload, '$._pwa.manifests') != "[]"
+GROUP BY
+  client,
+  display
+HAVING
+  display IS NOT NULL AND
+  freq > 1000
```
rviscomi, line 53: Is this intentionally different from 100 above?

sql/2021/pwa/top_manifest_icon_sizes.sql
```diff
@@ -0,0 +1,88 @@
...
+  return null;
+}
+''';
+
+SELECT
+  "PWA Sites" AS type,
+  _TABLE_SUFFIX AS client,
+  size,
+  COUNT(0) AS freq,
+  total,
```
rviscomi, line 18: Same optimization as the following query
```suggestion
  SUM(COUNT(0)) OVER (PARTITION BY _TABLE_SUFFIX) AS total,
```

sql/2021/pwa/top_manifest_properties.sql
```diff
@@ -0,0 +1,92 @@
...
+  return null
+}
+''';
+
+SELECT
+  "PWA Sites" AS type,
+  _TABLE_SUFFIX AS client,
+  property,
+  COUNT(0) AS freq,
+  total,
```
rviscomi, line 22:
```suggestion
  SUM(COUNT(0)) OVER (PARTITION BY _TABLE_SUFFIX) AS total,
```

rviscomi (changes requested):

sql/2021/pwa/install_events.sql
```diff
@@ -0,0 +1,85 @@
+#standardSQL
+# SW install events
+
+CREATE TEMPORARY FUNCTION getInstallEvents(windowEventListeners ARRAY<STRING>, windowPropertiesInfo ARRAY<STRING>)
+RETURNS ARRAY<STRING> LANGUAGE js AS '''
+try {
+  return [...new Set([...windowEventListeners ,...windowPropertiesInfo])];
+} catch (e) {
+  return [e];
```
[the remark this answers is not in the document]
tunetheweb: Is it not good to return these? So if a large proportion of pages return debug output instead of real data then that suggests you JavaScript might be fishy?

if only a small proportion return this then unlikely to impact results.

The concern with NOT returning this is errors like that get hidden.

sql/2021/pwa/install_events.sql
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+
+    var objectValues = Object.values(jsonObject);
+    if (typeof objectValues != 'string') {
+        objectValues = objectValues.toString();
+    }
+
+    // Double check we are not back to an empty string
+    if(field == '[]' || field == '') {
```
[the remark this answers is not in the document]
tunetheweb: @demianrenzulli can you take this one?

sql/2021/pwa/manifests_not_json_parsable.sql
```diff
@@ -0,0 +1,53 @@
...
+  JSON_EXTRACT(payload, '$._pwa.manifests') != "[]" AND
+  JSON_EXTRACT(payload, '$._pwa.serviceWorkerHeuristic') = "true"
+GROUP BY
+  client,
+  can_parse
+UNION ALL
+SELECT
+  'All Pages' AS type,
+  _TABLE_SUFFIX AS client,
+  canParseManifest(JSON_EXTRACT(payload, '$._pwa.manifests')) AS can_parse,
```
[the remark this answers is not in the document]
tunetheweb: Yup. There’s an awful lot. And we haven’t figured out why!!

sql/2021/pwa/popular_pwa_libraries.sql
```diff
@@ -0,0 +1,108 @@
...
+  /* 'importScriptsInfo' returns an array of script that might import other script
+     The final array of script comes from the combination of both */
+  var ObjKeys = Object.keys(JSON.parse(importScriptsInfo));
+  var ObjValues = Object.values(JSON.parse(importScriptsInfo));
+  if (typeof ObjKeys == 'string') {
+    ObjKeys = [ObjKeys];
+  }
+  if (typeof ObjValues == 'string') {
+    ObjValues = [ObjValues];
+  }
```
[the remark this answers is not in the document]
tunetheweb: @demianrenzulli ?

sql/2021/pwa/install_events.sql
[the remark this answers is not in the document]
rviscomi: It's ok return debugging info while you're still refining the query, similar to using the sample_data tables, but for the "official" queries they should only return the actual data.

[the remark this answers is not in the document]
tunetheweb: Well hopefully the exceptions should never be triggered if the custom metric and UDF are written correctly so it’s hopefully a moot point!

But still concerned with a full query run throwing some error that wasn’t seen in sample_data and it being hidden if this isn’t returned.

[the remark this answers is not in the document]
rviscomi: By all means, include debug output while you're iterating and revise the query as needed. When you're done though, remove the debug output before checking the query in and saving the results to Sheets.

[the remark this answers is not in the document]
tunetheweb: Or an error message instead of silently swallowing the error and pretending nothing's wrong.

But fine. They are gone.

sql/2021/pwa/sw_objects.sql
tunetheweb: Reran and got same numbers as sheet so must have just checked in the wrong code.

sql/2021/pwa/top_manifest_categories.sql
tunetheweb: To be honest there's not real reason for the first one (`JSON_EXTRACT(payload, '$._pwa') != "[]"`) since the second and third covers this. Will remove from all queries which will bring it down a bit. But still be a lot of these.

sql/2021/pwa/sw_events.sql
tunetheweb: Not really, this it just reflects the way the query was built up. Rewrote.

sql/2021/pwa/top_manifest_display_values.sql
tunetheweb: Nope. Changed. There only are 5 in our current data set anyway so 100 definitely enough 😁

sql/2021/pwa/top_manifest_icon_sizes.sql
tunetheweb: Actually we can't do this in this query and took a bit of digging into why. We need to be very careful with this syntax - it can only be used when the following applies:
- You are not using `HAVING` to restrict to top 100 (or whatever). `QUALIFY` is safe to use, as applied after.
- You are not doing any joins (including `UNNEST`)
- You want the total to reflect the `WHERE` clause.

In this case we are using both `UNNEST` and `HAVING`. We could switch the `HAVING` to `QUALIFY` but still have the `UNNEST`. So can't use this short cut, and have to use a `JOIN` to a total table.

In fact this has identified that I was using this total shortcut incorrect in `top_manifest_display_values` so have fixed that.

I wonder how many other people are using this incorrectly?

sql/2021/pwa/top_manifest_properties.sql
tunetheweb: As per above can't use this shortcut here due to `UNNEST` and `HAVING` clauses.

sql/2021/pwa/top_manifest_categories.sql
tunetheweb: As per above idea, this (and all the other queries run for both All and PWA Sites) could be refactored to do the main select in a CTE like this, but honestly think the `UNION ALL` is more easily understood. Last year these were separate queries but this year I decided to merge precisely because there was a lot of repetition.

```sql
#standardSQL
# Top manifest categories
CREATE TEMPORARY FUNCTION getCategories(manifest STRING)
RETURNS ARRAY<STRING> LANGUAGE js AS '''
try {
  var $ = Object.values(JSON.parse(manifest))[0];
  var categories = $.categories;
  if (typeof categories == 'string') {
    return [categories];
  }
  return categories;
} catch (e) {
  return null;
}
''';

WITH totals AS (
  SELECT
    _TABLE_SUFFIX,
    COUNT(0) AS total,
    COUNTIF(JSON_EXTRACT(payload, '$._pwa.serviceWorkerHeuristic') = "true") AS pwa_total
  FROM
  WHERE
    JSON_EXTRACT(payload, '$._pwa.manifests') != "[]"
  GROUP BY
    _TABLE_SUFFIX
),

manifests_categories AS (
  SELECT
    'All Sites' AS type,
    _TABLE_SUFFIX AS client,
    category,
    COUNT(0) AS freq,
    total,
    COUNT(0) / total AS pct,
    COUNTIF(JSON_EXTRACT(payload, '$._pwa.serviceWorkerHeuristic') = "true") AS pwa_freq,
    pwa_total,
    COUNTIF(JSON_EXTRACT(payload, '$._pwa.serviceWorkerHeuristic') = "true") / pwa_total AS pwa_pct
  FROM
    UNNEST(getCategories(JSON_EXTRACT(payload, '$._pwa.manifests'))) AS category
  JOIN
    totals
  USING (_TABLE_SUFFIX)
  WHERE
    JSON_EXTRACT(payload, '$._pwa.manifests') != "[]"
  GROUP BY
    client,
    category,
    total,
    pwa_total
  HAVING
    category IS NOT NULL
  ORDER BY
    type DESC,
    freq / total DESC,
    category,
    client
)

SELECT
  'PWA Sites' AS type,
  client,
  category,
  pwa_freq AS freq,
  pwa_total AS total,
  pwa_pct AS pct
FROM
  manifests_categories
UNION ALL
SELECT
  'All Sites' AS type,
  client,
  category,
  freq,
  total,
  pct
FROM
  manifests_categories
ORDER BY
  type DESC,
  pct DESC,
  category,
  client
```

sql/2021/pwa/top_manifest_icon_sizes.sql
obto: Yeah just learned that window functions are applied after HAVING:
```
FROM -> WHERE -> GROUP BY -> HAVING -> OVER -> QUALIFY
```

UNNEST generally isn't a problem in and of itself though, you just have to remember that you've fundamentally changed what you're querying now. A simple `COUNT(DISTINCT url)` can completely fix that problem though

sql/2021/pwa/manifests_not_json_parsable.sql
```diff
@@ -0,0 +1,52 @@
+#standardSQL
+# Manifests that are not JSON parsable for service worker pages and all pages
+
+CREATE TEMP FUNCTION canParseManifest(manifest STRING)
+RETURNS BOOLEAN LANGUAGE js AS '''
+try {
+  manifestJSON = Object.values(JSON.parse(manifest))[0];
```
obto, line 7:
```suggestion
  var manifestJSON = Object.values(JSON.parse(manifest))[0];
```

sql/2021/pwa/popular_pwa_libraries.sql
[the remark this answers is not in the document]
obto: Yes, they will only return an array. Plus if JSON.parse fails it throws and error and heads over to the catch block, so those two branches can be removed.

sql/2021/pwa/popular_pwa_libraries_helper.sql
```diff
@@ -0,0 +1,54 @@
...
+  /* 'importScriptsInfo' returns an array of libraries that might import other libraries
+     The final array of libraries comes from the combination of both */
+  var ObjKeys = Object.keys(JSON.parse(importScriptsInfo));
+  var ObjValues = Object.values(JSON.parse(importScriptsInfo));
+  if (typeof ObjKeys == 'string') {
+    ObjKeys = [ObjKeys];
+  }
+  if (typeof ObjValues == 'string') {
+    ObjValues = [ObjValues];
+  }
```
obto, line 16: See previous comment by Rick. These two branches aren't needed since `Object.keys` and `Object.values` always returns an array
```suggestion
```

sql/2021/pwa/top_manifest_properties.sql
```diff
@@ -0,0 +1,88 @@
+#standardSQL
+# Top manifest properties
+
+CREATE TEMP FUNCTION getManifestProps(manifest STRING)
+RETURNS ARRAY<STRING> LANGUAGE js AS '''
+try {
+  manifestJSON = Object.values(JSON.parse(manifest))[0];
```
obto, line 7:
```suggestion
  var manifestJSON = Object.values(JSON.parse(manifest))[0];
```

sql/2021/pwa/workbox_methods.sql
```diff
@@ -0,0 +1,58 @@
+#standardSQL
+#Workbox methods
+CREATE TEMPORARY FUNCTION getWorkboxMethods(workboxInfo STRING)
+RETURNS ARRAY<STRING> LANGUAGE js AS '''
+try {
+
```
obto, line 6:
```suggestion
```

sql/2021/pwa/workbox_packages.sql
```diff
@@ -0,0 +1,61 @@
+#standardSQL
+#Workbox package and methods
+CREATE TEMPORARY FUNCTION getWorkboxPackages(workboxInfo STRING)
+RETURNS ARRAY<STRING> LANGUAGE js AS '''
+try {
+
```
obto, line 6:
```suggestion
```

sql/2021/pwa/workbox_versions.sql
```diff
@@ -0,0 +1,61 @@
+#standardSQL
+#Workbox versions
+CREATE TEMPORARY FUNCTION getWorkboxVersions(workboxInfo STRING)
+RETURNS ARRAY<STRING> LANGUAGE js AS '''
+try {
+
```
obto, line 6:
```suggestion
```

obto (approved):
More or less an LGTM from me. Just small js formatting and one optimization

sql/2021/pwa/install_events.sql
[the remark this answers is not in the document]
tunetheweb: Rewrote this anyway so resolving.

sql/2021/pwa/top_manifest_icon_sizes.sql
tunetheweb: The problem is you are not doing a simple `COUNT(DISTINCT url)`, but a `SUM(COUNT(DISTINCT url))` so that's not necessarily distinct URLs.

tunetheweb:
OK addressed all your feedback @obto
 
@rviscomi there's a few points which haven't been addressed in yours but not sure we should change. Can you have another look at the open items and my feedback and let me know if you're good to merge as is or not?

sql/2021/pwa/lighthouse_pwa_audits.sql
```diff
@@ -0,0 +1,68 @@
...
+      `httparchive.pages.2021_07_01_mobile`
+    WHERE
+      JSON_EXTRACT(payload, '$._pwa') != "[]" AND
+      JSON_EXTRACT(payload, '$._pwa.serviceWorkerHeuristic') = "true" AND
+      JSON_EXTRACT(payload, '$._pwa.manifests') != "[]"
+  )
+USING (url)
+GROUP BY
+  audits.id
+UNION ALL
```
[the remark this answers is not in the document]
tunetheweb: Decided to go with the CTE method for these more complex queries. Have left the simpler ones as is. Resolving.

sql/2021/pwa/top_manifest_categories.sql
tunetheweb: Decided to go with the CTE method for these more complex queries. Have left the simpler ones as is.

sql/2021/pwa/top_manifest_icon_sizes.sql
tunetheweb: Resolving this as updated @rviscomi 's other concern on the COUNTs.

sql/2021/pwa/lighthouse_pwa_audits.sql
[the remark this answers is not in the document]
tunetheweb: BTW this was one of the simpler ones 😁

`LEFT JOIN` won't help as above and CTE won't really make the query any shorter or easier to understand IMHO.

Resolving.

PR #2273, opened 2021-07-12
Add Site Search to website

### LINKED ISSUE #559: Add Search Functionality
I think we should have the ability to search this amazing collection of data.

We could use Google Custom Search or similar. I’ve used SiteSearch360 a bit since Google stopped offering to accept money for ad-free version of GCS. Though lot of contributors and some sponsorship coming from Google so that’s probably not a runner.

Or alternatively just use main Google with a `site:https://almanac.httparchive.org/en/2019` option. That’s probably the simplest and would also make it easy to change the URL searched and option passed to serve different languages and/or years. For example French pages could search www.google.fr with `site:https://almanac.httparchive.org/fr/2019` option.

Or could do something server side?

Thoughts?

tunetheweb:
Fixes #559 

Just a draft PR for now as needs some more work but an initial implementation of Site Search using Google's Programmable Search (formerly known as Google Custom Search).

- [x] - Adds a new route (/en/search - and similar for other languages) with a search box and a search results
- [x] - Limits search to the currently selected language
- [x] - Adds a "Search in English" option for those translations (as most don't have full translations)
- [x] - Adds a filter for 2020 and 2019 results to limit results to those (need to remember to add 2021 on launch)
- [x] - Adds search menu to header and footer
- [x] - Adds pytest and npm tests
- [x] - No Ads
- [x] - Add `SearchAction` Schema.org to home page to hopefully [add search bar to Google SERPs](https://developers.google.com/search/docs/data-types/sitelinks-searchbox).

I've staged a version here for you to play: https://20210722t214744-dot-webalmanac.uk.r.appspot.com/en/search#gsc.q=JavaScript (note URL has been updated since PR was initially opened)

Let me know your thoughts.
Would this be useful?
Alternative approaches?


tunetheweb:
And if any @HTTPArchive/developers wants to help with the CSS to add a form field to the nav then let me know!

shantsis (commented):
Which part(s) of the page is customizable? For example, can we adjust the colors to avoid color contrast issues?
color contrast

Also can we adjust the search button to be the same height as the input?
search field

Is any autocomplete/suggested searches possible when focusing on the field?

BTW seems like the tab focus is partially cut off on the first two (I got this on click)
tab focus

tunetheweb:
Colours are completely customisable but other than that can seem to change much else (e.g. the button size you mention).

The other option is just to have a plain search text box on our page that directs back to this page on submit and not use their search box, so we have full control over the styling. Have done that before. Then just use their component for the search results (which picks the query up from the query param in the URL).

Auto complete is enabled but says it might take a few hours to start working. And obviously if we move to our search entry field then can’t have auto complete. Though to be honest not sure how important that is.

tunetheweb:
Feel try to suggest colours. Or DM me your email and will temporarily add you as an admin so you can play with the settings yourself.

hemanth (approved):
LGTM

shantsis:
If we want to use colors as close to the current, we just need to adjust the green to #00852C or something similar

tunetheweb:
> Only thing I would have changed would be to add evident demarcations for the ads

I'm hoping to be able to get rid of the ads completely.

> If we want to use colors as close to the current, we just need to adjust the green to #00852C or something similar

That's updated now.

tunetheweb:
I've uploaded a new version to https://20210720t193916-dot-webalmanac.uk.r.appspot.com/en/search with improved navigation integration and a few other fixes.

Think it's pretty much ready for review so taking it out of draft.

shantsis:
- Can we add some placeholder text to the search field? Either just "search" or some examples like "Search for topics, pages, authors etc"
- I didn't realize search was in the nav, and the dropdown is odd. What if it's just a link that goes to the search page, or opens the overlay that's there now? The search icon may help to find that action if someone is scanning instead of reading

tunetheweb:
> * Can we add some placeholder text to the search field? Either just "search" or some examples like "Search for topics, pages, authors etc"

Fraid Google Programmable Search doesn't allow that. Plus placeholder text gets [a very bad rep in the accessibility world](https://www.smashingmagazine.com/2018/06/placeholder-attribute/)!

> * I didn't realize search was in the nav, and the dropdown is odd. What if it's just a link that goes to the search page, or opens the overlay that's there now? The search icon may help to find that action if someone is scanning instead of reading

I just spent ages converting the link to add the field! 😀 It was a link previously.

I would say norm is to allow the search phrase to be entered on any page rather than having to go to search link and then enter the phase, and then click it? But yeah ideally we'd have a search field but except on very wide desktop there's not enough space in the nav for that so would need a bit of a redesign.

Open to alternative thoughts here though. Could go back to the plain link if people think it's better...

shantsis:
> Fraid Google Programmable Search doesn't allow that. Plus placeholder text gets a very bad rep in the accessibility world!
It does when it's used in place of a label. But we don't have either right now so visually it looks worse :(

> I just spent ages converting the link to add the field! 😀 It was a link previously.
Whoops my bad! Maybe it's just the chevron styling itself. Maybe just replacing that with the search icon might do the trick?

tunetheweb:
> It does when it's used in place of a label. But we don't have either right now so visually it looks worse :(

Where do you mean?

- On Desktop the search field is only shown when the Search button is clicked so I'd argue that's a sufficient label. Plus there is the magnifying glass on the search button.
- On Mobile I admit it's not as clear, but again has the magnifying glass on the search button.
- On the search results page, there is the big Search H1 just about the search box which again I'd argue is a pretty clear label!
 
> Whoops my bad! Maybe it's just the chevron styling itself. Maybe just replacing that with the search icon might do the trick?

The chevon is consistent with the other navigation menus to show it will expand/can be collapsed again.

Need to dash off so will have a think on it some more...

shantsis:
> Where do you mean?

I meant the full-page view but your points are valid.

BTW the search button isn't labelled for screen readers, it just reads as "button"
Screen Shot 2021-07-20 at 3 43 18 PM

tunetheweb:
Yeah I noticed that too. Though occasionally Voiceover did speak it as “Button Search” but not consistently. Again, that’s rendered by Google Programmable Search so again outside our control 😔

Not ideal I agree but on the other hand, not sure how useful it is to have the heading announced as “Search” the input announced as “Search” an then a button announced as “Search”. Seems bordering on overkill!

Anyway not how I’d personally code it but we’re going to have to live with it. We can make our own input to submit a request to `/en/search?q=JavaScript` which will work (as it’ll pick up the query param from the URL when that page loads). But then lose the handy auto complete on the input. So think it’s better to use theirs.

tunetheweb:
OK played with this some more. I really wasn't happy with the amount of compromises we had to make to have the Google Programmable Search widget in the header and footer:
- Limited styling options
- Loads a load of JS. It's `defer`-ed but still adds a lot to the page weight
- Takes a while to load
- Needed to significantly reduce out Content Security Policy restrictions

So changed it to use our own input field. Which means we can style as we want so added that placeholder. We do loss the auto complete but just don't think above compromises are worth that functionality.

Latest iteration here: https://20210721t190648-dot-webalmanac.uk.r.appspot.com

shantsis:
I love the linting comments 😄 

Can we add a border around the input field on the full page view? It's hard to see right now. Maybe #939393?

tunetheweb:
> I love the linting comments 😄

😄

> Can we add a border around the input field on the full page view? It's hard to see right now. Maybe #939393?

Done: https://20210721t212824-dot-webalmanac.uk.r.appspot.com

shantsis:
Looks great, thanks!

Issue #2174, opened 2021-04-27
Structured Data 2021


rviscomi:
# Part I Chapter 4: Structured Data

If you're interested in contributing to the Structured Data chapter of the 2021 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| ---- | ------- | --------- | -------- | ------- | ----------- |
| @jonoalderson | @jonoalderson @cyberandy | @kevinmarks @vdwijngaert @jvandriel @philbarker | @GregBrimble  | @jvandriel @JasmineDWillson | @rviscomi |

<details>
<summary><b>Expand for more information about each role</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
- **[Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)** are responsible for researching the stats and trends used throughout the Almanac. Analysts work closely with authors and reviewers during the planning phase to give direction on the types of stats that are possible from the dataset, and during the analyzing/writing phases to ensure that the stats are used correctly.
[...]

PR #2293, opened 2021-08-04
Structured Data 2021

GregBrimble:
Queries for the Structured Data chapter (#2174)


tunetheweb:
Hi @GregBrimble , I've just merged a SQL Linter into `main`. When you get a chance could you merge into this branch so your new queries will be linted according to the Web Almanac style?

tunetheweb:
Hey @GregBrimble hope you don't mind, but I've just merged `main` into this branch and fixed your linting errors. Keen to get all open branches on this linter.

jono-alderson:
Hey @GregBrimble , any updates from your side? Nervous that things have gone quiet in the Slack channel.

GregBrimble:
Okay, so, that's [the PR comment](https://github.com/HTTPArchive/almanac.httparchive.org/pull/2293#issue-703107336) now containing all analysis tasks for this chapter. I've got to ponder a bit about a couple of those last points under "Unknowns". Any thoughts are welcome.

I took a stab at doing as much analysis as I could in BigQuery, but as predicted, it was basically impossible for the complex queries. So, I've downloaded a local copy of the `rendered` structured data custom metrics (27GB) and have been playing around with it on my machine. I've completed all the non-JSON-LD queries, apart from the RDFa ones. Those just need a little more processing before they can be useful. I'll upload results and get back to the remaining queries shortly.

jono-alderson:
Nice work, @GregBrimble!

On the unknowns, some thoughts:

- For scalar values, Can we normalize any which match the pattern `https?://schema.org/([a-zA-Z]+)` into the lowercase'd path, and, convert any resultant 'true' or 'false' strings into bool `true`/`false`? E.g:
  - `https://schema.org/True` becomes `(str) true`, and then `(bool) true`.
  - `https://www.example.com/Pineappele` remains unmodified.
- For the relationship counts (range vs domain vs both), I think domain is more interesting.

rviscomi:
@GregBrimble do you feel comfortable that you'd be able to get these queries written by the milestone deadline (end of the month)? If not, anything we can help with?

rviscomi:
@GregBrimble anything I can do to help unblock this PR?

rviscomi:
Hi @GregBrimble just checking in again. Do you have an ETA when we can expect the analysis to be completed? Anything we can do to help?

For context, authors only have 4 more weeks to take the results of the analysis and write the entire chapter, so the sooner this is wrapped up the sooner they can start. Hoping not to block them much longer.

GregBrimble:
Sorry for dropping off a bit there. I've just re-written the queries I already had as JS UDFs, and great news that it has worked without a hitch! Those results are now in [the results sheet](https://docs.google.com/spreadsheets/d/1uEA217YjpX1xdRVBnIz9ekjpjIrpGIh5GwUxvnzJNig/edit#gid=453283266).

I appreciate any early feedback anyone has on the queries. And I'll be moving onto the remaining queries this week. Again, sorry for the delay. I'll try to get the JSON-LD ones complete by Wednesday evening.

I haven't done any normalization yet, @jono-alderson , but thanks for your suggestion. Fine with lowercasing (because I don't think I've ever seen that really matter) but I'm a bit worried about stripping the schema.org hosts. Some people might set a value of `true` but use that with a different context vocabulary, so you'll be making potentially different values seem the same.

Feel free to play with the results sheet. I've added a `isValidURL` column (D) to `microdata_item_types` as a quick demo of the type of meta-analysis you might want to include (e.g. "almost 0.8% of the `itemtype` values set for Microdata structured data are invalid"). If you've got any questions, happy to go over any of it with you.

sql/2021/structured-data/rdfa_type_ofs.sql
```diff
@@ -0,0 +1,36 @@
+# standardSQL
+  # Count RDFa Type Ofs
+CREATE TEMP FUNCTION
+  getRDFaTypeOfs(rendered STRING)
+  RETURNS ARRAY<STRING>
+  LANGUAGE js AS r"""
+  try {
+    rendered = JSON.parse(rendered);
+    return rendered.rdfa_typeofs.map(typeOf => typeOf.toLowerCase());
```
GregBrimble, line 9: TODO: I should be splitting these at whitespace (e.g. `Book Product` becomes [`Book`, `Product`]).

GregBrimble: From [Schema.org](https://schema.org/additionalType):

```html
<div vocab="https://schema.org/" resource="#record" typeof="Book Product">
    <h3 property="name">
        Le concerto
    </h3>
</div>
```

jono-alderson:
Oh wow, this is looking awesome. Nice work!

A though; What do you think about normalizing HTTP and HTTPS prefixes for schema.org and ogp.me?

GregBrimble:
Yup, that's a good idea. When I can parse it as a URL, I'll drop the protocol so we're just looking at host + path.

sql/2021/structured-data/classic_microformats_types.sql
```diff
@@ -0,0 +1,36 @@
+# standardSQL
+  # Count Classic Microformats types
+CREATE TEMP FUNCTION
+  getClassicMicroformatsTypes(rendered STRING)
+  RETURNS ARRAY<STRUCT<name STRING, count NUMERIC>>
+  LANGUAGE js AS r"""
```
rviscomi, line 6: I'm used to seeing `r''` for regex literals. What is `r` used for here?

sql/2021/structured-data/classic_microformats_types.sql
```diff
@@ -0,0 +1,36 @@
...
+    rendered = JSON.parse(rendered);
+    return rendered.microformats_classic_types.map(microformats_classic_type => ({name: microformats_classic_type.name, count: microformats_classic_type.count}));
+  } catch (e) {
+    return [];
+  }
+""";
+WITH
+  rendered_data AS (
+  SELECT
+    getClassicMicroformatsTypes(rendered) AS classicMicroformatsTypes
```
rviscomi, line 17: (throughout) prefer underscore naming convention for SQL field names
```suggestion
    getClassicMicroformatsTypes(rendered) AS classic_microformats_types
```

sql/2021/structured-data/classic_microformats_types.sql
```diff
@@ -0,0 +1,36 @@
...
+  FROM (
+    SELECT
+      JSON_EXTRACT(JSON_VALUE(JSON_EXTRACT(payload,
+            '$._structured-data')),
+        '$.structured_data.rendered') AS rendered
+    FROM
+      `httparchive.pages.2021_07_01_*`)
```
rviscomi, line 24: `_*` includes both desktop and mobile tables. We always want to keep these results separate since they have different characteristics.

Best to include a field `_TABLE_SUFFIX AS client` in the output and include that in any `GROUP BY` or `PARTITION BY` clauses.

sql/2021/structured-data/classic_microformats_types.sql
```diff
@@ -0,0 +1,36 @@
...
+)
+
+SELECT
+  classicMicroformatsType.name AS classicMicroformatsType,
+  SUM(classicMicroformatsType.count) AS count
```
rviscomi, line 29: Absolute counts are interesting but harder for readers to grasp the magnitude. Always also include relative percentages and a total field.
```suggestion
  SUM(classicMicroformatsType.count) AS count,
  SUM(SUM(classicMicroformatsType.count)) OVER (PARTITION BY client) AS total,
  SUM(classicMicroformatsType.count) / SUM(SUM(classicMicroformatsType.count)) OVER (PARTITION BY client) AS pct
```

sql/2021/structured-data/classic_microformats_types.sql
```diff
@@ -0,0 +1,36 @@
...
+  FROM (
+    SELECT
+      JSON_EXTRACT(JSON_VALUE(JSON_EXTRACT(payload,
+            '$._structured-data')),
+        '$.structured_data.rendered') AS rendered
+    FROM
+      `httparchive.pages.2021_07_01_*`)
+)
+
+SELECT
```
rviscomi, line 27:
```suggestion
SELECT
  client
```

sql/2021/structured-data/classic_microformats_types.sql
```diff
@@ -0,0 +1,36 @@
...
+  classicMicroformatsType.name AS classicMicroformatsType,
+  SUM(classicMicroformatsType.count) AS count
+FROM
+  rendered_data,
+  UNNEST(classicMicroformatsTypes) AS classicMicroformatsType
+GROUP BY
+  classicMicroformatsType
+ORDER BY
+  count DESC;
```
rviscomi, line 36: (throughout) Trailing commas are optional and by convention we omit them
```suggestion
  count DESC
```

sql/2021/structured-data/error_logs.sql
```diff
@@ -0,0 +1,13 @@
+#standardSQL
+  # A list of the error logs created in the run
```
rviscomi, line 2: @jono-alderson is this only useful for metadata about the success of the custom metric itself, or is this intended to be a reader-facing metric? If the former we can omit it from the list of queries.

If there are any opportunities to fix bugs reported in the logs, please file an issue to track it or PR to fix it.

sql/2021/structured-data/present_types.sql
```diff
@@ -0,0 +1,69 @@
+#standardSQL
+  # A count of pages which include each type of structured data
```
rviscomi, line 2: This query seems like it'd benefit from a custom JS function to return a list of structured data types present on the page to avoid the need to extract each value individually from the JSON. (instead, one JSON extraction per page)

That should reduce a lot of the redundancy in the SELECT.


Not exactly the same but very similar to my feedback on [this Performance query](https://github.com/HTTPArchive/almanac.httparchive.org/pull/2316/files#r721871823). See the linked Accessibility query as an example of the pattern I'm suggesting.

sql/2021/structured-data/summary_count.sql
```diff
@@ -0,0 +1,51 @@
...
+    ELSE
+    0
+  END
+    ) AS `no_success_no_errors`
+FROM (
+  SELECT
+    JSON_EXTRACT(`structured_data`,
+      '$.structured_data') AS `success`,
+    JSON_EXTRACT(`structured_data`,
+      '$.log') AS `errors`
```
rviscomi, line 45: I'd suggest evaluating these to booleans, which you can negate as needed in the outer SELECT, to avoid the need to repeatedly do the IS [NOT] NULL checks.

Also, are the backticks necessary?
```suggestion
    JSON_EXTRACT(`structured_data`, '$.structured_data') IS NOT NULL AS is_success,
    JSON_EXTRACT(`structured_data`, '$.log') IS NOT NULL AS has_errors
```

rviscomi (changes requested):
Looking good so far @GregBrimble. To avoid repeating myself in each query, take my feedback in the first query as general guidance for other queries.

For example, all queries should be sure to segment the results by desktop and mobile and include total and pct fields. Also style nits like using underscores for field names and omitting the trailing semicolon are applicable to all queries.

GregBrimble:
Just to confirm, @jono-alderson , if I've got the following:

```txt
Person {
  name "Greg Brimble"
  age 24
  worksFor Organization {
    name "Cloudflare"
  }
}
```

And we're counting the `worksFor` relationship, you're more interested in associating that with the `Person` entity than you are the `Organization` entity?

So we'd end up with something like:

| Entity:Relationship | Count |
| --- | --- |
| Person:name | 983 |
| Person:age | 651 |
| Person:worksFor | 35 |

Rather than:

| Relationship:Entity | Count |
| --- | --- |
| worksFor:Organization | 523 |
| worksFor:Corporation | 186 |
| worksFor:NGO | 32 |

Correct?

GregBrimble:
Yes. That's all queries completed, and I added another, `jsonld_max_depth` (finds the most nested entity type—18!) as a fun piece of trivia.

The results are on [the spreadsheet](https://docs.google.com/spreadsheets/d/1g_R64uLejGNaMYLOnKPyc7rPiof0gaEWO7R1CtqQzd0/edit) under their corresponding tabs.

I couldn't get figure out how to address those two remaining comments of yours, but they're just refactors anyway, so shouldn't affect any of the data.

PR #2298, opened 2021-08-06
Add sql linting

### LINKED ISSUE #991: Lint SQL

tunetheweb:
Fixes #991 as GitHub linter recently added [SQLFluff](), which includes [a lot of useful rules](https://docs.sqlfluff.com/en/stable/rules.html), to reduce chance of bad SQL getting in, and to enforce consistency in writing styles.

There's a lot of change in here but most are white space ([ignore those in the review](https://github.com/HTTPArchive/almanac.httparchive.org/pull/2298/files?diff=unified&w=1)).

Had a few issues (as well as cleaning up the whitespace complaints):
- I had to move the BigQuery UDF `OPTION` library import line further up (after `LANGUAGE js` and before `AS` line)
- Some complex `STRUCT` UDFs cause the linter to break so had to `--noqa` override them. Weirdly some `STRUCT` UDFs work fine but can't explain the difference of when it works and when it doesn't. Will dig some more and then raise an issue but for now need quite a few `--noqa` overrides for them.
- It didn't like the `DETERMINISTIC` UDF keyword. Will dig some more and then raise an issue but for now need a few `--noqa` overrides.
- It didn't like UDF Object references in `UNNEST` statements so again had to `--noqa` override them
- Can't specify a config file for linter to use in usual `.github/linters` location, but putting it in the `.sql` seems to work so a minor inconvenience (pricing it works in the GitHub action, which we're about to find out when I submit this).
- Had to turn off a lot of rules. Only one I miss and think we should probably have is L003 and I [raised an issue for this as it doesn't like how we indent `USING`](https://github.com/sqlfluff/sqlfluff/issues/1250).
- It didn't like `SUM(COUNT(0)) OVER () AS total,` syntax, but seems fine with `SUM(COUNT(0)) OVER (PARTITION BY 0) AS total,` which is equivalent so changed to that to save an exception.

More importantly it found a number of bugs in our code. Not many but definitely some there. So good we have both some syntax checking AND consistency checking.


sql/2019/cdn/17_19.sql
```diff
@@ -2,50 +2,50 @@
...
-  countif(ifnull(a.protocol, b.protocol) not in ('HTTP/0.9', 'HTTP/1.0', 'HTTP/1.1', 'HTTP/2')) as http_other,
-  countif(isSecure or ifnull(a.protocol, b.protocol) = 'HTTP/2') as tls_total,
-  round(100*countif(ifnull(a.protocol, b.protocol) = 'HTTP/0.9')/count(0), 2) as http09_pct,
-  round(100*countif(ifnull(a.protocol, b.protocol) = 'HTTP/1.0')/count(0), 2) as http10_pct,
-  round(100*countif(ifnull(a.protocol, b.protocol) = 'HTTP/1.1')/count(0), 2) as http11_pct,
-  round(100*countif(ifnull(a.protocol, b.protocol) = 'HTTP/2')/count(0), 2) as http2_pct,
-  round(100*countif(ifnull(a.protocol, b.protocol) not in ('HTTP/0.9', 'HTTP/1.0', 'HTTP/1.1', 'HTTP/2')) /count(0), 2) as http_other_pct,
-  round(100*countif(isSecure or ifnull(a.protocol, b.protocol) = 'HTTP/2')/count(0), 2) as tls_pct,
-  count(0) as total
+  countIF(IFNULL(a.protocol, b.protocol) = 'HTTP/0.9') AS http09,
```
rviscomi, line 5: Shou

tunetheweb: I've turned off function capitalisation checking (rule L030) as we don't use them for our UDF function names. Unfortunately it sees `COUNTIF` as a function rather than a keyword and can't differentiation capitalisation rules between inbuilt ones and UDFs.

Anyway fixed this oversight in https://github.com/HTTPArchive/almanac.httparchive.org/pull/2298/commits/08c5198f15d32146fc7bb8d424bdc282f614f9d5 but unfortunately errors like this won't be picked up automatically.

tunetheweb:
BTW I've not fixed everything as linting locally takes a long time for so many files. But have fixed a good few (mostly with global search and replaces), and run several sample chapters queries through the linter locally until they were clean.

So planning on leave it running overnight in this action and checking it out over the weekend and applying more changes. Wanted to know it would actually work before I spent any more time on it.

tunetheweb:
OK think this is good for merging now. Know there is a lot of change here but 

The biggest downside (other than all the change!) is the need to use `-- noqa: PRS` in some places to get the linter to skip a bit of code it can't handle.

15 instances when using `DETERMINISTIC`:

DETERMINISTIC code screenshot

This should be fixable in SQL Fluff and working on a patch. So might be able to remove some of these in future.

8 instances in UDF definitions with `STRUCT`s (note not all `STRUCT` UDFs are affected, but can't explain why some are at this time):

Function code screenshot

Looks more tricky to fix this one since it should in theory work.

And similar 7 Instances when calling UDFs and selecting object fields:

Object selection fields

Not sure how difficult to fix this one.

So 30 overrides across our 994 SQL files. More than I'd like, but less than I thought we'd need to be honest given how complex some of our SQL and functions are

I think the consistency this gives, and the bugs it's caught in this review, more than makes that minor inconvenience worthwhile, even if we have to add a few more of these in future.

Appreciate this is a lot to review, so don't think we can review every change but @rviscomi  if you're happy with the general direction of this and a few samples of what it has caught, then I'm going to merge this.

@HTTPArchive/analysts , FYI for this upcoming change and let me know if any of you have any comments or suggestions.

tunetheweb:
FYI opened a [PR to add DETERMINISTIC support to SQLFluff](https://github.com/sqlfluff/sqlfluff/pull/1252) but don't think we should wait for that as could be a while before that's accepted, released, and included in the GitHub SuperLinter.

tunetheweb:
Discovered the issue with some of the complex structs not parsing (on top of the `DETERMINISTIC` issue mentioned above). It only accepts `STRUCT` fields names beginning with `A-Z` or `a-z` so the following are not valid, hence why we've had to use the `-- noqa: PRS` override on these `STRUCT`s only:

Using ` in field name:

```
CREATE TEMPORARY FUNCTION getAttributeSelectorType(css STRING) -- noqa: PRS
-- SQL Linter cannot handle complex STRUCTs so needs noqa ignore command on previous line
RETURNS STRUCT<`=` BOOLEAN, `*=` BOOLEAN, `^=` BOOLEAN, `$=` BOOLEAN, `~=` BOOLEAN>
```

Nor having underscores in field names:

```
CREATE TEMPORARY FUNCTION getSummary(payload STRING) -- noqa: PRS
-- SQL Linter cannot handle complex STRUCTs so needs noqa ignore command on previous line
RETURNS STRUCT<requestId STRING, startedDateTime INT64, 
...
_cdn_provider STRING, _gzip_save INT64, type STRING, ext STRING, format STRING>
LANGUAGE js AS """
```

Can't find any BigQuery definition to confirm if this is a valid restriction (and it's pure luck that our code works).

Also can't find where this restriction is in SQLFluff if we want to change this to allow it.

Will keep digging later...

tunetheweb:
Good news. All the fixes have been merged into SQLFluff so hopefully will be included in a future release and not require the `-- noqa: PRS` exceptions in future:

- https://github.com/sqlfluff/sqlfluff/pull/1252
- https://github.com/sqlfluff/sqlfluff/pull/1254
- https://github.com/sqlfluff/sqlfluff/pull/1256

It's actually a very cool and generic SQL linter! So think we can improve upon it to enforce more of our preferred coding styles in future. I've raised these issues/PRs to do that:
- https://github.com/sqlfluff/sqlfluff/issues/1250 - PR: https://github.com/sqlfluff/sqlfluff/pull/1257
- https://github.com/sqlfluff/sqlfluff/issues/1260
- https://github.com/sqlfluff/sqlfluff/issues/1261

Will probably also submit a PR to get us listed on their ["Who uses SQLFluff" page](https://docs.sqlfluff.com/en/stable/inthewild.html).

PR #2301, opened 2021-08-10
added rank and security details to requests table

paulcalvano:
Progress on #1293


sql/util/requests.sql
```diff
@@ -1,10 +1,9 @@
-CREATE TEMPORARY FUNCTION getSummary(payload STRING) -- noqa: PRS
--- SQL Linter expects STRUCT field names to beging with a-z or A-Z so needs noqa ignore command on previous line
```
rviscomi, line 2: Isn't the `noqa` still needed for fields like `_cdn_provider`?

sql/util/requests.sql
```diff
@@ -92,37 +82,36 @@ LANGUAGE js AS """
...
   try {
     var $ = JSON.parse(payload);
     var mimeType = $.response.content.mimeType;
     var ext = getExt($.request.url);
     var prettyType = prettyType(mimeType, ext);
     var reqHeaders = ["accept", "accept-charset", "accept-encoding", "accept-language", "connection", "host", "if-modified-since", "if-none-match", "referer", "user-agent", "cookie"];
-    var respHeaders = ["accept-ranges", "age", "cache-control", "connection", "content-encoding", "content-language", "content-length", "content-location", "content-type", "date", "etag", "expires", "keep-alive", "last-modified", "location", "pragma", "server", "transfer-encoding", "vary", "via", "x-powered-by", "set-cookie"];
+    var respHeaders = ["accept-ranges","age","cache-control","connection","content-encoding","content-language","content-length","content-location","content-type","date","etag","expires","keep-alive","last-modified","location","pragma","server","transfer-encoding","vary","via","x-powered-by","set-cookie"];
     var startedDateTime = new Date($.startedDateTime).getTime();
+
```
rviscomi, line 109:
```suggestion
```

sql/util/requests.sql
```diff
@@ -178,21 +167,34 @@ _cdn_provider: $._cdn_provider,
...
 
 SELECT
-  CAST('2020-08-01' AS DATE) AS date,
+  CAST('2021-07-01' AS DATE) AS date,
   _TABLE_SUFFIX AS client,
   page,
-  url,
-  getSummary(payload).*, --  noqa: PRS, L013
+  rank,
+  req.url as url,
```
rviscomi, line 192:
```suggestion
  url,
```

sql/util/requests.sql
```diff
@@ -178,21 +167,34 @@ _cdn_provider: $._cdn_provider,
...
+  getSummary(payload).*,
+  JSON_EXTRACT(payload, "$.request.headers") AS request_headers,
+  JSON_EXTRACT(payload, "$.response.headers") AS response_headers,
   payload
 FROM
-  `httparchive.requests.2020_08_01_*`
+  `httparchive.requests.2021_07_01_*` req
+LEFT JOIN (SELECT DISTINCT url, rank FROM `httparchive.summary_pages.2021_07_01_*`) rank_data
+ON req.page = rank_data.url
```
rviscomi, line 200: I think we can omit `DISTINCT` since the summary_pages dataset only contains unique pages. Also reformatting and aliasing `url` to avoid table aliases.

```suggestion
  `httparchive.requests.2021_07_01_*
LEFT JOIN (
  SELECT
    url AS page,
    rank
  FROM
    `httparchive.summary_pages.2021_07_01_*`)
USING (page)
```

rviscomi (changes requested):

sql/util/requests.sql
paulcalvano: `url` would be ambiguous here, which is why I needed to use `req.url`

sql/util/requests.sql
paulcalvano: That query is collecting the distinct url, rank combination from both desktop and mobile tables.    elimating this would result in many sites having two entries in that joined table (which would artifically increase the size of the results).

tunetheweb: Oh, shouldn't we be adding `client` to this subquery and using the rank for that `client`? Are the ranks always the same for `desktop` and `mobile`?

sql/util/requests.sql
rviscomi: To disambiguate I suggested aliasing `pages.url` to `page`

sql/util/requests.sql
rviscomi: Sorry @paulcalvano is right that my suggestion doesn't work.

tunetheweb: Looks like it is the same rank as this returns zero rows:

```sql
SELECT
    d.url,
    d.rank as desktop_rank,
    m.rank AS mobile_rank
FROM
    `httparchive.summary_pages.2021_07_01_desktop` d,
    `httparchive.summary_pages.2021_07_01_mobile` m
WHERE 
    d.url = m.url AND 
    d.rank != m.rank
```

Not sure we should assume that though, in case it ever changes.

Any harm to adding the client to this?
And then also don't need the `DISTINCT`?

sql/util/requests.sql
tunetheweb: Added this back in https://github.com/HTTPArchive/almanac.httparchive.org/pull/2301/commits/6e551e2c0ad9f44d4c163de9b60bb7e5fcb42799

sql/util/requests.sql
tunetheweb: I added the client in (and also to the JOIN) just in case, and removed the DISTINCT.

tunetheweb:
OK let's merge this. I fixed up the linting errors and a couple of other things.

Any further comments or good to merge now?

tunetheweb (approved):

sql/util/requests.sql
tunetheweb: I changed it to a `USING` join instead to resolve this.

rviscomi (approved):

PR #2307, opened 2021-08-13
Add Search button to Contributors page

tunetheweb:
With the addition of Site Search in #2273 I thought it would be nice to use this to allow users to search for what chapters contributors contributed too, by adding a search to the social media section (note this was updated from when the original PR was opened):

Desktop view

Clicking on this shows all the chapters they'd contributed too:

Tony McCreath search results

I've staged a version here for you to play with: https://20210816t091305-dot-webalmanac.uk.r.appspot.com/en/2020/contributors

@rviscomi it's not to the scale of what you wanted in #1195 but think it's a nice feature along the way.

@HTTPArchive/designers appreciate any design tips or suggestions!

### DIFF
--- a/src/templates/base/contributors.html
+++ b/src/templates/base/contributors.html
@@ -98,7 +98,7 @@
   align-items: center;
   border-radius: 15px;
   box-shadow: 0 0 16px 0 rgba(78, 85, 100, 0.2);
-  padding: 24px 34px;
+  padding: 24px 24px;
   width: 230px;
   margin: 10px;
 }
@@ -141,7 +141,6 @@
   color: #757575;
   font-size: 14px;
   font-size: 0.875rem;
-  flex-grow: 1;
 }
 
 .contributor-team {
@@ -149,27 +148,6 @@
   font-size: 0.875rem;
   white-space: nowrap;
 }
-
-.contributor-search-btn {
-  margin-top: 20px;
-  padding: 5px 20px;
-  display: inline-block;
-  align-items: center;
-  border-radius: 29px;
-  text-decoration: none;
-  background-color: transparent;
-  border: 1px solid #757575;
-  cursor: pointer;
-  font-size: 13.6px;
-  font-size: 0.8em;
-}
-
-.contributor-search-btn:hover,
-.contributor-search-btn:focus {
-  background-color: #f7f779;
-  text-decoration: none;
-}
-
 
 {% for id in config.teams.keys() %}
 .contributors.{{ id }} .contributor.{{ id }} {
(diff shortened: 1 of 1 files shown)


rviscomi:
The button is quite prominent. What would you think about having a designated search icon that we add to the row of social links?

shantsis (commented):
Aren't we adding a profile page eventually? The search action under each user is odd

rviscomi:
Yes that's #1195 that @tunetheweb mentioned as a step in the right direction.

Another idea to provide the search functionality in a more subtle way would be to make the contributor names clickable links to the search results page, although that might be too subtle.

shantsis:
I like that better, or making that the card clickable action. Not sure if "search" for each user is the right word to display to site visitors, its more like showing them their contributions via the search

tunetheweb:
> The button is quite prominent.

Yeah I struggled with it myself. But then was wondering what the real intention of this page is? Authors get social media links in chapters (though admittedly other contributors don't) and figured showing their contributions was therefore quite an important/useful extra piece so I was OK with it being quite prominent. Could tone down the style a bit.

> What would you think about having a designated search icon that we add to the row of social links?

I did look at that, but unfortunately it doesn't quite fit in when someone has all social media icons:

image

Could probably make it fit though with a little tweaking if this is the preferred option. It probably is better.

We could use that icon below, but I thought the text was clearer::

image

> Another idea to provide the search functionality in a more subtle way would be to make the contributor names clickable links to the search results page, although that might be too subtle.

I'm not a fan of that - I think it's too subtle.

> I like that better, or making that the card clickable action.

I think there's quite complicated interactions here to be honest (clickable links within clickable links), and also don't think it's obvious, so wouldn't be a fan of that.

> Not sure if "search" for each user is the right word to display to site visitors, its more like showing them their contributions via the search

I did toy with "Show contributions" but the problem is it's not that. Some contributions (design, development...etc.) are not directly shown on the site so won't appear. Also some searches don't always show the contributions. Including a GitHub search link is also problematic as 1) Not everyone is on GitHub and 2) it's overwhelming. Could change it to "Chapter contribution search" but that's a bit long winded.

tunetheweb:
OK I changed it to use the icon. Looks better I think:

**Desktop:**

Desktop view

**Mobile:**

Mobile view

Staged version here: https://20210816t091305-dot-webalmanac.uk.r.appspot.com/en/2020/contributors

Very small change now so, although it might be made redundant if we get round to #1195 I still think this is a nice extra touch in the meantime.

Let me know your thoughts and will update first comment and then merge if all in agreement.

rviscomi (approved):
Icon LGTM

Issue #2162, opened 2021-04-27
HTTP 2021


rviscomi:
# Part IV Chapter 24: HTTP
![HTTP illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/http/hero_lg.jpg)

If you're interested in contributing to the HTTP chapter of the 2021 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| --- | ------- | --------- | -------- | ------- | -- |
| @dominiclovell  | @dominiclovell @rmarx | @ibnesayeed @tunetheweb | @tunetheweb | - | @rviscomi |

<details>
<summary><b>Expand for more information about each role</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
- **[Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)** are responsible for researching the stats and trends used throughout the Almanac. Analysts work closely with authors and reviewers during the planning phase to give direction on the types of stats that are possible from the dataset, and during the analyzing/writing phases to ensure that the stats are used correctly.
[...]

PR #2309, opened 2021-08-14
HTTP 2021 queries

tunetheweb:
Makes progress on #2162 

**Old queries:**
- [x] - Copy and update 2021 queries:
- [x] - Change old queries to use new header JSON columns as more efficient

**New queries:**

Headers:
- [x] - Total Header size growth
- [x] - Header usage
- [x] - Percentile size of each header
- [x] - Bad headers (to be defined but space and/or no-ASCII in header name)?

Version:
- [x] - HTTP/2 & 3 usage by rank percentile
- [x] - HTTP/3 support by rank percentile

Other:
- [x] - Status Codes including Redirects
- [x] - Favicons - 404s - number and size


tunetheweb:
@dominiclovell @rmarx @ibnesayeed I've run all the queries and placed the results here:
https://docs.google.com/spreadsheets/d/1pCdpndXTXexSIZLmVc-aUbp5PcHhZ83hbEEfmDyfD0U/edit#gid=2077755325

I've added a few visualisations for some of the stats, but can 

I've also created a `web-almanac-http` channel on [HTTP Archive Slack](https://join.slack.com/t/httparchive/shared_invite/zt-45sgwmnb-eDEatOhqssqNAKxxOSLAaA) and can answer any questions there.

sql/2021/http/count_of_preload_http_headers_with_nopush_attribute_set.sql
```diff
@@ -0,0 +1,37 @@
...
+  return headers.filter(h => h.name.toLowerCase() == header.toLowerCase()).map(h => h.value);
+
+} catch (e) {
+  return [];
+}
+ """;
+
+SELECT
+  client,
+  COUNTIF(link_header LIKE '%nopush%') AS num_nopush,
```
rviscomi, line 18: Very minor concern, but how do you feel about these detections being simple substring matches rather than tokenized? I wonder if HA should have better standard support for parsing headers.

sql/2021/http/http2_3_support_by_cdn_per_page_and_request.sql
```diff
@@ -0,0 +1,28 @@
...
+  CDN,
+  SAFE_DIVIDE(COUNTIF(http2_3 AND firstHTML), COUNTIF(firstHTML)) AS http2_3_page_pct,
+  COUNTIF(http2_3) / COUNT(0) AS http2_3_request_pct
+FROM
+  (
+    SELECT
+      client,
+      page,
+      firstHTML,
+      IFNULL(REGEXP_EXTRACT(_cdn_provider, r'^([^,]*).*'), '') AS CDN,
```
rviscomi, line 14: This is only the first CDN in the list. It's ok to ignore the rest?

sql/2021/http/http2_3rd_party_by_types.sql
```diff
@@ -0,0 +1,40 @@
...
+  FROM (
+    SELECT
+      client,
+      page,
+      url,
+      category,
+      protocol AS http_version
+    FROM
+      `httparchive.almanac.requests` r,
+      `httparchive.almanac.third_parties` tp
```
rviscomi, line 23: These tables are cross-joined together. Should they be a plain `JOIN` with `USING`? I'd expect that to use a lot less processing.

sql/2021/http/http2_http3_usage_by_rank.sql
```diff
@@ -0,0 +1,53 @@
...
+      firstHTML AND
+      LOWER(protocol) IN ('http/2', 'http/3', 'quic', 'h3-29', 'h3-q050')
+  ),
+  UNNEST([1000, 10000, 100000, 1000000, 10000000]) AS rank_grouping
+JOIN
+  (
+    SELECT
+      client,
+      rank_grouping,
+      COUNT(0) AS total
```
rviscomi, line 32: I think we can expect the total to equal the `rank_grouping` value, or at least very closely. So maybe this isn't necessary?

sql/2021/http/http2_http3_usage_by_rank.sql
```diff
@@ -0,0 +1,53 @@
+# standardSQL
+# Percentiles of sites that main page over HTTP/2 or HTTP/3 by rank
+SELECT
+  client,
+  rank_grouping,
+  CASE
+    WHEN rank_grouping = 10000000 THEN 'all'
+    ELSE FORMAT("%'d", rank_grouping)
```
rviscomi, line 8: Defer all formatting to the results sheet
```suggestion
    ELSE rank_grouping
```

sql/2021/http/pages_most_requests.sql
```diff
@@ -0,0 +1,18 @@
+# standardSQL
+# Pages loading most number of requests
```
rviscomi, line 2: Nit: This looks like a repeat from previous years, but I think it'd be better as a distribution with 100th percentile for "most requests" but also a sense of the rest of the curve.

sql/2021/http/percentiles_of_resources_loaded_over_HTTP2_or_better_per_site.sql
```diff
@@ -0,0 +1,25 @@
...
+    page,
+    COUNTIF(LOWER(protocol) IN ('http/2', 'http/3', 'quic', 'h3-29', 'h3-q050')) / COUNT(0) AS http2_3_pct
+  FROM
+    `httparchive.almanac.requests`
+  WHERE
+    date = '2021-07-01'
+  GROUP BY
+    client,
+    page),
+  UNNEST([5, 6, 7, 8, 9, 10, 25, 50, 75, 90, 95, 100]) AS percentile
```
rviscomi, line 19: What's the rationale for these unusual percentiles?

sql/2021/http/response_headers_largest.sql
```diff
@@ -0,0 +1,13 @@
+#standardSQL
+SELECT
+  IF(ENDS_WITH(_TABLE_SUFFIX, 'desktop'), 'desktop', 'mobile') AS client,
+  SUBSTRING(url, 1, 400) AS url,
+  respHeadersSize / 1024 AS respHeadersSizeKiB
+FROM
+  `httparchive.summary_requests.*`
+WHERE
+  respHeadersSize IS NOT NULL AND
+  _TABLE_SUFFIX LIKE "2021_07_01%"
```
rviscomi, line 10: Any reason not to do it this way?
```suggestion
  `httparchive.summary_requests.2021_07_01_*`
WHERE
  respHeadersSize IS NOT NULL
```

sql/2021/http/response_headers_type_percentile_size.sql
```diff
@@ -0,0 +1,46 @@
...
+    WHERE
+      date = '2021-07-01'
+  ),
+  UNNEST([10, 25, 50, 75, 90, 100]) AS percentile
+GROUP BY
+  client,
+  header,
+  percentile
+HAVING
+  COUNT(DISTINCT url) > 100000
```
rviscomi, line 42: Is this equivalent?
```suggestion
QUALIFY
  URLs > 100000
```

sql/2021/http/response_headers_type_percentile_size.sql
```diff
@@ -0,0 +1,46 @@
...
+
+  // Filter by header name (which is case insensitive)
+  // If multiple headers it's the same as comma separated
+  return headers.map(h => { h.name, h.value });
+
+} catch (e) {
+  return "";
+}
+ """;
+
```
rviscomi, line 16: AFAICT this function is unused in this query
```suggestion
```

sql/2021/http/response_number_and_size.sql
```diff
@@ -0,0 +1,32 @@
+# standardSQL
+# Number and Size of all responses by percentile
```
rviscomi, line 2: Is this a duplicate of `pages_most_requests.sql`?

rviscomi (approved):
Looks great! Mostly nits.

sql/2021/http/count_of_preload_http_headers_with_nopush_attribute_set.sql
tunetheweb: Well it's a substring, on the value of `link` headers only. So think chances of it being false positive are small. So don't think worth tokenising just for this. Plus `nopush` is non-standard (it's more a conventions followed by some) so wouldn't have anything official to tokenise for. So I think it's fine as is.

sql/2021/http/http2_3_support_by_cdn_per_page_and_request.sql
tunetheweb: Yes my understand is it should be in order, and we only the one the client connects to. I presume it's not in reverse order btw?

sql/2021/http/http2_3rd_party_by_types.sql
tunetheweb: I can use a `USING` as the column names don't match for the domain:

```sql
    WHERE
      r.date = '2021-07-01' AND
      tp.date = '2021-07-01' AND
      NET.HOST(url) = domain)
```

I guess I could use a subquery to normalise the names (but then is that really any more efficient?)

Or I could use `ON`:

```sql
ON r.date = tp.date AND NET.HOST(url) = domain
    WHERE
      r.date = '2021-07-01' AND
      tp.date = '2021-07-01'
      )
```

but I presume that's the same as a cross join anyway?

Either way this query uses 210.5 GB and takes 17 seconds to process so small in terms of BigQuery so don't think a big deal.

sql/2021/http/http2_http3_usage_by_rank.sql
tunetheweb: To be honest I like showing total (especially since discovering the UNNEST issue!) - it's a good validation that the total is what it's expected to be!

sql/2021/http/http2_http3_usage_by_rank.sql
tunetheweb: I'd agree with you, if I didn't already have a `rank_grouping` column (see line 5) so this is the text column of that (mainly so I can call the last bucket `all` so not trying to claim we have 10 millions websites, when we only have ~7 million) so see no harm in text formatting the other rank_groupings and makes it easier to see in BigQuery:

image

sql/2021/http/pages_most_requests.sql
tunetheweb: Actually this is a new one for this year! It's looking for extreme stats for one of those call outs of a memorable stat: "The most number of resources loaded is 17,923 requests!" I gave a few examples (top 100) so we could confirm the sites do use that many and not a temporarily crawling error.

sql/2021/http/percentiles_of_resources_loaded_over_HTTP2_or_better_per_site.sql
tunetheweb: Straight copy from last year. Will encourage to only use the "standard ones" if we do end up using this (at which point I might update the query to avoid this conversation next year!). For now I'd like to leave it the same to allow like for like comparison.

sql/2021/http/response_headers_largest.sql
tunetheweb: None whatsoever! Think I copied from the historical one (`response_headers_size_growth.sql`) and just limited to this one. Your way is better so will update (though think BigQuery is clever enough to effectively do the same so no query difference).

sql/2021/http/response_headers_type_percentile_size.sql
tunetheweb: It is in this case. But `HAVING` is more widely understood so not sure the real gain in going with that other (though does look a tiny bit neater and does show off QUALIFY knowledge)?

sql/2021/http/response_headers_type_percentile_size.sql
tunetheweb: Good spot! Will remove.

sql/2021/http/response_number_and_size.sql
tunetheweb: As per above `pages_most_requests.sql` is the top 100 worst offenders for an interesting call out. We list the sites to make sure they are real things and not crawl errors.

This is the percentile version of that (as you were suggesting I use above!) more for graph.

tunetheweb:
Went through the feedback @rviscomi . Actioned two of them and argued back the rest 😁

Let me know if you're good to merge or wanna fight the good fight on the ones I've pushed back on.

sql/2021/http/http2_3rd_party_by_types.sql
rviscomi: Fair enough!

sql/2021/http/http2_3_support_by_cdn_per_page_and_request.sql
rviscomi: I think that's correct

rviscomi (approved):

PR #2321, opened 2021-08-25
Upgrade SQLFluff and enable extra rules

tunetheweb:
Upgrades to latest version of SQL Fluff and enables the new rules it allows us to turn on thanks to some enhancements in there. Namely:

- Fix and remove all `noqa: PRS` overrides for parsing errors as all parsing errors are now fixed.
- Enable L003 - We can check for consistent spacing now they allow us to configure `UNION` to be spaced at same level as `FROM` as we use a lot.
- Enable L007 - We now force operators (e.g. `AND`) at end of line instead of beginning of previous line (though gotta say I personally prefer beginning of line as easier to comment out!)
- L048 you need space around quoted literals (e.g. not `SELECT 'foo'AS bar`). Previously this caused a problem with triple quotes used at beginning and end of UDFs
- Enforce `COUNT(0)` over `COUNT(1)` or `COUNT(*)` now that rule L047 allows this to be configured.
- Fix unnecessary use of `PARTITION BY 0` in `OVER` clauses as empty `OVER ()` is now supported.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -28,8 +28,8 @@ jobs:
         run: |
           echo "VALIDATE_ALL_CODEBASE=false" >> $GITHUB_ENV
       - name: Lint Code Base
-        uses: github/super-linter@v4.6.1
-        #uses: docker://github/super-linter:v4.6.1
+        uses: github/super-linter@v4.6.2
+        #uses: docker://github/super-linter:v4.6.2
         env:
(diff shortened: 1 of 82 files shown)


tunetheweb:
FYI @rviscomi / @obto as know you didn't like that `OVER (PARTITION BY 0)` workaround so it's now gone! If you update your branch after this is merged you can get rid of it in your 2021 queries too.

Issue #2152, opened 2021-04-27
Capabilities 2021


rviscomi:
# Part II Chapter 15: Capabilities

![Capabilities illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2020/capabilities/hero_lg.jpg)

If you're interested in contributing to the Capabilities chapter of the 2021 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| --- | ------- | --------- | -------- | ------- | -- |
| @christianliebel | @christianliebel | @tomayac @hemanth | @tomayac | @tunetheweb  | @obto |

<details>
<summary><b>Expand for more information about each role</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
- **[Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)** are responsible for researching the stats and trends used throughout the Almanac. Analysts work closely with authors and reviewers during the planning phase to give direction on the types of stats that are possible from the dataset, and during the analyzing/writing phases to ensure that the stats are used correctly.
[...]

PR #2322, opened 2021-08-30
Add Capabilities queries

tomayac:
Adds queries for mobile and desktop for all the APIs we have detection for.

Related to #2152.


tunetheweb:
@tomayac looks like you’ve lots of linting errors. You can fix most of these automatically if you have your Python env set up in your src directory by running:

```
sqlfluff fix sql/2021/capabilities
```

Gimme a shout if you need a hand.

tunetheweb:
Oh and similarly you can run:

```
sqlfluff lint sql/2021/capabilities
```

to make sure you’re all clean before committng if bored of waiting for GitHub Action to complete.

tomayac:
> Oh and similarly you can run:
> 
> ```
> sqlfluff lint sql/2021/capabilities
> ```
> 
> to make sure you’re all clean before committng if bored of waiting for GitHub Action to complete.

All checks green now. I formatted via the BigQuery front-end, but looks like we have other preferences here, which is fine, too. I like the lint rules here better that the front-end's…

tunetheweb:
> All checks green now. I formatted via the BigQuery front-end, but looks like we have other preferences here, which is fine, too. I like the lint rules here better that the front-end's…

Agree with most of them but not a big fan of `AND` at the end of line (makes commenting out more difficult!) and `COUNT(0)` over `COUNT(1)` (seems counter intuitive to me!), but it's the consistency that's more important so I'll live with it!

sql/2021/capabilities/desktop/absolute_orientation_sensor.sql
```diff
@@ -0,0 +1,21 @@
...
+FROM
+  `httparchive.pages.2021_07_01_desktop`,
+  UNNEST(getFuguAPIs(JSON_QUERY(payload, '$."_fugu-apis"'))) AS fuguAPI
+WHERE
+  JSON_QUERY(payload, '$."_fugu-apis"') != "[]" AND
+  fuguAPI = "Absolute Orientation Sensor"
+GROUP BY
+  fuguAPI,
+  url,
+  payload;
```
tunetheweb, line 21: Do we need to separate out desktop and mobile? Could we not just do this like we usually do?:

```suggestion
SELECT
  _TABLE_SUFFIX as client
  fuguAPI,
  url,
  payload
FROM
  `httparchive.pages.2021_07_01_*`,
  UNNEST(getFuguAPIs(JSON_QUERY(payload, '$."_fugu-apis"'))) AS fuguAPI
WHERE
  JSON_QUERY(payload, '$."_fugu-apis"') != "[]" AND
  fuguAPI = "Absolute Orientation Sensor"
GROUP BY
  client,
  fuguAPI,
  url,
  payload;
```

Also do we really want to return each url AND the full payload at that - not even restricting it to the '$."_fugu-apis"' payload? Aren't counts better? Could always give a sample of URLS if you want (have a look at [this example of how to do that](https://github.com/HTTPArchive/bigquery/blob/master/sql/histograms/htmlElementPopularity.sql).

Also it wouldn't run cause you've capitalised `data` on line 2 so need to revert that.

Finally I do wonder if it would be easier to run this as one big query and pivot on the fuguAPI in sheets? Might be easier...

tunetheweb (changes requested):

sql/2021/capabilities/desktop/absolute_orientation_sensor.sql
tunetheweb: Muuuuuch nicer in latest version!

tomayac: Thanks for the pointers :-)

sql/2021/capabilities/fugu.sql
```diff
@@ -0,0 +1,93 @@
...
+    'Gamepad',
+    'WebGPU',
+    'Window Controls Overlay',
+    'Web Share Target',
+    'Web Share Target (Files)',
+    'Shortcuts',
+    'Declarative Link Capturing',
+    'Tabbed Application Mode',
+    'URL Handlers',
+    'Protocol Handlers')
```
tunetheweb, line 84: Any reason to restrict here? Why not just show them all since I presume you're already restricting in the custom metric? Would make it future proof in case you add more in the custom metric.

```suggestion
  JSON_QUERY(payload, '$."_fugu-apis"') != "[]"
```

tomayac:
Latest data dump: https://docs.google.com/spreadsheets/d/1mwJv6l4h13HyK7VSGVN8B55l8S73dz-0aqIfQkNdlMY/edit?usp=sharing.

sql/2021/capabilities/fugu.sql
```diff
@@ -0,0 +1,93 @@
...
+GROUP BY
+  fuguAPI,
+  client,
+  total
+HAVING
+  COUNT(DISTINCT url) >= 10
+ORDER BY
+  pages / total DESC,
```
tunetheweb, line 92: Nit, but should either use the column as it's defined in the `SELECT`, or the alias it was given in the `SELECT`. Here you've a combination of of the two, whereby you're using the `pages` alias but then redefining the `pct` definition. It works fine, but could be cleaner to make it more obvious with the link to the `SELECT` columns.

```suggestion
  pct DESC,
```

tunetheweb (commented):
Looks great now. A couple of minor comments/nits.

Also is this the only query for the Capabilities chapter or will you be adding more? Noticed you had some blink counter queries last year. If adding more it would be good to add a checklist to the initial comment (see the other PRs as examples) so we can track progress for this chapter and see how far through we are in writing the queries.

tomayac:
Thanks for the nits :-) 

This will be the only query then, since the queries based on use counters already exist and are available in an evergreen [report](https://httparchive.org/reports/project-fugu). We can reference specific APIs if we need to ([example](https://httparchive.org/reports/project-fugu#getInstalledRelatedApps)). This year the idea was to focus less on quantitative analysis but focus more on qualitative aspects like how apps use these APIs.

tomayac:
Actually, literally hitting the **Comment** button I recall we wanted to do a fun analysis that would be quantitative: We wanted to determine the most Fugu page on the Internet by ordering pages by the number of Fugu APIs they use. Could you help with this? My SQL foo is a bit rusty (as you have no doubt noticed).

tunetheweb:
> Actually, literally hitting the **Comment** button I recall we wanted to do a fun analysis that would be quantitative: We wanted to determine the most Fugu page on the Internet by ordering pages by the number of Fugu APIs they use. Could you help with this? My SQL foo is a bit rusty (as you have no doubt noticed).

OK then think it's good to merge. Can you copy the results into the official sheet for this chapter: https://docs.google.com/spreadsheets/d/1b4moteB9EiLYkH1Ln9qfi1tnU-E4N2UQ87uayWytDKw/edit?usp=sharing

tunetheweb (approved):

tunetheweb:
> Actually, literally hitting the **Comment** button I recall we wanted to do a fun analysis that would be quantitative: We wanted to determine the most Fugu page on the Internet by ordering pages by the number of Fugu APIs they use. Could you help with this? My SQL foo is a bit rusty (as you have no doubt noticed).

What about something like this:

```sql
CREATE TEMP FUNCTION getFuguAPIs(data STRING)
RETURNS ARRAY<STRING>
LANGUAGE js AS '''
const $ = JSON.parse(data);
return Object.keys($);
''';
CREATE TEMP FUNCTION getFuguAPIs(data STRING)
RETURNS ARRAY<STRING>
LANGUAGE js AS '''
const $ = JSON.parse(data);
return Object.keys($);
''';
SELECT
  _TABLE_SUFFIX AS client,
  url,
  COUNT(DISTINCT fuguAPI) AS fuguAPIs
FROM
  UNNEST(getFuguAPIs(JSON_QUERY(payload, '$."_fugu-apis"'))) AS fuguAPI
WHERE
  JSON_QUERY(payload, '$."_fugu-apis"') != "[]"
GROUP BY
  client,
  url
HAVING
  COUNT(DISTINCT fuguAPI) >= 1
ORDER BY
  fuguAPIs DESC,
  url,
  client
LIMIT 100;
```

First 15 results are:

client | url | fuguAPIs
-- | -- | --
desktop | https://whatwebcando.today/ | 28
mobile | https://whatwebcando.today/ | 28
mobile | https://polisnotis.se/ | 10
desktop | https://system-scanner.net/ | 9
desktop | https://community.emlid.com/ | 8
mobile | https://community.emlid.com/ | 8
desktop | https://excalidraw.com/ | 8
mobile | https://excalidraw.com/ | 8
desktop | https://fstdesk.com/ | 8
mobile | https://fstdesk.com/ | 8
desktop | https://mirea.ninja/ | 8
mobile | https://mirea.ninja/ | 8
desktop | https://permission.site/ | 8
mobile | https://permission.site/ | 8

tomayac:
Amazing, thanks for that! Just committed this query to the repo.

tunetheweb:
> Amazing, thanks for that! Just committed this query to the repo.

Don't see it. Did you push?

tomayac:
Ooops, sorry. I pushed, but didn't see that I had to pull first in order for it to go through.

tomayac:
Results for both queries added to the official results sheet: https://docs.google.com/spreadsheets/d/1b4moteB9EiLYkH1Ln9qfi1tnU-E4N2UQ87uayWytDKw/edit?usp=sharing.

tunetheweb:
Cheers. I added the rest of the 100 since I had that tab open still.

Also added a pivot table. Why is GamePad used sooo much? Does some popular embed (YouTube?) used it?

Anyway think you can tick off chapter item **3 - Validate Results**! 🎉

tomayac:
> Cheers. I added the rest of the 100 since I had that tab open still.

Oh, thanks!

> Also added a pivot table. Why is GamePad used sooo much? Does some popular embed (YouTube?) used it?

Looking at [ChromeStatus](https://chromestatus.com/metrics/feature/timeline/popularity/1964), it looks not as popular (but it's looking at concrete events, not just the API _per se_. 

> Anyway think you can tick off chapter item **3 - Validate Results**! 🎉

Woohoo!

tunetheweb:
> > Also added a pivot table. Why is GamePad used sooo much? Does some popular embed (YouTube?) used it?
> 
> Looking at [ChromeStatus](https://chromestatus.com/metrics/feature/timeline/popularity/1964), it looks not as popular (but it's looking at concrete events, not just the API _per se_.

It may be worth investigating these further. For the PWA queries we specifically excluded YouTube embeds for some queries (see #2272 ) as we felt they were incorrectly clouding the stats for events that were never used. @demianrenzulli has more details.

tomayac:
I found a better [ChromeStatus stats entry](https://chromestatus.com/metrics/feature/timeline/popularity/1916) (this was hidden in plain sight). It looks like the `navigator.getGamepads()` API is used in tracking contexts, for example, by [Yandex](https://mc.yandex.ru/metrika/tag.js#:~:text=navigator.getgamepads()). This being a thing is also suggested by [Mozilla disabling the API in their resist fingerprinting mode](https://bugzilla.mozilla.org/show_bug.cgi?id=1337161).

PR #2344, opened 2021-09-26
Extra PWA queries

tunetheweb:
Added some extra queries at author's request.

Informed author that the first would treat all origins as of equal weight and does not take into account how often these notifications show, but that's in keeping with the rest of our analysis anyway.

Allows graphs like this:

Notification acceptance rates

Mobile notification acceptance rates over time

Desktop notification acceptance rates over time

### DIFF
--- a/sql/2021/pwa/pwa_notification_acceptance_rates_over_time.sql
+++ b/sql/2021/pwa/pwa_notification_acceptance_rates_over_time.sql
@@ -12,7 +12,6 @@ SELECT
 FROM
   `chrome-ux-report.materialized.device_summary`
 WHERE
-  date <= ('2021-07-01') AND
   (
     notification_permission_accept IS NOT NULL OR
     notification_permission_deny IS NOT NULL OR
--- a/sql/2021/pwa/pwa_notification_acceptance_rates_over_time.sql
+++ b/sql/2021/pwa/pwa_notification_acceptance_rates_over_time.sql
@@ -1,0 +1,28 @@
+#standardSQL
+# Total Noitification Acceptence Rates across all origins
+# Note: not weighted by domain popularity nor number of notifications shown
+
+SELECT
+  date,
+  IF(device = 'phone', 'mobile', device) AS client,
+  SUM(notification_permission_accept) / SUM(notification_permission_accept + notification_permission_deny + notification_permission_ignore + notification_permission_dismiss) AS accept,
+  SUM(notification_permission_deny) / SUM(notification_permission_accept + notification_permission_deny + notification_permission_ignore + notification_permission_dismiss) AS deny,
+  SUM(notification_permission_ignore) / SUM(notification_permission_accept + notification_permission_deny + notification_permission_ignore + notification_permission_dismiss) AS _ignore,
+  SUM(notification_permission_dismiss) / SUM(notification_permission_accept + notification_permission_deny + notification_permission_ignore + notification_permission_dismiss) AS dismiss
+FROM
+  `chrome-ux-report.materialized.device_summary`
+WHERE
+  date <= ('2021-07-01') AND
+  (
+    notification_permission_accept IS NOT NULL OR
+    notification_permission_deny IS NOT NULL OR
+    notification_permission_ignore IS NOT NULL OR
+    notification_permission_dismiss IS NOT NULL
+  ) AND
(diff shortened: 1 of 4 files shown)


sql/2021/pwa/pwa_notification_acceptance_rates_over_time.sql
```diff
@@ -0,0 +1,28 @@
...
+  date,
+  IF(device = 'phone', 'mobile', device) AS client,
+  SUM(notification_permission_accept) / SUM(notification_permission_accept + notification_permission_deny + notification_permission_ignore + notification_permission_dismiss) AS accept,
+  SUM(notification_permission_deny) / SUM(notification_permission_accept + notification_permission_deny + notification_permission_ignore + notification_permission_dismiss) AS deny,
+  SUM(notification_permission_ignore) / SUM(notification_permission_accept + notification_permission_deny + notification_permission_ignore + notification_permission_dismiss) AS _ignore,
+  SUM(notification_permission_dismiss) / SUM(notification_permission_accept + notification_permission_deny + notification_permission_ignore + notification_permission_dismiss) AS dismiss
+FROM
+  `chrome-ux-report.materialized.device_summary`
+WHERE
+  date <= ('2021-07-01') AND
```
rviscomi, line 15: The query LGTM but the chart in the [PR comment](https://github.com/HTTPArchive/almanac.httparchive.org/pull/2344#issue-1007314166) shows data up to 2021-08-01:

image

Was that based on a different query or might this not be working properly?

PS: It's unfortunate that there was a really interesting spike in ignore rates in 2021-08-01, omitting it would make us miss out on digging into why that might have happened. I'd still be curious to hear @demianrenzulli's take :)

rviscomi (approved):

sql/2021/pwa/pwa_notification_acceptance_rates_over_time.sql
tunetheweb: Good spot!

The query is correct but my Sheets-foo is poor. And took me a moment to figure out why! I only added that WHERE clause at the end to avoid looking like we added this one late. Then I copied and pasted the values in. Unfortunately I originally had them in ascending order and we were two lines shorter now and the last two lines were left from my first run (without the WHERE clause) and, as these were the in ascending order they were the August ones I'd removed from the query. I then included them in the chart without noticing.

You're right through, interesting change in last month. We should have September stacks available later today so might have look and see if this was a blip then or continues in September.

@demianrenzulli are you aware of any changes in this space that might explain this? As @rviscomi says it's (just) out of scope for the 2021 Web Almanac, but might be interesting to know why or note upcoming changes in this space.

rviscomi: Oh, one other suggestion would be to generate the full distribution of rates for all origins, as shown here:

image

From an [earlier post](https://dev.to/chromiumdev/yestifications-exploring-how-users-engage-with-notification-prompts-in-the-chrome-ux-report-4h7c#:~:text=let's%20zoom%20out%20and%20see%20how%20all%2030%2C000%20websites%20stack%20up.) I wrote about this dataset

demianrenzulli: I can confirm that both the assumptions we made before are correct:

- Desktop vs mobile differences are due to the mobile enrollment UI is much more intrusive  than the desktop one, leading to more users "denying". Also, in Desktop users have more space to click outside of the prompt to "ignore" it.
- The spike in ignore rates in August, is consistent to what we are seeing internally from some recent changes.

With that being said, this query seems to be ok! It would be interesting to take Rick's suggestions into account, in order to improve it, but as a V1, I think it's great!

tunetheweb: > We should have September stats available later today so might have look and see if this was a blip then or continues in September. Will keep this PR open until then as a reminder as no rush to merge.

Forgot this was based on CrUX data, and not HTTP Archive data so won't have September data for another couple of weeks.

rviscomi:
@tunetheweb is this ready for review?

tunetheweb:
@demianrenzulli think we’re done with new queries now? Presume so, so will make this as ready for review.

rviscomi (approved):

PR #2348, opened 2021-09-28
More PWA 2021 query tweaks

tunetheweb:
Going to keep this as draft for now in case there are any more.


tunetheweb:
Actually forgot #2344 was still open so will add this to that.

PR #2351, opened 2021-09-30
SEO 2021 queries

obto:
Queries for #2146

Note: refer to [last year's queries](https://docs.google.com/spreadsheets/d/1ram47FshAjzvbQVJbAQPgxZN7PPOPCKIK67VJZCo92c/edit) for additional comments


obto:
@jroakes @rvth @jessthebp @tunetheweb Here is the new PR that utilizes the `seo-2021` branch so you can all easily contribute.

rviscomi:
@obto would it be manageable to create a checklist of metrics for this PR? Hard to tell what the remaining work is and how close it is to being ready for review.

jroakes:
@rviscomi and @obto I have added all that I think I can.  If there are any other requests, or if @obto wants to do anymore with popularity.rank, that would be cool.  I think I am done for now, unless y'all have any tips for the new files.

tunetheweb:
I'm going to take this out of Draft then and give it a once over.

> ## New Metrics
> * [ ]  Robots values by Allow / Disallow (can't do)
>   This is not possible based on how [this is extracted](https://github.com/HTTPArchive/legacy.httparchive.org/commit/599db3b7222c7784740316eeaec213bff1cc2ac8).  There is no relationship between user-agents and what is allowed or disallowed.  Further, it is probably that some user-agents are allowed and disallowed.
...
> * [ ]  Robots.txt - Noindex usage (can't do)
>   I don't think this is possible.  The custom extraction only targets sitemaps, disallow, and allow, and lighthouse doesn't report `Noindex:` as an error, AFAICT.

It would be great if someone could update then [robots.txt custom metric](https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/robots_txt.js) to capture these so we can consider for next year.

jroakes:
@tunetheweb .  I gave a go at updating the robots_text script for 2022.  I figured it would be better go ahead since I would have loved to have the metrics.  Feedback welcome: https://github.com/HTTPArchive/legacy.httparchive.org/compare/master...jroakes:patch-1

Issue #2160, opened 2021-04-27
Compression 2021


rviscomi:
# Part IV Chapter 22: Compression

![Compression illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/compression/hero_lg.jpg)

If you're interested in contributing to the Compression chapter of the 2021 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| --- | ------- | --------- | -------- | ------- | -- |
| @lvandeve | @lvandeve @jyrkialakuijala | @paulcalvano | @paulcalvano | - | - | @rviscomi |

<details>
<summary><b>Expand for more information about each role</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
- **[Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)** are responsible for researching the stats and trends used throughout the Almanac. Analysts work closely with authors and reviewers during the planning phase to give direction on the types of stats that are possible from the dataset, and during the analyzing/writing phases to ensure that the stats are used correctly.
[...]

PR #2354, opened 2021-10-02
Compression 2021 queries

paulcalvano:
Queries for #2160

### DIFF
--- a/sql/2021/compression/compression_format_trend.sql
+++ b/sql/2021/compression/compression_format_trend.sql
@@ -1,0 +1,20 @@
+#standardSQL
+  # compression_format_trend.sql : What compression formats are being used (gzip, brotli, etc)
+SELECT
+  EXTRACT(YEAR from date) as year,
+  client,
+  IF(resp_content_encoding = "gzip", "Gzip", IF(resp_content_encoding = "br", "Brotli", IF(resp_content_encoding = "", "no text compression", "other"))) AS compression_type,
+  COUNT(0) AS num_requests,
+FROM
+  (
+    SELECT date, client, resp_content_encoding FROM `httparchive.almanac.requests_2020` WHERE DATE IN ('2020-08-01', '2019-07-01')
+    UNION ALL 
+    SELECT date, client, resp_content_encoding FROM `httparchive.almanac.requests` 
+  )
+
+GROUP BY
+  year,
+  client,
+  compression_type
+ORDER BY
+  compression_type DESC
--- a/sql/2021/compression/compression_format_trend.sql
+++ b/sql/2021/compression/compression_format_trend.sql
@@ -1,15 +1,15 @@
 #standardSQL
   # compression_format_trend.sql : What compression formats are being used (gzip, brotli, etc)
 SELECT
-  EXTRACT(YEAR from date) as year,
+  EXTRACT(YEAR FROM date) AS year,
   client,
   IF(resp_content_encoding = "gzip", "Gzip", IF(resp_content_encoding = "br", "Brotli", IF(resp_content_encoding = "", "no text compression", "other"))) AS compression_type,
-  COUNT(0) AS num_requests,
+  COUNT(0) AS num_requests
 FROM
   (
     SELECT date, client, resp_content_encoding FROM `httparchive.almanac.requests_2020` WHERE DATE IN ('2020-08-01', '2019-07-01')
-    UNION ALL 
-    SELECT date, client, resp_content_encoding FROM `httparchive.almanac.requests` 
+    UNION ALL
+    SELECT date, client, resp_content_encoding FROM `httparchive.almanac.requests`
   )
 
 GROUP BY
--- a/sql/2021/compression/compression_format_trend.sql
+++ b/sql/2021/compression/compression_format_trend.sql
@@ -3,7 +3,12 @@
 SELECT
   EXTRACT(YEAR FROM date) AS year,
   client,
-  IF(resp_content_encoding = "gzip", "Gzip", IF(resp_content_encoding = "br", "Brotli", IF(resp_content_encoding = "", "no text compression", "other"))) AS compression_type,
+  CASE
+    WHEN resp_content_encoding = 'gzip' THEN 'Gzip'
+    WHEN resp_content_encoding = 'br' THEN 'Brotli'
+    WHEN resp_content_encoding = '' THEN 'no text compression'
+    ELSE 'other'
+  END AS compression_type,
   COUNT(0) AS num_requests
 FROM
   (
--- a/sql/2021/compression/compression_format_trend.sql
+++ b/sql/2021/compression/compression_format_trend.sql
@@ -12,9 +12,7 @@ SELECT
   COUNT(0) AS num_requests
 FROM
   (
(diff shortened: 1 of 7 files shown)


rviscomi:
Thanks Paul. Could you take a look at the linter errors?

sql/2021/compression/compression_format_trend.sql
```diff
@@ -0,0 +1,20 @@
+#standardSQL
+  # compression_format_trend.sql : What compression formats are being used (gzip, brotli, etc)
+SELECT
+  EXTRACT(YEAR FROM date) AS year,
+  client,
+  IF(resp_content_encoding = "gzip", "Gzip", IF(resp_content_encoding = "br", "Brotli", IF(resp_content_encoding = "", "no text compression", "other"))) AS compression_type,
+  COUNT(0) AS num_requests
+FROM
+  (
+    SELECT date, client, resp_content_encoding FROM `httparchive.almanac.requests_2020` WHERE DATE IN ('2020-08-01', '2019-07-01')
```
rviscomi, line 10: Oh I just learned that `almanac.requests` now only includes 2021 data. I had intended to grow the same `almanac` tables with each year, but I think I understand why you did it this way. Does it have to do with the new fields and incompatible schema? If so I think it should work to add the 2019/2020 data back into `requests` with NULL values for those new fields.

The way it's organized now means that all of the previous years' queries that depend on this table will be broken, so it'd be good to fix it. I think the `summary_response_bodies` table has the same issue. Could you combine the respective tables so that 2019-2021 are all in one and update this query?

sql/2021/compression/compression_format_trend.sql
```diff
@@ -0,0 +1,20 @@
+#standardSQL
+  # compression_format_trend.sql : What compression formats are being used (gzip, brotli, etc)
+SELECT
+  EXTRACT(YEAR FROM date) AS year,
+  client,
+  IF(resp_content_encoding = "gzip", "Gzip", IF(resp_content_encoding = "br", "Brotli", IF(resp_content_encoding = "", "no text compression", "other"))) AS compression_type,
```
rviscomi, line 6: Nit: this if/else chain lends itself well to CASE/WHEN/ELSE statements instead and is a bit easier to read
```suggestion
  CASE
    WHEN resp_content_encoding = 'gzip' THEN 'Gzip'
    WHEN resp_content_encoding = 'br' THEN 'Brotli'
    WHEN resp_content_encoding = '' THEN 'no text compression'
    ELSE 'other'
  END AS compression_type,
```

sql/2021/compression/compression_formats.sql
```diff
@@ -0,0 +1,16 @@
+#standardSQL
+  # compression_formats.sql : What compression formats are being used (gzip, brotli, etc)
+SELECT
+  client,
+  resp_content_encoding,
+  COUNT(0) AS num_requests,
+  COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client) AS pct
```
rviscomi, line 7: Nit: also include the total for completeness

```suggestion
  SUM(COUNT(0)) OVER (PARTITION BY client) AS total,
  COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client) AS pct
```

sql/2021/compression/compression_formats_text_responses.sql
```diff
@@ -0,0 +1,24 @@
+#standardSQL
+# compression_formats_text_responses.sql : What compression formats are being used (gzip, brotli, etc) on text responses
+SELECT
+  client,
+  IF(resp_content_encoding = "gzip", "Gzip", IF(resp_content_encoding = "br", "Brotli", IF(resp_content_encoding = "", "no text compression", "other"))) AS compression_type,
```
rviscomi, line 5: Ditto on CASE statement

sql/2021/compression/compression_formats_text_responses.sql
```diff
@@ -0,0 +1,24 @@
...
+  COUNT(0) AS num_requests,
+  COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client) AS pct
```
rviscomi, line 7: Ditto on `total`

sql/2021/compression/lighthouse_compression_byte_savings.sql
```diff
@@ -0,0 +1,17 @@
+#standardSQL
+# lighthouse_compression_byte_savings.sql :  Text Based Compression Byte Savings
+SELECT
+  _TABLE_SUFFIX AS client,
+  ROUND(CAST(JSON_EXTRACT_SCALAR(report, "$.audits.uses-text-compression.details.overallSavingsBytes") AS INT64) / 1024 / 1024) AS mbyte_savings,
+  COUNT(0) AS num_pages,
+  ROUND(COUNT(0) * 100 / SUM(COUNT(0)) OVER (PARTITION BY _TABLE_SUFFIX), 2) AS pct_pages
```
rviscomi, line 7: Remove percent formatting, do in Sheets instead
```suggestion
  COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY _TABLE_SUFFIX) AS pct_pages
```

rviscomi, line 7: Ditto on `total`

sql/2021/compression/lighthouse_compression_byte_savings.sql
```diff
@@ -0,0 +1,17 @@
...
+FROM
+  `httparchive.lighthouse.2021_07_01_*`
+WHERE
+  JSON_EXTRACT_SCALAR(report, "$.audits.uses-text-compression.score") != "1"
+GROUP BY
+  client,
+  mbyte_savings
+ORDER BY
+  client,
+  mbyte_savings ASC
```
rviscomi, line 17: `ASC` is implied
```suggestion
  mbyte_savings
```

sql/2021/compression/lighthouse_compression_scores.sql
```diff
@@ -0,0 +1,17 @@
...
+  SUM(COUNT(0)) OVER (PARTITION BY _TABLE_SUFFIX) AS total,
+  COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY _TABLE_SUFFIX) AS pct_pages
+FROM
+  `httparchive.lighthouse.2021_07_01_*`
+GROUP BY
+  client,
+  text_compression_score
+ORDER BY
+  client,
+  text_compression_score ASC
```
rviscomi, line 16: Implied
```suggestion
  text_compression_score
```

rviscomi (changes requested):

sql/2021/compression/compression_format_trend.sql
paulcalvano: We added a few new columns to the requests dataset for this years almanac, so I added them to a new table.  It probably makes sense to rerun the query for generating the requests table with the 2019 and 2020 datasets so we have them all in one place.

paulcalvano: I updated the requests table to include the 2019-2020 data and updated this query to remove the JOIN in the inner query.

paulcalvano:
Sorry for the delay.  I updated the almanac.requests table to include the 2019 and 2020 data, and fixed the compression_trends query so that it runs against almanac.requests.

rviscomi (approved):

Issue #2154, opened 2021-04-27
CMS 2021


rviscomi:
# Part III Chapter 17: CMS

![CMS illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/cms/hero_lg.jpg)

If you're interested in contributing to the CMS chapter of the 2021 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| --- | ------- | --------- | -------- | ------- | -- |
| @alonkochba | @alonkochba @logicalphase | @saterchris @jdevalk | @TosinArasi | - | @logicalphase |

<details>
<summary><b>Expand for more information about each role</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
- **[Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)** are responsible for researching the stats and trends used throughout the Almanac. Analysts work closely with authors and reviewers during the planning phase to give direction on the types of stats that are possible from the dataset, and during the analyzing/writing phases to ensure that the stats are used correctly.
[...]

PR #2376, opened 2021-10-14
CMS 2021 queries (part 2)

rviscomi:
Progress on #2154 

- [x] CMS Adoption Rank Magnitude
- [x] Wordpress Plugin Resources - Page Builder
- [x] Total Page Weight (KB) per CMS
- [x] Median KB per resource Type
- [x] Popularity of image formats


logicalphase:
Thanks @alonkochba! I'm looking 1 over. Will report back on that by this evening.

2. @rviscomi use August sets? As for definition, use of page builders will get messy----just WordPress alone there are dozens of page builders (thinking Envato).

PR #2379, opened 2021-10-16
Sql linting fixes

tunetheweb:
Looks like some of the queries were on branches with the original rules, before we added other rules like indentation checks., so they fail the current sqlfluff tests. So I ran a quick `sqlfluff fix` on them.

Also added a change to run the super-linter on merge so at least we'll get notification on this after merging.


tunetheweb:
> Those CMS ones will be obsolete in #2376

OK want me to revert?

PR #2381, opened 2021-10-17
Updated A11Y metrics

foxdavidj:
- [x] Root body font-size
   - Note: This may be highly misleading. Since sites can easily specify explicit font sizes on `<p>` and `<div>` elements
   - Not going to IMPL
- [x] Root body line-height
   - Note: This may be highly misleading. Since sites can easily specify explicit font sizes on `<p>` and `<div>` elements
   - Not going to IMPL
- [x] Root body font-size units
   - Note: This may be highly misleading. Since sites can easily specify explicit font sizes on `<p>` and `<div>` elements
   - Not going to IMPL
- [x] Line-height and how they are declared (unit or unitless) on text elements. WCAG recommends 1.5 for paragraph elements, default is 1.2 so wondering how many people modify for the good or modify for the bad
   - Marking not feasible. What you really care about here is not what the root body size is, but what % of text on the page is above or below the standards and there's not a good way to be able to pull that off. We can see what the most popular way of specifying it is however from [this CSS chapter metric](https://docs.google.com/spreadsheets/d/12vQIA0xsC5Jr3J9Sh03AcAvgFjMAmP1xSS6Tjai9LF0/edit#gid=2127104776)
- [x] Sites intentionally hiding the native focus styles on buttons etc
   - `:focus, textarea:focus, input:focus, button:focus { outline: none; }`
   - https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2021/css/focus_outline_0.sql
- [x] Usage of https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast
- [x] `use-landmarks` lighthouse aXe metric not reporting data
- [x] Most commonly used elements
- [x] <title> length for SPAs vs non-SPA's
   - Not feasible.
- [x] Breakdown of tabindex values

### DIFF
--- a/sql/2021/accessibility/button_name_sources.sql
+++ b/sql/2021/accessibility/button_name_sources.sql
@@ -15,7 +15,8 @@ RETURNS ARRAY<STRING> LANGUAGE js AS '''
 
       if (tree_node.accessible_name.length === 0) {
         // No A11Y name given
-        accessible_name_sources("No accessible name");
+        accessible_name_sources.push("No accessible name");
+        continue;
       }
 
       if (tree_node.accessible_name_sources.length <= 0) {
--- a/sql/2021/accessibility/form_input_name_sources.sql
+++ b/sql/2021/accessibility/form_input_name_sources.sql
@@ -16,7 +16,8 @@ RETURNS ARRAY<STRING> LANGUAGE js AS '''
 
       if (tree_node.accessible_name.length === 0) {
         // No A11Y name given
-        accessible_name_sources("No accessible name");
+        accessible_name_sources.push("No accessible name");
+        continue;
       }
 
       if (tree_node.accessible_name_sources.length <= 0) {


obto:
`use-landmarks` is getting reported "correctly". The metric itself has a value of "null" for every single site. We'll have to create our own metric or not report it at all

obto:
Instead of looking for focus usage on specific elements, its going to be much cleaner to instead talk about the usage of `:focus` selectors with a `outline: none` style set inside. We can keep consistent and cross-link with the CSS chapter as well this way.

obto:
Usage of https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast:

This can be found in the CSS chapter data here: https://docs.google.com/spreadsheets/d/12vQIA0xsC5Jr3J9Sh03AcAvgFjMAmP1xSS6Tjai9LF0/edit#gid=1192142763

obto:
@rviscomi Can you give me some guidance on how you would determine the first 3 metrics related to the root body? Is there a similar metric you can point me to?

rviscomi:
@obto is this ready for review? Also have you heard back about whether the new parsed css queries are needed?

Issue #2414, opened 2021-10-30
Create HTTP Archive Discuss topic.


tunetheweb:
Even though we might be adding additional ways to comment on chapters in #2192 we should still have https://discuss.httparchive.org topics for each chapter as well I think.

Rick do you want to do the honours and create these as we're going to need these soon as people start converting to markdown.

rviscomi:
I was thinking that we would use web mentions instead of comments, rather than in addition to them. Engagement hasn't been high enough to justify keeping the comments IMO.

PR #2422, opened 2021-11-01
Hide discussion icons if no discuss metadata

tunetheweb:
As per #2414 we'll not be having the HTTP Archive Discussion topics this year so we should hide the icon if not set in Metadata.

### DIFF
--- a/src/templates/base/base_chapter.html
+++ b/src/templates/base/base_chapter.html
@@ -52,9 +52,11 @@
 {% endblock %}
 
 {% block scripts %}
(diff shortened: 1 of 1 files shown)


rviscomi (approved):
Are there any chapters with the `discuss` field set to something other than a topic ID?

tunetheweb:
> Are there any chapters with the `discuss` field set to something other than a topic ID?

Only the 2021 Capabilities and Third Party chapters but will fix those before merging.

PR #2429, opened 2021-11-03
Change icon ids to allow chapters to use them

tunetheweb:
The 2021 Third Party chapter has a heading called "Twitter" but this creates an `id` clash with the Twitter icon. So let's change that to `twitter-logo` and similar for other icons.


rviscomi (commented):
Why do the IDs exist? I assume for styling, so would there also need to be corresponding CSS changes?

tunetheweb:
Actually no, not for styling.

They are embedded SVGs and then referenced later in the HTML within SVGs using the ids (`xlink:href="#twitter-logo"`) to actually use them in the right place (e.g. on Contributors page, where we use them a lot, but also on header and footer of every other page).

Quite a cool system implemented by @catalinred meaning we don't need extra image calls.

PR #2451, opened 2021-11-08
Resource Hints queries - added imgLazy

kevinfarrugia:
Progress on #2158

This is copy-pasted from https://httparchive.org/reports/state-of-images#imgLazy but needed to reference the SQL script in the chapter.

### DIFF
--- a/sql/2021/resource-hints/imgLazy.sql
+++ b/sql/2021/resource-hints/imgLazy.sql
@@ -1,0 +1,17 @@
+#standardSQL
+SELECT
+  SUBSTR(_TABLE_SUFFIX, 0, 10) AS date,
+  UNIX_DATE(CAST(REPLACE(SUBSTR(_TABLE_SUFFIX, 0, 10), '_', '-') AS DATE)) * 1000 * 60 * 60 * 24 AS timestamp,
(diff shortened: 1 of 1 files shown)


sql/2021/resource-hints/imgLazy.sql
```diff
@@ -0,0 +1,17 @@
+#standardSQL
+SELECT
+  SUBSTR(_TABLE_SUFFIX, 0, 10) AS date,
+  UNIX_DATE(CAST(REPLACE(SUBSTR(_TABLE_SUFFIX, 0, 10), '_', '-') AS DATE)) * 1000 * 60 * 60 * 24 AS timestamp,
+  IF(ENDS_WITH(_TABLE_SUFFIX, 'desktop'), 'desktop', 'mobile') AS client,
+  ROUND(COUNT(DISTINCT IF(LOWER(attr) = '"lazy"', url, NULL)) * 100 / COUNT(DISTINCT url), 2) AS percent
```
rviscomi, line 6:
```suggestion
  COUNT(DISTINCT IF(LOWER(attr) = '"lazy"', url, NULL)) / COUNT(DISTINCT url) AS percent
```

PR #2458, opened 2021-11-09
Allow markdown to be used in more places

tunetheweb:
Some authors expect markdown to work in more places than it does currently including:
- Table captions
- Table `<td>` and `<th>`'s
- Our asides (`<p class="note">`).

This last one can be worked around with `<p class="note" data-markdown="1">` but not many people know that, and for the others this doesn't seem to work.

This PR adds some additional functionality to allow markdown to be used in these elements.


tunetheweb:
Note this fixes a number of broken things we've not spotted in the past:

2019 and 2020 Ecommerce chapter:

image

2019 Security chapter:

image

image

(plus loads of other examples)


All the changes look good. Also shows the need for this change!




2019 Mobile Web Chapter:

image

PR #2462, opened 2021-11-10
Reorder 2021 chapters

rviscomi:


tunetheweb (approved):
I think we should remove the `src/contents/en/2021/fonts.md placeholder too otherwise you get an error in the generation:

```
 Generating chapter: en, 2021, fonts
TypeError: Cannot read property 'todo' of undefined
    at generate_chapters (/Users/barry/almanac.httparchive.org/src/tools/generate/generate_chapters.js:95:42)
    at async /Users/barry/almanac.httparchive.org/src/tools/generate/index.js:6:5
  Failed to generate chapter, moving onto the next one. 
```

Will add to this branch.

Happy to leave the SQL there as is for next year.

PR #2464, opened 2021-11-10
Improve robot_txt_sensitive_disallow.sql filter and fix version-evolution-top-technologies.sql

GJFR:
The filtering of potentially sensitive endpoints in `robots.txt` is made much more strict. I'm aware that some truly sensitive endpoints will be missed (false negatives), but I think it would be better to consider this as a lower bound than to dilute our statistics with too many false positives. The query could be upgraded (in the long run) by adding edge cases that are false negatives in its current state.

I based my regex on the one described [here](https://github.com/HTTPArchive/legacy.httparchive.org/pull/230) and added the `secure` and `oauth` keywords (thanks for the suggestions @jroakes, please let me know if I changed or oversaw something that you really wanted in there for some reason).

@SaptakS @nrllh @tomvangoethem If there's any concerns / suggestions, please let me know.


rviscomi:
@GJFR is this PR ready for review? If not what other changes are needed?

rviscomi:
Closing this PR until it's clear what the next steps are to keep it moving

PR #2465, opened 2021-11-10
Fix Safari print view for chapters

tunetheweb:
Noticed this is how the chapters print in Safari:

image

Fixing this.


rviscomi:
Good catch!

PR #2478, opened 2021-11-13
Fix image generator to work on Windows

tunetheweb:
@dwsmart if you get a chance to test this on windows would be great!


dwsmart:
The regex on line 32 needs changing from:

```
: 'content\\\\' +  '(' + chapter_match.replace(/\\/g,')\\\\(') + ').md';
```
to
```
: 'content\\\\' +  '(' + chapter_match.replace(/\//g,')\\\\(') + ').md';
```

and it works a treat then when running `npm run figure-images en/2021/mobile-web`

tunetheweb:
Done. Thanks!

dwsmart (approved):
Excellent!

PR #2482, opened 2021-11-14
Fix search button on iOS

tunetheweb:
Search icon is white on iOS:

image

Weirdly it's fine on Safari on desktop 🤷‍♂️

### DIFF
--- a/src/static/css/almanac.css
+++ b/src/static/css/almanac.css
@@ -581,6 +581,7 @@ header .cta {
 
 .search-button svg {
   fill: #000;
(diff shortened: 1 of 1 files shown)


src/static/css/almanac.css
```diff
@@ -581,6 +581,7 @@ header .cta {
 
 .search-button svg {
   fill: #000;
+  color: #000
```
tunetheweb, line 584:
```suggestion
  color: #000;
```

PR #2486, opened 2021-11-14
Update Media 2021 contributors

rviscomi:
Progress on #2144


src/content/en/2021/media.md
```diff
@@ -2,7 +2,7 @@
 #See https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide#metadata-to-add-at-the-top-of-your-chapters
 title: Media
 description: TODO
-authors: []
+authors: [eeeps]
 reviewers: []
 analysts: []
```
rviscomi, line 7: @eeeps can you suggest edits to the Authors, Reviewers, and Analysts properties so that they contain everyone who contributed to the chapter, roughly ordered by those with the most significant contributions first? This is what readers will see at the top of the chapter so we want to be sure to give appropriate credit to everyone who helped.

rviscomi: Added yourself and @dougsillars as authors and analysts. Deferring on the reviewers and other coauthors.

tunetheweb (approved):
Approved. Suspect this will fail again since we now no longer generate chapters when TODO is set to false, so Lighthouse tests cannot run.

Ignore any Lighthouse error and merge.

Issue #2168, opened 2021-04-27
WebAssembly 2021


rviscomi:
# Part I Chapter 6: WebAssembly

If you're interested in contributing to the WebAssembly chapter of the 2021 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| ---- | ------- | --------- | -------- | ------- | ----------- |
| @RReverser | @RReverser | @jsoverson @carlopi | @RReverser | - | @rviscomi |

<details>
<summary><b>Expand for more information about each role</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
- **[Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)** are responsible for researching the stats and trends used throughout the Almanac. Analysts work closely with authors and reviewers during the planning phase to give direction on the types of stats that are possible from the dataset, and during the analyzing/writing phases to ensure that the stats are used correctly.
[...]

PR #2504, opened 2021-11-15
WebAssembly SQL updates

RReverser:
Progress on #2168


rviscomi (changes requested):
Why do some queries group by client but not others? We need to be very careful to avoid combining desktop and mobile results.

RReverser:
> Why do some queries group by client but not others? We need to be very careful to avoid combining desktop and mobile results.

Because in some cases - like section sizes or instruction kinds - the split was more of a distracting piece of information than serving a useful purpose. Avoiding grouping seems like the most natural way to aggregate those together.

As far as I understand it, the only risk is that some resources will be counted twice because they're popular enough to get into both datasets, while others (more rare) will be only counted once?

rviscomi:
When gathering the data, queries should always segment desktop and mobile. That doesn't necessarily mean that you need to show both in your charts or talk about both in the narrative for every stat, but this way you can look at the results before deciding how to present them.

Most websites are in both desktop and mobile datasets. See the [methodology](https://github.com/HTTPArchive/almanac.httparchive.org/blob/57f70bd6db3fc6056ccd0f33e1bc03ccc1d51aec/src/templates/en/2021/methodology.html#L120) for more info about the corpus.

RReverser:
> but this way you can look at the results before deciding how to present them

Sure, that's what I did and why I decided to aggregate some of those stats in this update but not others.

I could commit queries with `GROUP BY` kept in, but then the graph won't represent the numbers returned by the query. It seemed worse, or are you saying that's okay?

rviscomi:
What I'm saying is that it's invalid to _combine_ desktop and mobile results together. Either your queries to should have `WHERE client = 'mobile'` or `GROUP BY client` but preferably the latter because next year's author will want to rerun the 2021 queries and omitting data for a client can potentially hide meaningful results. It's totally fine to query both clients and only discuss the results for one, granted it's clear to readers which is being discussed.

RReverser:
Okay, I see. However what I meant by

> Because in some cases - like section sizes or instruction kinds - the split was more of a distracting piece of information than serving a useful purpose.

is that for those stats, where we collect percentages of total binary size or look at "top offenders" in terms of unstripped debug data, it really doesn't matter which client that data comes from.

For example, where we show percentage of each Wasm section type as fraction of total binary size, any website that appears in both client datasets - which, as you confirmed, is most of the websites - will contribute same size on both desktop and mobile. The resulting percentage will be still roughly the same, because `x+x/(x+x+y+y) = x/(x+y)`, that is, datasets contribute almost equally, and the difference between those two equations is just a small fluctuation, and not particularly interesting one in context of this stat.

rviscomi:
Yeah in some cases the magnitude of the difference will be small, but for correctness it's something that needs to be fixed.

RReverser:
Ok but I really don't want to distract the reader with desktop/mobile split for those few stats, and showing only one side (e.g. `desktop`) seems a bit worse than showing the averaged data using both sources as it ignores part of the data... I'm not sure what other choices are there.

RReverser:
(chatted privately and agreed I'll update this to GROUP BY and use desktop data)

sql/util/wasm_initial_set.sql
```diff
@@ -0,0 +1,7 @@
+SELECT DISTINCT
+  url
```
rviscomi, line 2:
```suggestion
  _TABLE_SUFFIX AS client,
  url
```

rviscomi, line 2:
```suggestion
  _TABLE_SUFFIX AS client,
  url
```

rviscomi (changes requested):

sql/util/wasm_initial_set.sql
RReverser: This one is def not supposed to be grouped because it's an input list for downloader which is meant to contain only distinct URLs.

PR #2507, opened 2021-11-15
Privacy 2021 mini edits

VictorLeP:


tunetheweb (approved):

tunetheweb:
Thanks @VictorLeP - been a long day!

VictorLeP:
Something weird happened to _all_ the captions - all the last characters seem to have dropped off :p Going over them now, afterwards it should be OK to merge.

tunetheweb:
> Something weird happened to _all_ the captions - all the last characters seem to have dropped off :p Going over them now, afterwards it should be OK to merge.

That'll be my fault. Noticed some of them were missing full stops so did a regex replace to add it. Obviously got that wrong :-(

VictorLeP:
No problem, it's the least we can do after all the work you've done for this chapter as well!

Captions have been fixed, I think we're good to merge.

VictorLeP:
How much of an issue is it that we're linking to an unfinished chapter, in our case the Security chapter? It throws a page with not much more than a big [TODO](https://20211112t221945-dot-webalmanac.uk.r.appspot.com/en/2021/security)

tunetheweb:
> How much of an issue is it that we're linking to an unfinished chapter, in our case the Security chapter? It throws up a page with not much more than a big [TODO](https://20211112t221945-dot-webalmanac.uk.r.appspot.com/en/2021/security) (which comes from how the Markdown files are [initialized](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/src/content/en/2021/security.md) - something like "Coming soon!" would be nicer, but the TODO is effectively hardcoded for now)

They should redirect back to the home page for now. Will check it out before the first launch.

Thanks for fixing all of that!

PR #2511, opened 2021-11-16
Add note about color contrast visualization issues to Accessibility Statement

tunetheweb:
What do people think of this wording to add to our Accessibility Statement given the color contrast discussions we've had?

> We recognize that some of the color choices for our visualizations do not meet WCAG color contrast requirements. We make a conscious effort to use the more contrasting colors and labels to reduce the impact of this. However, we also believe that with the detailed descriptions provided for each visualization, plus access to the underlying data itself, the takeaways from all the visualizations are still accessible. We aim to improve the accessibility of our visualization color schemes in in future years.


alextait1:
> What do people think of this wording to add to our Accessibility Statement given the color contrast discussions we've had?
> 
> > We recognize that some of the color choices for our visualizations do not meet WCAG color contrast requirements. We make a conscious effort to use the more contrasting colors and labels to reduce the impact of this. However, we also believe that with the detailed descriptions provided for each visualization, plus access to the underlying data itself, the takeaways from all the visualizations are still accessible. We aim to improve the accessibility of our visualization color schemes in in future years.

I don't think we can say they are "accessible" with WCAG failures (people with low vision or other vision conditions may not be able to see these clearly) but we could say "the takeaways from all the visualizations are accessible to many. We aim...."

tunetheweb:
Thanks for looking @alextait1 - updated the PR.

rviscomi (approved):

src/templates/en/accessibility_statement.html
```diff
@@ -66,6 +66,10 @@ <h2 id="conformance-status"><a href="#conformance-status" class="anchor-link">Co
...
         </p>
 
+        <p>
+          We recognize that some of the color choices for our visualizations do not meet WCAG color contrast requirements. We make a conscious effort to use the more contrasting colors and labels to reduce the impact of this. However, we also believe that with the detailed descriptions provided for each visualization, plus access to the underlying data itself, the takeaways from all the visualizations are still accessible to many. We aim to improve the accessibility of our visualization color schemes in in future years.
```
ericwbailey, line 70: >However, we also believe that with the detailed descriptions provided for each visualization, plus access to the underlying data itself, the takeaways from all the visualizations are still accessible to many.

This feels a little defensive to me. Maybe something like, "We hope the detailed descriptions, as well as access to the underlying data itself can help with this issue until we can correct this issue."? I worry that it could be interpreted that we feel delegating the work of parsing the description/data is equal to viewing a graph, which I don't think is true.

tunetheweb: Thanks @ericwbailey . I went with the shorter: "We hope the detailed descriptions, as well as access to the underlying data itself can help with this issue." as felt "until we can correct this issue" repeated the sentiment of the next line and suggested the descriptions and access might go once this is resolved, when we plan to keep them.

The full addition now reads as below:

> We recognize that some of the color choices for our visualizations do not meet WCAG color contrast requirements. We make a conscious effort to use the more contrasting colors and labels to reduce the impact of this. We hope the detailed descriptions, as well as access to the underlying data itself, can help with this issue until we can correct this. We aim to improve the accessibility of our visualization color schemes in in future years.

tunetheweb:
Anyone else any comments before I merge this?

siakaramalegos:
Repeated word at end. Rewrite is better thanks.

ericwbailey (approved):

siakaramalegos:
@tunetheweb sorry did you see my note? The word "in" is repeated twice at the end.

tunetheweb:
Saw your note, but thought that was referring to my point that "issue" was repeated in Eric's version. Will fix.

siakaramalegos:
Yeah sorry I was in the car and running late and wasn't clear enough :)

tunetheweb:
Fixing in #2543

PR #2526, opened 2021-11-16
Fix link and logicalphase's username

tunetheweb:
The block quote obscure's links so link doesn't work as reported by @shantsis in Slack:

image
  image


rviscomi:
Great catch @shantsis! And thanks @tunetheweb for the fix.

PR #2531, opened 2021-11-16
Compression 2021 contributors

rviscomi:
#2160 cc @lvandeve


src/config/2021.json
```diff
@@ -963,6 +963,13 @@
       "github": "philbarker",
       "twitter": "philbarker"
     },
+    "fischbacher": {
+      "name": "Philip Fischbacher",
```
lvandeve, line 967: It's Thomas actually

Issue #2142, opened 2021-04-27
Markup 2021


rviscomi:
# Part I Chapter 3: Markup

![Markup illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/markup/hero_lg.jpg)

If you're interested in contributing to the Markup chapter of the 2021 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| --- | ------- | --------- | -------- | ------- | -- |
| @AlexLakatos | @AlexLakatos | @j9t @ibnesayeed @meyerweb @brycehowitson | @kevinfarrugia | @shantsis | @rviscomi |

<details>
<summary><b>Expand for more information about each role</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
- **[Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)** are responsible for researching the stats and trends used throughout the Almanac. Analysts work closely with authors and reviewers during the planning phase to give direction on the types of stats that are possible from the dataset, and during the analyzing/writing phases to ensure that the stats are used correctly.
[...]

PR #2538, opened 2021-11-17
Change Markup queries to order by pct first

tunetheweb:
Makes progress on #2142 

Change to order by percentage descending so mobile is not cut off

Also added a total column.
And a couple of "(not sets)" to avoid having to manipulate the data too much in Sheets

All queries have been rerun and the Sheet updated

### DIFF
--- a/sql/2021/markup/attributes.sql
+++ b/sql/2021/markup/attributes.sql
@@ -22,6 +22,7 @@ SELECT
   _TABLE_SUFFIX AS client,
   almanac_attribute_info.name,
   SUM(almanac_attribute_info.freq) AS freq, # total count from all pages
+  SUM(SUM(almanac_attribute_info.freq)) OVER (PARTITION BY _TABLE_SUFFIX) AS total,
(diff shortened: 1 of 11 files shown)


sql/2021/markup/attributes.sql
```diff
@@ -30,6 +31,7 @@ GROUP BY
   client,
   almanac_attribute_info.name
 ORDER BY
+  pct_ratio DESC,
```
kevinfarrugia, line 34: Is sorting by pct_ratio preferable to using `ROW_NUMBER()` and then filtering on it?
```
SELECT
  ..
  ROW_NUMBER() OVER (PARTITION BY client ORDER BY COUNT(0) DESC) AS pos
FROM
  ..
WHERE pos <= 100
```

tunetheweb: Using `ROW_NUMBER() OVER...` does have the advantage of being able to have exactly 100 mobile and 100 desktop results. However in reality that rarely matters as we normally only report on the top 5, or 10, or 15. So as long as you have enough values (100 or 200 is usually a good enough number) and usage is high enough on both mobile and desktop then it rarely matters.

And just ordering by `pct` is a heck of a lot simpler 😁

kevinfarrugia (approved):

PR #2570, opened 2021-11-22
chore: Added extra contributor detail

soulcorrosion:
Hi, just updating my contributor info. Thank you


tunetheweb:
You also contributed in 2019. Do you want to update that too?

We really must merge these at some point to stop having to manage duplicate entries.

soulcorrosion:
Hi Barry, yes, I think the only difference was the website. I'll update 2019 to guarantee we have the same information. Thank you

tunetheweb (approved):

Issue #2192, opened 2021-05-05
Rethinking chapter comments


rviscomi:
I'd like to reevaluate our use of discuss.httparchive.org for chapter comments. It's been relatively underused and one thing I dislike is that the comments aren't visible on the chapter page itself.

WDYT about embedding a comments widget in each chapter where readers can give feedback without leaving the page? Something like [Disqus](https://disqus.com/).

Alternatively, we could go all-in on discuss.httparchive.org and use its API to more tightly integrate it with the chapter page. For example, on the [HTTP Archive website](https://httparchive.org/#:~:text=posts-,Featured,Discussions,-Topic) we use the API to list recent and top discussion threads and pull out some metadata like number of replies, title, and commenters. We could similarly extract the contents of each comment and show them on the page. The forum has built-in moderation and spam prevention, which is a nice bonus. But my biggest concern would be the registration flow for new users. Maybe embedding the comments is all we need to do and new comments and registrations can still happen on the forum itself?

bazzadp:
Yes it has been relatively underused, and it not being directly on the page is almost certainly an extra barrier.

However, in my opinion comments on posts seems to be a dying art. On my own (admittedly low-traffic-ed) blog, which uses Disqus, I've noticed them drop to basically zero. Even [popular articles I've published on high trafficed sites like Smashing Magazine](https://www.smashingmagazine.com/2021/04/complete-guide-measure-core-web-vitals/) only pick up a few comments. I really do think it's a bit of a dying medium to be honest though some sites like [Dev.to](https://dev.to/) and [CSS Tricks](https://css-tricks.com/) do seem to get a fair amount of comments, though the quality varies.

I wonder if we're better with something like Web Mentions ([example](https://nooshu.github.io/blog/2019/11/18/implementing-webmentions/)) that pick up from the likes of Twitter are a better option? Many often see them as more privacy preserving and performant too as Disqus has a bad name here.

But keen to hear other's thoughts!

rviscomi:
Web Mentions is a great idea and I'd be happy ditching comments in favor of that!

ibnesayeed:
I second the adoption of Web Mentions!

bazzadp:
I experimented a bit with Web Mentions in my own blog and have actually gone off the idea a bit. It makes the chapter seem a little messy to be honest.

[...]

PR #2572, opened 2021-11-22
Implementation of UI for webmention reactions.

SaptakS:
Refs #2192 

I have added the UI implementation for the webmention reactions corresponding to like, reposts, replies and mentions of the chapter link. The tabs are keyboard navigable and have aria-notations similar to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role

Screenshot for likes in 2021 restructured data:
image

Screenshot for reposts in 2021 restructured data:
image

Screenshot for replies in 2021 restructured data:
image

Screenshot for mention in 2021 restructured data:
image

Things that can still be improved or added:
- HTML sanitizing or DOM purifying: In the content for replies and mentions, we want to keep some HTML tags but santizie maybe the rest of them to prevent any kind of XSS attack. I know https://github.com/cure53/DOMPurify allows such implementation telling which tags and attributes to allow. Does anyone have any other suggestion?
- Pagination: If the number of replies or mentions become a lot, then it might cause a lot of scroll. Also, in likes and reposts, more number of likes mean more images to load. So might want to think about that?
- Homepage reactions: I and @tunetheweb had a little discussion that it might be worth adding webmentions (like replies and mentions) in the homepage in the form of testimonials like in https://www.11ty.dev/#dont-take-my-word-for-it-%F0%9F%8C%88

Staged URL: https://20211122t101742-dot-webalmanac.uk.r.appspot.com/en/2021/

### DIFF
--- a/src/static/css/page.css
+++ b/src/static/css/page.css
@@ -250,7 +250,8 @@
 }
 
 .authors h2,
-.chapter-links h2 {
+.chapter-links h2,
+.webmentions h2 {
   padding: 16px 0;
   padding: 1rem 0;
 }
@@ -258,7 +259,9 @@
 .authors,
 .authors h2,
 .chapter-links,
-.chapter-links h2 {
+.chapter-links h2,
+.webmentions,
+.webmentions h2 {
   border-bottom: 1px solid #1a2b490a;
   font-size: 17px;
   font-size: 1.0625rem;
@@ -856,6 +859,47 @@ pre {
 
 .code-block code {
   width: 100%;
+}
+
+/* Webmention code */
+.reaction-tabs [role="tab"][aria-selected="true"] {
+  z-index: 3;
+  font-weight: bold;
+  border-bottom: 0.5rem solid #a8caba;
+}
+.reaction-tabs [role="tab"] {
+  position: relative;
+  z-index: 1;
+  background: white;
+  border: 0;
+  padding: 0rem 0.75rem 0.25rem;
+}
+
+.reactions [role="tabpanel"] {
+  margin-top: 1.5rem;
+}
+
+.webmention-likes,
+.webmention-reposts {
+  display: flex;
+}
+
+.webmentions .reactions ul {
+  padding: 0;
+}
+
+.webmentions .reactions ul > li {
+  margin-right: 1.5rem;
+}
+
+.webmentions .reactions ul > li::before {
+  display: none;
+}
+
+.webmention-author .webmention-author-avatar {
+  width: 60px;
+  height: 60px;
+  border-radius: 50%;
 }
 
 /* Needed for Safari to hide the index properly */
--- a/src/static/css/page.css
+++ b/src/static/css/page.css
@@ -867,6 +867,7 @@ pre {
   font-weight: bold;
   border-bottom: 0.5rem solid #a8caba;
 }
+
 .reaction-tabs [role="tab"] {
   position: relative;
   z-index: 1;
--- a/src/static/css/page.css
+++ b/src/static/css/page.css
@@ -873,7 +873,7 @@ pre {
   z-index: 1;
   background: white;
   border: 0;
-  padding: 0rem 0.75rem 0.25rem;
+  padding: 0.6rem 0.75rem 0.6rem;
 }
 
 .reactions [role="tabpanel"] {
--- a/src/static/css/page.css
+++ b/src/static/css/page.css
@@ -880,6 +880,10 @@ pre {
   margin-top: 1.5rem;
 }
 
+.reaction-tabs > button > span {
+  pointer-events: none;
+}
+
 .webmention-likes,
 .webmention-reposts {
   display: flex;
@@ -923,16 +927,16 @@ pre {
   color: #515660;
 }
 
-.webmention-reply-item,
-.webmention-mention-item {
+.webmention-replies-item,
+.webmention-mentions-item {
   display: flex;
   flex-direction: column;
   position: relative;
   padding: 1rem 0 1rem 60px;
 }
 
-.webmention-reply-item .webmention-author .webmention-author-avatar,
-.webmention-mention-item .webmention-author .webmention-author-avatar {
+.webmention-replies-item .webmention-author .webmention-author-avatar,
+.webmention-mentions-item .webmention-author .webmention-author-avatar {
   position: absolute;
   top: 1rem;
   left: 0;
--- a/src/static/css/page.css
+++ b/src/static/css/page.css
@@ -882,6 +882,7 @@ pre {
 .webmention-likes,
 .webmention-reposts {
   display: flex;
+  flex-wrap: wrap;
 }
 
 .webmentions .reactions ul {
@@ -889,17 +890,51 @@ pre {
 }
 
 .webmentions .reactions ul > li {
-  margin-right: 1.5rem;
+  margin-right: 1rem;
 }
 
 .webmentions .reactions ul > li::before {
   display: none;
 }
 
+.webmentions a.webmention-author {
+  text-decoration: none;
+}
+
+.webmentions a.webmention-author:hover {
+  text-decoration: underline;
+}
+
 .webmention-author .webmention-author-avatar {
-  width: 60px;
-  height: 60px;
+  width: 48px;
+  height: 48px;
   border-radius: 50%;
+  object-fit: cover;
+  background: grey;
+}
+
+.webmention-meta {
+  font-size: 0.875rem;
+  color: #515660;
+}
+
+.webmention-meta a.webmention-source {
+  color: #515660;
+}
+
+.webmention-reply-item,
+.webmention-mention-item {
+  display: flex;
+  flex-direction: column;
+  position: relative;
+  padding: 1rem 0 1rem 60px;
+}
+
+.webmention-reply-item .webmention-author .webmention-author-avatar,
+.webmention-mention-item .webmention-author .webmention-author-avatar {
+  position: absolute;
+  top: 1rem;
+  left: 0;
 }
 
 /* Needed for Safari to hide the index properly */
--- a/src/static/js/webmentions.js
+++ b/src/static/js/webmentions.js
@@ -1,0 +1,170 @@
+// Code related to parsing and showing webmentions
+
+// Gets the webmentions json data from current URL
+async function getWebmentions(targetURL) {
+  const apiURL = `https://webmention.io/api/mentions.jf2?perPage=500&target=${targetURL}`;
+  let mentions = []
+  try {
+    const response = await window.fetch(apiURL);
+    if (response.status >= 200 && response.status < 300) {
+      json = await response.json();
+      mentions = json.children;
+    } else {
+      console.error("Could not parse response", response.statusText);
+    }
+  } catch(error) {
+    console.error("Request failed", error);
+  }
+  return mentions;
+}
+
+// Parse webmentions for individual category
+function parseMentions(webmentions, mentionType) {
+  let filteredMentions = []
+  webmentions.forEach(function(mention){
+    if (mention["wm-property"] == mentionType) {
+      filteredMentions.push(mention)
+    }
+  });
+  return filteredMentions;
+}
+
+// Renders webmention likes
+function renderLikes(likes) {
+  // Add the likes count to the likes-tab
+  document.querySelector('#likes-count').innerHTML = likes.length;
+  let likeHtmlElements = [];
+  likes.forEach(function(like) {
+    const likeHtml = `
+      <li class="webmention-likes-item">
+        <a class="webmention-author" href="${like["url"]}" aria-label="${like["author"]["name"]} liked this chapter">
+          <img class="webmention-author-avatar" src="${like["author"]["photo"]}" />
+        </a>
+      </li>
+    `;
+    likeHtmlElements.push(likeHtml);
+  })
+  document.querySelector("#likes-panel").innerHTML = `
+    <ul class="webmention-likes">
+      ${likeHtmlElements.join("\n")}
+    </ul>
+  `;
+}
+
+// Renders webmention reposts
+function renderReposts(reposts) {
+  // Add the reposts count to the reposts-tab
+  document.querySelector('#reposts-count').innerHTML = reposts.length;
+  let repostHtmlElements = [];
+  reposts.forEach(function(repost) {
+    const repostHtml = `
+      <li class="webmention-repost-item">
+        <a class="webmention-author" href="${repost["url"]}" aria-label="${repost["author"]["name"]} liked this chapter">
+          <img class="webmention-author-avatar" src="${repost["author"]["photo"]}" />
+        </a>
+      </li>
+    `;
+    repostHtmlElements.push(repostHtml);
+  })
+  document.querySelector("#reposts-panel").innerHTML = `
+    <ul class="webmention-reposts">
+      ${repostHtmlElements.join("\n")}
+    </ul>
+  `;
+}
+
+// Renders webmention replies
+function renderReplies(replies) {
+
+}
+
+// Parses and renders mentions into likes, reposts, replies and mentions
+async function renderWebmentions(targetURL) {
+  const webmentions = await getWebmentions(targetURL);
+  if (!webmentions.length) {
+    return;
+  }
+
+  const likes = parseMentions(webmentions, "like-of");
+  const reposts = parseMentions(webmentions, "repost-of");
+  const replies = parseMentions(webmentions, "in-reply-to");
+
+  if (likes.length) {
+    renderLikes(likes);
+  }
+
+  if (reposts.length) {
+    renderReposts(reposts);
+  }
+
+  if (replies.length) {
+    renderReplies(replies);
+  }
+}
+
+// Change tabs for webmentions UI
+function changeTabs(e) {
+  const target = e.target;
+  const parent = target.parentNode;
+  const grandparent = parent.parentNode;
+
+  // Remove all current selected tabs
+  parent
+    .querySelectorAll('[aria-selected="true"]')
+    .forEach(t => t.setAttribute("aria-selected", false));
+
+  // Set this tab as selected
+  target.setAttribute("aria-selected", true);
+
+  // Hide all tab panels
+  grandparent
+    .querySelectorAll('[role="tabpanel"]')
+    .forEach(p => p.setAttribute("hidden", true));
+
+  // Show the selected panel
+  grandparent.parentNode
+    .querySelector(`#${target.getAttribute("aria-controls")}`)
+    .removeAttribute("hidden");
+}
+
+function addTabListeners() {
+  const tabs = document.querySelectorAll('.reactions [role="tab"]');
+  const tabList = document.querySelector('.reactions [role="tablist"]');
+
+  // Add a click event handler to each tab
+  tabs.forEach(tab => {
+    tab.addEventListener("click", changeTabs);
+  });
+
+  // Enable arrow navigation between tabs in the tab list
+  let tabFocus = 0;
+
+  tabList.addEventListener("keydown", e => {
+    if (e.keyCode === 39 || e.keyCode === 37) {
+      tabs[tabFocus].setAttribute("tabindex", -1);
+      // Move right
+      if (e.keyCode === 39) {
+        tabFocus++;
+        // If we're at the end, go to the start
+        if (tabFocus >= tabs.length) {
+          tabFocus = 0;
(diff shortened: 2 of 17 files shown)


src/static/js/webmentions.js
```diff
@@ -0,0 +1,288 @@
...
+      filteredMentions.push(mention)
+    }
+  });
+  return filteredMentions;
+}
+
+// Renders webmention likes
+function renderLikes(likes) {
+  // Add the likes count to the likes-tab
+  document.querySelector('#likes-count').innerHTML = likes.length;
```
rviscomi, line 41: Could we reimplement every use of `innerHTML` with something like `createElement` and `appendChild`?

src/static/js/webmentions.js
```diff
@@ -0,0 +1,288 @@
+// Code related to parsing and showing webmentions
+
+// Gets the webmentions json data from current URL
+async function getWebmentions(targetURL) {
+  const apiURL = `https://webmention.io/api/mentions.jf2?perPage=500&target=${targetURL}`;
+  let mentions = []
+  try {
+    const response = await window.fetch(apiURL);
+    if (response.status >= 200 && response.status < 300) {
+      json = await response.json();
```
rviscomi, line 10:
```suggestion
      const json = await response.json();
```

src/static/js/webmentions.js
```diff
@@ -0,0 +1,288 @@
...
+  // Add a click event handler to each tab
+  tabs.forEach(tab => {
+    tab.addEventListener("click", changeTabs);
+  });
+
+  // Enable arrow navigation between tabs in the tab list
+  let tabFocus = 0;
+
+  tabList.addEventListener("keydown", e => {
+    if (e.keyCode === 39 || e.keyCode === 37) {
```
rviscomi, line 260: Can we use the key names for these codes, or set them to constants so it's clear what they represent?

src/static/js/webmentions.js
```diff
@@ -0,0 +1,288 @@
...
+  document.querySelector('#likes-count').innerHTML = likes.length;
+  let likeHtmlElements = [];
+  likes.forEach(function(like) {
+    const likeHtml = `
+      <li class="webmention-likes-item">
+        <a
+          class="webmention-author"
+          href="${like["url"]}"
+          title="${like["author"]["name"]}"
+          aria-label="${like["author"]["name"]} liked this chapter"
```
rviscomi, line 50: I wonder if it's worth decoupling the descriptive text ("... liked this chapter") from the script. For example, we could translate it by defining the text in the localized Jinja templates. To get the translation into the script, we could use `querySelector` to grab the invisible element out of the DOM.

@tunetheweb WDYT?

rviscomi (dismissed):

src/static/js/webmentions.js
tunetheweb: Yeah we normally have a `data-` attribute for this to allow translated text to be picked up from the template.

For example:
https://github.com/HTTPArchive/almanac.httparchive.org/blob/ea67e4b8a40c1c712394f7412c26f866f9f982a2/src/static/js/almanac.js#L158-L159

src/templates/en/base.html
```diff
@@ -88,6 +88,12 @@
 {% block author %}Author{% endblock %}
 {% block authors %}Authors{% endblock %}
 
+{% block reactions %}Reactions{% endblock %}
+{% block likes %}Likes{% endblock %}
+{% block reposts %}Reposts{% endblock %}
+{% block replies %}Replies{% endblock %}
+{% block mentions %}Mentions{% endblock %}
+
```
rviscomi, line 96: I think the linter is complaining because these blocks should also be added to the other lang dirs (in English for now until translated)

rviscomi:
I've staged the PR here: https://20211122t101742-dot-webalmanac.uk.r.appspot.com/en/2021/

@SaptakS @tunetheweb I'm getting 0 reactions on all chapters. Is that a dev env CORS issue? How did it work for the screenshots above?

image

SaptakS:
Not sure if it's available in all chapters. I tested in structured-data. And seems it's showing up there: https://20211122t101742-dot-webalmanac.uk.r.appspot.com/en/2021/structured-data

rviscomi:
Weird! Not appearing on that chapter for me.

image

rviscomi:
Very weird. It only seems to work some of the time. Here's a side by side comparison of the same URL loaded twice

image

src/static/js/webmentions.js
```diff
@@ -0,0 +1,288 @@
...
+        }
+      }
+
+      tabs[tabFocus].setAttribute("tabindex", 0);
+      tabs[tabFocus].focus();
+    }
+  });
+}
+
+window.addEventListener("DOMContentLoaded", () => {
```
VictorLeP, line 284: Is `DOMContentLoaded` "safe" to use here? Some cursory Stack Overflow browsing suggests that the event might fire before the event listener is added, meaning the callback function is never executed. This might explain the erratic behavior in loading the mentions.

rviscomi: Great catch! I agree this sounds like a likely culprit for the inconsistent behavior.

Given that webmentions.js is added to the DOM at the end of the `<body>`, I think we could safely remove the listener and immediately invoke the handler function.

src/templates/base/base_chapter.html
```diff
@@ -149,6 +150,60 @@ <h2 id="explore-results">
...
+  <div class="reactions">
+    <div class="reaction-tabs" role="tablist" aria-label="Reaction Tabs">
+      <button
+        role="tab"
+        aria-selected="true"
+        aria-controls="likes-panel"
+        id="likes-tab"
+        tabindex="0"
+      >
+        <span id="likes-count">0</span> {{ self.likes() }}
```
rviscomi, line 166: I think we'd need additional logic to handle singular and plural grammar, ie "1 like" vs "2 likes".

This is complicated for translations that follow different rules than simply 1 or not-1. @tunetheweb do we have any precedent for how to handle that in the translations?

src/static/js/webmentions.js
SaptakS: That's a great catch! Thanks for pointing that out @VictorLeP . Makes sense to me. I will make the function calls in the eventListener immediately without any eventListener then.

src/templates/base/base_chapter.html
tunetheweb: Yup:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/4b332cae91268b62fa5d208a5bc681a88242add6/src/static/js/almanac.js#L471-L479

Though suspect we need more than that for Slavic languages that have more than just singlar and plural. We have a `plural_ru` macro for that but looks like we never implemented that for the comment/comments. Can look at that later.

rviscomi: Ah ok. Let's get basic singular/plural support into this PR then we can worry about the more complex rules in a separate issue.

I also just came across [`Intl.PluralRules.select()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules/select) which looks like it could useful.

SaptakS:
I have tried to address most of the comments above.

tunetheweb (changes requested):
This is fantastic work @SaptakS !

Particularly appreciate the effort to make it accessible (and today I learned how to make accessible tablists!).

Got some requested changes though.

src/static/js/webmentions.js
```diff
@@ -0,0 +1,331 @@
...
+        >
+          <img
+            class="webmention-author-avatar"
+            src="${like["author"]["photo"]}"
+            loading="lazy" width="60" height="60"/>
+        </a>
+      </li>
+    `;
+    const parser = new DOMParser();
+	  const likeHtmlObject = parser.parseFromString(likeHtml, 'text/html');
```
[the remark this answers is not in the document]
rviscomi: What parts of the web mention are important to render as raw HTML? Why not play it safe and render text contents only?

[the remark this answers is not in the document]
tunetheweb: I thought about that, and there is a text field in the WebMentions data set. The only thing that renders is Links (including Twitter names). It's nice, but I'm not sure it's "have to add DomPurify to the site" nice.

HTML:

image

Text:

image

Though currently the way the whole section is rendered is as a string, which is then parsed to HTML. So that would have to be rewritten but think we probably shoud.

If we are going that route we'd need to ensure w're using [`textcontent`](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) and not `innerHTML` or `innerText`.

src/static/js/webmentions.js
```diff
@@ -0,0 +1,331 @@
...
+  document.querySelector('#likes-count').textContent = likes.length;
+  if (likes.length === 1) {
+    document.querySelectorAll('.like-singular').forEach(el => {
+      el.removeAttribute('data-translation');
+    });
+  } else {
+    document.querySelectorAll('.like-plural').forEach(el => {
+      el.removeAttribute('data-translation');
+    });
+  }
```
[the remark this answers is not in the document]
SaptakS: Makes sense! I kind of leaned on to the implementation for comments, but this makes more sense in this context.

src/static/js/webmentions.js
```diff
@@ -0,0 +1,331 @@
...
+  const webmentionLikesList = document.createElement("ul");
+  webmentionLikesList.setAttribute("class", "webmention-likes");
+  likes.forEach(function(like) {
+    const likeHtml = `
+      <li class="webmention-likes-item">
+        <a
+          class="webmention-author"
+          href="${like["url"]}"
+          title="${like["author"]["name"]}"
+          aria-label="${like["author"]["name"]} ${likeAriaText}"
```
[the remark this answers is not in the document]
SaptakS: > Shouldn't title and Aria Label be equal?

Not necessarily. Many assistive tools will announce both actually. aria-label in anchor tags are helpful to convey the actual purpose of the link when the anchor text doesn't. Hence my effort to add the text along with name. But I can see how it would complicate the translation. Still think maybe we should have an open ticket for such cases to brainstorm later?

src/static/js/webmentions.js
```diff
@@ -0,0 +1,331 @@
...
+        >
+          <img
+            class="webmention-author-avatar"
+            src="${like["author"]["photo"]}"
+            loading="lazy" width="60" height="60"/>
```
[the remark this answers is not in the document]
SaptakS: That was a bad miss on my part for some reason. 😅

src/static/js/webmentions.js
```diff
@@ -0,0 +1,331 @@
...
+        if (tabFocus < 0) {
+          tabFocus = tabs.length - 1;
+        }
+      }
+
+      tabs[tabFocus].setAttribute("tabindex", 0);
+      tabs[tabFocus].focus();
+    }
+  });
+}
```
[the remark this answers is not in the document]
SaptakS: Yes. I saw both the implementations in different places. So kind of decided to go with an enter key at end, but totally can move on the left and right keys.

src/static/js/webmentions.js
```diff
@@ -0,0 +1,331 @@
...
+// Parse webmentions for individual category
+function parseMentions(webmentions, mentionType) {
+  let filteredMentions = []
+  webmentions.forEach(function(mention){
+    if (mention["wm-property"] == mentionType) {
+      filteredMentions.push(mention)
+    }
+  });
+  return filteredMentions;
+}
```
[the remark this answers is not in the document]
SaptakS: We can do the order while fetching from webmentions link itself.

src/static/js/webmentions.js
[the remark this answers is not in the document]
SaptakS: As mentioned in the description, I was also thinking about how to sanitize html or to use the text itself. In case we don't care about links, I think we can safely just show the text, given we anyways link to the source. In that case I can rewrite the entire renders using createElement, appendChild and textContent. I will do that.

SaptakS:
@tunetheweb I have addressed all the comments.

src/static/js/webmentions.js
```diff
@@ -0,0 +1,211 @@
...
+  if (!webmentions.length) {
+    return;
+  }
+
+  renderReactions(webmentions, "likes");
+  renderReactions(webmentions, "reposts");
+  renderReactions(webmentions, "replies");
+  renderReactions(webmentions, "mentions");
+
+}
```
tunetheweb, line 137: Is there much value in this function? Especially since you need have the `reactionMap` hardcoded above anyway so we know there are 4 types? Would it be better to merge renderWebmentions and `renderReactions into one function?

tunetheweb (commented):
Looks much cleaner. Will have a deeper look later but a few quick comments for you upfront.

src/static/js/webmentions.js
SaptakS: That makes sense to me. I will do that.

src/static/js/webmentions.js
```diff
@@ -0,0 +1,211 @@
...
+  try {
+    const response = await window.fetch(apiURL);
+    if (response.status >= 200 && response.status < 300) {
+      const json = await response.json();
+      mentions = json.children;
+    } else {
+      gtag('event', 'error', { 'event_category': 'webmentions.js', 'event_label': response.statusText, 'value': 1 });
+    }
+  } catch(error) {
+    gtag('event', 'error', { 'event_category': 'webmentions.js', 'event_label': error, 'value': 1 });
```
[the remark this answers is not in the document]
SaptakS: Done

src/static/js/webmentions.js
SaptakS: Done

src/static/css/page.css
```diff
@@ -858,6 +861,87 @@ pre {
...
+  font-weight: bold;
+  border-bottom: 0.5rem solid #a8caba;
+}
+
+.reaction-tabs [role="tab"] {
+  position: relative;
+  z-index: 1;
+  background: white;
+  border: 0;
+  padding: 0rem 0.75rem 0.25rem;
```
tunetheweb, line 876: Solve tap target issue by making sure it's at least 44px high

```suggestion
  padding: 0.6rem 0.75rem 0.6rem;
```

src/static/js/webmentions.js
```diff
@@ -0,0 +1,209 @@
+// Code related to parsing and showing webmentions
+
+// Gets the webmentions json data from current URL
+async function getWebmentions(targetURL) {
+  const apiURL = `https://webmention.io/api/mentions.jf2?perPage=500&target=${targetURL}`;
```
tunetheweb, line 5: Oldest comment first

```suggestion
  const apiURL = `https://webmention.io/api/mentions.jf2?perPage=500&target=${targetURL}&sort-dir=up`;
```

src/static/js/webmentions.js
```diff
@@ -0,0 +1,209 @@
...
+  let mentions = []
```
tunetheweb, line 6: Fix code scanning issue:

```suggestion
  let mentions = [];
```

src/static/js/webmentions.js
```diff
@@ -0,0 +1,209 @@
...
+    console.error("Request failed", error);
+    gtag('event', 'error', { 'event_category': 'webmentions.js', 'event_label': error, 'value': 1 });
+  }
+  return mentions;
+}
+
+// Format published date into human readable form
+function formatDate(dateString){
+  const options = { year: "numeric", month: "long", day: "numeric" }
+  return new Date(dateString).toLocaleDateString(undefined, options)
```
tunetheweb, line 26: Show time as well:

```suggestion
  return new Date(dateString).toLocaleTimeString(undefined, options)
```

src/static/js/webmentions.js
```diff
@@ -0,0 +1,209 @@
...
+  // Show the selected panel
+  grandparent.parentNode
+    .querySelector(`#${target.getAttribute("aria-controls")}`)
+    .removeAttribute("hidden");
+}
+
+function addTabListeners() {
+  const tabs = document.querySelectorAll('.reactions [role="tab"]');
+  const tabList = document.querySelector('.reactions [role="tablist"]');
+
```
tunetheweb, line 171: Fix the fact it's broken for prior years:

```suggestion

  if (!tabs || !tabList) {
    return;
  }
```

src/templates/base/base_chapter.html
```diff
@@ -344,6 +408,16 @@ <h1 class="title title-lg">
             <span data-translation class="comment-plural visually-hidden-mobile">{{ self.comments() }}</span>
           </a>
         </div>
+        {% elif year | int >= 2021 %}
+        <div id="cta-container">
+          <a class="alt btn chapter-cta comment-cta" href="#reactions">
+            <svg width="22" height="22" role="img" aria-hidden="true">
+              <title id="discuss-this-chapter-cta">{{ self.discuss_this_chapter() }}:</title>
```
tunetheweb, line 415: Remove this as inaccurate (it's using `aria-hidden` anyway)

```suggestion
```

src/templates/base/base_chapter.html
```diff
@@ -149,6 +150,64 @@ <h2 id="explore-results">
   {% endif %}
 {% endmacro %}
 
+{% macro render_webmentions() %}
+  <h2 id="reactions">
+    <a href="#reactions" class="anchor-link">{{ self.reactions() }}</a>
+  </h2>
+  <div class="reactions" data-source="{{ self.source() }}">
+    <div class="reaction-tabs" role="tablist" aria-label="Reaction Tabs">
```
tunetheweb, line 158:
```suggestion
    <div class="reaction-tabs" role="tablist" aria-label="{{ self.reactions() }}">
```

tunetheweb (changes requested):
I'm pretty happy with this now. I think these last few suggestions will make CI pass and allow us to merge this.

tunetheweb (approved):
I think this is good to go! Great work on it.

@rviscomi you wanna have another look or happy to merge?

src/templates/nl/base.html
```diff
@@ -88,6 +88,18 @@
 {% block author %}Auteur{% endblock %}
 {% block authors %}Auteurs{% endblock %}
 
+{% block reactions %}Reacties{% endblock %}
+{% block source %}Bron bekijken{% endblock %}
+
+{% block like %}like{% endblock %}
+{% block repost %}repost{% endblock %}
+{% block reply %}antwoord{% endblock %}
+{% block mention %}vermeld{% endblock %}
```
VictorLeP, line 97:
```suggestion
{% block mention %}vermelding{% endblock %}
```

tunetheweb:
Great job @SaptakS ! Thanks a lot for this.

Btw there's another tab index dev issue if interested?: https://github.com/HTTPArchive/almanac.httparchive.org/issues/1325#issuecomment-836071476

src/templates/nl/base.html
tunetheweb: Missed this. Will open another PR. Happy with the rest of the Dutch? I just used Google Translate to be honest but figured an attempt was better than no attempt at all!

src/templates/fr/base.html
```diff
@@ -88,6 +88,18 @@
 {% block author %}Auteur·ice{% endblock %}
 {% block authors %}Auteur·ice·s{% endblock %}
 
+{% block reactions %}Réactions{% endblock %}
+{% block source %}Voir la source{% endblock %}
+
+{% block like %}comme{% endblock %}
```
VictorLeP, line 94:
```suggestion
{% block like %}j'aime{% endblock %}
```

src/templates/nl/base.html
VictorLeP: Dutch speakers will probably understand 'like' and 'repost' though it's not ideal. Also have some changes for the French translation.

src/templates/fr/base.html
tunetheweb: Going with `j’aime` as this won't be fixed by smart quotes library.

VictorLeP: Where do I comment? Here or in the new PR?

tunetheweb: New PR please. Will resolve these comments.

tunetheweb:
FYI @SaptakS I forgot to add you to the Developers team with this so added that to #2590 too.

SaptakS:
> Great job @SaptakS ! Thanks a lot for this.
> 
> Btw there's another tab index dev issue if interested?: [#1325 (comment)](https://github.com/HTTPArchive/almanac.httparchive.org/issues/1325#issuecomment-836071476)

I can take a stab at it after the security chapter related stuff is done with.

PR #1969, opened 2021-02-03
Handle Russian and Ukrainian plurals

bazzadp:
Fixes #1968 using [plural-ru](https://github.com/andrewp-as-is/plural-ru.py) function. Note this is not available as an importable function so added code directly to our code base (checked the licence first).

Home page intro:
Home page intro

Methodology text:
Methodology text

Contributors:
Contributors

@dtikhonov can you confirm I've used the right words?

English word|ends in 1 (except 11)| ends in 2,3 and 4 (except 12, 13, 14)|else
-----|-----|----|----
chapters|глава|главы|глав
contributors|участник|участника|участников
people|человек|человека|человек

It does look like we used "главах" below but Google Translate didn't come up with that at all so I guess double check that?

@tymosh does Ukrainian follow the same rules? Should we do the same for those templates?

PR #2590, opened 2021-11-25
Web mentions translations

tunetheweb:
@HTTPArchive/translators in #2572 we implemented Web Mentions UI to the chapters

This necessitated some new words in our `base.html` template:

```
{% block reactions %}Reactions{% endblock %}
{% block source %}View source{% endblock %}

{% block like %}like{% endblock %}
{% block repost %}repost{% endblock %}
{% block reply %}reply{% endblock %}
{% block mention %}mention{% endblock %}
{% block likes %}likes{% endblock %}
{% block reposts %}reposts{% endblock %}
{% block replies %}replies{% endblock %}
{% block mentions %}mentions{% endblock %}
```

Which are shown like below:

image

I used Google Translate to translate those to our various dialects, but undoubtedly that will have errors. However I felt that was good enough to merge and didn't know if/when you'd all have time to review.

I'm opening this new PR so you can can all tell me the better words to use! I've already got a couple of corrections in there thanks to @VictorLeP but feel free to add new ones directly to this branch, or in the comments in you prefer.

I will merge what we have for the 1st December release.


tymosh:
Hi @tunetheweb. In Ukrainian we have more than 2 forms. This issue has already been addressed in #1969. Can you do the same for this template?

tunetheweb:
Yeah thought this might come up. We addressed it on the template side, but this is client side generated code. So will need to reimplement. Can you comment with a table of the three forms for each affected link (the likes, comments, mentions and reposts I presume?) and will look into it.

VictorLeP:
fyi: the main issue I saw in the Dutch/French translations was that Google Translate decided on its own which part of speech it thought the word was. "like" can be "I gave it a _like_", "do you _like_ this", "this is _like_ that" - similar for "a/to reply". So some translations were completely off because it picked the wrong one.

tunetheweb:
Yeah language is tough. I'm still in two minds whether auto translating is better than leaving in English because of this.

On the one hand, at least making an attempt seems admirable, and often it will be right, and when it's not, at least it's easy to search for in the code base to find it and correct it (reusing English one will be more difficult to find for those that don't know the code base, and also potentially make people think translating that phrase is not supported). Also do think it encourages native speakers to correct them 😉

On the other it often gets it wrong and looks poor. And many languages (Dutch for example) will be more than familiar with ceertain commonly used English phrases like this on-line and may prefer the English to a bad translation.

Any thoughts on this? 🤔

I definitely don't want to hold up PRs on all the translations since not every supported language has an active translator on the project anymore. So it;'s a tough call...

VictorLeP:
This feels to me as if it is based on an assumption that anyone who reads the Web Almanac is a possible contributor (even specifically translator): those who read it and see something wrong will search the code base and correct things. I don't know how realistic that is though... and if it's something we want. It might be quite useful to have a bit more involvement in the Web Almanac to do good translations. For example, I changed a few translations of roles, and I felt like I could only do that because I somewhat know what a reviewer and an editor actually do. (a "reviewer" comments and gives suggestions on the text, they're not handing out 5 stars to a movie)

So, from a viewpoint from a read-only reader, I would feel that these kinds of messy translations will just be very confusing and hard to "back-translate" to English to understand what was meant. I think that we reach a community with already a very good grasp of English (necessary to be involved in web development, imho), so even beyond languages that are used to English loanwords, readers will probably not have an issue with some English words here or there.

You could highlight any untranslated bits on the page and have some call to action to contribute to the translation, but that might be too invasive.

tymosh:
@tunetheweb here's the table:

English word|ends in 1 (except 11)| ends in 2,3 and 4 (except 12, 13, 14)|else
-----|-----|----|----
likes|лайк|лайки|лайків
reposts|репост|репости|репостів
replies|відповідь|відповіді|відповідей
mentions|згадка|згадки|згадок

tymosh:
Another option might be to change text to something what Facebook does (if plugin allows):
Screenshot from 2021-11-26 14-16-22

But use it like:

likes - 1|reposts - 0|replies - 7|mentions - 2
-----|-----|----|----

In this case we can use 1 plural form in Ukrainian/Russian.

Don't know how it affects other languages though.

tunetheweb:
Fixed in last push. Can you confirm these look good:

image

image

image

image

image

image

tymosh:
'ends in 2,3 and 4 (except 12, 13, 14)' case doesn't work. For example, '24 likes' should be '24 лайки'.

tunetheweb:
Sorry works fine. A combo of using the English page and UK plurals.

It looks like this actually:

image

tymosh:
Yeah, this one is correct. I'm not sure about capitalized 'Лайки'. Is that a screenshot issue? No need to capitalize any form is any case.

tunetheweb:
Fixed:

image

tymosh:
Perfect!

tymosh (approved):

src/templates/nl/base.html
```diff
@@ -143,17 +147,17 @@
...
+{% block ebook_title %}E-book{% endblock %}
+{% block ebook_download_short %}E-book PDF ({{ ebook_size_in_mb }}MB){% endblock %}
 {% block ebook_download %}Download de volledige {{ year }} Web Almanac in PDF-formaat ({{ ebook_size_in_mb }}MB){% endblock %}
-{% block ebook_download_note %}(generated with <a href="https://www.princexml.com/">www.princexml.com</a>){% endblock %}
+{% block ebook_download_note %}(gegenereerd met <a href="https://www.princexml.com/">www.princexml.com</a>){% endblock %}
 
 {%
   set localizedPartTitles = {
-    "I. Page Content": "I. Pagina Inhoud",
+    "I. Page Content": "I. Paginainhoud",
```
noah-vdv, line 157:
```suggestion
    "I. Page Content": "I. Inhoudsopgave",
```

VictorLeP: This is a translation for [this part name](https://almanac.httparchive.org/en/2021/table-of-contents#part-1), I don't think this means "table of contents"?

noah-vdv: True, then it'd be "Pagina inhoud"

VictorLeP: But it's a compound noun ("inhoud van een pagina", "content of a page"), so it should be written as one word, no?

noah-vdv: True but a & i can't be right next to eachother because of 'klinkerbotsing' so then it should be Pagina-inhoud

VictorLeP: Yep, you're right! Now it should be good :)

CYBAI (commented):
@tunetheweb Because I cannot use GitHub suggestion on these lines, could you please help me update them? 🙏 Thank you very much!

```diff
- {% block unedited %}[Unedited]{% endblock %}
+ {% block unedited %}[尚未編輯]{% endblock %}
```

```diff
- {% block source %}查看源代碼{% endblock %}
+ {% block source %}查看原始碼{% endblock %}
```

```diff
- {% block reply %}條回复{% endblock %}
- {% block mention %}提及{% endblock %}
- {% block likes %}個贊{% endblock %}
+ {% block reply %}則回覆{% endblock %}
+ {% block mention %}提及{% endblock %}
+ {% block likes %}個喜歡{% endblock %}
```

src/templates/zh-TW/base.html
```diff
@@ -99,6 +99,10 @@
 {% block reposts %}轉帖{% endblock %}
 {% block replies %}回复{% endblock %}
 {% block mentions %}提及{% endblock %}
+{% block likesAlt %}{% endblock %}
+{% block repostsAlt %}{% endblock %}
+{% block repliesAlt %}{% endblock %}
+{% block mentionsAlt %}{% endblock %}
```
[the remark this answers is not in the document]
tunetheweb: These are not needed for Chinese. They are used for alternative plurals for Russian and Ukrainian which have more than just singular and plural. So will leave them blank.

tunetheweb:
Thanks you @CYBAI ! I've updated the zh-TW file. If you get a chance could you recheck to make sure I did it correctly?

tunetheweb:
Going to merge this now. Thanks to all that contributed better translations!

CYBAI:
> Thanks you @CYBAI ! I've updated the zh-TW file. If you get a chance could you recheck to make sure I did it correctly?

Sorry for late reply! Yes! I confirmed they are correct! Thank you very much! 🙏

PR #2599, opened 2021-11-27
TinyPNG Capabilities images

tunetheweb:
Great savings here!! Again a pity that the Calibre Action isn't more aggressive for those willing to lose some of the colour palette.

CSS chapter is next on my list 😁


rviscomi (commented):
Some of these images are over 2000px wide 🤔

tunetheweb:
Some close to 3000!

Now, as screenshots that will be shown on desktop too, they are different than our regular figures which are taken at 1200 as I do notice our figures are not as sharp as the Sheet embeds when viewed on on desktop (using the `?print` param) on a 4k screen. That's OK for figures as desktop users should mostly get Sheets embeds (unless on a slow connection, or behind a company proxy that blocks Sheets, or on data-saver in which case it's right to accept a dip in quality), but for non-figure images (like most of these are), 1200px (or 2x of display width) may not be enough. Probably 2000px is fine, and maybe 3000px is overkill but it's not as extreme as it sounds to be honest.

Either way this PR has a good saving without spending time individually resizing the to see what's acceptable.

tunetheweb:
Total images for the chapter goes from 4.1MB to 1.5MB with this PR. Probably more to save as you say, but that's a good saving for two mins work so will take it.

rviscomi:
Yeah maybe capping images at 2 dpr is something to consider as a best practice for the future. Agreed that these are good savings in this PR and good to merge.

PR #2601, opened 2021-11-28
Fix back ticks usage in Russian PWA translation

pepelsbey:
Original:

image

Translation, looks broken:

image


tunetheweb (approved):
Good spot!

tunetheweb:
Thanks for that @pepelsbey . We'll include that fix in the next release on Wednesday.

PR #2602, opened 2021-11-28
Fix webmentions for sources that do not provide full meta data

tunetheweb:
We've got a couple of web mentions for Performance and Structured Data chapters from a couple of sources who don't give quite as much data as Twitter do. So getting JavaScript console errors. This PR adds some more error checking and fall backs so they render what we have without error:

image


SaptakS (approved):
Looks good to me.

PR #2609, opened 2021-11-29
User singular team role for contibutor cards

tunetheweb:
Makes the team roles singular on the cards as had a few people raise this:

image


tunetheweb:
@HTTPArchive/translators if any of you get a chance, I've had my Google Translate button out again!

Though at least had the plural ones here to go upon so just needed to figure out the singular. So if a Team role is wrong, then could you also check if the existing plural Team name is also wrong?

src/templates/nl/base.html
```diff
@@ -206,3 +206,16 @@
   "translators": "Vertalers",
   }
 %}
+
+{% set localizedTeamRole = {
+  "analysts": "Analist",
+  "authors": "Auteur",
+  "brainstormers": "Brainstormers",
```
VictorLeP, line 213:
```suggestion
  "brainstormers": "Brainstormer",
```

PR #2612, opened 2021-11-29
Update 2021 config to launch remaining chapters

tunetheweb:
Add back in the three chapters removed for pre-launch and remove all the TODOs (except for those chapters which don't have an PR yet).

Keeping in draft for now until ready to merge.


src/config/2021.json
```diff
@@ -124,55 +116,62 @@
...
           "part": "IV",
           "chapter_number": "21",
+          "title": "Compression",
+          "slug": "compression"
+        },
+        {
+          "part": "IV",
+          "chapter_number": "22",
+          "title": "Caching",
+          "slug": "cachine"
```
rviscomi, line 168:
```suggestion
          "slug": "caching"
```

PR #2639, opened 2021-12-01
Fix column scoping on CSS chapter

tunetheweb:
The PDF is not being accessibly tagged.

The CSS chapter has a mixed table heading as it has an empty cell (which must be `<td></td>` according to PrinceXML:

```html
    <tr>
      <td></td>
      <th>Keyword</th>
      <th>Desktop</th>
      <th>Mobile</th>
    </tr>
   </thead>
```

When we do that we need to add scope to the `<th>` headings to explain to PrinceXML what's going on (this is not necessary for tables with all `<th>` heading columns - only ones mixed with `<td>` and `<th>`).

```html
    <tr>
      <td></td>
      <th scope="col">Keyword</th>
      <th scope="col">Desktop</th>
      <th scope="col">Mobile</th>
    </tr>
   </thead>
```

### DIFF
--- a/src/content/en/2021/css.md
+++ b/src/content/en/2021/css.md
@@ -525,9 +525,9 @@ In color formats where the value has historically used commas inside a functiona
    <thead>
     <tr>
       <td></td>
-      <th>Keyword</th>
-      <th>Desktop</th>
-      <th>Mobile</th>
+      <th scope="col">Keyword</th>
+      <th scope="col">Desktop</th>
+      <th scope="col">Mobile</th>
     </tr>
    </thead>
    <tbody>


src/content/en/2021/css.md
```diff
@@ -525,9 +525,9 @@ In color formats where the value has historically used commas inside a functiona
    <thead>
     <tr>
       <td></td>
```
rviscomi, line 527: Should this be `th`?

tunetheweb: Nope.

If it was this wouldn't be an issue. Yeaah!
However then PrinceXML complains about an empty heading. Booo

The correct, accessible markup an empty heading cell is apparently `<td></td>` Which looks weird, but that's what you supposed to do apparenlty! Remember researching this when the ebook first launched as had to set a few of them to that.

The only problem is that automatically turns it into a complex table in PrinceXML's eyes, hence need to explicitly define the roles for the `<th>`'s.

PR #2652, opened 2021-12-01
Remove Emojis from Capabilities chapters

tunetheweb:
Sorry @christianliebel and @tomayac but they don't generate properly in the ebook 😞


tomayac (approved):
LGTM. 

A bit sad, but your reasoning makes sense of course.

PR #2653, opened 2021-12-01
Methodology nits

VictorLeP:
Micro nits on the methodology. These aren't just typos, so feel free to reject if you disagree!

- remove sentence about "BQ billed by the TB" - the billing unit does not mean much to me? If they charged $0.0001 per TB, they'd be billing by the TB but it would still be cheap :)
- update example query to one from 2021 - showcasing the latest iteration of the query style guide ;)

Also noticed that under "Planning" it refers to 23 (+2) chapters, but don't know if that needs to be changed.

### DIFF
--- a/src/templates/en/2021/methodology.html
+++ b/src/templates/en/2021/methodology.html
@@ -81,25 +81,25 @@
         </p>
 
         <p class="note">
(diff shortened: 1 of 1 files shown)


rviscomi (approved):

tunetheweb:
Thanks @VictorLeP 

> Also noticed that under "Planning" it refers to 23 (+2) chapters, but don't know if that needs to be changed.

That is what happened. Just one unfortunately one didn't make it to the end.

PR #2670, opened 2021-12-01
Stop CDN chapter failing CI

### LINKED ISSUE #2668: Keep getting a Lighthouse error in CI for Webmentions CDN
Weird that this happens only for the CDN chapter.

image

@SaptakS any ideas?

tunetheweb:
Fixes #2668 

As mentioned in https://github.com/HTTPArchive/almanac.httparchive.org/issues/2668#issuecomment-984036189 this is (hopefully) a short term measure.


tunetheweb:
FYI @SaptakS if you do end up increase spacing, then might be worth trying to remove this and including a dummy edit of CDN chapter (so Lighthouse is run) to see if it still fails.

SaptakS:
Sounds good!

PR #2674, opened 2021-12-02
Edits to 2021 Compression chapter

iulia-m-comsa:
Minor wording fixes in the Introduction.

### DIFF
--- a/src/content/en/2021/compression.md
+++ b/src/content/en/2021/compression.md
@@ -22,9 +22,9 @@ featured_stat_label_3: Websites that pass the Lighthouse audit with maximum scor
 
 ## Introduction
 
-A user's time is valuable, so they shouldn't have to wait a long time for a web page to load. The HTTP protocol allows to compress responses and this decreases the time needed to transfer the content. Compression often leads to significant improvement in the user experience. It can reduce [page weight](./page-weight), improve [web performance](./performance) and boost search rankings. As such it's an important part of [Search Engine Optimization](./seo).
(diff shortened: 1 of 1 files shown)


src/content/en/2021/compression.md
```diff
@@ -22,9 +22,9 @@ featured_stat_label_3: Websites that pass the Lighthouse audit with maximum scor
...
+A user's time is valuable, so they shouldn't have to wait a long time for a web page to load. The HTTP protocol allows the responses to be compressed, which decreases the time needed to transfer the content. Compression often leads to significant improvement in the user experience. It can reduce [page weight](./page-weight), improve [web performance](./performance) and boost search rankings. As such, it's an important part of [Search Engine Optimization](./seo).
 
-This chapter discusses lossless compression applied on a HTTP response. Lossy and lossless compression used in [media](../2020/media) formats such as images, audio and video are equally if not more important for increasing page speed. However these are not in the scope of this chapter as they usually are part of the file format itself.
+This chapter discusses lossless compression applied to a HTTP response. Lossy and lossless compression used in [media](../2020/media) formats such as images, audio and video are equally (if not more) important for increasing the page loading speed. However, these are not in the scope of this chapter, as they usually are part of the file format itself.
```
tunetheweb, line 27: I'm going to suggest reverting back the change in the first sentence (the second sentence changes look good). The compression is allied on a resource, which then forms part of the HTTP response, rather than on the HTTP response itself. A minor quibble but think it makes the original text more accurate than the proposed change.

```suggestion
This chapter discusses lossless compression applied on a HTTP response. Lossy and lossless compression used in [media](../2020/media) formats such as images, audio and video are equally (if not more) important for increasing the page loading speed. However, these are not in the scope of this chapter, as they usually are part of the file format itself.
```

tunetheweb (changes requested):
Thanks for this. One minor reversion but the rest looks good.

FYI @lvandeve

lvandeve:
looks good to me, thanks!

tunetheweb (approved):
Thanks again for submitting!

PR #2684, opened 2021-12-03
Fix Markup social graph and other minor edits

### LINKED ISSUE #2681: Markup structured data stats are incorrectly marked as % of page
https://almanac.httparchive.org/en/2021/markup#fig-26 says it's % of pages, when it's not - it's percentage of meta data.

Twitter thread: https://twitter.com/tunetheweb/status/1466735564946870276?s=20

This is a problem of the `UNNEST` issue [as discussed here](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide#a-warning-on-totals).

We should either update the label or correct the query.
Personally I think % of pages is more useful and so we should update the query.

tunetheweb:
Fixes #2681 

Changed graph to percentage of pages.

New versus old:

Also did a couple of other minor edits.


AlexLakatos (approved):
LGTM. Thanks for catching and fixing this. It's indeed less confusing!

tunetheweb:
I'm confused by the difference between `og:title` and `twitter:title` (for example) between the two graphs. They should be the same. Looking into it...

tunetheweb:
I don't trust that social-meta query (it's using some fancy `reduce` JS which always makes my head explore) so am instead using the other one, but adding a page count to that. So can also see for other nodes too so more interesting data 😀

AlexLakatos:
There are still minor differences in percentages when compared with the structured data chapter. are we worried about those, or is that an acceptable error margin?

https://almanac.httparchive.org/en/2021/structured-data#fig-4
https://almanac.httparchive.org/en/2021/structured-data#fig-5

tunetheweb:
Looks like they are collected form two different custom metrics. Both look to do roughly the same thing (one collects all meta, the other collects them individually). Not ideal but I think it's fine as roughly the same.

PR #2688, opened 2021-12-03
Source Have I Been Pwned more

### LINKED ISSUE #2687: Issue with the 2021 third-parties chapter
The link is wrong for "Privacy" in [introduction](https://almanac.httparchive.org/en/2021/third-parties#introduction)

tunetheweb:
I feel we should cite Have I Been Pwned more as, unlike other data (e.g. Who Tracks Me data) that is not combined with our data to create original results, but really is just their data. I know we reference it in the text but think we should be doing more here.

As such I've updated the chart (fallback image attached) and also add a source link:

image

I've also renamed the text HaveIBeenPwned to Have I Been Pwned since [that seems to be the preferred way of writing it](https://haveibeenpwned.com/About).

Finally I've also corrected a link in Third Parties chapter to fix #2687 

WDYT?


VictorLeP:
Looks good, good catch on the way it is written. Thanks!

VictorLeP (approved):

PR #2690, opened 2021-12-04
Filter out duplicates in data for SSG detection

denar90:


sql/2021/jamstack/framework_libraries.sql
```diff
@@ -55,7 +55,7 @@ total_ssg_app AS (
 ),
 
 js AS (
-  SELECT
+  SELECT DISTINCT
```
denar90, line 58: I assume we also need that for libs and frameworks

PR #2705, opened 2021-12-07
2021 Media markdown

eeeps:
Markdownifying the google doc. WIP!

Makes progress on #2144 

Staging: https://20211207t160821-dot-webalmanac.uk.r.appspot.com/en/2021/media


src/content/en/2021/media.md
```diff
@@ -1,20 +1,900 @@
...
-description: TODO
+description: Media chapter of the 2021 Web Almanac covering how images and videos are currently being encoded, embedded, styled, and delivered on the web
 authors: [eeeps, dougsillars]
-reviewers: []
-analysts: [eeeps, dougsillars]
+reviewers: [Navaneeth-akam, tpiros, akshay-ranganath, addyosmani]
+analysts: [eeeps, dougsillars, akshay-ranganath]
 editors: []
 translators: []
+eeeps_bio: Eric Portis is a Web Platform Advocate at <a hreflang="en" href="https://cloudinary.com/">Cloudinary</a>.
```
tunetheweb, line 10: @eeeps I copied this profile from a previous year. Feel free to update it.

src/content/en/2021/media.md
```diff
@@ -1,20 +1,900 @@
...
+dougsillars_bio: Doug Sillars is a freelance digital nomad working on the intersection of performance and media. He tweets [@dougsillars](https://twitter.com/dougsillars), and blogs regularly at <a hreflang="en" href="https://dougsillars.com">dougsillars.com</a>.
```
tunetheweb, line 11: @dougsillars I copied this profile from a previous year. Feel free to update it.

src/content/en/2021/media.md
```diff
@@ -1,20 +1,900 @@
 ---
 #See https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide#metadata-to-add-at-the-top-of-your-chapters
 title: Media
-description: TODO
+description: Media chapter of the 2021 Web Almanac covering how images and videos are currently being encoded, embedded, styled, and delivered on the web
 authors: [eeeps, dougsillars]
-reviewers: []
-analysts: [eeeps, dougsillars]
+reviewers: [Navaneeth-akam, tpiros, akshay-ranganath, addyosmani]
+analysts: [eeeps, dougsillars, akshay-ranganath]
```
tunetheweb, line 7: @eeeps are these all up to date and have all the reviewers reviewed the draft (either in Google Docs or here)?

src/content/en/2021/media.md
```diff
@@ -1,20 +1,900 @@
...
+{{ figure_markup(
+  content="99.93%",
+  caption="Percentage of pages that generated at least one request for an image resource",
+  classes="big-number",
+  sheets_gid="1062090109",
+  sql_file="at_least_one_image_request.sql"
+)
+}}
+
+Put simply, almost every page contains image content and effectively all pages serve up _some_ sort of imagery (even if it's just a background or favicon).
```
tunetheweb, line 49: We've two big figures one after the other, which I don't think looks the best. Should we lose one?

src/content/en/2021/media.md
```diff
@@ -1,20 +1,900 @@
...
+  classes="big-number",
+  sheets_gid="1594311632",
+  sql_file="image_srcset_sizes.sql"
+)
+}}
+
+{{ figure_markup(
+  image="srcset-descriptor-adoption.png",
+  caption="`srcset` descriptor adoption.",
+  description="A bar chart showing... TODO",
```
tunetheweb, line 391: Missed one @eeeps

src/content/en/2021/media.md
```diff
@@ -1,20 +1,900 @@
...
+  chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vQM9deg869BD9knNdVhFNbFnUdVXeyuwzUEIgSW-2XgOBEbALtVnoFapQ5JsDxzzepj6mVoepKBmN_m/pubchart?oid=89579817&format=interactive",
+  sheets_gid="1030564653",
+  sql_file="TODO.sql"
```
tunetheweb, line 394: And the SQL file @eeeps

src/content/en/2021/media.md
```diff
@@ -1,20 +1,900 @@
...
+      <tr>
+        <th>`w` descriptors</th>
+        <td class="numeric">24%</td>
+      </tr>
+    </tbody>
+  </table>
+  <figcaption>{{ figure_link(caption="`srcset` descriptor adoption.", sheets_gid="1030564653", sql_file="image_srcset_descriptor.sql") }}</figcaption>
+</figure>
+
+{# TODO I was wondering if this was desktop or mobile, then discovered you had a graph which showed both, so repalced this table with that. I see the table is back now. Is it needed now we have the graph which In think is more interesting? #}
```
tunetheweb, line 421: What do you think @eeeps ?

Also we have a big figure, a figure, and then a table. While we like to be stats driven that's three figures in a row before we get to any text. If we are keeping all three should we add some text in between?

src/content/en/2021/media.md
```diff
@@ -1,20 +1,900 @@
...
+      </tr>
+      <tr>
+        <td>`["video/mp4;","video/webm;"]`</td>
+      </tr>
+    </tbody>
+  </table>
+  <figcaption>{{ figure_link(caption="The most common ordered pairs of `type` values, when there are two `source` elements within a `video` element.", sheets_gid="1549760253", sql_file="video_source_formats_mobile.sql") }}</figcaption>
+</figure>
+
+{# TODO I'm not sure this is such an interesting table. Consider adding some stats to make it more interesting? % of usage on mobile and desktop? #}
```
tunetheweb, line 836: @dougsillars should this be a figure table? If so should we add some stats?

Or should it be a bullet list instead?

src/content/en/2021/media.md
```diff
@@ -1,20 +1,900 @@
...
+
+In six of the top 10 examples, the MP4 is listed as the first source. <a hreflang="en" href="https://caniuse.com/mpeg4">MP4 support on the web is at 98.4%</a>, and the browsers that do not support MP4 generally do not support the `<video>` tag at all. This implies that these sites do not need two sources, and could save some storage on their web servers by removing their WebM or Ogg video sources. (Or, they could reverse the order of the videos, and browsers that support WebM will download the WebM).
+
+{# TODO - second time you meantioned and links to the 9* MP4 stat, but this time it's 98.4% instead fo 98.1%. Do we need more references and do both need to be links? If so can we at least be consistent. #}
```
tunetheweb, line 840: @eeeps / @dougsillars ?

src/content/en/2021/media.md
```diff
@@ -1,20 +1,900 @@
...
+      </tr>
+      <tr>
+        <td>`["video/mp4","video/webm","video/ogv"]`</td>
+      </tr>
+    </tbody>
+  </table>
+  <figcaption>{{ figure_link(caption="The most common ordered triplets of `type` values, when there are three `source` elements within a `video` element.", sheets_gid="1549760253", sql_file="video_source_formats_mobile.sql") }}</figcaption>
+</figure>
+
+{# TODO I'm not sure this is such an interesting table. Consider adding some stats to make it more interesting? % of usage on mobile and desktop? #}
```
tunetheweb, line 887: @dougsillars same questions here as previous - should this be a figure table? If so should we add some stats?

Or should it be a bullet list instead?

tunetheweb (changes requested):
Not had a chance to go through this fully yet, but have some light feedback already so might as well get you started on those!

src/content/en/2021/media.md
tunetheweb: FYI you can see what it looks like here: https://20211207t160821-dot-webalmanac.uk.r.appspot.com/en/2021/media#images

src/content/en/2021/media.md
tunetheweb: What it looks like here: https://20211207t160821-dot-webalmanac.uk.r.appspot.com/en/2021/media#feature-adoption

src/content/en/2021/media.md
tunetheweb: What it looks like here: https://20211207t160821-dot-webalmanac.uk.r.appspot.com/en/2021/media#feature-adoption

src/content/en/2021/media.md
tunetheweb: First reference:

image

second reference:

image

src/content/en/2021/media.md
eeeps: LGTM!

src/content/en/2021/media.md
eeeps: Yes, these are all up to date and all of the reviewers listed here contributed comments / helped shape the final doc.

src/content/en/2021/media.md
eeeps: You have a point. Perhaps I'll move a sentence between them to break them up.

src/content/en/2021/media.md
dougsillars: i think the link is not the right figure, but I found it.  

I can add stats into the table.... like a percentage?

tunetheweb: Yup. Just seems like a one-column table is more a list than a table. So either change to list or make it at least two columns (or three if wanna show Desktop and mobile). A table does have the advantage that it links to data and SQL query so the latter I think is better.

src/content/en/2021/media.md
eeeps: The chart and the table are duplicitous. Deleting the table and adding a sentance to break up the back-to-back figures.

eeeps: Done.

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+So: one in ten `sizes` attributes is being authored on the client by a Javascript library, and at least one in four is inaccurate enough that the error is likely to impact resource selection. Both of these facts represent significant opportunities for either <a hreflang="en" href="https://github.com/ausi/respimagelint">existing tooling</a> or <a hreflang="en" href="https://github.com/whatwg/html/issues/4654">new web platform features</a> to help more authors get `sizes` right.
+
+##### `<picture>`
+
+The `<picture>` element serves a couple of use cases:
+
+1. Art direction, with the [`media` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture#the_media_attribute)
+2. Type-switching, via the [`type` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture#the_type_attribute)
```
tunetheweb, line 533: Should this be format switching instead of type switching?

```suggestion
2. Format-switching, based on mime-type, via the [`type` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture#the_type_attribute)
```

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+{{ figure_markup(
+  caption="The percentage of pages which use `<picture>`.",
+  content="6%",
+  classes="big-number",
+  sheets_gid="620965569",
+  sql_file="picture_distribution.sql"
+)
+}}
+
+`<picture>` is used much less frequently than `srcset`, reflecting its somewhat-niche pair of use-cases. Here's how usage breaks down between those two use cases:
```
tunetheweb, line 544: Oi! I like `<picture>` ;-)

But seriously those isn't format switching a non-niche case? especially with the number of new formats springing up and varying browser support?

Or do most hold off from that or handle it a different way?

Curious on your thoughts.

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+  image="picture-feature-usage.png",
+  caption="`<picture>` feature usage.",
+  description="A bar chart showing the percent of pages which use the `media` and `type` attributes on `source` elements, in conjunction with the `picture` element. `media` is used with 48% of `picture` elements on mobile, and just over 44% on desktop. `type` is used with 37% of `picture` elements on mobile, and close to 39% of `picture` elements on desktop.",
+  chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vQM9deg869BD9knNdVhFNbFnUdVXeyuwzUEIgSW-2XgOBEbALtVnoFapQ5JsDxzzepj6mVoepKBmN_m/pubchart?oid=1613173002&format=interactive",
+  sheets_gid="2031063502",
+  sql_file="picture_switching.sql"
+  )
+}}
+
+Art direction appears a bit more popular than type-switching. Notably, these two features are not mutually exclusive, and, put together, the usage percentages here do not add up to 100%. This suggests that at least 15% of `<picture>` elements do not leverage either of these attributes, making those `<picture>`s  functionally equivalent to a `<span>`.
```
tunetheweb, line 556: That's interesting! Did you look at any examples to see if any ideas WHY they did this?

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+### Layout
+
+Once you've embedded an image on a page, you must lay it out amongst the rest of the page's contents. There are many, many ways to do this, but we can derive a few insights about how it's generally done by zooming out and answering a couple of big-picture questions.
+
+#### Intrinsic vs extrinsic sizing
+
+As [replaced elements](https://developer.mozilla.org/en-US/docs/Web/CSS/Replaced_element), images have a natural, ["intrinsic" size](https://developer.mozilla.org/en-US/docs/Glossary/Intrinsic_Size). This is the size that they will render at by default, if there are no CSS rules placing "extrinsic" layout constraints upon them.
+
+How many images are extrinsically vs extrinsically sized?
```
tunetheweb, line 566: Is this correct or did you mean:

```suggestion
How many images are intrinsically vs extrinsically sized?
```

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+{{ figure_markup(
+  image="intrinsic-and-extrinsic-image-sizing.png",
+  caption="Intrinsic and extrinsic image sizing.",
+  description="A stacked bar chart showing the percentage of images whose width and height are determined based on extrinsic vs extrinsic sizing. There is a (for now) mysterious third category: both. The distributions between intrinsic, extrinsic, and both are quite different for height and width. For height, 60% of images are intrinsically sized, and 30% extrinsic; the remainder fall under both. For width, just 10% of images are intrinsically sized, and 66% are extrinsic; the remaining 24% are both.",
+  chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vQM9deg869BD9knNdVhFNbFnUdVXeyuwzUEIgSW-2XgOBEbALtVnoFapQ5JsDxzzepj6mVoepKBmN_m/pubchart?oid=769042260&format=interactive",
+  sheets_gid="576119731",
+  sql_file="image_sizing_extrinsic_intrinsic.sql"
+  )
+}}
```
tunetheweb, line 576: Again maybe don't use red unless you mean bad?

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+  sheets_gid="1864781514",
+  sql_file="video_durations_mobile.sql"
+  )
+}}
+
+Most videos on the web are short: ~ 60% of videos are under 30 seconds long on both mobile and desktop. However, 12-13% are between one and two minutes, and 10% of videos are over two minutes long.
+
+### Video: formats
+
+What types of files are being delivered as video?  We queried all files with "video" in the MIME type, and then sorted by the file extension.
```
tunetheweb, line 655:
```suggestion
What types of files are being delivered as video?  We queried all files with `video` in the MIME type, and then sorted by the file extension.
```

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+  chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vQM9deg869BD9knNdVhFNbFnUdVXeyuwzUEIgSW-2XgOBEbALtVnoFapQ5JsDxzzepj6mVoepKBmN_m/pubchart?oid=2066681624&format=interactive",
+  sheets_gid="1430902287",
+  sql_file="video_ext.sql"
+  )
+}}
+
+By far, the #1 video format on the web is the mp4, since the <a hreflang="en" href="https://caniuse.com/mpeg4">mp4 h264 format has 98.4% support in all modern browsers</a>, and the 1.9% of browsers that do not support mp4 have no video support, so the number is really 100% coverage. Interestingly, the mp4 usage has dropped by ~15% YOY on both desktop and mobile. WebM support also dropped significantly YOY (50% drop on both mobile and desktop). (data from [Web Almanac 2020](../2020/media#videos)
```
tunetheweb, line 669:
```suggestion
By far, the #1 video format on the web is the `mp4` (or MPEG-4), since the <a hreflang="en" href="https://caniuse.com/mpeg4">mp4 h264 format has 98.4% support in all modern browsers</a>, and the 1.9% of browsers that do not support `mp4` have no video support, so the number is really 100% coverage. Interestingly, the mp4 usage has dropped by ~15% YOY on both desktop and mobile. WebM support also [dropped significantly YOY](../2020/media#videos) (50% drop on both mobile and desktop).
```

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+Where we see the growth are files with no extension (these are often from YouTube or other streaming platforms), and in web streaming. Ts files are segments used in HTTP Live Streaming (HLS)  where we see a 4% jump in usage. .m4s are MPEG Dynamic Adaptive Streaming over HTTP (MPEG-DASH) video segments. M4s files grew by 50% from 2.3% to 3.3% YOY.
```
tunetheweb, line 671:
```suggestion
Where we see the growth are files with no extension (these are often from YouTube or other streaming platforms), and in web streaming. `ts` files are segments used in HTTP Live Streaming (HLS)  where we see a 4% jump in usage. `.m4s` are MPEG Dynamic Adaptive Streaming over HTTP (MPEG-DASH) video segments. M4s files grew by 50% from 2.3% to 3.3% YOY.
```

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+### Video CSS: `display`
+
+To begin, let's look at how the video will appear on the page by looking at the CSS `display` property for the video. What we find is that approximately half of all videos use a display value of `block`—placing the video on its own line and allowing for height and width values to be set for the video. The `inline-block` value also  allows height and width to be specified—for a total of two third's of all videos.
```
tunetheweb, line 675:
```suggestion
To begin, let's look at how the video will appear on the page by looking at the CSS `display` property for the video. What we find is that approximately half of all videos use a display value of `block`—placing the video on its own line and allowing for height and width values to be set for the video. The `inline-block` value also allows height and width to be specified—for a total of two third's of all videos.
```

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+{{ figure_markup(
+  image="video-element-attributes.png",
+  caption="Video element attributes.",
+  description="A bar chart showing the number of times various attributes on the HTML video element were found, on both desktop and mobile. With the exception of `controls`, which is slightly more common on mobile, every attribute is more common on desktop by 5-30%. The tallest pair of bars is `preload`; `buffered`s are the shortest. None of the bars are labelled, but their desktop numbers are approximately: `autoplay`: 100,000; `buffered`: 0; `controls`: 45,000; `controllist`: 1,000; `crossorigin`: 30,000; `height`: 3,000; `loop`: 50,000; `muted`: 55,000; `playsinline`: 52,000; `poster`: 24,000; `preload`: 115,000; `src`: 90,000; width: 60,000.",
+  chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vQM9deg869BD9knNdVhFNbFnUdVXeyuwzUEIgSW-2XgOBEbALtVnoFapQ5JsDxzzepj6mVoepKBmN_m/pubchart?oid=1099468813&format=interactive",
+  sheets_gid="1827909463",
+  sql_file="video_attribute_count_mobile.sql"
+  )
+}}
```
tunetheweb, line 703: Can we change these to percentage of videso? The raw numbers are kind of meaningless without context and less easily compared year on year.

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+#### Preload
+
+The most commonly used attribute is preload. The preload attribute gives the browser a hint on the best way to handle the video download. There are four possible options: `auto`, `metadata`, `none`, and an empty response = `auto`.
```
tunetheweb, line 707:
```suggestion
The most commonly used attribute is preload. The preload attribute gives the browser a hint on the best way to handle the video download. There are four possible options: `auto`, `metadata`, `none`, and an empty response (which uses the default of `auto`).
```

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+  chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vQM9deg869BD9knNdVhFNbFnUdVXeyuwzUEIgSW-2XgOBEbALtVnoFapQ5JsDxzzepj6mVoepKBmN_m/pubchart?oid=1322480946&format=interactive",
+  sheets_gid="7133313",
+  sql_file="video_attribute_count_mobile.sql"
+  )
+}}
+
+Interestingly, we see a large push away from `preload=auto` on both mobile and desktop, with most of the movement to `preload=none.`  While it is possible that this changed for many videos, it could just be that the new videos added to the web over the last year utilize the "none" parameter more than in the past. From a page weight perspective—this is a large win for the web.
```
tunetheweb, line 719:
```suggestion
Interestingly, we see a large push away from `preload=auto` on both mobile and desktop, with most of the movement to `preload=none.`  While it is possible that this changed for many videos, it could just be that the new videos added to the web over the last year utilize the `none` parameter more than in the past. From a page weight perspective—this is a large win for the web.
```

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+#### Autoplay
+
+The next most commonly used attribute is "autoplay."  This tells the browser that the video should play as soon as possible (and because of this, autoplay will actually override the preload attribute).
```
tunetheweb, line 723:
```suggestion
The next most commonly used attribute is `autoplay`.  This tells the browser that the video should play as soon as possible (and because of this, autoplay will actually override the preload attribute).
```

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+  image="video-element-attributes.png",
+  caption="Video element attributes.",
+  description="A bar chart showing the number of times various attributes on the HTML video element were found, on both desktop and mobile. With the exception of `controls`, which is slightly more common on mobile, every attribute is more common on desktop by 5-30%. The tallest pair of bars is `preload`; `buffered`s are the shortest. None of the bars are labelled, but their desktop numbers are approximately: `autoplay`: 100,000; `buffered`: 0; `controls`: 45,000; `controllist`: 1,000; `crossorigin`: 30,000; `height`: 3,000; `loop`: 50,000; `muted`: 55,000; `playsinline`: 52,000; `poster`: 24,000; `preload`: 115,000; `src`: 90,000; width: 60,000.",
+  chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vQM9deg869BD9knNdVhFNbFnUdVXeyuwzUEIgSW-2XgOBEbALtVnoFapQ5JsDxzzepj6mVoepKBmN_m/pubchart?oid=1099468813&format=interactive",
+  sheets_gid="1827909463",
+  sql_file="video_attribute_count_mobile.sql"
+  )
+}}
+
+#### Preload
```
tunetheweb, line 705:
```suggestion
#### `preload`
```

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+  chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vQM9deg869BD9knNdVhFNbFnUdVXeyuwzUEIgSW-2XgOBEbALtVnoFapQ5JsDxzzepj6mVoepKBmN_m/pubchart?oid=1322480946&format=interactive",
+  sheets_gid="7133313",
+  sql_file="video_attribute_count_mobile.sql"
+  )
+}}
+
+Interestingly, we see a large push away from `preload=auto` on both mobile and desktop, with most of the movement to `preload=none.`  While it is possible that this changed for many videos, it could just be that the new videos added to the web over the last year utilize the "none" parameter more than in the past. From a page weight perspective—this is a large win for the web.
+
+#### Autoplay
```
tunetheweb, line 721:
```suggestion
#### `autoplay`
```

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+The next most commonly used attribute is "autoplay."  This tells the browser that the video should play as soon as possible (and because of this, autoplay will actually override the preload attribute).
+
+The autoplay attribute is a boolean attribute, meaning that its presence by fault means true. So for the 190 sites that use `autoplay="false"`, we're sorry to tell you that is not going to work.
+
+#### Width
```
tunetheweb, line 727:
```suggestion
#### `width`
```

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+The width attribute is also one of the top video attributes. It tells the browser how wide the video player should be (note that height is very rarely used, since the aspect ratio of the video will decide the space with just one value.)
```
tunetheweb, line 729:
```suggestion
The `width` attribute is also one of the top video attributes. It tells the browser how wide the video player should be (note that `height` is very rarely used, since the aspect ratio of the video will decide the space with just one value.)
```

Actually I discovered recently that this is not the case! And unless you override with CSS a default aspect-ration of 2:1 will be used: https://github.com/whatwg/html/issues/3090

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+The width can be presented as a percentage, or a width in pixels.
+
+* When a percentage width is defined, the value "100%" is used in 99% of cases.
```
tunetheweb, line 733:
```suggestion
* When a percentage width is defined, the value `100%` is used in 99% of cases.
```

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+{{ figure_markup(
+  image="video-widths.png",
+  caption="Video widths.",
+  description="A bar chart showing the distribution of width attribute values. The values for desktop and mobile are very similar throughout most of the range, until the end. Starting at the left: values between 100-200 pixels were used ~800 times. As we go wider, there's a sharp increase, as values between 200-300 and 300-400 are both used around 4,000 times. After that, there's a slow, steady decline for a while, with a large exception in the 600-700 pixel range. This range has far more usage than any other: 7,000 uses. There's another much smaller bump for the 1,200-1,300 range, which looks about twice as high as its preceeding neighbor, measuring 1,200 uses. By the time we hit values of 1700-1800, usage looks close to zero, but then it shoots back up again. On desktop, values between 1800-1900 are found more than 4,000 times, whereas on mobile this range accounts for 2,400 uses. Finally, values between 1900-2000 are also fairly common (and more common on desktop than they are on mobile): 2,800 uses on desktop, and 600 uses on mobile.",
```
tunetheweb, line 739:
```suggestion
  description="A bar chart showing the distribution of `width` attribute values. The values for desktop and mobile are very similar throughout most of the range, until the end. Starting at the left: values between 100-200 pixels were used ~800 times. As we go wider, there's a sharp increase, as values between 200-300 and 300-400 are both used around 4,000 times. After that, there's a slow, steady decline for a while, with a large exception in the 600-700 pixel range. This range has far more usage than any other: 7,000 uses. There's another much smaller bump for the 1,200-1,300 range, which looks about twice as high as its preceeding neighbor, measuring 1,200 uses. By the time we hit values of 1700-1800, usage looks close to zero, but then it shoots back up again. On desktop, values between 1800-1900 are found more than 4,000 times, whereas on mobile this range accounts for 2,400 uses. Finally, values between 1900-2000 are also fairly common (and more common on desktop than they are on mobile): 2,800 uses on desktop, and 600 uses on mobile.",
```

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+  <figcaption>{{ figure_link(caption="The most common ordered triplets of `type` values, when there are three `source` elements within a `video` element.", sheets_gid="1549760253", sql_file="video_source_formats_mobile.sql") }}</figcaption>
+</figure>
+
+{# TODO I'm not sure this is such an interesting table. Consider adding some stats to make it more interesting? % of usage on mobile and desktop? #}
+
+Of course, these implementations will just play the MP4 file, and the WebM and Ogg files will be ignored.
+
+The incorporation of video on the web has grown immensely over the last year—jumping from 1-2% of webpages to 4-5%. We expect this growth to continue. Interestingly, the "king of video", MP4, while still the king, is having its market share eroded by video streaming formats (that feature responsive and adaptive video sizing).
+
+We do see movement to more performant usage of the `<video>` tag—with less use of `preload=auto`—and more use of `preload=none` as well as we see behaviors in the width attribute that indicate that videos are being modified (or removed) for smaller screens.
```
tunetheweb, line 868:
```suggestion
We do see movement to more performant usage of the `<video>` tag—with less use of `preload=auto`—and more use of `preload=none` as well as we see behaviors in the `width` attribute that indicate that videos are being modified (or removed) for smaller screens.
```

tunetheweb (commented):
Well that was a fascinating read! Well done both. Really strong chapter and looking forward to launching!

I do have some feedback. I also have done a fully copy edit, and spelling and grammar check but will merge this first after you review the feedback and do that after. @eeeps just to warn you I might cut back on some of your use of emphasis and complex sentences with many clauses with semi-colons, em-dashes and parenthesis! But we'll worry about that later.

In the meantime if you could review the feedback below, and the previous ones on the video tables then I think we're good to merge this.

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+description: Media chapter of the 2021 Web Almanac covering how images and videos are currently being encoded, embedded, styled, and delivered on the web
 authors: [eeeps, dougsillars]
-reviewers: []
-analysts: [eeeps, dougsillars]
+reviewers: [Navaneeth-akam, tpiros, akshay-ranganath, addyosmani]
+analysts: [eeeps, dougsillars, akshay-ranganath]
 editors: []
 translators: []
+eeeps_bio: Eric Portis is a Web Platform Advocate at <a hreflang="en" href="https://cloudinary.com/">Cloudinary</a>.
+dougsillars_bio: Doug Sillars is an leader in developer relations, and a digital nomad working on the intersection of performance and media. He tweets [@dougsillars](https://twitter.com/dougsillars), and blogs regularly at <a hreflang="en" href="https://dougsillars.com">dougsillars.com</a>.
```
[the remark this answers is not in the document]
dougsillars: lgtm

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+Every image on the web is passed around as a file. What can we say out about the files themselves, and the image data that they contain?
+
+#### Dimensions
+
+Let’s start by looking at their pixel dimensions. We'll start small.
+
+##### Single-pixel images
+
+Many `<img>` elements don't actually represent contentful <a hreflang="en" href="https://www.merriam-webster.com/dictionary/image">images</a>; instead, they contain a only a single–likely invisible–pixel:
```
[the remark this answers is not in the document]
eeeps: Sure! The nesting came from my outline/organizing my thoughts, but in a prose piece, I think it's fine to make some of the empty-ish shells-of-headings implicit.

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+      <tr>
+        <td>Desktop</td>
+        <td>47.1%</td>
+      </tr>
+    </tbody>
+  </table>
+  <figcaption>{{ figure_link(caption="Data URI single pixel images.", sheets_gid="1851007461", sql_file="image_1x1.sql") }}</figcaption>
+</figure>
+
+The single-pixel `<img>`s containing data URIs are almost certainly being used for layout. The remaining ~54% which generate a request might be there for layout or they might be tracking pixels; we can't tell.
```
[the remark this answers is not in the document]
eeeps: Good point.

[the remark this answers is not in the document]
eeeps: Added a link (to one of the original resources on this topic!)

src/content/en/2021/media.md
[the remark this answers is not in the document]
eeeps: Re-orged...it still gets really deep during the responsive images section, though that perhaps needs it?

eeeps:
@tunetheweb I welcome all copy editing! I know I often put voice and cadence over clarity, to a fault.

Accepted your changes and addressed your comments. And with that I'm offline until Sunday.

Thank you!

tunetheweb:
> Accepted your changes and addressed your comments. And with that I'm offline until Sunday.

Not sure you saw all the chances as some still open?

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+Note that throughout the rest of this analysis, we have excluded single-pixel `<img>`s from the results. We're interested in `<img>` elements that are presenting visual information to the user, not tracking pixels or layout hacks.
+
+##### Multiple-pixel images
+
+When `<img>`s contain more than one pixel–how many pixels do they contain?
+
+{{ figure_markup(
+  image="image-pixel-counts.png",
+  caption="Distribution of image pixel counts.",
+  description="Bar chart showing the distribution of megapixels per image. Both the desktop and mobile bars are shown, but there is little difference between them, and only the mobile bars are labeled. At the 10th percentile, images contain 0.001 megapixels; at the 25th, 0.009; at the 50th, 0.043; at the 75th, 0.170, and at the 90th, 0.514.",
```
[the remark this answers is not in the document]
eeeps: I went back and forth on this. Kilopixels make the nicest numbers but no one has ever heard of a kilopixel. Could be convinced to switch to pixels (to match the prose discussion).

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+{{ figure_markup(
+  image="image-orientations.png",
+  caption="Image orientations.",
+  description="A pair of stacked bars showing what percentage of images are portrait-oriented, landscape-oriented, and square, on both desktop and mobile. The desktop breakdown is: 12.7% portrait, 32.9% square, and 54.4% landscape. The mobile breakdown is very similar: 12.6% portrait, 32.7% square, and 54.5% landscape.",
+  chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vQM9deg869BD9knNdVhFNbFnUdVXeyuwzUEIgSW-2XgOBEbALtVnoFapQ5JsDxzzepj6mVoepKBmN_m/pubchart?oid=1361616235&format=interactive",
+  sheets_gid="478089289",
+  sql_file="portrait_landscape_square.sql"
+  )
+}}
```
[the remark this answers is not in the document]
eeeps: Is there a suggested color pallete for three non-judgemental categories? Or should I use my judgement?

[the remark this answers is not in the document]
tunetheweb: Here's our standard colours:

Pale green - #A8CABA #A8CABA - this is first colour used for desktop stats
Dark grey - #62718B #62718B - this is the second colour used for mobile stats. I think of it more as our primary colour over the green, despite it being the second colour in our charts as we normally highlight mobile more than desktop.
Brown - #4A3244 #4A3244 - this one is quite dark so tend not to use it
Pale yellow - #FFE599 #FFE599
Pale blue - #9FC5E8 #9FC5E8
Pale orange - #F6B26B #F6B26B
Pale pink - #EA9999 #EA9999
Yellow - #FFD966 #FFD966
Blue - #6D9EEB #6D9EEB
Green - #93C47D #93C47D
Orange - #E69138 #E69138

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+  description="Bar chart showing the distribution of megapixels per image. Both the desktop and mobile bars are shown, but there is little difference between them, and only the mobile bars are labeled. At the 10th percentile, images contain 0.001 megapixels; at the 25th, 0.009; at the 50th, 0.043; at the 75th, 0.170, and at the 90th, 0.514.",
+  chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vQM9deg869BD9knNdVhFNbFnUdVXeyuwzUEIgSW-2XgOBEbALtVnoFapQ5JsDxzzepj6mVoepKBmN_m/pubchart?oid=493015352&format=interactive",
+  sheets_gid="1999710461",
+  sql_file="bytes_and_dimensions.sql"
+  )
+}}
+
+The median `<img>` loads ~40,000 pixels on both desktop and mobile. I found this number surprisingly small. Just under half of crawled`<img>`s (excluding the ones that loaded single-pixel images, or nothing at all) contain fewer pixels than a 200x200 image. When you consider the number of `<img>` elements per page, though, this statistic is less surprising. Most pages contain more than 15 images, so while images with more than half-a-megapixel might only account for one in ten `<img>` elements, they are not at all uncommon, as we navigate across pages.
```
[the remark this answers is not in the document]
eeeps: Yes, I re-wrote this a few times and I think that showed. Your suggestion spells things out nicer.

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+      </tr>
+      <tr>
+        <td>8:5</td>
+        <td class="numeric">0.5%</td>
+        <td class="numeric">0.5%</td>
+      </tr>
+    </tbody>
+  </table>
+  <figcaption>{{ figure_link(caption="A ranked list of the top ten image aspect ratios.", sheets_gid="914161583", sql_file="top_aspect_ratios.sql") }}</figcaption>
+</figure>
```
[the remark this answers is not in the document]
eeeps: I prefered it as a table. The relative difference between the ten was less important than the dominance of 1:1, and a list of the top 10 - a table seemed to convey that better.

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+{{ figure_markup(
+  image="distribution-of-image-byte-sizes.png",
+  caption="Distribution of image byte sizes.",
+  description="A bar chart showing the distribution of image byte sizes on both desktop and mobile. At the tenth percentile, both the desktop and mobile bars are so small they're effectively invisible; the place where the mobile bar would otherwise be is labeled 252 (bytes). At the 25th percentile both bars are visible and tiny; the mobile bar is labeled 1,902. At the 50th percentile, both bars still look about the same and are still rather short; the mobile bar is labeled 10,290. At the seventy fifth percentile, the mobile bar is slightly taller than the desktop one, and is labelled 41,590. Finally, the 90th percentile bars tower over the rest. The desktop bar is slightly ahead this time; the mobile bar is labelled 130,662.",
+  chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vQM9deg869BD9knNdVhFNbFnUdVXeyuwzUEIgSW-2XgOBEbALtVnoFapQ5JsDxzzepj6mVoepKBmN_m/pubchart?oid=717078449&format=interactive",
+  sheets_gid="1999710461",
+  sql_file="bytes_and_dimensions.sql"
+  )
+}}
```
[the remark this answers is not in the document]
eeeps: Yes, kilobytes would probably be better. I'll change it!

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+First: tooling. JPEG encoders vary incredibly widely, ranging from hardware encoders in cameras which don't spend much effort compressing images well, to ancient copies of <a hreflang="en" href="https://en.wikipedia.org/wiki/Libjpeg">libjpeg</a> installed decades ago, to bleeding-edge, best-practice-by-default encoders like MozJPEG. In short, there are a lot of old, badly-compressed JPEGs out there, but every WebP and AVIF has been compressed with modern tooling.
+
+Also, anecdotally, <a hreflang="en" href="https://developers.google.com/speed/webp/download">cwebp</a> is relatively aggressive about quality/compression, and returns lower-quality, more-compressed results by default than most common JPEG tooling.
```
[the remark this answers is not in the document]
eeeps: I think cwebp might be the only WebP encoder. In any case it's the reference implementation: I'll call it that as it it conveys its importance a bit more than "one of" here.

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+#### Accessibility
+
+When you embed contentful images on webpages, you should make their content as accessible as possible for non-visual users. I note this only to [defer you to the Accessibility chapter](./accessibility#images), whose in-depth analysis of image accessibility on the web found small year-over-year progress in this area, but mostly: a whole lot of room for improvement.
+
+#### Responsive images
+
+In 2013, a suite of features enabling adaptive image loading on responsive websites landed, to much fanfare. Eight years in, how are responsive image features being used?
+
+##### Srcset
```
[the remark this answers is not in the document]
eeeps: Ok

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+{{ figure_markup(
+  caption="Percent of pages that use `srcset`.",
+  content="31%",
+  classes="big-number",
+  sheets_gid="1594311632",
+  sql_file="image_srcset_sizes.sql"
+)
+}}
+
+And `w` descriptors, which allow browsers to select a resource based on both varying layout widths and varying screen densities, are four times more popular than `x` descriptors, which only enable DPR-adaptation.
```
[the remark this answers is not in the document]
eeeps: I'll add a link!

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+  chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vQM9deg869BD9knNdVhFNbFnUdVXeyuwzUEIgSW-2XgOBEbALtVnoFapQ5JsDxzzepj6mVoepKBmN_m/pubchart?oid=403058700&format=interactive",
+  sheets_gid="1410495845",
+  sql_file="image_srcset_densities.sql"
+  )
+}}
+
+The mobile crawler here saw higher densities than the desktop crawler, which is expected. Viewport-relative `sizes` values resolve to smaller values on mobile viewports, resulting in higher densities for the same resources. Taken as a whole, [given that most devicePixelRatios fall somewhere between 1x-3x](https://twitter.com/TheRealNooshu/status/1397862141894529027), this appears to be a healthy range of densities. It would be interesting to perform a deeper analysis that counted how many `srcsets` _didn't_ fully cover a "reasonable" ~1x-2x range; this is left as an exercise to the reader (or next year's analysts!).
```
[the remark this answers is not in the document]
eeeps: Right, this is crawler relative. See above about deleting this section.

src/content/en/2021/media.md
```diff
@@ -1,20 +1,875 @@
...
+
+{{ figure_markup(
+  image="distribution-of-image-srcset-candidate-densities.png",
+  caption="Distribution of image srcset candidate densities.",
+  description="A bar chart showing the distribution of image srcset candidate densities, on both desktop and mobile. Only the mobile bars (which are always higher) are labelled. At the tenth percentile, mobile candidates have a density of 0.7x; the desktop bar appears to be about 0.4x. At the 25th, 1.0x on mobile, and close to 0.8x on desktop. At the median (50th percentile), candidates on mobile have a density of 1.5x; the desktop density appears to be about 1x. At the 7th percentile, the mobile density is 2.7x and the desktop density appears to be about 2.0x. Lastly, at the 90th percentile, mobile candidates had 4.3x and desktop candidates were around 3.15x.",
+  chart_url="https://docs.google.com/spreadsheets/d/e/2PACX-1vQM9deg869BD9knNdVhFNbFnUdVXeyuwzUEIgSW-2XgOBEbALtVnoFapQ5JsDxzzepj6mVoepKBmN_m/pubchart?oid=403058700&format=interactive",
+  sheets_gid="1410495845",
+  sql_file="image_srcset_densities.sql"
+  )
+}}
```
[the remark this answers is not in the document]
eeeps: Every candidate has a density. That density possibly changes with context (resolved sizes value). This is a chart of the distribution of all of those densities, in each context.

This whole section requires a rather deep understanding of the srcset mechanism to understand, and, as I point out at the end, isn't really even as useful as it could be. Another iteration or two of analysis would have gotten me to a clear conclusion (srcsets are/aren't covering the ranges they need to). This is a step short, in a way that takes a lot of effort to explain/understand.

What would you think about deleting this whole section, rather than leading folks through a long-winded explanation of the srcset process, only to get to a conclusion that I wish was stronger?

src/content/en/2021/media.md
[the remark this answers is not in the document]
eeeps: As long as I'm making changes to other charts, I'll change this too.

src/content/en/2021/media.md
eeeps: You’re right that it's a feature with a lot of potential benefit, that's perhaps underutilized. Art direction, too (see the above aspect-ratio-on-mobile discussion).

I’ll re-work my language in order to say that there’s a lot of potential (but underutilized) utility, here. Rather than making it sound like there’s not much need. "Niche use case" isn't quite right.

A common alternative way to do format switching is with server-side content negotiation. I’ll work that in as well, (but have no data on how common it is — very hard to gather).

src/content/en/2021/media.md
eeeps: If I had to guess I’d say as apart of a template or as a styling hook, but no, I did not look at specific cases. Worth speculating here (or taking the time to investigate)?

src/content/en/2021/media.md
eeeps: Ok!

src/content/en/2021/media.md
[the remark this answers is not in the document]
eeeps: Done

src/content/en/2021/media.md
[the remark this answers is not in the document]
eeeps: Added a few.

src/content/en/2021/media.md
eeeps: Applied these edits.

src/content/en/2021/media.md
eeeps: Applied

src/content/en/2021/media.md
eeeps: I think this is a good idea. Thoughts, @dougsillars?

src/content/en/2021/media.md
eeeps: Applied

src/content/en/2021/media.md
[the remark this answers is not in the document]
eeeps: Done

src/content/en/2021/media.md
eeeps: Thoughts, @dougsillars?

eeeps:
@tunetheweb oops, I missed the "30 hidden conversations" Github collapsed in the middle of your review!

Stole some time today to address all of the easy ones in the Images section. I'll try to make as much headway on the meatier things as I can over the next couple of days before I'm fully back on Monday.

I'll ping @dougsillars about your proposed changes to Video.

Thanks again @tunetheweb !! Extremely valuable feedback! Should we credit you as editor?

src/content/en/2021/media.md
dougsillars: looks good to me

src/content/en/2021/media.md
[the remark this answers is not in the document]
eeeps: Updated the chart in the sheet and the description. Embeds appear to have uploaded. I'll try my hand at running the pupetteer script once updates to the other charts are completed

src/content/en/2021/media.md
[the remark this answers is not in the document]
eeeps: Done.

src/content/en/2021/media.md
eeeps: Done

src/content/en/2021/media.md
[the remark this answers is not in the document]
eeeps: Went with #A8CABA #62718B #FFE599 for all three-category charts

src/content/en/2021/media.md
eeeps: Done

src/content/en/2021/media.md
eeeps: @dougsillars We should absolutely use ticks around the attribute names, but I think @tunetheweb was calling the accuracy of your statement about aspect ratios into question. If they do indeed get a default aspect ratio, rather then filling out their natural aspect ratio, you should re-write this.

eeeps:
Ok, from my perspective, the final remaining tasks are:

- [ ] decide whether to delete the "srcset resource densities" section. I'm in favor, but @tunetheweb if you think there's valuable insight here that's worth clarifying, I'm game to give it another go
- [ ] possibly convert the video chart showing absolute counts to use percentages @dougsillars 
- [ ] possibly re-write the bit about default video aspect ratios @dougsillars 
- [ ] re-run the chart-PNG-generating Puppeteer script, and replace the updated charts in `src/static/images/2021/media/`

src/content/en/2021/media.md
tunetheweb: Added the stats in https://github.com/HTTPArchive/almanac.httparchive.org/pull/2705/commits/0b9689ecd8d3a77cfc6de9d8d4de3127933a90c1

src/content/en/2021/media.md
tunetheweb: Added the stats in https://github.com/HTTPArchive/almanac.httparchive.org/pull/2705/commits/0b9689ecd8d3a77cfc6de9d8d4de3127933a90c1

src/content/en/2021/media.md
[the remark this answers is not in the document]
tunetheweb: Ran the pupetteer script and tinypng'd the images

src/content/en/2021/media.md
tunetheweb: Changed to percentages.

tunetheweb:
OK I've taken care of a couple of those things and left TODOs in place for the other two. We can pick them up in editing PR.

@dougsillars I had rewrite some of your queries and also had to change some of your preload text as it looks like there wasn't a switch from auto but more a move away from using preload at all.

I've updated the staging link at the top if you want to have a look.

I'm now going to merge this now, then do a quick copy edit later today or tomorrow and open the PR and then, if you accept that, we're done and can launch!

PR #2720, opened 2021-12-13
Complete srcset resource densities section re-write

eeeps:
The old section was unintelligible to people without an intimate knowledge of how the srcset selection process works, and the analysis left the most important questions unanswered.

So I wrote a new query and re-wrote the section. If this is still unintelligible even to fairly technical folks, we should *definitely* just delete the section. I, personally, think it's much better than v1.


eeeps:
Moving to a PR against @tunetheweb's copy edit branch

PR #2726, opened 2021-12-14
A few nits

eeeps:
A few final (knock on wood) changes


eeeps:
fault → default is the only one that really, really, really needs to land

tunetheweb (approved):

PR #2730, opened 2021-12-15
Add Poppins Bold Italic fonts

VictorLeP:
My eagle eyes just _knew_ something was off :wink: 

Added the italic versions of the Poppins Bold fonts, as used in table headers. This means we do not fall back to the browser just slanting the regular version, which leads to uglier results.

Before:
poppins_italic_before

After:
poppins_italic_after


VictorLeP:
I only added the _bold_ italic versions for now, because lazy, and because I didn't see Poppins being used with any other font weights.

tunetheweb:
Fonts are too hard. Roll on variable fonts!

Thanks for this.

tunetheweb (approved):

PR #2740, opened 2021-12-18
CSS percentage correction

### LINKED ISSUE #2739: Issue with the 2021 css chapter
In the fragment

> While CSS-in-JS adoption grew to 3% of all pages crawled (a 1% jump from last year) [...]

you probably meant a _growth by 1 **percentage point**_, instead of _growth by 1%_. It's different.

GeekBoySupreme:
Updating the introduction to resolve #2739 
Replacing 1% with 1 percentage point.


tunetheweb:
@ksakae1216 I tried to add the same change in this PR to the Japanese translation using Google Translate. Could you check if I got it right?

ksakae1216 (approved):
Thanks, No problem.
LGTM!!

PR #2748, opened 2021-12-29
2021 SEO Chapter: fix rule in robots.txt

kartashovio:
I think this is a typo. User-agent set with `User-agent`, not with `Disallow`. [Documentation](https://developers.google.com/search/docs/advanced/robots/create-robots-txt).


tunetheweb (approved):
Agreed. At best it's a shortening of:

```
User-Agent: Googlebot
Disallow: /

User-Agent: *
Allow: /
```

But that's a bit of a mouthful to get into a sentence so agree with your proposed correction.

@patrickstox any concerns before I merge this?

PR #2787, opened 2022-01-26
2021 SEO Chapter: change figure description

kartashovio:
Invalid bar chart description. [Link to the bar chart](https://almanac.httparchive.org/en/2021/seo#image-loading-attributes).

`invalid` is not the value of the `loading` attribute, it is the case when an invalid attribute value was specified in the `img` tag. If you try searching the internet for `loading="invalid"`, you won't find anything.

Same with `loading="blank"`. In fact, we are talking about an empty attribute value.


tunetheweb (approved):
Agree. Thanks for that.

kartashovio:
@tunetheweb , thanks, it's interesting!

PR #2811, opened 2022-02-09
2021 SEO Chapter: typos

kartashovio:
1. Word «Глава», whats translating like «Chapter»
I see, that this is automatically added in description: 
image

2. Some typos


tunetheweb (approved):
LGTM.

The descreiption is set in `src/templates/ru/base_chapter.html` and is set to the following:

```jinja
{% block description %}{{ 'Глава ' + metadata.get('description',metadata.get('title') + ' в Web Almanac за ' + year + ' год разбирает использование ' + metadata.get('description',metadata.get('title')) + ' в вебе.') }}{% endblock %}
```

Let me know if you're OK with that or want to change that in this PR.

tunetheweb:
Actually that looks wrong.

For English we have the following:

```html
{% block description %}{{ metadata.get('description',metadata.get('title') + ' chapter of the ' + year + ' Web Almanac probing into the use of ' + metadata.get('description',metadata.get('title')) + ' on the web.') }}{% endblock %}
```

This  basically means this:

Use this by default:
```
metadata.get('description')
```
Or if that doesn't exist then this:
```
metadata.get('title') + ' chapter of the ' + year + ' Web Almanac probing into the use of ' + metadata.get('description',metadata.get('title')) + ' on the web.'
```

That is, we use the description only form the chapter, if it exists, and use only that, and only if it doesn't exist we fall back to an attempt to make up a description based on the title.

The Russian version doesn't look like it's doing that and instead if adding to the description that is set in the chapter. That is wrong.

So instead of that being:

```
{% block description %}{{ 'Глава ' + metadata.get('description',metadata.get('title') + ' в Web Almanac за ' + year + ' год разбирает использование ' + metadata.get('description',metadata.get('title')) + ' в вебе.') }}{% endblock %}
```

I think it should be:

```
{% block description %}{{metadata.get('description', 'Глава ' + metadata.get('title') + ' в Web Almanac за ' + year + ' год разбирает использование ' + metadata.get('description',metadata.get('title')) + ' в вебе.') }}{% endblock %}
```

Could you change that in this PR and revert your description change to add  'Глава '  back in to chapter description?

tunetheweb (approved):

tunetheweb:
Thanks for making that change. Will merge tests complete.

FYI, I will do a follow up PR for all the templates as the second `metadata.get('description',metadata.get('title'))` is a bit pointless (we know there's no `description` at this point or we wouldn't be in this section!). But as that's all templates leave that to me to sort out.

tunetheweb:
FYI, follow up PR: #2812

PR #2909, opened 2022-04-12
Update progress-tracker.yml

rviscomi:


.github/workflows/progress-tracker.yml
```diff
@@ -2,9 +2,9 @@ name: Progress Tracker
 
 env:
   # Update these environment variables every year
```
tunetheweb, line 4:
```suggestion
  # Update these environment variables every year
  # And check the tracker is enabled at: 
  # https://github.com/HTTPArchive/almanac.httparchive.org/actions/workflows/progress-tracker.yml
```

tunetheweb (approved):
LGTM.

Don't forget to re-enable the GitHub Action (I disable it at the end of every year to save it from needlessly running every time someone creates or updates and issue). Added a suggested comment.

PR #2918, opened 2022-04-23
Upgrade to SQLFluff 0.13.0 and add consistency check for quoted literals

tunetheweb:
We mostly use single quotes for literals, but not always. This PR changes to be consistent to prevent a new SQLFluff rule flagging (the other alternative is to turn off that rule, but think we should have consistency).

### DIFF
--- a/sql/.sqlfluff
+++ b/sql/.sqlfluff
@@ -130,3 +130,9 @@ blocked_words = None
 extended_capitalisation_policy = upper
 # Comma separated list of words to ignore for this rule
 ignore_words = None
+
+[sqlfluff:rules:L064]
+# Consistent usage of preferred quotes for quoted literals
+preferred_quoted_literal_style = single_quotes
+# Disabled for dialects that do not support single and double quotes for quoted literals (e.g. Postgres)
+force_enable = False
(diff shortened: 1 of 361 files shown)


rviscomi (commented):
I think there may be some valid cases to use double quotes, for example see JSON functions on fields with hyphens. Worth double checking that this doesn't break those types of queries.

rviscomi:
Actually it looks like the linter did a very nice job of keeping some valid double quote uses, like sql/2019/performance/07_20.sql

```sql
IFNULL(JSON_EXTRACT(payload, "$['_cpu.EvaluateScript']"), "0")
```

became:

```sql
IFNULL(JSON_EXTRACT(payload, "$['_cpu.EvaluateScript']"), '0')
```

rviscomi (approved):

tunetheweb:
Yeah it basically keeps any quotes within the quotes, and checks if it's shorter to escape or not:

```sql
IFNULL(JSON_EXTRACT(payload, "$['_cpu.EvaluateScript']"), '0')
IFNULL(JSON_EXTRACT(payload, '$[\'_cpu.EvaluateScript\']'), '0')
```

These should be equivalent but the first is shorter, and therefore easier to read, so it keeps that, even though it's not the preferred quoting style (single quotes).

There may well be some that has messed up, but I couldn't find any from some spot checks.

PR #2924, opened 2022-04-27
Ignore shortcut keys in search box

### LINKED ISSUE #2922: Previous/Next shortcut keys triggered in search box
We have a search box on our site.

We also have short cut keys to quickly skip through chapters:
- `p` goes to previous chapter
- `n` goes to the next chapter

Unfortunately these don’t play well together (on desktop at least) and you can’t search for a term including either of these characters on a chapter page without triggering the navigation.

We should capture key presses and prevent them bubbling up to the document when in the input.

Should be a quick one for any of the @HTTPArchive/developers wanting to get their first PR under their belt!

tunetheweb:
Fixes #2922

### DIFF
--- a/src/static/js/almanac.js
+++ b/src/static/js/almanac.js
@@ -700,9 +700,9 @@ function addShortKeyEventListers() {
     }
   });
 
-  // Prevent the search box from propogating key up events
+  // Prevent the search box and other inputs from propogating key up events
   // so the above are not activated when typing words with them.
-  document.querySelectorAll('.search-input').forEach(searchInput => {
+  document.querySelectorAll('input').forEach(searchInput => {
     searchInput.addEventListener("keyup", function onPress(event) {
       event.stopPropagation();
     });
--- a/src/static/js/almanac.js
+++ b/src/static/js/almanac.js
@@ -699,6 +699,14 @@ function addShortKeyEventListers() {
       });
     }
   });
+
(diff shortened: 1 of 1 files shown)


src/static/js/almanac.js
```diff
@@ -700,9 +700,9 @@ function addShortKeyEventListers() {
     }
   });
 
-  // Prevent the search box from propogating key up events
+  // Prevent the search box and other inputs from propogating key up events
   // so the above are not activated when typing words with them.
-  document.querySelectorAll('.search-input').forEach(searchInput => {
+  document.querySelectorAll('input').forEach(searchInput => {
     searchInput.addEventListener("keyup", function onPress(event) {
       event.stopPropagation();
```
rviscomi, line 707: Alternatively, WDYT about adding a check at the top of this event handler for something like this?

```js
if (isInputField(event.currentTarget)) {
  return;
}
```

tunetheweb: Will give it a go. Probably a bit more efficient - especially for our large DOM chapters!

Guessing I’ll need `target` though rather than `currentTarget`?

rviscomi: Maybe, I get them mixed up and coded that off the top of my head :)

tunetheweb: Done!

rviscomi (approved):

Issue #2883, opened 2022-04-12
Fonts 2022


rviscomi:
# Fonts 2022

![Fonts illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/fonts/hero_lg.jpg)

If you're interested in contributing to the Fonts chapter of the 2022 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| ---- | ------- | --------- | -------- | ------- | ----------- |
| @bramstein | @bramstein | @RoelN @svgeesus @jmsole @alexnj | @konfirmed | @shantsis | @siakaramalegos  |

<details>
<summary><b>Expand for more information about each role 👀</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
- **[Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)** are responsible for researching the stats and trends used throughout the Almanac. Analysts work closely with authors and reviewers during the planning phase to give direction on the types of stats that are possible from the dataset, and during the analyzing/writing phases to ensure that the stats are used correctly.
[...]

PR #2938, opened 2022-05-21
Fonts 2022 queries

konfirmed:
Progress on #2883 

## Usage

- [x] Usage of web fonts over time.
- [x] Usage of web fonts hosted vs. self-hosted.
- [x] Usage of hosted web fonts split out by service
- [x] Usage by country
- [x] Font Format Usage
- [x] Popular Fonts by foundry
- [x] Most popular foundries

## Color Fonts

- [x] Growth in color font usage since 2020
- [x] The popularity of color font formats 
- [x] Color font palettes (number of palettes and palette entries)
- [x] Types of color fonts: emoji (and other icon fonts) vs. display (or text) fonts.

## Variable Fonts

- [x] Growth in variable fonts served since their introduction.
- [x] Number of variable fonts served (non-Google Fonts)
- [x] Types of variable fonts served (CFF2/TT)
- [x] Registered axis usage vs custom axis usage
- [x] Animations including variable fonts (through font-feature-settings/width/weight)

## Performance
- [x] Use of priority hints
- [x] Usage of system generic font families
- [x] Font display usage
- [x] CSS Font Loading API usage

## Font data & Usage
- [x] Number, type, and complexity of OpenType features
- [x] Popularity of outline formats
- [x] Usage of font-feature-settings vs. font-variant-*
- [x] Quantified language and script support in the font
- [x] Are size-adjust and ascent-override used
- [x] How common are PUA characters
- [x] How common is usage of -webkit-font-smoothing?

## Optional
These queries would be nice to have but are not essential to the chapter.
- [ ] Are control characters used on a lot of web pages
- [ ] Are all the localisations in the font used on the page
- [ ] Does the font have optional features (and are they used on the page.)
- [x] Font sizes with and without services
- [x] Size of various OT tables (i.e. where does a font spend most of its bytes; does it differ by language/script or type of font?)
- [ ] Mixed usage of system fonts and web fonts on a page
- [ ] Axis location settings through font-variations-settings vs registered axis CSS properties
- [ ] Usage of font-feature-settings in @font-face vs. in general styles


siakaramalegos:
@konfirmed @bramstein How are things going? Do you think you can finish the queries and start making the charts before the end of the month? We want to keep the writing on track for publication.

bramstein:
@siakaramalegos I plan to put some more time into it next week.

rviscomi:
@konfirmed @bramstein is this PR ready for review?

bramstein:
@rviscomi Not yet. I hope to wrap up the last queries this week. Though feel free to review the ones that are already there.

rviscomi:
Thanks @bramstein. In that case I'd suggest opening the PR for review now, getting early feedback, and merging what we have so far. Then later in the week when the remaining queries are written, we can push the smaller PR through.

sql/2022/fonts/font_palettes_colors.sql
```diff
@@ -0,0 +1,35 @@
...
+  client,
+  entries,
+  pages,
+  total,
+  pages / total AS pct
+FROM (
+  SELECT
+    client,
+    JSON_EXTRACT_SCALAR(payload,
+      '$._font_details.color.numPaletteEntries') AS entries,
```
rviscomi, line 11: Same question. Use percentiles for high cardinality data.

sql/2022/fonts/hosted_vs_self_hosted.sql
```diff
@@ -0,0 +1,38 @@
+#standardSQL #self_hosted_vs_hosted
+SELECT
+  client,
+  CASE
+    WHEN pct_self_hosted_hosted = 1 THEN 'self-hosted'
+    WHEN pct_self_hosted_hosted = 0 THEN 'external'
+    ELSE 'both'
+  END AS font_host,
+  COUNT(DISTINCT page) AS pages,
+  SUM(COUNT(DISTINCT page)) OVER (PARTITION BY client) AS total,
```
rviscomi, line 10: Similar comment about the total
Not sure of the purpose of the join

sql/2022/fonts/outline_formats.sql
```diff
@@ -0,0 +1,34 @@
...
+  SELECT
+    _TABLE_SUFFIX AS client,
+    COUNT(0) AS total_page
+  FROM
+    `httparchive.pages.2022_06_01_*`
+  GROUP BY
+    _TABLE_SUFFIX)
+USING
+  (client),
+  UNNEST(REGEXP_EXTRACT_ALL(JSON_EXTRACT(payload, '$._font_details.table_sizes'), '(?i)(CFF |glyf|SVG|CFF2)')) AS format
```
rviscomi, line 28: Is the space after `CFF` intentional?

sql/2022/fonts/popular_typeface.sql
```diff
@@ -0,0 +1,48 @@
...
+FROM (
+  SELECT
+    client,
+    font_family,
+    COUNT(DISTINCT page) AS pages
+  FROM
+    `httparchive.almanac.parsed_css`,
+    UNNEST(getFontFamilies(css)) AS font_family
+  WHERE
+    date = '2022-06-01'
```
rviscomi, line 31: Need to use July for this table

sql/2022/fonts/popular_typeface.sql
```diff
@@ -0,0 +1,48 @@
...
+  GROUP BY
+    client,
+    font_family)
+JOIN (
+  SELECT
+    _TABLE_SUFFIX AS client,
+    COUNT(0) AS total
+  FROM
+    `httparchive.summary_pages.2022_06_01_*`
```
rviscomi, line 40: And use July here for consistency

sql/2022/fonts/usage_of_system_families.sql
```diff
@@ -0,0 +1,311 @@
+CREATE TEMPORARY FUNCTION getSystemFamilies(data STRING)
+RETURNS ARRAY < STRING > LANGUAGE js AS '''
+// Font Family parser taken from https://github.com/bramstein/css-font-parser (BSD license)
```
rviscomi, line 3: Oh this is cool! Let's add this library to [sql/lib](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/sql/lib) and publish it on GCS so we can include it in queries more easily. Similar to how we include CSS utils in many of the CSS queries ([example](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2022/css/all_properties.sql#L5)).

That would help decouple the analysis code from the library code in this query.

sql/2022/fonts/variable_font_usage.sql
```diff
@@ -0,0 +1,33 @@
...
+    _TABLE_SUFFIX)
+USING
+  (client)
+GROUP BY
+  client,
+  total_page
+HAVING
+  freq_vf > 100
+ORDER BY
+  freq_vf DESC
```
rviscomi, line 33: Are these needed? There are only 2 clients and even if one of them has very few pages with variable fonts, we'd want to see that reflected in the results rather than filtered out.

rviscomi (changes requested):
This looks great nice work @bramstein! A few minor comments.

bramstein:
I moved some of the queries to a nice-to-have list. I'll focus on wrapping up the queries that are essential to the chapter this week and hopefully get around to some of the optional ones as well.

bramstein:
@siakaramalegos @rviscomi I completed all the required queries and two optional ones. I think I'll leave it at this and focus on the data and writing now.

I'll go through the queries tomorrow and address the review notes.

sql/2022/fonts/usage_of_system_families.sql
bramstein: Oh, I didn't know that exists! I can add that in a separate PR. How do I deploy that library? It looks like there is a utility script to copy the files to a bucket, but I assume that requires credentials.

sql/2022/fonts/outline_formats.sql
bramstein: Yes, table tags in OpenType are always 4 characters. The table tags aren't trimmed in the font metadata extractor in wptagent, so it needs the trailing space.

sql/2022/fonts/font_writing_scripts.sql
```diff
@@ -0,0 +1,100 @@
+CREATE TEMP FUNCTION detect(codepoints ARRAY<STRING>)
+RETURNS ARRAY<STRING>
+LANGUAGE js AS r"""
+
+# This function and data taken from https://github.com/bramstein/detect-writing-script (BSD-3-Clause)
```
rviscomi, line 5: Similar to my other comment, util code like this could be side-loaded via GCS

sql/2022/fonts/usage_of_system_families.sql
rviscomi: One of the project maintainers would need to deploy it, but if you check the library into this PR I can stage it on GCS and show how to integrate it with the query.

sql/2022/fonts/font_writing_scripts.sql
bramstein: Whoops, good catch. Thanks!

sql/2022/fonts/font_size_quantiles_by_opentype_table.sql
```diff
@@ -0,0 +1,42 @@
...
+  COUNT(0) AS total,
+  APPROX_QUANTILES(bytes, 1000)[OFFSET(percentile * 10)] AS bytes
+FROM (
+  SELECT
+    client,
+    percentile,
+    key AS table,
+    value AS bytes
+  FROM
+    `httparchive.almanac.requests` TABLESAMPLE SYSTEM (0.1 PERCENT),
```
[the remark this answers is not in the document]
bramstein: Whoops, good catch. Thanks!

sql/lib/text-utils.js
```diff
@@ -0,0 +1,68 @@
+const writingScriptDetector = (() => {
```
bramstein, line 1: I named this `text-utils` because I can see myself adding a couple of other methods if I get around to writing the other optional queries. Happy to rename if it is too generic.

sql/2022/fonts/usage_of_system_families.sql
bramstein: I added both libraries to this PR (one for parsing font-family/font properties, and the other for detecting writing scripts.) Let me know how I can help with staging it on GCS.

rviscomi: Ok I've pushed the lib files onto GCS. IIUC there's still a bit of code needed to use the library, so the final query might look like the snippet below. All I've done is add `OPTIONS (library = "gs://httparchive/lib/css-font-parser.js")` and remove any boilerplate provided by the library.

```sql
CREATE TEMPORARY FUNCTION getSystemFamilies(data STRING)
RETURNS ARRAY < STRING > LANGUAGE js
OPTIONS (library = "gs://httparchive/lib/css-font-parser.js")
AS '''
function parseCss(rule, acc) {
  if (rule.hasOwnProperty('rules')) {
    rule.rules.forEach(rule => parseCss(rule, acc));
  }

  if (rule.hasOwnProperty('declarations')) {
    rule.declarations.forEach(decl => {
      if (decl.property === 'font-family') {
        const fonts = parseFontFamily(decl.value);

        if (fonts) {
          fonts.forEach(font => acc.push(font));
        }
      } else if (decl.property === 'font') {
        const value = parseFont(decl.value);

        if (value) {
          value['font-family'].forEach(font => acc.push(font));
        }
      }
    });
  }
}

const system = [
  'serif',
  'sans-serif',
  'cursive',
  'fantasy',
  'monospace',
  'system-ui',
  'emoji',
  'math',
  'fangsong',
  'ui-serif',
  'ui-sans-serif',
  'ui-monospace',
  'ui-rounded'
];

function getSystemFonts(json) {
  try {
    const css = JSON.parse(json);
    const result = [];

    parseCss(css.stylesheet, result);

    return result.filter(font => system.includes(font));
  } catch (e) {
    return [];
  }
}

return getSystemFonts(data);
''';

SELECT
  client,
  font_family,
  pages,
  total,
  pages / total AS pct
FROM (
  SELECT
    client,
    font_family,
    COUNT(DISTINCT page) AS pages
  FROM
    UNNEST(getSystemFamilies(css)) AS font_family
  WHERE
    date = '2022-07-01'
  GROUP BY
    client,
    font_family)
JOIN (
  SELECT
    _TABLE_SUFFIX AS client,
    COUNT(0) AS total
  FROM
    `httparchive.summary_pages.2022_07_01_*`
  GROUP BY
    client)
USING
  (client)
ORDER BY
  pct DESC
```

bramstein: Thanks, I'll update the queries to use the libraries.

sql/2022/fonts/hosted_vs_self_hosted.sql
bramstein: This file was copied over from 2020. It is replaced by the usage by services query, so I ended up just deleting this one.

sql/2022/fonts/font_palettes.sql
```diff
@@ -0,0 +1,35 @@
...
+  client,
+  num,
+  pages,
+  total,
+  pages / total AS pct
+FROM (
+  SELECT
+    client,
+    JSON_EXTRACT_SCALAR(payload,
+      '$._font_details.color.numPalettes') AS num,
```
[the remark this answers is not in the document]
bramstein: The number of palettes has low cardinality, so I just cast it to an integer and kept the rest of the query the same.

sql/2022/fonts/font_palettes_colors.sql
bramstein: This one also has low cardinality. I think I'll just mention that most fonts only have a single palette and most palettes have only a handful of colors. I simplified the queries a bit, but I don't think percentiles would be super helpful until there is a lot more cardinality.

bramstein:
@rviscomi I've addressed all comments. Let me know if you spot anything else or if the PR can be merged. Thanks!

sql/2022/fonts/font_palettes_colors.sql
```diff
@@ -1,35 +1,16 @@
...
-    COUNT(0) AS total
-  FROM
-    `httparchive.summary_pages.2022_06_01_*`
-  GROUP BY
-    client)
-USING
-  (client)
+  SAFE_CAST(JSON_EXTRACT_SCALAR(payload, '$._font_details.color.numPaletteEntries') AS INT64) AS entries,
+  COUNT(0) AS total,
+  COUNT(0) * 1.0 / SUM(COUNT(0)) OVER(PARTITION BY client) AS pct
```
rviscomi, line 5:
```suggestion
  COUNT(0) AS freq,
  SUM(COUNT(0)) OVER(PARTITION BY client) AS total,
  COUNT(0) / SUM(COUNT(0)) OVER(PARTITION BY client) AS pct
```

rviscomi (changes requested):
This is great! Just a couple of small suggestions

rviscomi (approved):

Issue #2903, opened 2022-04-12
HTTP 2022


rviscomi:
# HTTP 2022

![HTTP illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/http/hero_lg.jpg)

If you're interested in contributing to the HTTP chapter of the 2022 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| ---- | ------- | --------- | -------- | ------- | ----------- |
| @paivaspol | @paivaspol | @tunetheweb @rmarx @LPardue | @paivaspol | | @paulcalvano |

<details>
<summary><b>Expand for more information about each role 👀</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
- **[Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)** are responsible for researching the stats and trends used throughout the Almanac. Analysts work closely with authors and reviewers during the planning phase to give direction on the types of stats that are possible from the dataset, and during the analyzing/writing phases to ensure that the stats are used correctly.
[...]

PR #2939, opened 2022-05-22
HTTP 2022 queries

paivaspol:
This PR tracks the query progress for #2903 . All metrics will be computed in both absolute and relative (i.e., percentage) terms. They should also include breakdown by desktop and mobile.

# Adoption
Compare between 2021 and 2022:
- [x] (2022) % of websites using HTTP/2+ vs HTTP/1.1
- [x] (2022) Distribution of number and % of resources on HTTP/2 vs HTTP/1.1 broken down by whether the resource was requested via CDN.
- [x] (2022) Distribution of number and % of resources on HTTP/2 vs HTTP/1.1
- [x] (2021) % of websites using HTTP/2+ vs HTTP/1.1
- [ ] (2021) Distribution of number and % of resources on HTTP/2 vs HTTP/1.1
- [ ] (2021) Distribution of number and % of resources on HTTP/2 vs HTTP/1.1 broken down by whether the resource was requested via CDN.

# Benefits from HTTP/2

## Connections
- [x] Median number of connections per page load.
- [ ] Average number of connections per page load.
- [x] Distribution of connections across websites.

## Prioritization
- [x] Number of sites using priority hints. Custom metrics already [implemented](https://github.com/HTTPArchive/custom-metrics/blob/4c6bc0493b1b51dc021d6dcf7471b80224dc9228/dist/almanac.js#L215-L219).
- [x] Distribution of number of resources with priority hints.
- [x] _Not querying as most websites have not adopted priority hints._ Breakdown of priority hint levels per page. For example, CDF across websites of the number of resources with priority hints broken down in two lines: `high` and `low`. We should only execute this only if there is enough data for this analysis to make sense.-

## Push
- [x] Number of pushed resources
- [ ] Number of Link rel="preload" in HTTP response headers--signal for push for CDNs and proxies.
- [x] Number of sites using Early Hints. Sites with 103 responses.

# HTTP/3
- [ ] Breakdown of server side supported versions of protocols. Distribution and median value.

### DIFF
--- a/sql/2022/http/connections_per_page_load_dist.sql
+++ b/sql/2022/http/connections_per_page_load_dist.sql
@@ -1,0 +1,44 @@
+#standardSQL
+
+# Measure the distribution of TCP Connections per site.
+
+SELECT
+  percentile,
+  client,
+  http_version_category,
+  COUNT(0) AS num_pages,
+  APPROX_QUANTILES(_connections, 1000)[OFFSET(percentile * 10)] AS connections
+FROM (
+  SELECT
+    client,
+    page,
+    CASE
+      WHEN LOWER(protocol) = 'quic' OR LOWER(protocol) LIKE 'h3%' THEN 'HTTP/2+'
+      WHEN LOWER(protocol) = 'http/2' OR LOWER(protocol) = 'http/3' THEN 'HTTP/2+'
+      WHEN protocol IS NULL THEN 'Unknown'
+      ELSE 'Non-HTTP/2'
+    END AS http_version_category
+  FROM
+    `httparchive.almanac.requests`
+  WHERE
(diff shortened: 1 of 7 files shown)


paivaspol:
@rviscomi We can start the review. I follow up with adding queries for the ones without check marks by mid next week.

As for the HTTP/2 adoption in 2021 for comparison with 2022, I plan to refer to the values from last year's spreadsheet, so we may not have to create new queries for them. However, if you would like me to include queries for them, please let me know. Happy to include queries for those.

tunetheweb (approved):

tunetheweb:
The queries look good to me so far so happy for you to run this. Shall we merge these and add more later in a new PR to make it easier to see what's changed?

Issue #2886, opened 2022-04-12
Third Parties 2022


rviscomi:
# Third Parties 2022

![Third Parties illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/third-parties/hero_lg.jpg)

If you're interested in contributing to the Third Parties chapter of the 2022 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| ---- | ------- | --------- | -------- | ------- | ----------- |
| @imeugenia | @imeugenia | @tunetheweb @housseindjirdeh @pepelsbey @kevinfarrugia @alexnj  | @kevinfarrugia | - | @siakaramalegos  |

<details>
<summary><b>Expand for more information about each role 👀</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
- **[Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)** are responsible for researching the stats and trends used throughout the Almanac. Analysts work closely with authors and reviewers during the planning phase to give direction on the types of stats that are possible from the dataset, and during the analyzing/writing phases to ensure that the stats are used correctly.
[...]

PR #2940, opened 2022-05-23
Third Parties 2022 queries

kevinfarrugia:
Progress on #2886.

## Prevalence and structure
- [x] Third party prevalence by rank
- [x] Content-types
- [x] Third-party domains
- [x] Third-party requests by type and rank

## Performance
- [ ] ~Using third-party domains versus self-hosting~
- [x] Popular third-party embeds and their performance impact
  - [x] Popular third-parties and their impact on render
  - [x] Popular third-parties and their impact on main thread
- [x] Timing-Allow-Origin header prevalence
- [x] Page weight impact
  - [x] Heaviest third-parties
  - [x] Cacheability
  - [x] Large redirects
- [ ] Early-loaders
  - [x] Are third party scripts loaded after first-party content, i.e. what is their position in <head/> tag? 
  - [ ] Is async or defer used for third party scripts? 
- [x] Are third parties serving compressed/optimized resources?
  - [x] JavaScript (minified/compressed)
  - [x] CSS (minified/compressed)
  - [x] Images (bytes/pixel)

## Best practices
- [ ] Lazy-loading
- [x] Usage of third-party facades
  - [x] Usage of Lite-youtube
- [x] Legacy APIs
  - [x] unload event
  - [x] document.write
  - [x] legacy JavaScript
- [x] Usage of Partytown
- [ ] Correlation between CLS and 3P

## Sustainability
- [ ] CO2 emissions based on 3rd party script weight (Source: https://sustainablewebdesign.org/calculating-digital-emissions/)

## Security and privacy
- [x] Permissions-Policy header
- [x] CSP usage

## Accessibility
- [x] a11y tech usage
- [x] a11y overall usage by domain and rank
- [x] a11y usage by domain and rank


kevinfarrugia:
Yes, I will update all queries to June 2022, verify the data and update the PR by mid next week.

siakaramalegos:
Hi @kevinfarrugia @imeugenia and team - great job so far! How close do you think you are to finishing up any queries, adding the data to the results sheet and making the charts? Trying to see if we're still on track for the Aug 1 milestone. Again, thanks so much for your hard work so far

kevinfarrugia:
@siakaramalegos most of the queries are already written and the result sheet is updated. A handful remain which I will be discussing with @imeugenia tomorrow.

kevinfarrugia:
I will be covering the remaining 3 queries in a separate PR.

tunetheweb:
Cool. I’ll give this one a review tomorrow since I wrote last years queries.

Thanks for your fantastic work on this! Especially since you’re helping out with lots of other chapters too.

sql/2022/third-parties/distribution_of_lighthouse_unminified_css_by_3p.sql
```diff
@@ -0,0 +1,46 @@
...
+  });
+} catch (e) {
+  return [];
+}
+''';
+
+WITH base AS (
+  SELECT
+    page,
+    SUM(IF(is_3p, potential_savings, 0)) AS potential_savings
```
tunetheweb, line 19:
```suggestion
    SUM(IF(is_3p, potential_savings, 0)) AS potential_third_party_savings
```

sql/2022/third-parties/distribution_of_lighthouse_unminified_css_by_3p.sql
```diff
@@ -0,0 +1,46 @@
...
+      `httparchive.lighthouse.2022_06_01_mobile` AS lighthouse,
+      UNNEST(getUnminifiedCssUrls(JSON_EXTRACT(report, "$.audits['unminified-css']"))) AS data
+  )
+  GROUP BY
+    page
+)
+
+SELECT
+  percentile,
+  APPROX_QUANTILES(potential_savings, 1000)[OFFSET(percentile * 10)] AS potential_savings_bytes
```
tunetheweb, line 39:
```suggestion
  APPROX_QUANTILES(potential_savings, 1000)[OFFSET(percentile * 10)] AS potential_third_party_savings_bytes
```

sql/2022/third-parties/distribution_of_lighthouse_unminified_css_by_3p.sql
```diff
@@ -0,0 +1,46 @@
...
+    page,
+    SUM(IF(is_3p, potential_savings, 0)) AS potential_savings
+  FROM (
+    SELECT
+      lighthouse.url AS page,
+      NET.HOST(data.url) IS NOT NULL AND NET.HOST(data.url) IN (
+        SELECT domain
+        FROM `httparchive.almanac.third_parties`
+        WHERE date = '2022-06-01' AND category != 'hosting'
+      ) AS is_3p,
```
tunetheweb, line 27: This kind of reads weird. I wonder if better done in a `WHERE` clause with a `LEFT OUTER JOIN` if you want to include the sites with non-3p savings? Took me a bit to get what you were doing here.

sql/2022/third-parties/distribution_of_lighthouse_unminified_css_by_3p.sql
```diff
@@ -0,0 +1,46 @@
+#standardSQL
+# Third-party pages with unminified CSS
```
tunetheweb, line 2:
```suggestion
# Pages with unminified third-party CSS
```

sql/2022/third-parties/distribution_of_lighthouse_unminified_js_by_3p.sql
```diff
@@ -0,0 +1,46 @@
+#standardSQL
+# Third-party pages with unminified JavaScript
```
tunetheweb, line 2:
```suggestion
# Pages with unminified third-party JavaScript
```

sql/2022/third-parties/usage_of_lite_youtube_embed.sql
```diff
@@ -0,0 +1,14 @@
...
+  _TABLE_SUFFIX AS client,
+  COUNT(DISTINCT IF(app = 'lite-youtube-embed', url, NULL)) AS freq,
+  COUNT(DISTINCT url) AS total,
+  COUNT(DISTINCT IF(app = 'lite-youtube-embed', url, NULL)) / COUNT(DISTINCT url) AS pct
+FROM
+  `httparchive.technologies.2022_06_01_*`
+GROUP BY
+  client
+ORDER BY
+  client
```
tunetheweb, line 14: It's possible a page has no technologies detected by Wappalyzer so this would be more accurate as the following:

```suggestion
WITH totals AS (
  SELECT
    _TABLE_SUFFIX AS client,
    COUNT(DISTINCT url) AS total_pages
  FROM
    `httparchive.summary_pages.2022_06_01_*`
  GROUP BY
    client
),
youtube_embed AS (
SELECT
  _TABLE_SUFFIX AS client,
  COUNT(DISTINCT url) AS youtube_embed_pages
FROM
  `httparchive.technologies.2022_06_01_*`
WHERE
  app = 'lite-youtube-embed'
GROUP BY
  client
)

SELECT
  client,
  youtube_embed_pages,
  total_pages,
  youtube_embed_pages / total_pages AS pct_youtube_embed_pages
FROM
  totals
JOIN
  youtube_embed
USING (client)
ORDER BY
  client
```

sql/2022/third-parties/usage_of_partytown.sql
```diff
@@ -0,0 +1,14 @@
...
+  _TABLE_SUFFIX AS client,
+  COUNT(DISTINCT IF(app = 'Partytown', url, NULL)) AS freq,
+  COUNT(DISTINCT url) AS total,
+  COUNT(DISTINCT IF(app = 'Partytown', url, NULL)) / COUNT(DISTINCT url) AS pct
+FROM
+  `httparchive.technologies.2022_06_01_*`
+GROUP BY
+  client
+ORDER BY
+  client
```
tunetheweb, line 14: Same comment as above.

tunetheweb (changes requested):

sql/2022/third-parties/content_encoding_by_content_type.sql
```diff
@@ -0,0 +1,67 @@
...
+      WHEN resp_content_type LIKE '%plain%' THEN 'plain'
+      WHEN resp_content_type LIKE '%video%' THEN 'video'
+      WHEN resp_content_type LIKE '%audio%' THEN 'audio'
+      WHEN resp_content_type LIKE '%image%' THEN 'image'
+      WHEN resp_content_type LIKE '%font%' THEN 'font'
+      WHEN resp_content_type LIKE '%css%' THEN 'css'
+      WHEN resp_content_type LIKE '%javascript%' THEN 'javascript'
+      WHEN resp_content_type LIKE '%json%' THEN 'json'
+      ELSE 'other'
+    END AS content_type
```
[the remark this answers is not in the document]
kevinfarrugia: I saw that this used elsewhere (ex: sql/2021/javascript/most_requested_content_type.sql) and I assumed it was done intentionally to aggregate similar types together. I will update the query and review the results.

> Note that the JavaScript is listed as script and JSON type is also listed as a script in this case so can't separate them. Not sure if that's an issue for you?

I think one should be showing `javascript` and the other should be showing `json`.

> I don't think the intent of this query is to actually look at response content type per se. And to be honest, I think what you're getting here is covered by percent_of_third_party_requests_and_bytes_by_category_and_content_type.sql

The query you reference shows the `content-type` by 3P category. This query shows the `content-encoding` by `content-type` for 3Ps. For example, are 3Ps compressing images? Are 3Ps using `gzip` for JavaScript?

sql/2022/third-parties/csp_allowed_host_frequency.sql
```diff
@@ -0,0 +1,68 @@
...
+  total_pages,
+  pct
+FROM (
+  SELECT
+    client,
+    csp_allowed_host,
+    COUNT(DISTINCT page) AS freq,
+    total AS total_pages,
+    COUNT(DISTINCT page) / MIN(total) AS pct,
+    RANK() OVER (PARTITION BY client ORDER BY COUNT(DISTINCT page) DESC) AS rank
```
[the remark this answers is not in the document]
kevinfarrugia: It is ordered by `ORDER BY COUNT(DISTINCT page) DESC` and I am using `RANK()` to get the top 100 for desktop and the first 100 for mobile.

sql/2022/third-parties/content_encoding_by_content_type.sql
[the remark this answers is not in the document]
kevinfarrugia: Using the `mimeType` fragments the data into smaller groups, for example :

```
| mobile | application/font-woff     | NULL             |        95511 |   164404 |    0.5809530181747403 |
| mobile | application/font-woff2    | NULL             |      1384667 |  1394934 |    0.9926397951444298 |
| mobile | image/gif                 | NULL             |     33980072 | 36297076 |    0.9361655467784787 |
| mobile | image/jpeg                | NULL             |     20143073 | 20341833 |    0.9902290024699347 |
```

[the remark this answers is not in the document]
tunetheweb: I suggested `type` rather than `mimetype`. It groups them more: https://github.com/HTTPArchive/data-pipeline/blob/08cabedf000dfea42ea8ececebe7adc491224866/modules/utils.py#L51

sql/2022/third-parties/distribution_of_lighthouse_unminified_css_by_3p.sql
kevinfarrugia: I think this gives me the flexibility of changing the query to use the is_3p in the select. For example, on a separate ticket we are looking at including 1P too.

Then I can have,

```
  SUM(IF(is_3p, potential_savings, 0)) AS potential_3p_savings,
  SUM(potential_savings) AS potential_total_savings,
```

What do you think?

tunetheweb: Could do as a `LEFT OUTER JOIN` and then do:

```sql
  SUM(IF(third_parties.domain, potential_savings, 0)) AS potential_3p_savings,
  SUM(IF(third_parties.domain IS NULL, potential_savings, 0)) AS potential_1p_savings,
  SUM(potential_savings) AS potential_total_savings,
```

That would say you potentially having to repeat the subquery in both `potential_3p_savings` and `potential_1p_savings` columns.

Probably want to do a:
```sql
WITH third_parties AS (
  SELECT
    DISTINCT NET.HOST(domain)
  FROM
    `httparchive.almanac.third_parties`
)
```
 CTE in that case though to avoid joining multiple times if a domain is in there multiple times when NET.HOSTing it.

Or we could just go with that you have.

WDYT?

kevinfarrugia: Understood. Will see if I can improve it.

kevinfarrugia: Updated. Agreed on readability. Also updated some other similar queries to remove the nested `IN`.

The numbers varied very slightly. Any ideas?

If you're happy, it's good to go from my end.

sql/2022/third-parties/content_encoding.sql
```diff
@@ -0,0 +1,51 @@
...
+    requests r
+  ON NET.HOST(r.url) = NET.HOST(tp.domain)
+  WHERE
+    date = '2022-06-01' AND
+    category != 'hosting'
+  GROUP BY
+    domain
+  HAVING
+    page_usage >= 50
+)
```
tunetheweb, line 30: Should we NET.HOST it here in the CTE to avoid any duplicates in the later join? Note sure if it makes a difference or if the thirdparties table is already basically at NET.HOST level but might be safer.

```suggestion
third_party AS (
  SELECT
    NET.HOST(domain) AS domain,
    COUNT(DISTINCT page) AS page_usage
  FROM
    `httparchive.almanac.third_parties` tp
  JOIN
    requests r
  ON NET.HOST(r.url) = NET.HOST(tp.domain)
  WHERE
    date = '2022-06-01' AND
    category != 'hosting'
  GROUP BY
    NET.HOST(domain)
  HAVING
    page_usage >= 50
)
```

tunetheweb (commented):
Looks pretty good to me. One question about the thirdparties CTE. I got this wrong last year but not even sure it's an issue.

sql/2022/third-parties/content_encoding.sql
kevinfarrugia: Updated the query and re-ran it. No changes.

Also I think the `NET.HOST(domain)` in the `GROUP BY` is incorrect since the `GROUP BY` uses the `SELECT` fields (so `domain` is already `NET.HOST`-ed).

tunetheweb: True!

tunetheweb (approved):
LGTM!

PR #2947, opened 2022-05-23
Interoperability 2022 queries

rviscomi:
Progress on #2887 

## Compat 2021

- Flexbox
  - [x] Adoption of flexbox
  - [x] Adoption of `flex-basis: content`
  - [x] Popularity of `justify-content`, `align-items`, and `align` property values
  - [ ] Adoption of absolutely positioned flex item ❓ 
- [x] Adoption of `position: sticky`
- [x] Adoption of `aspect-ratio`
- [x] Adoption of `transform: perspective(none)`
- [x] Adoption of `transform-style: preserve-3d`

## Interop 2022

- [x] Adoption of cascade layers
- [x] Distribution of the number of `@layer` blocks per page
- [x] Adoption of color spaces
- [x] Adoption of color functions
- [x] Popularity of viewport units
- [x] Adoption of `<dialog>`
- [x] Adoption of `::backdrop`
- [x] Adoption of `contain` property
- [x] Popularity of `contain` values
- [x] Adoption of `@container` ruleset
- [x] Adoption of subgrid
- [ ] ❓ Forms
- [ ] ❓ Scrolling
- [ ] ❓ Typography and encodings

cc @bkardell


kevinfarrugia:
@bkardell could you help clarify what is the meaning of "Adoption of absolutely positioned flex item question" with an example please?

sql/2022/interop/popularity_of_contain_values.sql
```diff
@@ -0,0 +1,70 @@
...
+
+    var contain = rule.declarations.find(
+      (d) => d.property == "contain"
+    );
+
+    if (!contain) {
+      return values;
+    }
+
+    return values.concat(contain.value.replace("!important", "").split(" ").filter(Boolean));
```
rviscomi, line 23: @kevinfarrugia @bkardell should we split the values of the `contain` property? Or should we preserve the entire value?

For example if we're looking for values of `layout inline-size` for [container queries](https://css-tricks.com/next-gen-css-container/) we would only get instances of `layout` and instances of `inline-size` separately, but no stats on when they're used together and in order.

rviscomi:
> @bkardell could you help clarify what is the meaning of "Adoption of absolutely positioned flex item" with an example please?

I think it's referring to this: https://chenhuijing.com/blog/flexbox-and-absolute-positioning/#%F0%9F%91%BE

It'll be tricky to measure because I'm not sure we'll reliably know that a given element with absolute positioning is a flex item. Sometimes the declarations will also include a flex-related property like `align-items` in the example.

@bkardell WDYT?

bkardell:
> @bkardell could you help clarify what is the meaning of "Adoption of absolutely positioned flex item" with an example please?
> 
> I think it's referring to this: https://chenhuijing.com/blog/flexbox-and-absolute-positioning/#%F0%9F%91%BE
> 
> It'll be tricky to measure because I'm not sure we'll reliably know that a given element with absolute positioning is a flex item. Sometimes the declarations will also include a flex-related property like `align-items` in the example.
> 
> @bkardell WDYT?

Yeah, it's basically that - there are a slew of a few hundred tests for this in https://wpt.fyi/results/css/css-flexbox/abspos?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&view=interop&q=label%3Ainterop-2021-flexbox. 

But yeah, it's probably very hard to know/quantify this tbh - I'm not sure how we can, and if we can't I think that that is fine. Again there is going to be a lot in this chapter that it is difficult to measure this way or quantify and predict so I expect  we'll lean on pointing to interop tests and such as much as we can.

sql/2022/interop/popularity_of_contain_values.sql
bkardell: I would keep them together -- they are easy enough to see that that is both but impossible to reassmble

rviscomi: Updated the query and [results](https://docs.google.com/spreadsheets/d/1w3GzzTNeKxafFODmjDs6OC2dseNEDDKwUV8KeSgRI1Y/edit#gid=1251142331)

sql/2022/interop/align_items_values.sql
```diff
@@ -0,0 +1,70 @@
...
+
+    var alignItems = rule.declarations.find(
+      (d) => d.property == "align-items"
+    );
+
+    if (!alignItems) {
+      return values;
+    }
+
+    return values.concat(alignItems.value.replace("!important", "").split(" ").filter(Boolean));
```
kevinfarrugia, line 23: Do we want to `split` for `align-items`?

We could have values such as `align-items: first baseline;`

sql/2022/interop/backdrop.sql
```diff
@@ -0,0 +1,48 @@
+#standardSQL
```
kevinfarrugia, line 1: Do you think we should add a comment what the query is looking for?

sql/2022/interop/backdrop.sql
```diff
@@ -0,0 +1,48 @@
...
+}
+''';
+
+WITH detections AS (
+  SELECT
+    client,
+    page,
+    SUM(countSelectors(css)) > 0 AS detected
+  FROM
+    `httparchive.almanac.parsed_css`
```
kevinfarrugia, line 32: Does the `parsed_css` table contain one row per page? i.e. there is no need to `JOIN` with `httparchive.summary_pages.2022_07_01_*` to get the totals?

sql/2022/interop/color_spaces.sql
```diff
@@ -0,0 +1,200 @@
+#standardSQL
+CREATE TEMPORARY FUNCTION getP3Usage(css STRING)
```
kevinfarrugia, line 2: I understand there are other color spaces, such as HSL, HSLA, LAB and LCH. https://wpt.fyi/results/css/css-color?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&view=interop&q=label%3Ainterop-2022-color 

Are we looking at all these or only p3? (I'm not very knowledgeable on this topic TBF)

sql/2022/interop/layer_adoption.sql
```diff
@@ -0,0 +1,48 @@
+#standardSQL
```
kevinfarrugia, line 1: How come for the CSS chapter you used a RegEx (faster ?) and here you use `walkSelectors`?

sql/2022/interop/align_items_values.sql
rviscomi: Yes good call

sql/2022/interop/backdrop.sql
rviscomi: Will add

sql/2022/interop/backdrop.sql
rviscomi: No but this CTE is grouped by page, so there will always be one row per page in the subsequent query

sql/2022/interop/layer_adoption.sql
rviscomi: I mention in the comment for the CSS query that the parser doesn't extract `@layer` natively, so regex was more straightforward.

However this chapter has a few metrics looking at `@`-rules so I tried a new general purpose solution. `walkSelectors` is slightly better because it ensures we're testing the regex on only the selectors, and it includes inline style blocks.

No reason we shouldn't do this for CSS as well, besides being more work 😁

sql/2022/interop/color_spaces.sql
rviscomi: I'm not sure either, I left a comment for @bkardell about this in the Slack channel.

@bkardell what do you envision this query looking like?

rviscomi:
Queries updated and rerun based on feedback. There's an open question for @bkardell on color spaces. I'll merge this PR as-is and we can reopen a new one if any additional changes are needed.

PR #2966, opened 2022-06-10
Add fetch priority to hero images

tunetheweb:
Doesn't make a huge difference to be honest, but I still like to use the Web Almanac to demo best practices (and to act as a reminder for me on how to do things 😁).

WebPageTest Comparison: https://www.webpagetest.org/video/compare.php?tests=220610_AiDcT1_BT4%2C220610_BiDcKC_BTE&thumbSize=200&ival=100&end=visual

Without this change:
hero image request is delayed

With this change:
hero image is requested earlier

Note the hero image is requested a lot earlier (even though it finishes arriving at pretty much the same time).


pmeenan (approved):
LGTM

tunetheweb:
What do you reckon to `loading="eager"` @pmeenan ? It doesn't do anything AFAIK so should we just remove it as part of this, since it could be confusing to people (again using the Web Almanac source as a best practice example)?

pmeenan:
Doesn't do anything (it's the default) but it also doesn't hurt.

tunetheweb:
I've thought it over and I'm going to remove it. I think the term `eager` (falsely) implies it will be fetched sooner (I'm [not keen on the name eager](https://twitter.com/tunetheweb/status/1508512540120264710?s=20&t=8Rf3zzcs-o1TTmo6RQXQLg)). Now we have `fetchpriority`, we've a better way to give that signal so think using it is confusing.

PR #2981, opened 2022-06-20
Flag potentially incorrect data sets being used

tunetheweb:
This will update SQLFluff config to flag potentially incorrect datasets being used.

Note, this won't take effect until the next release of SQLFluff (1.0.1 or 1.1.0) as this feature's just been newly added in https://github.com/sqlfluff/sqlfluff/pull/3467.

Note, also for those chapters using secondary datasets this can be overridden with `--noqa: L063` on the given line like so:

```sql
-- Get secondary pages
SELECT
  ...
FROM
  `httparchive.summary_pages.2022_06_09_*` --noqa: L063
WHERE
  ....
```

### DIFF
--- a/sql/.sqlfluff
+++ b/sql/.sqlfluff
@@ -123,11 +123,13 @@ additional_allowed_characters = "-."
 
 [sqlfluff:rules:L062]
 # Comma separated list of blocked words that should not be used
+blocked_words = None
+# Regex of blocked SQL that should not be used.
+# Can be overridden with `-- noqa: L062` for those chapters using secondary pages
 # Block 2022_05_12 (contains secondary pages)
 # Block 2022_06_09 (contains secondary pages)
 # Block 2022_07_01 (probably forgot to update month to June for 2022)
 # Block 2021_06_01 (probably forgot to update month to July for 2021)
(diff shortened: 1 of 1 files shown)


rviscomi:
What does the warning message say to an analyst who queries one of these tables? Curious how clear it is that it's ok under intentional circumstances.

tunetheweb:
It’s a linting failure - same as the rest. There are no “warnings” and “errors”. And if there were the warnings would just be ignored 😄

rviscomi:
Ok sure, they're errors. But for example [this](https://github.com/HTTPArchive/almanac.httparchive.org/runs/6956348625?check_suite_focus=true) linter error contains the message "Unnecessary trailing whitespace". I'm curious what the corresponding message would look like for this error.

```
2022-06-19 16:32:54 [ERROR]   Error code: 65. Command output:
------
== [/github/workspace/sql/2022/markup/buttons.sql] FAIL
L:  31 | P:  54 | L001 | Unnecessary trailing whitespace.
```

rviscomi:
SGTM

rviscomi (approved):

PR #2988, opened 2022-06-21
Add 2022 base templates

tunetheweb:
This adds the base 2022 templates for all languages, except for:
1. Forward - commented out as a TODO for now
2. Methodology page - more complicated so left out for now.

This also sends the holding page for 2022 live, but leaves the default year as 2021.

To send 2022 fully live we should set the `DEFAULT_YEAR` in `config.py` to 2022. We should have at least one completed chapter before doing this.


src/templates/en/2022/base.html
```diff
@@ -0,0 +1,10 @@
+{% extends "%s/base.html" % lang %}
+
+{% block methodology_stat_1 %}8.36M{% endblock %}
+{% block methodology_stat_2 %}43.88 TB{% endblock %}
+{% block total_websites %}nearly 8.4 million{% endblock %}
```
tunetheweb, line 5: @rviscomi any thoughts on secondary pages for these stats? At the moment they aren't included.

If I include them it goes up to 81.47 TB

rviscomi: Let's omit their impact for now and update if they become a significant part of this edition

rviscomi:
> To send the "Coming Soon" holding page live we can set the settings->is_live flag to True in 2022.json. Also should follow instructions on line 51 of set_lighthouse_urls.sh to re-enable Lighthouse checks for 2022.

I'd be ok publishing the coming soon page now too

rviscomi (approved):

Issue #2898, opened 2022-04-12
Jamstack 2022


rviscomi:
# Jamstack 2022

![Jamstack illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2020/jamstack/hero_lg.jpg)

If you're interested in contributing to the Jamstack chapter of the 2022 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| ---- | ------- | --------- | -------- | ------- | ----------- |
| @seldo | @seldo @whitep4nth3r | @GregBrimble @nutlope | @seldo | - | @tunetheweb |

<details>
<summary><b>Expand for more information about each role 👀</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
- **[Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)** are responsible for researching the stats and trends used throughout the Almanac. Analysts work closely with authors and reviewers during the planning phase to give direction on the types of stats that are possible from the dataset, and during the analyzing/writing phases to ensure that the stats are used correctly.
[...]

PR #2992, opened 2022-06-27
Jamstack 2022 Queries

whitep4nth3r:
Progress on #2898 

# 2022 Jamstack queries

Jamstack is a hard category to define, since the definition is architectural -- a decoupled back end and front end -- which is not a distinction visible to a crawler. So we have gone with a number of proxy metrics, in combination. See [draft](https://docs.google.com/document/d/15RLaaTVqoqb5AuDrlBt6L0J_BMx1ltKW4t8VWX-sN_g/edit#heading=h.8z91yaf1dmft) for more detail on methodology.

1. Jamstack sites are fast

- <strike>Lighthouse Performance score</strike>
  * We found [distribution of Lighthouse performance scores](lighthouse_distribution_mobile.sql) in 2022
    * Mobile median score was 0.3
    * Better means: anything with this or larger
  * Lighthouse performance scores were not reliably available in earlier years
    * And omitting the score in the 2022 data did not make a big difference to the resulting set of URLs (since they were performant on 3 other measures already) so we omitted this.
- Largest Contentful Paint using user timings
  * Distribution of LCP times on [mobile](distribution_lcp_mobile.sql) and [desktop](distribution_lcp_desktop.sql)
    * Mobile median is 5.5 seconds
    * Desktop median is 3.7 seconds
  * Better means: anything with this or lower

2. Jamstack sites are resilient (and pre-rendered)

- How much the content changes post-load, i.e. Cumulative Layout Shift
  * Have distribution of CLS scores on [mobile](distribution_cls_mobile.sql) and [desktop](distribution_cls_desktop.sql)
    * Mobile median is 0.058
    * Desktop median is 0.023
  * Better means: anything with this or lower

3. Jamstack sites are cached for a long time

- Age header
  * Have distribution of Age headers on [mobile](distribution_age_headers_mobile.sql) and [desktop](distribution_age_headers_desktop.sql)
    * Median 21 hours in both cases
  * Better means: anything with this or longer


tunetheweb:
@seldo @GregBrimble any progress on this? Getting quick concerned with the Jamstack chapter falling behind. Let me know if there’s anything I can do to help.

seldo:
Sorry Barry, this was done as far as I'm concerned, and simply not merged -- my plan next week was to actually run these queries against some previous years. Is there more you'd expect prior to merging this?

tunetheweb:
Ah that’s great news.

If you could add the queries to this branch when you get a chance, we’ll give them a quick code review first, to save you having to rerun if we spot any problems.

seldo:
Sure thing. I'm assuming what we should be doing is adding SQL files to `sql/2022/jamstack` along the lines of https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2021/javascript/frameworks_libraries.sql or is there a more sophisticated way to do this? (And is it okay if that's next week?)

tunetheweb:
Yup that’s it. Next week is more than fine now I know you’ve actually worked/are working on them. Was just a bit worried it was all too quiet and worried nothing was happening.

So next milestone is 1st August to have them added, reviewed, merged and run. And then comes the writing about all that lovely data!

seldo:
Okay, I've done the initial queries and created a combined query that allows us to find all the URLs that match our various criteria and then a join that finds all the sites that match all of them. I've been inspecting the resulting URLs and they certainly all seem very jamstacky! There are quite a small number so I imagine picking "must exceed median" is quite stringent (as it excludes anybody at, or close to the median on one or more metrics).

Next steps:
* Examine the matching URLs for false positives (expecting very few)
* Try relaxing one or more parameters to include URLs close to the thresholds and see when we start getting a lot of false positives
* Having picked thresholds we like, do a longitudinal pass across previous years to see if this group is growing or shrinking and how fast

GregBrimble:
Hey folks, sorry, I've been pretty preoccupied with other stuff this year, and @seldo has already done a great job with the queries so far. Looks like it's pretty much just going to be relatively subjective tweaking of thresholds now. If it's alright, I'll bow out and let you finish them off. Sorry again, and good luck with the chapter!

sql/2022/jamstack/candidate_urls.sql
```diff
@@ -0,0 +1,88 @@
+-- step 1.1: get URLs with age headers
+with headers as (
+  SELECT 
+    url,
+    JSON_EXTRACT_ARRAY(payload, '$.response.headers') as headers_array
+  FROM `httparchive.requests.2022_07_01_mobile` 
+),
```
tunetheweb, line 7: This is available as a field in `httparchive.almanac.requests` which makes it significantly cheaper to query. The `almanac` dataset is an optimised version of the monthly datasets with certain keyfields (like this one) pulled out form the `payload` body to make them easier (and quicker and cheaper) to query.

Also we are using June not July for the almanac this year.

Plus we should look at both mobile and desktop.

```suggestion
with headers as (
  SELECT
    client,
    url,
    response_headers as headers_array
  FROM
    `httparchive.almanac.requests`
  WHERE
    date = '2022-06-01'
),
```

This is 1 TB compared to 12 TB.

And you will probably be able to merge this CTE into the next one now.

sql/2022/jamstack/candidate_urls.sql
```diff
@@ -0,0 +1,88 @@
...
+  where age > 86400 -- 1 day
+),
+
+-- step 2.1: get URLs with LCP times
+-- step 3.2: get URLs with CLS times
+lighthouse_audits as (
+  SELECT  
+    url,
+    CAST(JSON_EXTRACT(report, '$.audits.largest-contentful-paint.numericValue') as NUMERIC) as lcp_ms,
+    CAST(JSON_EXTRACT(report, '$.audits.cumulative-layout-shift.numericValue') as NUMERIC) as cls
```
tunetheweb, line 41: We should not use the lighthouse data for this. Instead either use the CrUX data (in the `pages` data set) if wanting home page data, or the CrUX data in `chrome-ux-report.materialized.device_summary` dataset if wanting to use the whole origin data.

sql/2022/jamstack/candidate_urls.sql
```diff
@@ -0,0 +1,88 @@
...
+  FROM `httparchive.lighthouse.2022_07_01_mobile`
```
tunetheweb, line 42: If you are using lighthouse (which, as I say I don't think we should use for LCP or CLS), then note we have desktop lighthouse data this year so should look to query for both.

sql/2022/jamstack/candidate_urls.sql
```diff
@@ -0,0 +1,88 @@
...
+  FROM `httparchive.lighthouse.2022_07_01_mobile` 
+),
+
+-- step 2.2 & 3.2: filter URLs with LCP and CLS at our chosen levels
+cls_and_lcp_filtered as (
+  select 
+    distinct(url) as url
+  from lighthouse_audits
+  where lcp_ms > 6600
+  and cls < 0.06
```
tunetheweb, line 51: Why are these levels chosen? Why not the CWV levels that more people are likely to know?

sql/2022/jamstack/candidate_urls.sql
```diff
@@ -0,0 +1,88 @@
...
+
+performance_scores as (
+  select 
+    url,
+    round(performance_score,2) as perf_rounded
+  from performance_scores_raw
+  where performance_score is not null
+),
+
+performance_filtered as (
```
tunetheweb, line 70: This three CTEs could easily be merged into one.

sql/2022/jamstack/candidate_urls.sql
```diff
@@ -0,0 +1,88 @@
...
+  select
+    distinct(url) as url
+  from performance_scores
+  where perf_rounded > 0.31
```
tunetheweb, line 74: Again, why this score? Seems arbitary.

sql/2022/jamstack/distribution_age_headers.sql
```diff
@@ -0,0 +1,33 @@
+-- getting ages to calculate median age (or any other percentage threshold)
+-- we're using 2022-07-01 run as the baseline because values change every year
```
tunetheweb, line 2: Shouldn't this only be looking at the HTML page request, rather than all requests?

sql/2022/jamstack/distribution_cls.sql
```diff
@@ -0,0 +1,22 @@
+-- getting bucketed distribution of CLS scores to find median and other percentiles
+with cls_values as (
+  SELECT  
+    url,
+    CAST(JSON_EXTRACT(report, '$.audits.cumulative-layout-shift.numericValue') as NUMERIC) as cls
+  FROM `httparchive.lighthouse.2022_07_01_mobile`
```
tunetheweb, line 6: Again better to use CrUX data.

sql/2022/jamstack/distribution_lcp.sql
```diff
@@ -0,0 +1,16 @@
+-- getting distribution of LCP times to calculate median LCP time
+-- or whatever other threshold we decide looks reasonable
+with lcp_times as (
+  SELECT  
+    url,
+    CAST(JSON_EXTRACT(report, '$.audits.largest-contentful-paint.numericValue') as NUMERIC) as lcp_ms
+  FROM `httparchive.lighthouse.2022_07_01_mobile`
```
tunetheweb, line 7: Use CrUX data.

tunetheweb (changes requested):

tunetheweb:
@seldo any update on this? We're now nearly a week past when analysis was due so a little concerned that we're not going to make the publish date...

sql/2022/jamstack/distribution_age_headers_desktop.sql
```diff
@@ -0,0 +1,33 @@
+-- getting ages to calculate median age (or any other percentage threshold)
+-- we're using 2022-06-01 run as the baseline because values change every year
```
tunetheweb, line 2: This looks at the age of ALL requests. Including analytics pixesl and the like which may be short.

We should restrict this to home page only surely? Using the `almanac dataset` makes this easier as it has a `firstHTML` field and also pulls the `age` field out for querying.

We also usually include both desktop and mobile in the one query:

```sql
SELECT
  client,
  date,
  COUNT(0) AS total,
  COUNTIF(resp_age IS NOT NULL AND resp_age != '') AS age,
  COUNTIF(resp_age IS NULL OR resp_age = '') AS no_age,
  COUNTIF(resp_age IS NOT NULL AND resp_age != '') / COUNT(0) AS pct_age,
  COUNTIF(resp_age IS NULL OR resp_age = '') / COUNT(0) AS pct_no_age,
FROM
WHERE
  firstHtml
GROUP BY
  client,
  date
ORDER BY
  client,
  date
```

sql/2022/jamstack/candidate_urls_count.sql
```diff
@@ -0,0 +1,64 @@
+{% set SAMPLE_MONTH = "2022-06-01" %}
+
+-- step 2.1: get URLs and LCP times from chrome user timings
+-- step 3.1: get URLs and CLS times from chrome user timings
+WITH lighthouse_audits AS (
+  SELECT
+    url,
+    CAST(JSON_EXTRACT(payload, "$['_chromeUserTiming.LargestContentfulPaint']") AS NUMERIC) AS lcp_ms,
+    CAST(JSON_EXTRACT(payload, "$['_chromeUserTiming.CumulativeLayoutShift']") AS NUMERIC) AS cls
```
tunetheweb, line 9: As discussed suggestion is to use CrUX data, rather than Lighthouse.

sql/2022/jamstack/r1_jamstackishness_over_time/candidate_urls_count_2020_06_01_desktop.sql
```diff
@@ -0,0 +1,62 @@
+-- step 2.1: get URLs and LCP times from chrome user timings
+-- step 3.1: get URLs and CLS times from chrome user timings
```
tunetheweb, line 2: We expect all queries to be in top level folder, rather than sub-folders.

sql/2022/jamstack/generate_queries.mjs
```diff
@@ -0,0 +1,52 @@
+// The queries in r1_jamstackishness_over_time are all very similar
+// This script generated them from a template
+// But the generated files are also checked in
```
tunetheweb, line 3: You can just either use the `almanac` data set (which by definition only has almanac years), or use this:"

```sql
SELECT
  IF(ENDS_WITH(_TABLE_SUFFIX, 'desktop'), 'desktop', 'mobile') AS client,
  REGEXP_REPLACE(_TABLE_SUFFIX, r'(\d+)_(\d+)_(\d+).*', r'\1-\2-\3') AS date,
  xxx
FROM
  `httparchive.requests.*
WHERE
  (_TABLE_SUFFIX like '2020_08_01%' OR _TABLE_SUFFIX like '2020_07_01%' OR _TABLE_SUFFIX like '2022_06_01%')
```

This only charges you for the tables you query.

tunetheweb (changes requested):

tunetheweb (approved):

PR #3012, opened 2022-07-14
Correct names of 2021 Privacy images and reference ebooks from GitHub

tunetheweb:
The 2021 Privacy chapter had some bad image names (I think IIRC they changed the query names at last minute for some reason).

This means a number of them don't on when not using interactive embeds (e.g. mobile, or ebook).

For example: https://almanac.httparchive.org/en/2021/privacy?print#fig-3

This PR corrects the names so they do generate.

Unfortunately however, we were right to the GCP App Engine single file storage limit (32Mb) for the 2021 ebook, and this knocks us over the edge. So I've added them to .gcloudignore so they are not deployed and changed links to GitHub.

On the plus side releases will be quicker as no longer need to upload these massive files to GCP!

### DIFF
--- a/src/server/routes.py
+++ b/src/server/routes.py
@@ -236,4 +236,4 @@ def redirect_old_hero_images(folder, image):
     '/static/pdfs/<pdf>'
 )
 def redirect_pdfs(pdf):
-    return redirect("https://github.com/HTTPArchive/almanac.httparchive.org/raw/main/src/static/pdfs/%s" % (pdf)), 301
+    return redirect("https://cdn.httparchive.org/almanac/ebooks/%s" % (pdf)), 301
--- a/src/server/routes.py
+++ b/src/server/routes.py
@@ -229,3 +229,11 @@ def redirect_old_images(folder, image):
 )
 def redirect_old_hero_images(folder, image):
     return redirect("/static/images/2020/%s/%s" % (folder, image)), 301
+
+
+# Redirect requests for the pdfs to GitHub
+@app.route(
+    '/static/pdfs/<pdf>'
+)
+def redirect_pdfs(pdf):
+    return redirect("https://github.com/HTTPArchive/almanac.httparchive.org/raw/main/src/static/pdfs/%s" % (pdf)), 301
--- a/src/server/tests/routes_test.py
+++ b/src/server/tests/routes_test.py
@@ -195,8 +195,7 @@ def test_render_pdf_redirect(client):
         client,
         "/static/pdfs/web_almanac_2019_en_cover_A5.pdf",
         301,
-        "https://github.com/HTTPArchive/almanac.httparchive.org/raw/main/"
-        + "src/static/pdfs/web_almanac_2019_en_cover_A5.pdf",
+        "https://cdn.httparchive.org/almanac/ebooks/web_almanac_2019_en_cover_A5.pdf",
     )
 
 
--- a/src/server/tests/routes_test.py
+++ b/src/server/tests/routes_test.py
@@ -195,7 +195,8 @@ def test_render_pdf_redirect(client):
         client,
         "/static/pdfs/web_almanac_2019_en_cover_A5.pdf",
         301,
-        "https://github.com/HTTPArchive/almanac.httparchive.org/raw/main/src/static/pdfs/web_almanac_2019_en_cover_A5.pdf",
+        "https://github.com/HTTPArchive/almanac.httparchive.org/raw/main/"
+        + "src/static/pdfs/web_almanac_2019_en_cover_A5.pdf",
     )
 
 
@@ -206,6 +207,7 @@ def test_render_old_http_image_dir_redirect(client):
         301,
         "/static/images/2020/http/random.png",
     )
+
 
 def test_render_old_hero_image_dir_redirect(client):
     assert_route(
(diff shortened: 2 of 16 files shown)


rviscomi:
Could you clarify why we don't need to upload the ebooks to GCP?

tunetheweb:
The only reason to upload them, is to be able to download them as we reference them in links:

```html
    <li class="nav-dropdown-list-chapter ebook">
      <a href="/static/pdfs/web_almanac_2021_en.pdf" data-event="ebook-click" data-label="toc-menu">Ebook PDF (31MB)</a>
    </li>
```

As you can see, in this PR, I've changed those links to go to GitHub instead.

You can see it in action here: https://20220714t222418-dot-webalmanac.uk.r.appspot.com/en/2021/table-of-contents#ebook

The only slight issue with that, that I've just thought about, is they will no longer appear in Google Search:

image

Will need to remove from site map too then since they will 404.

src/templates/base/base.html
```diff
@@ -244,7 +244,7 @@
     {% endif %}
 
     {% if ebook_size_in_mb and ebook_size_in_mb > 0 %}
-    <option value="/static/pdfs/web_almanac_{{ year }}_{{ lang }}.pdf" data-event="ebook-click">
+    <option value="https://github.com/HTTPArchive/almanac.httparchive.org/raw/main/src/static/pdfs/web_almanac_{{ year }}_{{ lang }}.pdf" data-event="ebook-click">
```
rviscomi, line 247: Have you considered uploading the PDFs to Cloud Storage and directing users to cdn.httparchive.org?

tunetheweb: Is there a benefit to that, versus GitHub? It's still off our app engine.

Though it would be nice to get them out of GitHub as they are big and do make your repo quick big too every time we regenerate them. But to do that we'd need GitHub action to have Cloud Storage Access to be able to upload to there on generation and we've never been comfortable enough to add that to the repo given how many contributors we have.

tunetheweb: Well that turned out to be easier than I thought - done!

src/server/routes.py
```diff
@@ -229,3 +229,11 @@ def redirect_old_images(folder, image):
...
 def redirect_old_hero_images(folder, image):
     return redirect("/static/images/2020/%s/%s" % (folder, image)), 301
+
+
+# Redirect requests for the pdfs to GitHub
+@app.route(
+    '/static/pdfs/<pdf>'
+)
+def redirect_pdfs(pdf):
+    return redirect("https://github.com/HTTPArchive/almanac.httparchive.org/raw/main/src/static/pdfs/%s" % (pdf)), 301
```
rviscomi, line 239: Update redirect

src/server/tests/routes_test.py
```diff
@@ -190,6 +190,16 @@ def test_render_old_image_dir_redirect(client):
     )
 
 
+def test_render_pdf_redirect(client):
+    assert_route(
+        client,
+        "/static/pdfs/web_almanac_2019_en_cover_A5.pdf",
+        301,
+        "https://github.com/HTTPArchive/almanac.httparchive.org/raw/main/"
```
rviscomi, line 198: Update test

rviscomi (changes requested):

src/tools/generate/generate_sitemap.js
```diff
@@ -6,7 +6,6 @@ const { get_static_lang_files } = require('./shared');
 const min_publish_date = '2019-11-11';
 const sitemap_template = `templates/sitemap.ejs.xml`;
 const sitemap_path = `templates/sitemap.xml`;
-const ebook_path = "static/pdfs/web_almanac_";
```
rviscomi, line 9: Should we keep the ebooks in the sitemap, but use their absolute URLs on the CDN? Not sure if that's a thing.

tunetheweb: AFAIK that’s not a thing. However Google should still crawl them anyway. Just might take a little longer to recrawl without a site map to give a clear indication they’ve been updated.

rviscomi (approved):

rviscomi:
One last thing to check is the cache policy for assets served on the CDN. Not sure what the defaults are.

tunetheweb:
Default looks to be 3 hours - same as we had on the almanac site itself:

```
< HTTP/2 200 
< x-guploader-uploadid: ADPycdsTg1FBOusHW3YUF9Vd73JC3LEMK-N0AeBD9qVffd1KwS73uPKkvUNjzHlIbTbMJ3Q8MJaW2BqXKmOi3i6C2Iwm4Q
< x-goog-generation: 1658263341925280
< x-goog-metageneration: 1
< x-goog-stored-content-encoding: identity
< x-goog-stored-content-length: 17394461
< content-language: en
< x-goog-hash: crc32c=C+PORg==
< x-goog-hash: md5=gICwIx6ZzMpzLx1UokgbkQ==
< x-goog-storage-class: STANDARD
< accept-ranges: bytes
< vary: Origin
< server: UploadServer
< date: Tue, 19 Jul 2022 20:45:28 GMT
< expires: Tue, 19 Jul 2022 21:45:28 GMT
< cache-control: public, max-age=3600
< last-modified: Tue, 19 Jul 2022 20:42:22 GMT
< etag: "8080b0231e99ccca732f1d54a2481b91"
< content-type: application/pdf
< content-length: 17394461
< age: 3544
< alt-svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
```

Issue #586, opened 2019-12-06
Translate content to Japanese


bazzadp:
These are the chapters to be translated, and they exist in the [content directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content):

**2020**
- [ ] capabilities

**2019**
- [x] accessibility - @MSakamaki 
- [x] caching - @ksakae1216
- [x] cdn - @ksakae1216
- [x] cms - @ksakae1216 (requires review)
- [x] compression - @ksakae1216 (requires review)
- [x] css - @ksakae1216 (requires review)
- [x] ecommerce - @ksakae1216 (requires review)
- [x] fonts - @ksakae1216  (requires review)
- [x] http2 - @ksakae1216 - #661
- [x] javascript - @ksakae1216 (requires review) 
- [x] markup - @MSakamaki 
- [x] media - @ksakae1216 (requires review)
- [x] mobile-web - @ksakae1216 (requires review)
- [x] page-weight- @ksakae1216 (requires review)
- [x] performance - @MSakamaki 
- [x] pwa - @ksakae1216 (requires review)
- [x] resource-hints - @ksakae1216 (requires review)
- [x] security - @ksakae1216 (requires review)
- [x] seo - @MSakamaki 
- [x] third-parties - @ksakae1216 (requires review)

Additionally the following pages need translated too in the [language specific templates directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/templates):
- [x] base.html - @ksakae1216 (requires review)
- [x] base_chapter.html - @ksakae1216 (requires review)
- [x] contributors.html - @ksakae1216 (requires review)
- [x] error.html
- [x] index.html - @ksakae1216 (requires review)
- [x] featured_chapters.html  - @ksakae1216 (requires review)
- [x] methodology.html - @ksakae1216 (requires review)
- [x] table_of_contents.html - @ksakae1216 (requires review)
- [x] accessibility_statement.html - @ksakae1216 (requires review)

There is no need to translate the chapters HTML pages are they are generated off the markdown combined with the above templates.

Common notes for writing consistency are here: https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Translators'-Guide. Feel free to edit that and/or add Japanese-specific extras by editing this comment.

Japanese specific extra advice:

* Nayori of foreign words
https://www.jtca.org/standardization/katakana_guide_3_20171222.pdf
[...]

PR #3023, opened 2022-07-26
Translation of 2021 CDN chapter to Japanese

ksakae1216:
Please review for me.


tunetheweb (approved):
LGTM!

I've just merged a fix for the test failure and merged that in here now. Unrelated to your change - a python dependence upgrade has a bug.

PR #3031, opened 2022-08-01
Pin werkzeug to work around bug

tunetheweb:


tunetheweb:
See: https://github.com/pallets/werkzeug/issues/2481

PR #3039, opened 2022-08-05
Fix sheets GID & SQL file name for video_track_usage figure of 2021 accessibility chapter

thibaudcolas:
Spotted this small issue while reviewing the contents of the 2021 chapter. On the [Video](https://almanac.httparchive.org/en/2021/accessibility#video) section, the figure dropdown links to the data / SQL query for the preceding _audio_ section.

The data and content of the section is all correct, it’s really just the links that are off.

- SQL file: https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2021/accessibility/video_track_usage.sql
- Sheet link: https://docs.google.com/spreadsheets/d/1WjAM5ZnHjMQt-rKyHvj2eVhU_WdzzFTjpoYWMr_I0Cw/edit#gid=1261793459

PR #3046, opened 2022-08-07
Fix translations for 3039 and 3040

tunetheweb:
Applies fixes for #3039 and #3040 to Japanese translations

FYI @thibaudcolas / @rviscomi


thibaudcolas:
Thank you! I hadn’t realised there were translations.

PR #3043, opened 2022-08-07
Fix SQL query link for 2021 accessibility "Language identification"

thibaudcolas:
Similar to #3039 / #3040. This time the sheet link is correct, but the SQL query isn’t.

- Section: https://almanac.httparchive.org/en/2021/accessibility#language-identification
- Sheet: https://docs.google.com/spreadsheets/d/1WjAM5ZnHjMQt-rKyHvj2eVhU_WdzzFTjpoYWMr_I0Cw/edit#gid=2009310389
- SQL file: https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2021/accessibility/valid_html_lang.sql

PR #3049, opened 2022-08-08
Fix labelling of valid_html_lang query results across 2020 and 2021 accessibility chapters

thibaudcolas:
Follow-up to #3043 (includes that commit too). The "Language identification" section’s metric is reported as a Desktop figure, even though the query is only fetching data from the `_mobile` table suffix. This is the case both in 2020 and 2021.

I’d have preferred to change the number like in #3048, unfortunately that’s not possible since we only fetch mobile data.

- SQL query 2020: https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2020/accessibility/valid_html_lang.sql
- SQL query 2021: https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2021/accessibility/valid_html_lang.sql

The translations are my best guess. For Dutch, it seems "mobiel" is used for the noun form, and "mobiele" as an adjective? For Japanese, I replaced "デスクトップ" (Desktop) with "モバイル" (Mobile) to match usage in other parts of the content and made sure the result wasn’t completely off in Google Translate.


tunetheweb (approved):
LGTM! Thanks.

PR #3062, opened 2022-08-10
Third Parties 2022 queries - New queries

kevinfarrugia:
Progress on #2886 

- [x] Unused CSS by 3P
- [x] Unused JavaScript by 3P
- [x] Third-party facades (from Lighthouse)
- [x] Compressed images by 3P

### DIFF
--- a/sql/2022/third-parties/compressed_images_by_3p.sql
+++ b/sql/2022/third-parties/compressed_images_by_3p.sql
@@ -1,5 +1,5 @@
 #standardSQL
-# Compressed images by third parties
+# Compressed images (excluding SVG) by third parties
 
 WITH requests AS (
   SELECT
@@ -14,6 +14,9 @@ WITH requests AS (
     type = 'image' AND (
       resp_content_encoding = 'gzip' OR
       resp_content_encoding = 'br'
+    ) AND (
+      resp_content_type NOT LIKE 'image/svg%' OR
+      ENDS_WITH(url, '.svg')
     )
 ),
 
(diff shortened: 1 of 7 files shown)


sql/2022/third-parties/compressed_images_by_3p.sql
```diff
@@ -0,0 +1,73 @@
...
+  total,
+  pct
+FROM (
+  SELECT
+    client,
+    content_encoding,
+    domain,
+    COUNT(0) AS num_requests,
+    SUM(COUNT(0)) OVER (PARTITION BY client, type) AS total,
+    COUNT(0) / SUM(COUNT(0)) OVER (PARTITION BY client, type) AS pct,
```
tunetheweb, line 52: What's this pct telling me? That Facebook serves 5.9% of all images used third parties used by all home pages that use br? I'm not sure how useful that is really? Or did you mean it as a percentage of each third party's own images?

Also why do we have any br images? Shouldn't the image format include compression and it's normally a bad practice to further compress. Except for SVGs I guess. In which case I further question the usefulness of this stat as some third-parties will use SVGs and some  not.

Confused.

kevinfarrugia: The 5% `pct` on Facebook refers to 5% of all compressed images are Facebook's. While it may be more appropriate (if possible) to show that 5% of all of Facebook's images are compressed.

I have also updated the query to exclude SVGs.

I have gone through the results and they seem correct, (e.g. https://www.s-coop-sumai.net/wp-content/uploads/sites/288/2021/01/fearure_3.png which is 58.5KB gzip compressed and 57.9KB uncompressed). Having said that, the increase in KBs is small and most of the images are tiny, with many 1x1.

tunetheweb: Maybe should limit it to images bigger than  X (for some value of X)? To ignore the 1x1 pixels where it probably doesn’t make a difference (some of these may even be empty IIRC?).

kevinfarrugia: I have added the image size as a column in the results and filter to the largest 100 third-parties for each content-encoding.

rviscomi (approved):

tunetheweb (approved):

PR #3063, opened 2022-08-11
Point to correct SQL files for overlay figures in 2021 accessibility chapter

thibaudcolas:
In [Accessibility overlays](https://almanac.httparchive.org/en/2021/accessibility#accessibility-overlays), two successive charts point to the correct sheets, but not to the correct query for their data. You can tell based on the column names.

1. Accessibility app usage by rank: [sheet](https://docs.google.com/spreadsheets/d/1WjAM5ZnHjMQt-rKyHvj2eVhU_WdzzFTjpoYWMr_I0Cw/edit#gid=2077755325), [correct query: `a11y_technology_usage_by_domain_rank.sql`](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2021/accessibility/a11y_technology_usage_by_domain_rank.sql). Note the `pct_sites_with_app` column.
2. Pages using accessibility apps by rank: [sheet](https://docs.google.com/spreadsheets/d/1WjAM5ZnHjMQt-rKyHvj2eVhU_WdzzFTjpoYWMr_I0Cw/edit#gid=827309922), [correct query: `a11y_technology_usage_by_domain_rank.sql`](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2021/accessibility/a11y_overall_tech_usage_by_domain_rank.sql). Note the `pct_sites_with_a11y_tech` column.

This is all pretty confusing due to the naming of the charts, sheets, queries.

1. The first query counts how much usage each of the accessibility overlays gets by domain rank (and the chart only shows 5 pre-selected tools)
2. The second query counts total usage of accessibility overlays by domain rank


tunetheweb (approved):
Thanks for the continue corrections on these @thibaudcolas ! Think most of these were my errors - added in a rush to publish last year! 😢

thibaudcolas:
My pleasure 😄 I honestly had no problem tracking down those corrections, and the data itself so far all seems flawless, so don’t worry too much about it!

PR #2958, opened 2022-06-06
Ecommerce 2022 queries

jrharalson:
Progress on #2897 

## eComm Platforms

- [x] Top platforms - top_vendors.sql
- [x] Top platforms growth - pct_ecommsites_bydevice_compare20202122.sql
- [x] Top platforms by 1M, 100k, 10k - top_vendors_crux_rank.sql
- [x] Top Enterprise targetted platforms - top_vendors.sql filtered

## Growth trends - overall
- [x] eComm growth - ecomm_growth.sql
- [x] platform growth - ecomm_vendors_growth.sql

## Server Stack
- [x] CDN Provider - top_cdn_bydevice_vendor_cdn.sql top_cdn_bydevice_vendor_wapp.sql

## Frontend Tech
- [x] JS Frameworks - top_jslibs_by_device.sql
- [x] JS Libraries - top_jsframework_providers_by_device.sql

## User Experience
- [x] CWV - Platform - Passing grade 10k vs 100k vendor - core_web_vitals_passingmetrics_byvendor_bydevice.sql
- [x] CWV - Top x Platforms - CLS, FCP, LCP, FID, TTFB - core_web_vitals_distribution_byvendor_bydevice.sql
- [x] Lighthouse - Top x Platforms - 5 scores - median_lighthouse_score_ecommsites.sql

## Page Anatomy
- [x] Page Reqs and Weight - pagestats_percentiles_bydevice.sql
- [x] Page Weight - HTML payload - pagestats_html_bydevice.sql
- [x] Page Reqs and Weight - images - pagestats_image_bydevice.sql
- [x] Page Weight - 3rd party domain reqs and weight - pct_3pusage_bydevice.sql

## Tools and Misc
- [x] Analytics usage - top_analytics_providers_bydevice_wapp.sql
- [x] Tag Manager usage - percent_of_ecommsites_using_each_tag_managers.sql
- [x] AMP usage - pct_ampusage_bydevice_vendor.sql
- [x] Consent Management usage - percent_of_ecommsites_using_cmp.sql
- [x] CSP usage - review latest security query
- [x] Web Push usage - webpushstats_ecommsites.sql

PR #3064, opened 2022-08-14
Linting fixes for already merged queries.

tunetheweb:
Includes following changes
- Correction to ecommerce queries checked in, in #2958 - only noticed after merging as lastest linting only ran then.
- Correction to CSS queries to noqa July datasets as using non-standard month

The dataset check was only recently released in SQLFluff and many branches were started before that update. Might see a few more of these.


tunetheweb:
FYI @rviscomi

PR #3085, opened 2022-08-23
Draft: Generate image exports for the 2022 accessibility chapter

thibaudcolas:
Do not merge. This contains all expected charts for the 2022 accessibility chapter along with current progress on the content. Just opening this as a PR so others can review all visuals in one place.


tunetheweb:
This is great work @thibaudcolas and everyone else! Looks like we've pulled this chapter back from the brink 🎉

A couple of comments though:
1. It might be easier to open this as a branch in the almanac repo. That way everyone else can edit without you having to permission them for your own fork. As contributors you should all have access to open branches on the main repo, but let me know if that is not the case and I can permission you.
2. Did you run all the images through tinypng.com? That often makes them significantly smaller. We have an image compressor GitHub Action but it's not as efficient as tinypng.com so only use that as a fallback. Also it doesn't work from forks (another reason not to use a fork for chapter submissions).
3. You've a few linting errors. If your IDE supports editorconfig (it needs a plugin for VSCode) then it will automatically clean up a lot of the issues by installing that, and then saving the file.
4. Could you remove the `todo: true` from `src/config/2022.json` for this chapter as part of this PR, so the chapter generates?

tunetheweb:
I moved this to #3087 - thanks again for the enormous effort on this @thibaudcolas !

Issue #2880, opened 2022-04-12
JavaScript 2022


rviscomi:
# JavaScript 2022

![JavaScript illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/javascript/hero_lg.jpg)

If you're interested in contributing to the JavaScript chapter of the 2022 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| ---- | ------- | --------- | -------- | ------- | ----------- |
| @malchata | @malchata @ibnesayeed | @mgechev @pankajparkar @NishuGoel @seldo @housseindjirdeh @kevinfarrugia | @NishuGoel @kevinfarrugia | @DesignrKnight @rviscomi | @siakaramalegos |

<details>
<summary><b>Expand for more information about each role 👀</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
[...]

PR #3102, opened 2022-09-02
Updated FID and INP queries to not exclude 0ms

kevinfarrugia:
Progress on #2880

### DIFF
--- a/sql/2022/javascript/fid.sql
+++ b/sql/2022/javascript/fid.sql
@@ -1,10 +1,6 @@
 #standardSQL
 # FID by device
 
-CREATE TEMP FUNCTION IS_NON_ZERO (p75_fid FLOAT64, p75_fid_origin FLOAT64) RETURNS BOOL AS (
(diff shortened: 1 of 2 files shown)


sql/2022/javascript/fid.sql
```diff
@@ -15,8 +11,7 @@ base AS (
     `chrome-ux-report.materialized.device_summary`
   WHERE
     device IN ('desktop', 'phone') AND
-    date IN ('2022-06-01') AND
-    IS_NON_ZERO(p75_fid, p75_fid_origin)
```
rviscomi, line 19: We should still filter out `NULL` FID values. Can do that in the `APPROX_QUANTILES` functions:

```sql
APPROX_QUANTILES(p75_fid, 1000 IGNORE NULLS)[OFFSET(percentile * 10)] AS p75_fid
```

rviscomi (changes requested):

rviscomi (approved):

PR #3108, opened 2022-09-04
HTTP 2022: add queries for getting more preload stats and HTTP/3 support

paivaspol:
Add more stats for link preload hints and HTTP/3 support for #2903.

### DIFF
--- a/sql/2022/http/sites_using_link_preload_header.sql
+++ b/sql/2022/http/sites_using_link_preload_header.sql
@@ -1,0 +1,48 @@
+#standardSQL
+
+# Number of <link rel="preload">.
+
+CREATE TEMPORARY FUNCTION getNumLinkRelPreloadRespHeader(HTTPheaders STRING, header STRING)
+RETURNS FLOAT64 LANGUAGE js AS """
+try {
+  var headers = JSON.parse(HTTPheaders);
+
+  // Filter by header name (which is case insensitive)
+  // If multiple headers it's the same as comma separated
+  const allLinkHeaderValues = headers.filter(h => h.name.toLowerCase() == header.toLowerCase()).map(h => h.value);
+  let numPreload = 0;
+  for (let linkHeader of allLinkHeaderValues) {
+    const linkHeaderTokens = linkHeader.split(',');
+    for (let linkHeaderToken of linkHeaderTokens) {
+      const linkTokenProps = linkHeaderToken.split(';');
+      for (let linkTokenProp of linkTokenProps) {
+        if (linkTokenProps.includes('rel=preload')) {
+          numPreload++;
+        }
+      }
+    }
+  }
+  return numPreload;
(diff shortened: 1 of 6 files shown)


sql/2022/http/sites_using_link_preload_header.sql
```diff
@@ -0,0 +1,48 @@
...
+      for (let linkTokenProp of linkTokenProps) {
+        if (linkTokenProps.includes('rel=preload')) {
+          numPreload++;
+        }
+      }
+    }
+  }
+  return numPreload;
+} catch (e) {
+  return [];
```
rviscomi, line 27: The return value is numeric so this type should be comparable
```suggestion
  return 0;
```

sql/2022/http/sites_using_link_preload_header.sql
```diff
@@ -0,0 +1,48 @@
+#standardSQL
+
+# Number of <link rel="preload">.
+
+CREATE TEMPORARY FUNCTION getNumLinkRelPreloadRespHeader(HTTPheaders STRING, header STRING)
+RETURNS FLOAT64 LANGUAGE js AS """
```
rviscomi, line 6:
```suggestion
RETURNS INT64 LANGUAGE js AS """
```

sql/2022/http/sites_using_link_preload_header.sql
```diff
@@ -0,0 +1,48 @@
...
+FROM (
+  SELECT
+    client,
+    page,
+    getNumLinkRelPreloadRespHeader(response_headers, 'link') AS num_link_rel_preload_resp_header
+  FROM
+    `httparchive.almanac.requests`
+  WHERE
+    date = '2022-06-01' AND
+    firstHtml
```
rviscomi, line 45: Need to group by page in order to count page-level stats in the outer query.
```suggestion
  SELECT
    client,
    page,
    SUM(getNumLinkRelPreloadRespHeader(response_headers, 'link')) AS num_link_rel_preload_resp_header
  FROM
    `httparchive.almanac.requests`
  WHERE
    date = '2022-06-01' AND
    firstHtml
  GROUP BY
    client,
    page
```

paivaspol: done

sql/2022/http/sites_using_link_preload_header.sql
paivaspol: done

sql/2022/http/sites_using_link_preload_header.sql
paivaspol: done

rviscomi: I don't think we want -1 since an error would undo 1 valid count

tunetheweb (approved):

PR #3109, opened 2022-09-05
Sustainability 2022: Updated queries

fershad:
This PR fixes the `green_third_party_requests.sql` query to align with the results of the Third Parties chapter.

I can be left open while the chapter is under review, in case other modifications are required.

### DIFF
--- a/sql/2022/sustainability/green_third_party_requests.sql
+++ b/sql/2022/sustainability/green_third_party_requests.sql
@@ -32,7 +32,6 @@ pages AS (
 third_party AS (
   SELECT
     domain,
-    canonicalDomain,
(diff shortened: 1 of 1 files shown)


tunetheweb:
@fershad , not sure if you’ve seen but the Third Parties chapter have added both variants (domain and canonicalDomain) based on your findings. Not sure which they are actually going to use at this stage to be honest but will see if I can get an answer for you.

fershad:
@tunetheweb thanks! Yep, I caught up with their changes. I decided to take the approach of using `domains` rather than `canonicalDomains` for our data as you suggested in the spreadsheet.

tunetheweb (approved):

PR #3110, opened 2022-09-05
Third Parties 2022 - content-encoding trends

kevinfarrugia:
Progress on #2886 

- Added third party queries for content-type trends
- Fixed linting issues

### DIFF
--- a/sql/2022/third-parties/trends_content_encoding.sql
+++ b/sql/2022/third-parties/trends_content_encoding.sql
@@ -1,0 +1,131 @@
+#standardSQL
+#party of content encoding
+
+WITH third_party_domains AS (
+  SELECT DISTINCT
+    NET.HOST(domain) AS domain
+  FROM
+    `httparchive.almanac.third_parties`
+),
+
+base AS (
+  SELECT
+    "2022" AS year,
+    client,
+    domain,
+    resp_content_encoding
+  FROM (
+    SELECT
+      "2022" AS year,
+      client,
+      third_party_domains.domain AS domain,
+      resp_content_encoding
+    FROM (
+      SELECT
+        client,
+        url AS page,
+        NET.HOST(url) AS domain,
+        resp_content_encoding,
+      FROM
+        `httparchive.almanac.requests`
+      WHERE
+        date = '2022-06-01'
+    ) AS potential_third_parties
+    LEFT OUTER JOIN
+      third_party_domains
+    ON
+      potential_third_parties.domain = third_party_domains.domain
+    GROUP BY
+      client,
+      page,
+      domain,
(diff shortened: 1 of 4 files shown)


sql/2022/third-parties/trends_content_encoding.sql
```diff
@@ -0,0 +1,131 @@
+#standardSQL
+#party of content encoding
+
+WITH third_party_domains AS (
+  SELECT DISTINCT
+    NET.HOST(domain) AS domain
+  FROM
+    `httparchive.almanac.third_parties`
+),
```
tunetheweb, line 9: I don't think it ever happens but a third-party domain _could_in theory drop out of this dataset (e.g. if it was misclassified as third-party).

Also, to replicate what the 2021 / 2020 numbers we should use the appropriate dataset for that year. Though that does run the risk of the third party dataset being the reason for any increase, but still think it's what we should do (for parity with those year's chapters).

tunetheweb (changes requested):

sql/2022/third-parties/trends_content_encoding.sql
kevinfarrugia: Pushed an update, let me know.

kevinfarrugia: I have noticed that there is also some variance compared to the query in the JavaScript chapter. :thinking: 

[sql/2022/javascript/compression_method_by_3p.sql](https://github.com/HTTPArchive/almanac.httparchive.org/blob/ce0261506cb5461dec6ae37167d65badf964119e/sql/2022/javascript/compression_method_by_3p.sql)

tunetheweb: Did you get this figured out @kevinfarrugia ? Is this good to merge?

kevinfarrugia: Haven't had the chance yet. The method for checking is slightly different, but in theory it should be equal. Will take another look tomorrow.

kevinfarrugia: @tunetheweb the difference comes because in this query I am treating each 3P URL as distinct. For example, if 100 different websites request `example.com/script.js`, then it is treated as one 3P request rather than 100 3Ps. I'm not sure if that's correct though, so I might need to push an update.

kevinfarrugia: Pushed an update. Numbers add up with JavaScript chapter. Also query is much faster (~7s compared to ~117s) but I will not change the JavaScript chapter's query at this late stage.

tunetheweb (approved):

PR #3116, opened 2022-09-09
Additional queries authored while writing

eeeps:
Last queries for #2884 I promise.

### DIFF
--- a/sql/2022/media/video_adoption.sql
+++ b/sql/2022/media/video_adoption.sql
@@ -2,7 +2,8 @@
 # How many pages use <video>?
 
 SELECT
-  client,
+  SUBSTR(_TABLE_SUFFIX, 0, 10) AS date,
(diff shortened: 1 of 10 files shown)


sql/2022/media/video_adoption.sql
```diff
@@ -11,6 +12,10 @@ FROM (
     _TABLE_SUFFIX AS client,
     CAST(JSON_VALUE(JSON_VALUE(payload, '$._media'), '$.num_video_nodes') AS INT64) AS num_video_nodes
   FROM
-    `httparchive.pages.2022_06_01_*`)
+    `httparchive.pages.*`)
```
rviscomi, line 15: This depends on the `media` custom metric, which didn't exist prior to the 2020 edition. So querying beyond that would waste a lot of bytes. 

```suggestion
    `httparchive.pages.*`
  WHERE
    _TABLE_SUFFIX >= '2020-08-01')
```

rviscomi (changes requested):

rviscomi (approved):

PR #3122, opened 2022-09-12
[review] Fix font specific queries and new queries

bramstein:
Progress on #2883 

This PR contains some fixes to font internals specific queries. The biggest change is that the font queries now no longer use pages, but instead only use requests. 

The reason for that is that a popular font can be included in the requests table many times which would skew the results. Instead the queries that look at font-internal data now look at the unique font requests, which produces a more much accurate reflection of what internal font formats are being created by type designers.

This PR also contains a couple of new queries that were requested during the writing process or the technical review.

The data sheet and chapter has already been updated with the most recent data.

### DIFF
--- a/sql/2022/fonts/color_fonts.sql
+++ b/sql/2022/fonts/color_fonts.sql
@@ -4,6 +4,7 @@ WITH
 fonts AS (
   SELECT
     url,
+    client,
     JSON_EXTRACT(payload, '$._font_details.color.formats') AS payload
   FROM
     `httparchive.almanac.requests`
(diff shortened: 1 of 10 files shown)


rviscomi (changes requested):
Queries look fine but a few of them are no longer grouping by `client`. This is required, otherwise we'd be double-counting requests on pages that exists for both client types, which skews the results. These queries would need to be updated and rerun, which will affect the results.

bramstein:
@rviscomi That is on purpose: for these queries I'm only interested in what is supported by unique font files. Wouldn't grouping by `url` achieve that?

rviscomi:
Yeah, it's just for consistency across chapters that we require stats to be scoped to either desktop or mobile pages. Even if your metric is % of requests, they must be distinctly sourced from either desktop or mobile, but not both. It's fine to pick one and report on it in the chapter (typically mobile), rather than sharing two similar stats. See the [query conventions](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide#query-conventions) for more info.

bramstein:
@rviscomi: Thanks for the explanation! I've updated the queries that didn't include client. I've updated the sheet and chapter as well (there were minimal changes as the numbers only changed by a couple decimal points at most.)

rviscomi (approved):
Looks great, thanks for making the changes!

Issue #2896, opened 2022-04-12
CMS 2022


rviscomi:
# CMS 2022

![CMS illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/cms/hero_lg.jpg)

If you're interested in contributing to the CMS chapter of the 2022 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| ---- | ------- | --------- | -------- | ------- | ----------- |
| @sirjonathan | @sirjonathan | @alexdenning @dknauss @alonkochba @honzasladek | @csliva | @dknauss | @tunetheweb  |

<details>
<summary><b>Expand for more information about each role 👀</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
- **[Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)** are responsible for researching the stats and trends used throughout the Almanac. Analysts work closely with authors and reviewers during the planning phase to give direction on the types of stats that are possible from the dataset, and during the analyzing/writing phases to ensure that the stats are used correctly.
[...]

PR #3153, opened 2022-09-24
Further edits to the CMS chapter

dknauss:
Edits for #2896 

Mostly edits for typos, brevity, and clarity. Noted default WebP is *not* going into WP 6.1 but eventually will, we can safely assume. @sirjonathan


tunetheweb (changes requested):
All looks good except we use KB (1024) not kB (1000).

Technically it's KiB but [see discussion here](https://github.com/HTTPArchive/httparchive.org/issues/643) as to why we're not using that.

tunetheweb:
Also you have some merge conflicts if you could look at those.

tunetheweb (approved):
Reverted the kB to KB change but rest looks good so merging in time for launch.

PR #3162, opened 2022-09-25
Update bramstein bio with a more accurate one.

bramstein:
@tunetheweb: As promised, the updated bio. Sorry I just missed the deadline.


tunetheweb (approved):
LGTM if CI passes!

PR #3169, opened 2022-09-26
Enable ebook for 2022 English

tunetheweb:
@rviscomi I presume we should enable the ebook even without some chapters? They just will be excluded.


tunetheweb:
The test will fail as ebook PDFs don’t exist. Should be fine to merge anyway and then prerelease script will create them.

rviscomi:
IIRC we've held off on the ebook until all chapters are published to prevent people from unintentionally saving incomplete/stale versions. Let's hold off until the rest of the chapters are in.

PR #3178, opened 2022-09-26
Fix 2022 logo

### LINKED ISSUE #3176: "2022" hangs off the edge on the home hero
The right edge of "2022" should stay within the boundaries of the browser graphic below, but it hangs over:

image

See 2019 for reference:

image

tunetheweb:
Fixes #3176

### DIFF
--- a/src/templates/base/index.html
+++ b/src/templates/base/index.html
@@ -68,16 +68,16 @@
     </div>
     <div class="intro-image-wrapper">
       {% if year == "2022" %}
-      <svg viewBox="0 0 53.5 16">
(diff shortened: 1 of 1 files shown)


tunetheweb:
Am sure there must be a better way to do this, rather than tweak each year, but this will do for now.

tunetheweb:
New:

image

Current Prod:

image

Also tweaked 2020 while I was at it.

rviscomi (approved):

yuseyhan:
The new logo looks good in terms of proportion and alignment.
resim

PR #3188, opened 2022-09-29
refactor: compress images (ImageOptim, defensive settings)

j9t:
Result: “Saved 3,8 MB out of 98,8 MB. 6.3% per file on average (up to 58.2%)”

(Context: Routinely do this, with newly forked repos… seemed it would shave some weight here.)

Signed-off-by: Jens Oliver Meiert <jens@meiert.com>


tunetheweb:
Controversially I’m going to turn down this PR @j9t - despite the fact I very much appreciate you working on it!

We use TinyPNG.com to optimise our images* and, while ImageOptum May squeeze a few more bytes it’s not as easy for all our contributors to use.

So why not just take this PR and do a similar sweep up at the end of every year? Good question but what I like about the TinyPNG option is it’s simple, repeatable and works with git diffs. Particularly when I’m editing chapters I can remove ALL the images for that chapter, retake them all (with the script we have for this), tinypng them, and then git diff and it only shows figures that have changed. This saves me tracking which figures I’m changing during editing when I inevitably miss one.

I feel bringing in another tool, that’s not quite as say to use (well it IS easier if installed and know what settings to use, but most contributors won’t have it installed), compared to drag and drop, will reduce efficiencies in the long run.

So thanks, and normally would jump on any free webperf wins, but think best not to merge this.

* FYI we also use the Calibre GitHub Action as a fall back, but it doesn’t compress anywhere near as good as TinyPng never mind ImageOptum so it’s really a fallback to ensure that at least SOME compression is done. I normally retake them and TinyPNG them when I says this happening when I get time.

PR #3241, opened 2022-10-23
Lighthouse desktop config

### LINKED ISSUE #3240: Update Lighthouse methodology to include desktop testing
This line from the [Lighthouse](https://almanac.httparchive.org/en/2022/methodology#lighthouse) section of the Methodology page is no longer applicable:

> HTTP Archive runs the latest version of Lighthouse for all of its mobile web pages — desktop pages are not included because of limited resources.

We should update this text to reflect that desktop tests were included in the June dataset and add the desktop configuration.

@pmeenan could you confirm the desktop config? For reference, here's what we show for the mobile config:

Config | Value
-- | --
CPU slowdown | 1x/4x
Download throughput | 1.6 Mbps
Upload throughput | 0.768 Mbps
RTT | 150 ms

rviscomi:
Fixes #3240

image

### DIFF
--- a/src/templates/en/2022/methodology.html
+++ b/src/templates/en/2022/methodology.html
@@ -254,7 +254,7 @@
       </p>
 
       <p>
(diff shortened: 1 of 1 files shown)


pmeenan (approved):
LGTM

PR #3293, opened 2022-11-29
Internationalise MDN links

tunetheweb:
MDN redirects to most appropriate language so let's use that rather than manually maintaining these. This is also what web.dev does.


tunetheweb:
WDYT @rviscomi ? It's what we do on web.dev, and would save translators having to check if a local version exists. Plus those get out of date anyway.

rviscomi (approved):
It's a small performance penalty for the reader but seems reasonable.

Issue #585, opened 2019-12-05
Translate content to Italian


bazzadp:
These are the core templates - without which we cannot release any translated chapters. They are in the [language specific templates directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/templates):
- [ ] base.html - 
- [ ] base_chapter.html - 
- [ ] error.html - 

**2020**
- [ ] base.html - 
- [ ] contributors.html - 
- [ ] index.html  - 
- [ ] table_of_contents.html - 

**2019**
- [ ] base.html - 
- [ ] contributors.html - 
- [ ] index.html  - 
- [ ] table_of_contents.html - 

These are the chapters to be translated, in rough order of popularity. They exist in the [content directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content):

**2020**
- [ ] javascript - 
- [ ] seo - 
- [ ] markup - 
- [ ] accessibility -
- [ ] cms - 
- [ ] performance
- [ ] http2 - 
- [ ] capabilities -
- [ ] resource-hints - 
- [ ] compression - 

**2019**
- [ ] javascript - 
- [ ] css - 
- [ ] seo - 
- [ ] performance - 
- [ ] http2 - 
- [ ] markup - 
- [ ] third-parties - 
- [ ] fonts - 
- [ ] accessibility - 
- [ ] pwa - 
- [ ] cms - 
- [ ] security - 
- [ ] media - 
- [ ] mobile-web - 
- [ ] cdn - 
- [ ] ecommerce - 
- [ ] page-weight - 
- [ ] caching - 
- [ ] compression - 
- [ ] resource-hints - 

Additionally the following pages need translated too in the [language specific templates directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/templates), but less important than core templates and chapters:
- [ ] accessibility_statement.html - 
- [ ] base_ebook.html- 
- [ ] 2019/methodology.html - 
- [ ] 2020/methodology.html - 

Please include "Makes progress on #585" in all pull requests so a link is created from the PR to this issue.

There is no need to translate the chapters HTML pages in `src/templates/en/2019/chapters/` nor the `src/templates/en/2019/ebook.html` file as they are generated off the markdown combined with the above templates.

---

[...]

PR #3297, opened 2022-12-06
2022 css chapter - Italian

webmatter-it:
Make progress on #585 - I have translated the 2022 CSS chapter


tunetheweb:
LGTM! Thanks for the translation!

PR #3303, opened 2022-12-12
Adds Share CTA and Reaction count

### LINKED ISSUE #2617: Add Web Share API to chapters
To encourage readers to share chapters, let's add support for the [Web Share API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API). Browser support is mostly mobile-only, so we should use feature detection before adding the UI.

We should also add an analytics event when someone uses the share button.

tunetheweb:
Fixes #2617 

This PR makes two changes:
- Adds a reaction count (if reactions are > 0) to the current CTA
- Adds a Share button (if the browser support navigator.share integration) as detailed in @tomayac 's [post on this](https://web.dev/patterns/advanced-apps/share/). That post suggests falling back to Twitter sharing when this isn't supported (and this is also what web.dev does), but with the current uncertainty of Twitter that might not be what some users want, so decided just to hide it.

On iOS and Safari, this looks like this:

image

On Android it looks like this:

image

Note this is not supported on Chrome on desktop yet. If on an Mac, you can try it on Safari, where it does work.

I added this all to the webmentions.js code as allows it to only render the CTA after the number of reactions have been got, rather than have CLS when that comes in.

It's staged here so you can try on your phone: https://20221212t204134-dot-webalmanac.uk.r.appspot.com/en/2022/http#reactions


rviscomi:
LGTM on Android

I also viewed the demo in Chrome on MacOS and the share button is visible but errors on click (`TypeError navigator.share is not a function`). Is that expected, or should it be hidden?

image

tunetheweb:
> I also viewed the demo in Chrome on MacOS and the share button is visible but errors on click (`TypeError navigator.share is not a function`). Is that expected, or should it be hidden?

Doh! Missed the final deploy so currently has commented out code:

image

Also added GA events, that also aren't in the deployed version.

Redeployed now and should be good by the time you wake up.

rviscomi (approved):

PR #3328, opened 2023-01-12
Translation of the 2022 Structured Data chapter to Italian

webmatter-it:


tunetheweb (approved):
LGTM after some small tweaks. Thanks!!

PR #3332, opened 2023-01-14
fix: correct lastnames (Matuzović, Soueidan), remove unnecessary ellipses before “etc.”

j9t:


tunetheweb (approved):

tunetheweb:
I think "... etc." must be a UK - English thing so al my fault 😁

PR #3343, opened 2023-02-06
Add inert to main content when mobile menu is open

tunetheweb:
See https://twitter.com/simevidas/status/1621794018358185984?s=20&t=OIektJsa65qGkx-SKT-Bcg

Will be interested to see if we see any INP regression as per https://github.com/GoogleChrome/web.dev/pull/9409. I'm not seeing a massive recalc layout style like we did on web.dev - only 6ms on my machine. FYI @brendankenny . You can compare [staged](https://20230206t115221-dot-webalmanac.uk.r.appspot.com/en/2022/css) to [prod](https://almanac.httparchive.org/en./2022/css) to see if you can spot anything?


rviscomi (approved):
Seems to work

PR #3387, opened 2023-04-15
Upgrade SQLFluff to latest

tunetheweb:
SQLFluff v2 contains a lot of breaking changes in the config, and also basically a complete rewrite of the formatting rules.

A previous attempt in #3364 couldn't be completed as we weren't happy with some of the changes necessary to support both SQLFluff v1 (as used by the Super Linter used by our GitHub Action ) and v2. Now the SQLFluff Linter has been upgraded we can do a (slightly!) cleaner and smaller change, though there are still a lot of changes.

The alternative is to stay on old, unsupported version but that's not a great idea for many reasons.


sql/2019/accessibility/09_06.sql
```diff
@@ -22,7 +22,8 @@ FROM (
     client,
     total,
     page,
-    id)
+    id
+)
```
tunetheweb, line 26: Unfortunatelty v2 doesn't allow trailing closing brackets. But think that's OK as more clear now when line 8 closes.

sql/2019/cms/14_15.sql
```diff
@@ -15,15 +15,17 @@ FROM (
...
     date = '2019-07-01' AND
     NET.HOST(url) IN (SELECT domain FROM `httparchive.almanac.third_parties` WHERE date = '2019-07-01' AND category != 'hosting')
   GROUP BY
     client,
-    page),
+    page
+),
   UNNEST([10, 25, 50, 75, 90]) AS percentile
 GROUP BY
   percentile,
```
tunetheweb, line 31: This is potentially a problem for us due to the "faux [K&R](https://en.wikipedia.org/wiki/Indentation_style)" style we're following here.

Basically if you take this simplified version of this query:

```sql
SELECT
  *
FROM (
  SELECT * FROM a table
),
  UNNEST([...])
```

The `FROM` clause is made up of two tables: 1) a subquery and 2) an `UNNEST` clause.

SQLFluff wants to close the opening bracket on the same indentation it was opened (i.e. where the `FROM` clause starts), but then the indentation looks weird to me. This is all because the opening bracket at the `FROM` clause is not encompassing the whole `FROM` clause - it's only there for the subquery. This is why I call it "faux K&R" because the brackets are not braces, even though we basically pretend they are for our formatting style.

Our choices are:

1) Accept this weird indentation for any `UNNEST` statements. We usually use `JOIN` clauses instead of command lists when joining tables, but `UNNEST` is the one place we don't do that ([and there's a difference between doing it that way](https://stackoverflow.com/questions/47039541/bigquery-sql-is-it-better-to-unnest-in-select-or-join)).

2) Switch to this style for `UNNEST` queries:

```sql
SELECT
  *
FROM
  (
    SELECT * FROM a table
  ),
  UNNEST([...])
```

But know you weren't a fan of that @rviscomi , even if I am as I think it better reflects what the brackets represent.

3) Persuade SQLFluff to support the other version. I'm still a maintainer on that project (though not really active on it anymore) but know they explicitly removed "hanging indents" support in v2 due to a number of bugs with it, so don't see us persuading them to bring it back.

4) Stop using the SQLFluff linter. I think it's helped us a lot so not keen on that.

5) Stay on old versions of SQLFluff and GitHub Supper Linter - NOT a fan of this at all and would prefer to do 4) if we're adamant not to have this.

6) Find a better SQL Linter which does support our needs. But none do as well as SQLFluff and they will likely have their own opinions here.

What's your thoughts @rviscomi ?

tunetheweb:
OK but then I'd like to turn off dependabot (at least for GitHub Actions and Python depedencies) to stop getting PRs opened automatically. WDYT?

rviscomi:
Yeah that SGTM

PR #3388, opened 2023-04-16
Updates figure 2.1 in the JavaScript 2022 chapter

malchata:
Changes:

- Updates figure 2.1 in the JavaScript 2022 chapter to include data labels for desktop.


rviscomi:
The new image is half the size, but I think we want to keep the bigger resolution for high DPI

tunetheweb:
Retook the image.

tunetheweb (approved):

PR #3409, opened 2023-05-03
Super linter slim

max-ostapenko:
From standard to [slim](https://github.com/github/super-linter#slim-image) docker image.
Pulling image takes 1:10 instead of 1:40 min now.


tunetheweb (approved):
Nice! Hadn't noticed them adding this.

PR #3410, opened 2023-05-04
add pdf artifact upload step

max-ostapenko:
- [x] generated PDFs are temporarily stored in Action Artifacts
- [x] deployment script uses artifact ZIP as a source for deployed PDFs
- [x] PDF files are ignored in the repo moving forward


tunetheweb:
How's this looking @max-ostapenko ? About to do a release, but will hold off for a day or two if you think this is close to being completed and wanna test it out on a release?

Still not sure how you can download these assets locally?

max-ostapenko:
@tunetheweb in progress, checking both artifacts and LFS.
Will adapt `src/tools/scripts/deploy.sh` and then you can test it out for your release.

Any other scripts impacted?

tunetheweb:
Cool will hold off on release then.

I think it should just be that script and the action which generates the PDF.

src/tools/scripts/deploy.sh
```diff
@@ -98,10 +98,30 @@ if [ "$(pgrep -f 'python main.py')" ]; then
...
+
+# Download latest e-books from Github Action artifacts
+echo "Downloading latest e-books from Github Action artifacts"
+# Set your GitHub repository and workflow name
+REPO_OWNER=HTTPArchive
+REPO_NAME=almanac.httparchive.org
+WORKFLOW_ID=predeploy.yml
+# Set your Github personal access token
+export $(cat ../.env | xargs)
+#read -sp 'Enter your Github personal access token: ' GITHUB_TOKEN
```
max-ostapenko, line 115: @tunetheweb please adjust to whatever is preferred option to pass GITHUB_TOKEN value.
Here I tested with `.env` file and user input.

src/tools/scripts/deploy.sh
```diff
@@ -98,10 +98,31 @@ if [ "$(pgrep -f 'python main.py')" ]; then
...
+# Download latest e-books from Github Action artifacts
+echo "Downloading latest e-books from Github Action artifacts"
+# Set your GitHub repository and workflow name
+REPO_OWNER=HTTPArchive
+REPO_NAME=almanac.httparchive.org
+WORKFLOW_ID=predeploy.yml
+# Set your Github personal access token
+# shellcheck disable=SC1091
+source .env
+#read -sp 'Enter your Github personal access token: ' GITHUB_TOKEN
```
tunetheweb, line 116: Let's just get deployers to set a GITHUB_TOKEN env variable themselves and remove these lines.

Can we add this to here some where? https://github.com/HTTPArchive/almanac.httparchive.org/blob/bed8e5eab20f28633dd2040562379023b7628723/src/README.md?plain=1#L321

Should also explain how to generate an access token and the permission it requires (looks to me `public_repo`).

Also might need to update these steps: https://github.com/HTTPArchive/almanac.httparchive.org/blob/bed8e5eab20f28633dd2040562379023b7628723/src/README.md?plain=1#L359-L372

tunetheweb (changes requested):
This is a VERY nice solution to the problem. Thanks SO much for figuring this out.

Just a few small comments and then think good to merge.

src/README.md
```diff
@@ -360,8 +368,9 @@ The deploy script will do the following:
 - Ask you to confirm you've run the pre-deploy script via GitHub Actions
 - Switch to the production branch
 - Merge changes from main
-- Do a clean install
-- Run the tests
+- Do a clean install (remove generated chapters and e-books)
+- Downloade latest e-books from Github Action artifacts
```
tunetheweb, line 372:
```suggestion
- Download latest e-books from Github Action artifacts
```

src/README.md
```diff
@@ -332,6 +332,14 @@ You must first do some setup locally:
 gcloud init
 ```
 
+3. Set Github personal access token to GITHUB_TOKEN environment variable
+
+Use existing or create a new token [in GitHub.com](https://github.com/settings/personal-access-tokens/new). In token configuration select *Repository access = Public Repositories (read-only)*.
```
tunetheweb, line 337:
```suggestion
Use an existing token or create a new token [in GitHub.com](https://github.com/settings/personal-access-tokens/new). In token configuration select *Repository access = Public Repositories (read-only)*.
```

src/README.md
```diff
@@ -332,6 +332,14 @@ You must first do some setup locally:
 gcloud init
 ```
 
+3. Set Github personal access token to GITHUB_TOKEN environment variable
+
+Use an existing token or create a new token [in GitHub.com](https://github.com/settings/personal-access-tokens/new). In token configuration select *Repository access = Public Repositories (read-only)*.
+
+```bash
+export GITHUB_TOKEN=github_pat_...
```
tunetheweb, line 340: My token begins with `ghp_` not `github_pat_`:

```suggestion
export GITHUB_TOKEN=...
```

tunetheweb:
Thanks again for this @max-ostapenko . I'm actually travelling for the next few days so will pick this up after that if you don't mind. But looks good to merge to me.

max-ostapenko:
Enjoy the travel, and let me know if I can assist further.

tunetheweb (approved):

PR #3404, opened 2023-04-28
.devcontainer for Github Codespaces and VSCode

max-ostapenko:
#3150

- [x] Added base devcontainer.json (we can improve on with post create commands, IDE configurations, extensions, ...)
- [x] Added "Open in GitHub Codespaces" badge to README

Issue #3150, opened 2022-09-24
Set up .devcontainer for codespaces


rviscomi:
https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/introduction-to-dev-containers

I'm primarily using codespaces now for Web Almanac development. To help standardize the container used for cloud development and encourage more contributions, let's create this config file to make it even easier to get up and running.

DesignrKnight:
If this is not urgent, I'll be happy to volunteer for this after the editing work for 2022 Edition is over.

tunetheweb:
Closed with #3424

PR #3424, opened 2023-05-16
.devcontainer for Github Codespaces and VSCode (reopened)

max-ostapenko:
#3150

#3404 reopened

- [x] Added base devcontainer.json (we can improve on with post create commands, IDE configurations, extensions, ...).
- [x] Added "Open in GitHub Codespaces" badge to README. [Test it out with this branch](https://codespaces.new/HTTPArchive/almanac.httparchive.org/tree/development-container?quickstart=1).

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -25,7 +25,7 @@
         },
         "codespaces": {
             "openFiles": [
-                "README.md"
+                "src/README.md"
             ]
         }
     }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,32 @@
+{
+    "name": "Dev Container",
+    "build": {
+        "dockerfile": "../src/Dockerfile",
+        "context": "../src/",
+        "args": {
+            "IS_DEV_CONTAINER": "true",
+            "SKIPGC": "false"
+        }
+    },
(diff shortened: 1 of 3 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,32 @@
...
+            "extensions": [
+                "ms-python.python",
+                "ms-python.vscode-pylance"
+            ]
+        },
+        "codespaces": {
+            "openFiles": [
+                "README.md"
+            ]
+        }
```
tunetheweb, line 30: Do we need this one? Do we want a default open while?

If we need one, it might be better to open this one:

```suggestion
        "codespaces": {
            "openFiles": [
                "src/README.md"
            ]
        }
```

tunetheweb (changes requested):

.devcontainer/devcontainer.json
max-ostapenko: Yes, makes sense. I also moved a badge there.
I like to be reminded of guidelines when developing something (existing ones and potential for improvement).

tunetheweb (approved):

tunetheweb:
FYI if you say "Closes #XXXX" or "Fixes #XXXX" in the initial comment it will close the issue when merged. This didn't auto close as you only said "#XXXX". But manually closed it now.

PR #3430, opened 2023-05-18
Node16 + stages for docker container

max-ostapenko:
- [x] node version upgraded to 16 (to run scripts successfully in container)
- [x] found a few more places where super-linter image wasn't updated to slim
- [x] split Dockerfile to stages to differentiate between `docker run` and `devcontainer` use cases


tunetheweb (approved):
LGTM!

PR #3441, opened 2023-06-07
Fix two typos in JS 2022 Chapter

ajcwebdev:
Hello 👋, very minor change here but I noticed two small typos in the conclusion of the JavaScript 2022 Chapter.

Thanks for all the hard work on this awesome project!


tunetheweb (approved):
LGTM!

PR #3448, opened 2023-06-19
More GA4 changes

tunetheweb:
- Set web-vitals info to better names
- Have Stories send to both properties
- Remove report uri as causing tests to fail as we’ve reached out limit

### DIFF
--- a/src/server/__init__.py
+++ b/src/server/__init__.py
@@ -78,7 +78,6 @@ app.jinja_env.filters["accentless_sort"] = accentless_sort
 talisman = Talisman(
     app,
     content_security_policy=csp.csp,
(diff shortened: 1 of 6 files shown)


rviscomi (commented):
See my comments in https://github.com/HTTPArchive/httparchive.org/pull/745

tunetheweb:
> See my comments in [HTTPArchive/httparchive.org#745](https://github.com/HTTPArchive/httparchive.org/pull/745)

Updated in https://github.com/HTTPArchive/almanac.httparchive.org/pull/3448/commits/fe5ae8680242f6aaf377b18cb9f239b397892c13.

Waiting to see about scrollY before merging.

PR #3480, opened 2023-07-26
fix/word-error-base

HakaCode:
I've been analyzing the site for untranslated words and I noticed this punctuation/word error.


tunetheweb:
Thanks. I actually noticed that one too and thought I'd pushed a fix to one of your PRs yesterday, but looks like I forgot to push it!

PR #3667, opened 2024-05-22
Font 2024 queries

konfirmed:
## To be added 

- [ ] Font Usage
- [ ] ...

PR #3696, opened 2024-07-01
Fonts 2024 Queries

IvanUkhov:
In this PR, the queries from 2022 are copied to 2024 and are updated as follows:

## Changes

* The data source is exclusively `httparchive.all`.
* All queries are annotated with the question they answer.
* All queries follow the same structure and style.
* Queries are split by the section in the chapter where they are used.
* Repeated parts are extracted into common functions.

## Structure

The queries are split by the section where they are used:

* `design/` is about foundries and families,
* `development/` is about tools and technologies, and
* `performance/` is about hosting and serving.

Each file name starts with one of the following prefixes indicating the primary subject of the corresponding analysis:

* `fonts_` is about font files,
* `pages_` is about HTML pages,
* `scripts_` is about JavaScript scripts, and
* `styles_` is about CSS style sheets.

The prefix is followed by the property studied given in singular, potentially extended one or several suffixes narrowing down the scope, as in `fonts_size_by_table.sql` and `pages_link_relation.sql`.

## Contents

Each query starts with a preamble indicating the section, question, and normalization type:

```sql
-- Section: Performance
-- Question: What is the distribution of the file size broken down by table?
-- Normalization: Websites
```

Many queries rely on temporary functions for convenience and clarity. The functions appear in several queries are extracted into a common file: `common.sql`. Whenever any of the functions defined in `common.sql` is used by a query, the query has the following line at the top:

```sql
-- INCLUDE https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2024/fonts/common.sql
```

It signalizes that, prior to executing the query, `common.sql` has to be inlined.

### DIFF
--- a/sql/2024/fonts/README.md
+++ b/sql/2024/fonts/README.md
@@ -43,7 +43,7 @@ functions appear in several queries are extracted into a common file:
 query, the query has the following line at the top:
 
 ```sql
--- INCLUDE ../common.sql
+-- INCLUDE https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2024/fonts/common.sql
 ```
 
 It signalizes that, prior to executing the query, `common.sql` has to be
--- a/sql/2024/fonts/README.md
+++ b/sql/2024/fonts/README.md
@@ -34,7 +34,7 @@ normalization type:
 ```sql
 -- Section: Performance
 -- Question: What is the distribution of the file size broken down by table?
--- Normalization: Websites
+-- Normalization: Pages
 ```
 
 Many queries rely on temporary functions for convenience and clarity. The
--- a/sql/2024/fonts/README.md
+++ b/sql/2024/fonts/README.md
@@ -34,7 +34,7 @@ normalization type:
 ```sql
 -- Section: Performance
 -- Question: What is the distribution of the file size broken down by table?
--- Normalization: Sites
+-- Normalization: Websites
 ```
 
 Many queries rely on temporary functions for convenience and clarity. The
--- a/sql/2024/fonts/common.sql
+++ b/sql/2024/fonts/common.sql
@@ -77,12 +77,17 @@ CREATE TEMPORARY FUNCTION COLOR_FORMATS(payload STRING) AS (
   )
 );
 
--- Check if it is a color font given its payload.
+-- Check if the font is a color font given its payload.
 CREATE TEMPORARY FUNCTION IS_COLOR(payload STRING) AS (
   ARRAY_LENGTH(COLOR_FORMATS(payload)) > 0
 );
 
--- Check if it is a variable font given its payload.
+-- Check if the font was successfully parsed given its payload.
+CREATE TEMPORARY FUNCTION IS_PARSED(payload STRING) AS (
+  JSON_EXTRACT(payload, '$._font_details.table_sizes') IS NOT NULL
+);
+
+-- Check if the font is a variable font given its payload.
 CREATE TEMPORARY FUNCTION IS_VARIABLE(payload STRING) AS (
   REGEXP_CONTAINS(
     JSON_EXTRACT(payload, '$._font_details.table_sizes'),
--- a/sql/2024/fonts/common.sql
+++ b/sql/2024/fonts/common.sql
@@ -1,11 +1,23 @@
 -- Normalize a name. Used in FAMILY.
 CREATE TEMPORARY FUNCTION FAMILY_INNER(name STRING) AS (
-  IF(LENGTH(TRIM(name)) < 3, NULL, NULLIF(TRIM(name), ''))
+  CASE
+    WHEN REGEXP_CONTAINS(name, r'(?i)font\s?awesome') THEN 'Font Awesome'
+    ELSE IF(LENGTH(TRIM(name)) < 3, NULL, NULLIF(TRIM(name), ''))
+  END
 );
 
 -- Extract the family name from a payload.
 CREATE TEMPORARY FUNCTION FAMILY(payload STRING) AS (
-  FAMILY_INNER(JSON_EXTRACT_SCALAR(payload, '$._font_details.names[1]'))
+  FAMILY_INNER(
+    REGEXP_REPLACE(
+      COALESCE(
+        JSON_EXTRACT_SCALAR(payload, '$._font_details.names[16]'),
+        JSON_EXTRACT_SCALAR(payload, '$._font_details.names[1]')
+      ),
+      r'(?i)\s+(extra|semi|ultra)?\s?(condensed)?\s?(bold|book|heavy|light|medium|thin)?$',
+      ''
+    )
+  )
 );
 
 -- Extract the file format from an extension and a MIME type.
--- a/sql/2024/fonts/common.sql
+++ b/sql/2024/fonts/common.sql
@@ -14,7 +14,7 @@ CREATE TEMPORARY FUNCTION FAMILY(payload STRING) AS (
         JSON_EXTRACT_SCALAR(payload, '$._font_details.names[16]'),
         JSON_EXTRACT_SCALAR(payload, '$._font_details.names[1]')
       ),
-      r'(?i)([\s-]?(black|bold|book|cond(ensed)?|demi|extra|heavy|italic|light|medium|regular|semi|thin|ultra))+$',
+      r'(?i)([\s-]?(black|bold|book|cond(ensed)?|demi|ex(tra)?|heavy|italic|light|medium|narrow|regular|semi|thin|ultra|wide|\d00|\d+pt))+$',
       ''
     )
   )
--- a/sql/2024/fonts/common.sql
+++ b/sql/2024/fonts/common.sql
@@ -32,16 +32,18 @@ CREATE TEMPORARY FUNCTION FILE_FORMAT(extension STRING, type STRING) AS (
   LOWER(IFNULL(REGEXP_EXTRACT(type, '/(?:x-)?(?:font-)?(.*)'), extension))
 );
 
+-- Normalize a foundry name. Used in FOUNDRY.
+CREATE TEMPORARY FUNCTION FOUNDRY_INNER(name STRING) AS (
+  CASE UPPER(name)
+    WHEN 'ADBO' THEN 'ADBE'
+    WHEN 'PFED' THEN 'AWSM'
+    ELSE NULLIF(TRIM(REGEXP_REPLACE(name, r'[[:cntrl:]]+', '')), '')
+  END
+);
+
 -- Extract the foundry name from a payload.
 CREATE TEMPORARY FUNCTION FOUNDRY(payload STRING) AS (
-  NULLIF(
-    TRIM(REGEXP_REPLACE(
-      JSON_EXTRACT_SCALAR(payload, '$._font_details.OS2.achVendID'),
-      r'[[:cntrl:]]+',
-      ''
-    )),
-    ''
-  )
+  FOUNDRY_INNER(JSON_EXTRACT_SCALAR(payload, '$._font_details.OS2.achVendID'))
 );
 
 -- Infer scripts from codepoints. Used in SCRIPTS.
--- a/sql/2024/fonts/common.sql
+++ b/sql/2024/fonts/common.sql
@@ -90,11 +90,37 @@ CREATE TEMPORARY FUNCTION SERVICE(url STRING) AS (
   END
 );
 
+-- Extract the color formats from a formats payload and remove spurious entries
+-- via a table-sizes payload.
+--
+-- When nonempty, it is expected that
+--
+-- * `CBDT` is larger than 2 + 2 bytes,
+-- * `COLR` is larger than 2 + 2 + 4 + 4 + 2 (+ 4 + 4 + 4 + 4 + 4) bytes,
+-- * `SVG ` is larger than 2 + 4 + 4 + 2 bytes, and
+-- * `sbix` is larger than 2 + 2 + 4 + 4 bytes.
+--
+-- For simplicity, the threshold is set to 50 bytes.
+CREATE TEMPORARY FUNCTION COLOR_FORMATS_INNER(jsonFormats STRING, jsonSizes STRING)
+RETURNS ARRAY<STRING>
+LANGUAGE js AS '''
+try {
+  const formats = JSON.parse(jsonFormats);
+  const sizes = JSON.parse(jsonSizes);
+  return formats.filter((format) => {
+    const table = `${format}    `.slice(0, 4);
+    return sizes[table] > 50;
+  });
+} catch (e) {
+  return [];
+}
+''';
+
 -- Extract the color formats from a payload.
 CREATE TEMPORARY FUNCTION COLOR_FORMATS(payload STRING) AS (
-  REGEXP_EXTRACT_ALL(
+  COLOR_FORMATS_INNER(
     JSON_EXTRACT(payload, '$._font_details.color.formats'),
-    '(?i)(sbix|CBDT|COLRv0|COLRv1|SVG)'
+    JSON_EXTRACT(payload, '$._font_details.table_sizes')
   )
 );
 
--- a/sql/2024/fonts/common.sql
+++ b/sql/2024/fonts/common.sql
@@ -14,7 +14,7 @@ CREATE TEMPORARY FUNCTION FAMILY(payload STRING) AS (
         JSON_EXTRACT_SCALAR(payload, '$._font_details.names[16]'),
         JSON_EXTRACT_SCALAR(payload, '$._font_details.names[1]')
       ),
-      r'(?i)\s+(extra|semi|ultra)?\s?(condensed)?\s?(bold|book|heavy|light|medium|thin)?$',
+      r'(?i)([\s-]?(black|bold|book|cond(ensed)?|demi|extra|heavy|italic|light|medium|regular|semi|thin|ultra))+$',
       ''
     )
   )
--- a/sql/2024/fonts/common.sql
+++ b/sql/2024/fonts/common.sql
@@ -1,21 +1,28 @@
--- Normalize a name. Used in FAMILY.
-CREATE TEMPORARY FUNCTION FAMILY_INNER(name STRING) AS (
+-- Normalize a family name. Used in FAMILY_INNER.
+CREATE TEMPORARY FUNCTION FAMILY_INNER_INNER(name STRING) AS (
   CASE
     WHEN REGEXP_CONTAINS(name, r'(?i)font\s?awesome') THEN 'Font Awesome'
     ELSE IF(LENGTH(TRIM(name)) < 3, NULL, NULLIF(TRIM(name), ''))
   END
 );
 
+-- Normalize a family name. Used in FAMILY.
+CREATE TEMPORARY FUNCTION FAMILY_INNER(name STRING) AS (
+  FAMILY_INNER_INNER(
+    REGEXP_REPLACE(
+      name,
+      r'(?i)([\s-]?(black|bold|book|cond(ensed)?|demi|ex(tra)?|heavy|italic|light|medium|narrow|regular|semi|thin|ultra|wide|\d00|\d+pt))+$',
+      ''
+    )
+  )
+);
+
 -- Extract the family name from a payload.
 CREATE TEMPORARY FUNCTION FAMILY(payload STRING) AS (
   FAMILY_INNER(
-    REGEXP_REPLACE(
-      COALESCE(
-        JSON_EXTRACT_SCALAR(payload, '$._font_details.names[16]'),
-        JSON_EXTRACT_SCALAR(payload, '$._font_details.names[1]')
-      ),
-      r'(?i)([\s-]?(black|bold|book|cond(ensed)?|demi|ex(tra)?|heavy|italic|light|medium|narrow|regular|semi|thin|ultra|wide|\d00|\d+pt))+$',
-      ''
+    COALESCE(
+      JSON_EXTRACT_SCALAR(payload, '$._font_details.names[16]'),
+      JSON_EXTRACT_SCALAR(payload, '$._font_details.names[1]')
     )
   )
 );
--- a/sql/2024/fonts/design/fonts_designer.sql
+++ b/sql/2024/fonts/design/fonts_designer.sql
@@ -2,6 +2,8 @@
 -- Question: Which designers are popular?
 -- Normalization: Sites
 
+-- INCLUDE ../common.sql
+
 WITH
 designers AS (
   SELECT
@@ -14,7 +16,8 @@ designers AS (
   WHERE
     date = '2024-07-01' AND
     type = 'font' AND
-    is_root_page
+    is_root_page AND
+    IS_PARSED(payload)
   GROUP BY
     client,
     designer
--- a/sql/2024/fonts/design/fonts_designer.sql
+++ b/sql/2024/fonts/design/fonts_designer.sql
@@ -1,6 +1,6 @@
 -- Section: Design
 -- Question: Which designers are popular?
--- Normalization: Sites
+-- Normalization: Websites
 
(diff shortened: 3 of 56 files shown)


sql/2024/fonts/development/fonts_variable_by_service.sql
```diff
@@ -0,0 +1,43 @@
+-- Section: Development
+-- Question: Who is serving variable fonts?
+
+CREATE TEMPORARY FUNCTION SERVICE(url STRING) AS (
+  CASE
+    WHEN REGEXP_CONTAINS(url, r'((use|fonts)\.typekit\.(net|com))|webfonts\.creativecloud\.com') THEN 'Adobe'
```
IvanUkhov, line 6: @bramstein, is the last one attributed correctly? It was under `edgefonts.net` in 2022.

IvanUkhov: And in general, does this list need any update since 2022?

bramstein: Edge Fonts is a separate (free) web font service by Adobe, but it has been deprecated. As far as I know it still serves fonts however.

I'm not aware of any new web font service used at a scale significant enough that it would show up here.

IvanUkhov: I see. Where should I put `webfonts.creativecloud.com`? Is it part of Edge Fonts?

IvanUkhov: Or we can group Edge Fonts with Adobe if appropriate.

sql/2024/fonts/performance/fonts_service.sql
bramstein, line 1: In the 2022 chapter we showed a graph of service usage over time:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2022/fonts/font_usage_by_service.sql

I think that would be interesting to show as it seems there has been some change in 2023.

sql/2024/fonts/performance/fonts_service.sql
```diff
@@ -0,0 +1,45 @@
...
+    date = '2024-07-01' AND
+    type = 'font'
+  GROUP BY
+    client,
+    service
+)
+
+SELECT
+  client,
+  service,
```
bramstein, line 35: The 2022 query also included data on whether a page used a combination of self-hosted or hosted fonts. Can we include that again? It can paint a very different picture...

sql/2024/fonts/common.sql
```diff
@@ -0,0 +1,94 @@
+-- Normalize a name. Used in FAMILY.
+CREATE TEMPORARY FUNCTION FAMILY_INNER(name STRING) AS (
+  IF(LENGTH(TRIM(name)) < 3, NULL, NULLIF(TRIM(name), ''))
+);
+
+-- Extract the family name from a payload.
+CREATE TEMPORARY FUNCTION FAMILY(payload STRING) AS (
+  FAMILY_INNER(JSON_EXTRACT_SCALAR(payload, '$._font_details.names[1]'))
```
bramstein, line 8: We should probably pull this from the CSS as in 2022, as it is very common to remove the name table entries for web fonts (i.e. why a lot of the entries in the data are empty).

sql/2024/fonts/common.sql
```diff
@@ -0,0 +1,94 @@
...
+);
+
+-- Extract the file format from a URL and a Content-Type header.
+CREATE TEMPORARY FUNCTION FILE_FORMAT(url STRING, header STRING) AS (
+  LOWER(COALESCE(
```
bramstein, line 13: This is different from previous years, what's the rationale for the change? Are they functionally equivalent? If not, we should make a note in the chapter explaining why it has been changed.

bramstein: The percentages are quite different from 2022, so we should definitely look into it.

sql/2024/fonts/performance/pages_link_relationship.sql
bramstein, line 1: This should check if the resource hint is for a font, as resource hints can be used for multiple resources. If I'm reading this correctly, this just summarises all resource hint usage instead of all resource hint usage for fonts.

sql/2024/fonts/performance/styles_font_display.sql
bramstein, line 1: Would love to have the 2023 data here; seems there has been a big change...

sql/2024/fonts/performance/scripts_font_face.sql
bramstein, line 1: Would be interesting to get the 2023 data as well. I don't think we included this in the chapter in 2022, but if there's been a significant shift it would be interesting to include.

sql/2024/fonts/common.sql
IvanUkhov: I did not change this aspect. There are queries where the name is taken from the CSS, and there are queries where the name is taken from the naming table. This function is used in the latter queries.

sql/2024/fonts/common.sql
IvanUkhov: You mean this, right?

https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2022/fonts/font_format_usage.sql#L5

There is no `mimeType` in `requests` anymore, but there are request headers. So I take `Content-Type`:

https://github.com/IvanUkhov/almanac.httparchive.org/blob/iu-fonts/sql/2024/fonts/performance/fonts_format_file.sql#L18-L19

IvanUkhov: And there is no more `ext` either. So I take the show URL.

https://github.com/IvanUkhov/almanac.httparchive.org/blob/iu-fonts/sql/2024/fonts/performance/fonts_format_file.sql#L8

IvanUkhov: Added two pages showing the same two queries but with the original regular expression. Please take a look:

Skärmavbild 2024-08-12 kl  09 43 00

JannisBush: @IvanUkhov 
`mimeType` and `ext` still exists, it moved to `summary` in requests: https://har.fyi/reference/structs/request-summary/
So yo can do `JSON_VALUE(summary, '$.mimeType') as mimeType` to access the mimeType and any other value of `summary`.

IvanUkhov: Thank you, Jannis! Totally missed that. I have now updated both the two queries and the two pages in the spreadsheet. I will also update the queries regarding the file size; I also took that from the headers. Instead of the summary field.

sql/2024/fonts/performance/pages_link_relationship.sql
IvanUkhov: Yes, I removed that on purpose. Here is the original query:

https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2022/fonts/resource_hints_usage.sql

The way I interpret is that it extracts all hints from all pages and then zooms in on those pages where there are fonts, which is probably not what you have in mind.

IvanUkhov: Maybe we want to then update the logic by strictly focusing on the HTML content and picking only those hints where `as="font"` or something?

sql/2024/fonts/performance/fonts_service.sql
IvanUkhov: I have added data for July 2022 and July 2023. The table was not backfilled for dates prior to 2022.

sql/2024/fonts/performance/fonts_service.sql
IvanUkhov: I have updated the query to show the actual combinations. Please take a look.

sql/2024/fonts/common.sql
IvanUkhov: More specifically, it is used in the following queries, which are based solely on the font files:

- design/fonts_family_by_foundry.sql,
- design/fonts_family_by_script.sql,
- development/fonts_color_family.sql,
- development/fonts_variable_family.sql, and
- performance/fonts_family_by_service.sql.

sql/2024/fonts/performance/pages_link_relationship.sql
IvanUkhov: Pushed a change to check `as` as well as `href` in the case of `dns-prefetch` and `preconnect`.

sql/2024/fonts/performance/scripts_font_face.sql
IvanUkhov: Added July 2022 and July 2023.

tunetheweb:
What's the difference between this one and #3667 ?

IvanUkhov:
As I explained there, I did not notice there was a PR open and created a new one. And even if I did, I would probably have to do that anyway to have more control. In that other PR, the queries from 2022 are copied over to the folder for 2024 with no change. So there is no progress there.

sql/2024/fonts/performance/styles_font_display.sql
IvanUkhov: Turns out `parsed_css` was not backfilled either. So it gave no data for 2022 and 2023.

Skärmavbild 2024-08-15 kl  10 39 43

sql/2024/fonts/design/fonts_foundry.sql
```diff
@@ -0,0 +1,48 @@
...
+    rank <= 100
+),
+pages AS (
+  SELECT
+    client,
+    COUNT(DISTINCT page) AS total
+  FROM
+    `httparchive.all.requests`
+  WHERE
+    date = '2024-07-01'
```
mgifford, line 31: Are we using the '2024-07-01' or '2024-06-01'?

IvanUkhov: The former for fonts. We did some crawler changes after the latter.

sql/2024/fonts/performance/fonts_size.sql
```diff
@@ -0,0 +1,33 @@
+-- Section: Performance
+-- Question: What is the distribution of the file size?
+
+WITH
+fonts AS (
+  SELECT
+    client,
+    url,
+    AVG(PARSE_NUMERIC(JSON_EXTRACT_SCALAR(summary, '$.respBodySize'))) AS size
```
bramstein, line 9: We should probably take format into account here as just using WOFF vs. WOFF2 would throw off comparisons.

bramstein: (and in the other file size related queries as well)

sql/2024/fonts/performance/fonts_format_file.sql
```diff
@@ -0,0 +1,28 @@
+-- Section: Performance
+-- Question: Which file formats are used?
+
+-- INCLUDE ../common.sql
+
+SELECT
```
bramstein, line 6: The numbers are *very* different from 2022, has anything else changed?

sql/2024/fonts/development/fonts_compiler.sql
```diff
@@ -0,0 +1,48 @@
+-- Section: Design
+-- Question: Which compilers are used?
+
+CREATE TEMPORARY FUNCTION COMPILER(version STRING) AS (
+  CASE
+    WHEN REGEXP_CONTAINS(version, r'(?i)(Core \d|PS \d|hotconv|makeotf)') THEN 'Adobe Font Development Kit for OpenType'
+    WHEN REGEXP_CONTAINS(version, r'(?i)FontCreator') THEN 'FontCreator'
+    WHEN REGEXP_CONTAINS(version, r'(?i)Fontself') THEN 'Fontself Maker'
+    WHEN REGEXP_CONTAINS(version, r'(?i)ttfautohint') THEN 'FreeType'
```
bramstein, line 9: I don't think we should attribute ttfautohint to FreeType, as neither are font compilers and the actual "compiling" has been done by one of the other tools.

sql/2024/fonts/development/fonts_compiler.sql
```diff
@@ -0,0 +1,48 @@
...
+    WHEN REGEXP_CONTAINS(version, r'(?i)(FEAKit|Glyphs)') THEN 'Glyphs.app'
+    WHEN REGEXP_CONTAINS(version, r'(?i)gftools') THEN 'Google Fonts Tools'
```
bramstein, line 11: Maybe this should be fontmake?

sql/2024/fonts/development/styles_font_variant.sql
```diff
@@ -0,0 +1,71 @@
+-- Section: Development
+-- Question: Which features are used via font-variant in CSS?
+
+CREATE TEMPORARY FUNCTION PROPERTIES(json STRING)
```
bramstein, line 4: The previous version used to compare `font-variant` settings vs. `font-feature-settings`. The `font-feature-settings` data seems to be missing.

bramstein: Nevermind, I see that is now in `style_feature_control.sql`.

sql/2024/fonts/development/fonts_compiler.sql
IvanUkhov: Hm, that is true. Removed.

sql/2024/fonts/development/fonts_compiler.sql
IvanUkhov: Renamed. Renamed also “Adobe Font Development Kit” to simply “makeotf.”

sql/2024/fonts/performance/fonts_format_file.sql
IvanUkhov: Changed the normalization to “Fonts on pages.”

sql/2024/fonts/performance/styles_font_display.sql
IvanUkhov: You mean this, right?

Skärmavbild 2024-09-05 kl  14 34 50

IvanUkhov: `swap` is now around 45%, and `block` 23%. It was noted in 2022, that the growth was 11%.

JannisBush:
Are all the `force-pushes` intentionally?
image
Clicking on my notifications I see this which is a bit annoying.

sql/2024/fonts/performance/fonts_size.sql
```diff
@@ -0,0 +1,35 @@
...
+)
+
+SELECT
+  client,
+  percentile,
+  COUNT(0) AS count,
+  ROUND(APPROX_QUANTILES(size, 1000)[OFFSET(percentile * 10)]) AS size
+FROM
+  fonts,
+  UNNEST([10, 25, 50, 75, 90, 99]) AS percentile
```
bramstein, line 29: @IvanUkhov We used `10, 25, 50, 75, 90, 100` in the 2022 query. I think this is better though, but we should make a note of it in the chapter.

IvanUkhov: Yeah, the 100th percentile is kind of meaningless. It just picks up outliers, which one is not going to show in the report anyway. I am even regretting I did not go lower to, say, 95 in all queries calculating percentiles.

tunetheweb: We normally show 10, 25, 50, 75, 90. The 100 is for the top stat which can some times make for a funny throw away line (“Shockingly the largest amount of fonts loaded by a single page is 123,456”) but we don’t normally include them in the actual charts (in part because they often skew the access).

bramstein: Yea, we didn't use the 100th percentile in the 2022 article, but it was in the spreadsheet. I was just pointing out the difference in the query as we were internally discussing the difference between the 2022 data and 2024 data which I think can be explained by this change.

tunetheweb: Well, given my understanding for what this is used for in the past, I'm not sure 99 (or 95) is better than 100. But  up to you all!

sql/2024/fonts/README.md
```diff
@@ -1,20 +1,54 @@
...
+
+```sql
+-- INCLUDE ../common.sql
+```
 
-- [📄 Planning doc][~google-doc]
-- [📊 Results sheet][~google-sheets]
-- [📝 Markdown file][~chapter-markdown]
+It signalizes that, prior to executing the query, `common.sql` has to be
+inlined.
```
tunetheweb, line 50: The queries are linked from the Web Almanac and this won’t be understood.

I think we should remove the common.sql files and just inline them in the files.

tunetheweb (changes requested):

sql/2024/fonts/common.sql
```diff
@@ -0,0 +1,123 @@
...
+AS r"""
+if (codepoints && codepoints.length) {
+  return detectWritingScript(codepoints.map((character) => parseInt(character, 10)), 0.05);
+} else {
+  return [];
+}
+""";
+
+-- Infer scripts from a payload.
+CREATE TEMPORARY FUNCTION SCRIPTS(payload STRING) AS (
```
tunetheweb, line 61: I worry about people confusing scripts with JavaScript (I think I did earlier in the chat for one!). Should we call this something else? WRITING_SCRIPT? FONT_SCRIPT? CODEPOINTS?

sql/2024/fonts/design/fonts_family_by_script.sql
```diff
@@ -0,0 +1,47 @@
...
+    FAMILY(payload) AS family,
+    COUNT(0) OVER (PARTITION BY client) AS total
+  FROM
+    `httparchive.all.requests`
+  WHERE
+    date = '2024-07-01' AND
+    type = 'font' AND
+    is_root_page AND
+    IS_PARSED(payload) AND
+    RAND() > 0.5
```
tunetheweb, line 21: Why? Can you add a comment to explain?

sql/2024/fonts/design/fonts_family_by_script.sql
```diff
@@ -0,0 +1,47 @@
...
+)
+
+SELECT
+  client,
+  script,
+  family,
+  2 * COUNT(0) AS count,
+  2 * total AS total,
```
tunetheweb, line 29: Why? Can you add a comment to explain (same for line above).

sql/2024/fonts/design/fonts_designer.sql
```diff
@@ -0,0 +1,52 @@
...
+designers AS (
+  SELECT
+    client,
+    NULLIF(TRIM(JSON_EXTRACT_SCALAR(payload, '$._font_details.names[9]')), '') AS designer,
+    COUNT(DISTINCT page) AS count,
+    ROW_NUMBER() OVER (PARTITION BY client ORDER BY COUNT(DISTINCT page) DESC) AS rank
+  FROM
+    `httparchive.all.requests`
+  WHERE
+    date = '2024-07-01' AND
```
tunetheweb, line 17: I thought we were using 2024-06-01 for web Almanac. Is there a reason we can’t use this?

sql/2024/fonts/design/fonts_metric.sql
```diff
@@ -0,0 +1,133 @@
...
+  percentile,
+  COUNT(0) AS count,
+  APPROX_QUANTILES(
+    IF(name = 'granularity', value, SAFE_DIVIDE(value, granularity)),
+    1000
+  )[OFFSET(percentile * 10)] AS value
+FROM
+  fonts,
+  UNNEST(metrics) AS metric,
+  UNNEST([10, 25, 50, 75, 90, 99]) AS percentile
```
tunetheweb, line 125: As discussed in chat would 100 be better than 99? It’s kinda meaningless at this extremes percentile anyway but highlighting the “worst use” can sometimes lead to interesting insights whereas I don’t think there’s anything of interest in p99 (since it’s highly volatile but also not that worse case).

sql/2024/fonts/development/fonts_color_color.sql
```diff
@@ -0,0 +1,63 @@
...
+  return [];
+}
+''';
+
+WITH
+fonts AS (
+  SELECT
+    client,
+    url,
+    COLORS(JSON_EXTRACT(ANY_VALUE(payload), '$._font_details.color.palettes')) AS colors,
```
tunetheweb, line 34: Why ANY_VALUE?

sql/2024/fonts/development/fonts_color_entry.sql
```diff
@@ -0,0 +1,45 @@
...
+-- INCLUDE ../common.sql
+
+WITH
+fonts AS (
+  SELECT
+    client,
+    url,
+    SAFE_CAST(
+      JSON_EXTRACT_SCALAR(
+        ANY_VALUE(payload),
```
tunetheweb, line 14: Ditto

sql/2024/fonts/development/fonts_color_color.sql
IvanUkhov: So we group by the client and URL. For that combination, there are typically many entries corresponding to requests performed from different pages and at different times. It is then assumed the server's response does not depend on the location and time, and we just take any one here and in all other queries with the same grouping.

sql/2024/fonts/design/fonts_designer.sql
IvanUkhov: We changed the crawler after the 2024-06-01 data were collected.

sql/2024/fonts/design/fonts_family_by_script.sql
IvanUkhov: BigQuery was running out of resources, and I was downsampling a little to make it go through. I have just rerun the query without the downsampling, and it succeeded. Perhaps it was something temporary. I have removed this code now.

sql/2024/fonts/common.sql
IvanUkhov: We have only one query dealing with JavaScript scripts (`performance/scripts_font_face.sql`). I think there will be little confusion what _script_ refers to when browsing the queries for the Fonts chapter. I am also assuming that in the article, one will be careful to explain.

sql/2024/fonts/design/fonts_designer.sql
tunetheweb: @nrllh FYI, just making sure you're cool with this?

sql/2024/fonts/README.md
IvanUkhov: Is it a hard rejection, or can this be discussed? We arguably want the queries to be easy to understand and easy to maintain. Duplicating the common functions at the top of each query will clutter the code and make it difficult to maintain, as one would have to keep track of the functions and keep them in sync.

If you want, we can replace

```sql
-- INCLUDE ../common.sql
```

with

```sql
-- INCLUDE https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2024/fonts/common.sql
```

Another alternative found be do create permanent SQL functions in a namespace in BigQuery dedicated to the chapter. Someone with the necessary permissions would have to do that.

sql/2024/fonts/design/fonts_metric.sql
IvanUkhov: I think everything above the 90th will be ignored in the body of the chapter. I lowered it just to have something more sensible there; there will always be a handful of corky fonts with unrealistic parameters set.

sql/2024/fonts/development/fonts_color_color.sql
tunetheweb: Oh OK. I expected to look at fonts per site, but here we're looking at fonts overall (no matter which site requests them). OK.

sql/2024/fonts/README.md
tunetheweb: I think that full path of the include would be a bit better.

tunetheweb (approved):

tunetheweb:
Let me know when good to merge.

sql/2024/fonts/design/fonts_designer.sql
nrllh: That's okay IMO

IvanUkhov:
@tunetheweb, Bram and I concluded that it can be merged now. Thank you.

tunetheweb (approved):

Issue #3602, opened 2024-03-02
Performance 2024


nrllh:
# Performance 2024

![Performance illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/performance/hero_lg.jpg)

If you're interested in contributing to the Performance chapter of the 2024 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide). You might be interested in exploring the changes to this year's version [here](https://github.com/HTTPArchive/almanac.httparchive.org/discussions/3619).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| ---- | ------- | --------- | -------- | ------- |    -------  |
| @ines-akrap  | @ines-akrap, @imeugenia   | @rviscomi, @siakaramalegos   | @kevinfarrugia, @guaca   | @Cherry  | @turban1988  | 

<details>
<summary><b>Expand for more information about each role 👀</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
[...]

PR #3712, opened 2024-08-02
Performance 2024 Queries

kevinfarrugia:
Makes progress on #3602

### DIFF
--- a/sql/2024/performance/lcp_resource_type.sql
+++ b/sql/2024/performance/lcp_resource_type.sql
@@ -7,7 +7,7 @@ WITH lcp AS (
     # Parse anchors out of LCP URLs.
     REGEXP_EXTRACT(JSON_VALUE(custom_metrics, '$.performance.lcp_elem_stats.url'), r'([^#]*)') AS url
   FROM
-    `httparchive.all.pages` TABLESAMPLE SYSTEM (0.01 PERCENT)
+    `httparchive.all.pages`
   WHERE
     date = '2024-06-01' AND
     is_root_page


sql/2024/performance/web_vitals_by_technology.sql
```diff
@@ -0,0 +1,208 @@
...
+),
+
+tech AS (
+  SELECT
+    _TABLE_SUFFIX AS client,
+    category,
+    app AS technology,
+    url AS page
+  FROM
+    `httparchive.technologies.2024_06_01_*`
```
mgifford, line 65: This should be using `httparchive.all.pages` as it has better data and is being maintained going ahead.

mgifford:
These tests seem to calculate their own web core vitals data, rather than just using the score that is provided by Google Lighthouse.  Is there a reason to calculate them separately?   
  
I think it could be easily added to so that we're pulling the performance data at the same time as accessibility  sql/2024/accessibility/lighthouse\_a11y\_score.sql  
https://github.com/HTTPArchive/almanac.httparchive.org/pull/3749/files#diff-6d6ec13b6d0025840195d847d1e13775c6fecb660c65f79ab449209cab0d8589

kevinfarrugia:
> These tests seem to calculate their own web core vitals data, rather than just using the score that is provided by Google Lighthouse.  Is there a reason to calculate them separately? 
> 
> I think it could be easily added to so that we're pulling the performance data at the same time as accessibility  sql/2024/accessibility/lighthouse_a11y_score.sql https://github.com/HTTPArchive/almanac.httparchive.org/pull/3749/files#diff-6d6ec13b6d0025840195d847d1e13775c6fecb660c65f79ab449209cab0d8589

Sorry I didn't notice this earlier. We should not be calculating our own web vitals data but either using the page-level data from `httparchive.all.pages` (field: `payload._CrUX`) or origin level data from the CrUX dataset, namely `chrome-ux-report.materialized.device_summary`.

tunetheweb:
@kevinfarrugia is this good to merge now?

sql/2024/performance/bfcache_cachecontrol_nostore.sql
```diff
@@ -0,0 +1,29 @@
...
+    LOGICAL_OR(HAS_NO_STORE_DIRECTIVE(JSON_VALUE(payload, '$._cacheControl'))) AS includes_ccns
+  FROM
+    `httparchive.all.requests`
+  WHERE
+    date = '2024-06-01' AND
+    is_main_document
+  GROUP BY
+    client,
+    page
+)
```
tunetheweb, line 17: Should we group by `root_page` rather than page? That way you can report by site rather than per page (given HTTP Archive is limited to two pages per site).

sql/2024/performance/bfcache_unload.sql
```diff
@@ -0,0 +1,41 @@
...
+  SELECT
+    client,
+    page,
+    rank,
+    getUnloadHandler(JSON_EXTRACT(lighthouse, '$.audits.deprecations')) AS has_unload
+  FROM
+    `httparchive.all.pages`
+  WHERE
+    date = '2024-06-01' AND
+    is_root_page
```
tunetheweb, line 21: So for unload you've used home pages, but for CCNS you've used all pages. Should they use the same methodology?

tunetheweb (commented):
Overall LGTM but a count of comments. Let me know when good to merge.

sql/2024/performance/cls_animations.sql
```diff
@@ -0,0 +1,31 @@
+# TODO: Research Lighthouse audits.non-composited-animations.details.items
+# TODO: What is the meaning of pages and total columns
```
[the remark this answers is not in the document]
kevinfarrugia: Will clean up.

sql/2024/performance/bfcache_cachecontrol_nostore.sql
kevinfarrugia: Updated to use `root_page`. Updated results: https://docs.google.com/spreadsheets/d/15038wEIoqY53Y_kR8U6QWM-PBO31ZySQGi147ABTNBc/

kevinfarrugia: Change in plans. Will use `page` and update the unload handler to use `page`. This is more aligned with the previous version (2022) and the writing.

sql/2024/performance/bfcache_unload.sql
kevinfarrugia: Agreed and updated.

tunetheweb (approved):

PR #3713, opened 2024-08-05
Uploading Image loading property usage SQL

henryp25:
Adding first SQL query for image-loading-property-usage-2024


tunetheweb (changes requested):
Left some comments, but can you add an overall SEO 2024 PR and leave it in draft and add each of the SQL to that one PR? Then we can review once all done.

tunetheweb:
Looks like this is included in #3791

PR #3717, opened 2024-08-07
robots-txt-user-agent-usage.sql

cnichols013:


tunetheweb (commented):
Similar to comments in #3713 can we move this to a common SEO 2024 PR? Can leave in draft for now until all queries are ready.

tunetheweb:
`TABLESAMPLE` is used to restrict rows for quicker testing. The submitted queries shouldn't use this as then it won't be the full dataset.

tunetheweb:
Looks like this is included in #3791

Issue #3596, opened 2024-03-02
Media 2024


nrllh:
# Media 2024

![Media illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/media/hero_lg.jpg)

If you're interested in contributing to the Media chapter of the 2024 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide). You might be interested in exploring the changes to this year's version [here](https://github.com/HTTPArchive/almanac.httparchive.org/discussions/3619).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| ---- | ------- | --------- | -------- | ------- |    -------  |
| @stefanjudis | @stefanjudis | @svgeesus, @nhoizey, @eeeps   | @foolip, @nucliweb, @eeeps, @scottjehl| @MichaelLewittes  | @turban1988  | 

<details>
<summary><b>Expand for more information about each role 👀</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
[...]

PR #3738, opened 2024-08-17
Media 2024 queries

stefanjudis:
This PR includes the updated queries for the media chapter. #3596

### DIFF
--- a/sql/2024/media/animated_gif_bpp.sql
+++ b/sql/2024/media/animated_gif_bpp.sql
@@ -1,0 +1,82 @@
+#standardSQL
+# what is the bpp of animated vs non-animated GIFs?
+# animated_gif_bpp.sql
+
+
+# ImageMagick reports big images as having, e.g., "1.29097M" pixels. This means ~1.2 million pixels, but BigQuery doesn't know that.
+CREATE TEMPORARY FUNCTION magickPixels(imageMagickNumberString STRING)
+RETURNS INT64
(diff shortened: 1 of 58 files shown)


mgifford:
How do you get around the of blocked regex 'TABLESAMPLE' error? I don't actually know what it does mind you...

stefanjudis:
@mgifford I'm sorry, I don't understand. What error?

tunetheweb:
`TABLESAMPLE SYSTEM (0.001 PERCENT)` says only run this on 0.001% of the rows in the table. This is good during query development as faster (and cheaper!) to run than on the full dataset.

However, eventually we will want all queries run on the full dataset. Therefore there is a linting rule warning this is being used and we should not merge this PR until that is fixed.

So it's fine to ignore it for now (especially as this PR is still in Draft status) if you're still working on those queries. It just means you won't see a nice green tick in the linting check for now and also that you might miss other linting errors unless you explicitly look at those checks to see if anything else needs fixing too.

stefanjudis:
Aaaah, I didn't realize there was a linting error. Thanks @tunetheweb!

I've run most of the queries against the full data set already. and will double check that this is true!

tunetheweb (approved):
Looks pretty good to me. Still using some older tables in some queries but let's not perfection be the enemy of the good!

Let me know if good to merge, but I think you might still be working on some queries?

PR #3740, opened 2024-08-18
python 3.12

max-ostapenko:
When adding analytical dependencies to the `src/requirements.txt` I [failed tests](https://github.com/HTTPArchive/almanac.httparchive.org/actions/runs/10396895786/job/28791696553) as the repo is on Python version 3.8, which is not compatible with some recent package versions (e.g. `pandas`).

3.8 will [soon be deprecated](https://devguide.python.org/versions/).

I adjusted the version.

### DIFF
--- a/.github/workflows/code-static-analysis.yml
+++ b/.github/workflows/code-static-analysis.yml
@@ -33,11 +33,11 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-      - name: Set up Python 3.8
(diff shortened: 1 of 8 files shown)


tunetheweb (approved):
LGTM

Any reason not to go to 3.12?

max-ostapenko:
Yeah, you're right.
At the first sight 'bugfix' version sounded too sketchy, but it's what they call a stable release.

tunetheweb (approved):

PR #3653, opened 2024-05-03
Privacy 2024 queries

max-ostapenko:
Makes progress on #3603

[Analysis plan details](https://docs.google.com/document/d/1WJT9kfKHxwNl5HNAhIddefWC0vCvH3DUIxajT1gQWao/edit)

### Queries
  
  Bounce tracking:
  - [x] number_of_websites_with_bounce_tracking.sql
  
  CNAME
  - [x] most_common_cname_domains.sql
  
  IAB consent frameworks:
  - [x] most_common_countries_for_iab_tcf_v2.sql
  - [x] most_common_referrer_policy.sql
  - [x] most_common_strings_for_iab_usp.sql
  - [x] number_of_websites_with_iab.sql
   
  GPC prevalence:
  - [x] number_of_websites_with_gpc.sql
  
  CMPs presence
  - [x] most_common_cmps_for_iab_tcf_v2.sql
  
  ads.txt & sellers.json:
  - [x] ads_and_sellers_graph.sql
  - [x] ads_lines_amount.sql
  - [x] ads_seller_accounts_by_type.sql
  - [x] common_ads_variables.sql
  - [x] top_direct_sellers.sql
  
  Privacy Sandbox:
  - [x] number_of_websites_with_related_origin_trials.sql
  - [x] privacy-sandbox-adoption-by-third-parties-by-publishers.sql
  - [x] number_of_privacy_sandbox_attested_domains.sql
  - [x] number_of_ara_destinations_registered_by_third_parties_and_publishers.sql
  - [x] top_ara_destinations_registered_by_most_publishers.sql
  - [x] top_ara_destinations_registered_by_most_third_parties.sql

  CCPA:
  - [x] ccpa_most_common_phrases.sql
  - [x] ccpa_prevalence.sql 

  Fingerprinting:
  - [x] fingerprinting_most_common_apis.sql
  - [x] fingerprinting_most_common_scripts.sql
  - [x] fingerprinting_script_count.sql

  Cookies:
  - [x] cookies_top_first_party.sql
  - [x] cookies_top_third_party.sql

  Other:
  - [x] number_of_websites_with_dnt.sql
  - [x] most_common_client_hints.sql
  - [x] number_of_websites_per_tracking_technology.sql
  - [x] number_of_websites_with_client_hints.sql
  - [x] number_of_websites_with_privacy_service.sql
  - [x] number_of_websites_with_referrerpolicy.sql
  - [x] number_of_websites_with_related_origin_trials.sql
[...]

Issue #3617, opened 2024-03-02
Cookies 2024


nrllh:
# Cookies 2024

If you're interested in contributing to the Cookies chapter of the 2024 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide). You might be interested in exploring the changes to this year's version [here](https://github.com/HTTPArchive/almanac.httparchive.org/discussions/3619).

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| ---- | ------- | --------- | -------- | ------- |    -------  |
| @ydimova  | @yohhaan, @ydimova, @samdutton    |  @samdutton  @rowan-m |  @yohhaan   | - | @tomvangoethem |

<details>
<summary><b>Expand for more information about each role 👀</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
- **[Analysts](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide)** are responsible for researching the stats and trends used throughout the Almanac. Analysts work closely with authors and reviewers during the planning phase to give direction on the types of stats that are possible from the dataset, and during the analyzing/writing phases to ensure that the stats are used correctly.
[...]

PR #3741, opened 2024-08-18
Cookies sql 2024

yohhaan:
Queries for the Cookies Chapter 2024 (#3617)

### Extract cookies into intermediate table to reduce size of data processed

- [x] 0_create_desktop_cookies.sql
- [x] 0_create_mobile_cookies.sql

### Prevalence Cookies type and attributes
- [x] prevalence_attributes_per_type.sql
- [x] prevalence_type_attributes_per_rank.sql

### Top Cookies of each type and top domains setting the most cookies
- [x]  top_20_domains_setting_cookies.sql
- [x]  top_20_first_party_cookies.sql
- [x]  top_20_third_party_cookies.sql

### Nb cookies 
- [x]  nb_cookies_cdf.sql
- [x]  nb_cookies_per_type_quantiles.sql
- [x]  nb_cookies_quantiles.sql

### Size cookies
- [x]  size_cookies_cdf.sql
- [x]  size_cookies_per_type_quantiles.sql
- [x]  size_cookies_quantiles.sql
- [x]  size_extract_largest.sql

### Age of cookies 
- [x]  age_expire_cookies_per_type_quantiles.sql
- [x]  age_expire_cookies_quantiles.sql
- [x] age_expires_cookies_cdf.sql

### New Privacy Sandbox APIs

- CHIPS:
   - [x]  CHIPS_top_20_first_party_cookies.sql
   - [x]  CHIPS_top_20_third_party_cookies.sql

- who is using them?
   - [x] see status in #3653
   
- RWS & Attestation File
   - [x] get list of domains with potentially the corresponding well-known file set - see  #3653
   - [x] Use of this custom crawler  from https://privacysandstorm.com/datasets_tools/well_known_crawler/  to parse and check that well-known files are actually valid 
   
   
   cc @ydimova, @shaoormunir, @samdutton @ChrisBeeti


sql/2024/cookies/0_create_desktop_cookies.sql
```diff
@@ -0,0 +1,39 @@
...
+    rank,
+    JSON_VALUE(summary, '$.startedDateTime') AS startedDateTime,
+    cookie
+  FROM
+    `httparchive.all.pages`,
+    UNNEST(JSON_EXTRACT_ARRAY(custom_metrics, '$.cookies')) AS cookie
+  WHERE
+    date = '2024-06-01' AND
+    client = 'desktop' AND
+    rank <= 1000000
```
tunetheweb, line 19: Any reason to restrict it like this? Can we add a comment to explain why?

sql/2024/cookies/0_create_mobile_cookies.sql
```diff
@@ -0,0 +1,39 @@
...
+    rank,
+    JSON_VALUE(summary, '$.startedDateTime') AS startedDateTime,
+    cookie
+  FROM
+    `httparchive.all.pages`,
+    UNNEST(JSON_EXTRACT_ARRAY(custom_metrics, '$.cookies')) AS cookie
+  WHERE
+    date = '2024-06-01' AND
+    client = 'mobile' AND
+    rank <= 1000000
```
tunetheweb, line 19: Ditto.

sql/2024/cookies/0_create_desktop_cookies.sql
```diff
@@ -0,0 +1,39 @@
+-- Create an intermediate table containing all cookies that were set during the
+-- <DATE> crawl on <CLIENT> when visiting sites of rank <= <RANK>. This table
+-- can then be reused in consecutive queries without having to reextract the
+-- data every time
+-- Export the table as httparchive.almanac.DATE_CLIENT_RANK_cookies
+
+WITH intermediate_cookie AS (
+  SELECT
+    page,
```
tunetheweb, line 9: Do we want this to restrict this to just root_pages? Or sum up cookies across both home and secondary pages for the same `root_page` so we can count sites using cookies?

I see you use `NET.HOST(page)` later on but `root_page` would perhaps be better and allow this filtering up front.

sql/2024/cookies/prevalence_type_attributes_per_rank.sql
```diff
@@ -0,0 +1,26 @@
...
+  SUM(IF(session = 'true', 1, 0)) / COUNT(0) AS session,
+  SUM(IF(sameParty = 'true', 1, 0)) / COUNT(0) AS sameParty,
+  SUM(IF(sameSite = 'Lax', 1, 0)) / COUNT(0) AS sameSiteLax,
+  SUM(IF(sameSite = 'None', 1, 0)) / COUNT(0) AS sameSiteNone,
+  SUM(IF(sameSite = 'Strict', 1, 0)) / COUNT(0) AS sameSiteStrict,
+  SUM(IF(sameSite IS NULL, 1, 0)) / COUNT(0) AS sameSiteNull,
+  SUM(IF(partitionKey IS NOT NULL, 1, 0)) / COUNT(0) AS partitionKey,
+  SUM(IF(partitionKeyOpaque IS NOT NULL, 1, 0)) / COUNT(0) AS partitionKeyOpaque
+FROM `httparchive.almanac.DATE_CLIENT_RANK_cookies`,
+  UNNEST([1000, 5000, 10000, 50000, 100000, 500000, 1000000]) AS rank_grouping
```
tunetheweb, line 19: We normally don't include the mid-steps in Web Almanac and just report in the "10s":

```suggestion
  UNNEST([1000, 10000, 100000, 1000000, 10000000, 100000000]) AS rank_grouping
```

Also added two more groupings just in case this is expanded above.

sql/2024/cookies/0_create_desktop_cookies.sql
yohhaan: This was to reduce the amount of data processed/parsed; without any cap on the rank, bigquery was returning an error "Response too large to return" or something like that. A solution is to specify a table where to output the results, but I don't have the permissions on the HTTP Archive project to create that. 

So, I created the table on a personal project/dataset, but then because I am on the free plan, there are some limits on amount of data I can store, and cookies on top 1M for both mobile and desktop were right below that limit, and top 1M seem plenty of enough data already.

Ideally, these cookies should not be in custom metrics but in their own separate table in my opinion. I believe I saw at some point an issue/discussion on HTTPArchive project where it was proposed to break down some fields in custom metrics in their table/column, instead of the current big blob.

sql/2024/cookies/0_create_desktop_cookies.sql
yohhaan: I tried to be as general as possible when extracting the cookies table, and I did not know up front if we were going to only stick to `root_page` or no for this year's queries, same for in the future.

sql/2024/cookies/0_create_desktop_cookies.sql
tunetheweb: I can this table, which should be much cheaper to query:

```sql
-- Create an intermediate table containing all cookies that were set during the
-- <DATE> crawl on <CLIENT> when visiting sites of rank <= <RANK>. This table
-- can then be reused in consecutive queries without having to reextract the
-- data every time
-- Export the table as httparchive.almanac.DATE_CLIENT_RANK_cookies

CREATE TABLE `httparchive.almanac.cookies` 
(
  date DATE,
  client STRING,
  page STRING,
  root_page STRING,
  rank INTEGER,
  startedDateTime STRING,
  firstPartyCookie BOOL,
  name STRING,
  domain STRING,
  path STRING,
  expires STRING,
  size STRING,
  httpOnly STRING,
  secure STRING,
  session STRING,
  sameSite STRING,
  sameParty STRING,
  partitionKey STRING,
  partitionKeyOpaque STRING
)
PARTITION BY date
CLUSTER BY
  client, rank, page

AS

WITH intermediate_cookie AS (
  SELECT
    date,
    client,
    page,
    root_page,
    rank,
    JSON_VALUE(summary, '$.startedDateTime') AS startedDateTime,
    cookie
  FROM
    `httparchive.all.pages`,
    UNNEST(JSON_EXTRACT_ARRAY(custom_metrics, '$.cookies')) AS cookie
  WHERE
    date = '2024-06-01'
)

SELECT
  date,
  client,
  page,
  root_page,
  rank,
  startedDateTime,
  ENDS_WITH(NET.HOST(page), NET.REG_DOMAIN(JSON_VALUE(cookie, '$.domain'))) AS firstPartyCookie,
  JSON_VALUE(cookie, '$.name') AS name,
  JSON_VALUE(cookie, '$.domain') AS domain,
  JSON_VALUE(cookie, '$.path') AS path,
  JSON_VALUE(cookie, '$.expires') AS expires,
  JSON_VALUE(cookie, '$.size') AS size,
  JSON_VALUE(cookie, '$.httpOnly') AS httpOnly,
  JSON_VALUE(cookie, '$.secure') AS secure,
  JSON_VALUE(cookie, '$.session') AS session,
  JSON_VALUE(cookie, '$.sameSite') AS sameSite,
  JSON_VALUE(cookie, '$.sameParty') AS sameParty,
  JSON_VALUE(cookie, '$.partitionKey') AS partitionKey,
  JSON_VALUE(cookie, '$.partitionKeyOpaque') AS partitionKeyOpaque
FROM intermediate_cookie
```

Could you update this query to this:

```sql
INSERT INTO `httparchive.almanac.cookies` 
WITH intermediate_cookie AS (
  SELECT
    date,
...
```

 But no need to run as I've run it.

Then use this new `httparchive.almanac.cookies` table in your other queries? I presume it won't change your results much.

yohhaan: Nice! Thank you! Looking into the new table you created and updating the queries for the cookies chapter accordingly.

sql/2024/cookies/prevalence_type_attributes_per_rank.sql
tunetheweb: Yeah. More confusing than anything else so let’s be consistent.

yohhaan:
@tunetheweb I just updated the SQL queries for the Cookies 2024 chapter to use the table you created. Thanks!

PR #3789, opened 2024-10-10
Adds 2024 base templates

SaptakS:
* Removes methodology for no
* Add 2024 base template files
* Updated link of last chapter to 2022 instead of 2023
* Does not update stats or dates

PS: i have not made 2024 as the default year, so the link will still take to 2022, but if someone visits the URL with `/2024/` path, then this will work.

@tunetheweb I think I still need to update the stats text in base.html and the month/year when the crawler was run?

### DIFF
--- a/src/config/2024.json
+++ b/src/config/2024.json
@@ -1,8 +1,8 @@
 {
   "settings": [
     {
-      "is_live": false,
-      "supported_languages": ["en"],
+      "is_live": true,
+      "supported_languages": ["en","es","fr","hi","it","ja","nl","pt","ru","tr","uk","zh-CN","zh-TW"],
       "ebook_languages": []
     }
   ],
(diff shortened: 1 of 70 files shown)


tunetheweb:
Updated stats based on:

```sql
DECLARE thedate DATE DEFAULT '2024-06-01';

# Run this to get number of sites analysed
SELECT "ALL" AS client, COUNT(DISTINCT root_page) AS num_sites FROM `httparchive.all.pages` where date = thedate
UNION ALL
SELECT  client, COUNT(DISTINCT root_page) FROM `httparchive.all.pages` where date = thedate group by client;

# Don't run this, but just note the BigQuery estimate
SELECT * FROM `httparchive.all.pages` WHERE date = thedate;
SELECT * FROM `httparchive.all.pages` WHERE date = thedate;
SELECT * FROM `chrome-ux-report.all.202406`;
SELECT * FROM `chrome-ux-report.materialized.metrics_summary`;
SELECT * FROM `chrome-ux-report.materialized.device_summary`;
```

PR #3794, opened 2024-10-18
Sustainability2024 Chapter

ldevernay:
Fixes #3611

Staged version: https://sustainability-2024-dot-webalmanac.uk.r.appspot.com/en/2024/sustainability

This is a first version of the markdown file for the Sustainability chapter. 
We're still missing : 
- [x] Some bios from contributors
- [x] Descriptions for diagrams (left as "TODO")
- [x] Some diagrams that could not be found in the spreadsheet : responsiveness/srcset and non-composited animations per page (both identified as "TODO")
- [x] sql_file for all diagrams until PR is available for queries

PR #3799, opened 2024-10-19
Update contributors.json

mgifford:
Adding myself to this list.


tunetheweb:
You can add 2024 roles without an equivalent chapter as it checks to make sure they match. So let’s leave this til the chapter PR. I’ve already included you in #3794 so let’s close this.

PR #3812, opened 2024-10-26
Update contributors.json

charlesberret:
Added my new contributor entry under "cberret."


tunetheweb:
Can you add this to the Fonts chapter PR? Until that's submitted the repo doesn't see you as a contributor so the CI checks fail as you are down as an Author but without a corresponding chapter that you authored.

Closing this for now.

PR #3822, opened 2024-10-31
Chart resizing script

max-ostapenko:
Just moving this script from my gist to the repo.


tunetheweb (changes requested):
Can we exclude shares with a width of 600 but a different height? There some legitimate use cases for that (for example [this one](https://almanac.httparchive.org/en/2022/performance#fig-7)). But width != 600 is definitely a red flag.

tunetheweb (approved):
Haven't tested it but happy to merge if you are.

PR #3831, opened 2024-11-06
Adds analysts section in the chapters

SaptakS:
As discussed with @nrllh in slack, I am adding an analyst section below the author section. I went with this way, instead of adding a tab so that it is easier to link to it, similar to the authors, and can maintain a similar document flow.

PS: Right now, it gets added to all previous year chapters as well. Happy to add a condition so that it shows up for only this year, if that's what we want.

image

### DIFF
--- a/src/templates/base/base_chapter.html
+++ b/src/templates/base/base_chapter.html
@@ -336,6 +336,112 @@ window.discussion_url="https://discuss.httparchive.org/t/{{ metadata.get('discus
 </ul>
 {% endmacro %}
 
+{% macro render_analysts() %}
+  {% for analyst in metadata.get('analysts') %}
+    {% if loop.index == 1 %}
+    <h2 id="analysts">
+      <a href="#analysts" class="anchor-link">
+      {% if loop.length == 1 %}{{ self.analyst() }}{% endif -%}
+      {% if loop.length > 1 and loop.index == 1 %}{{ self.analysts() }}{% endif %}
+      </a>
+    </h2>
+    <ul>
+    {% endif %}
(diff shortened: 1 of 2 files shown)


tunetheweb:
I've a concern when the Analysts is the exact same as the author as the bio will be repeated twice. For example for `/en/2021/third-parties` I see this:

image

So at the very least we'd need something to dedupe that. That could be "Author and analyst" in this case but gets more complicated when they are mixed (e.g. 3 authors, 1 of whom was also the analyst).

I also don't think we should show this for previous years since we won't have any bios for them.

Even with those, I'm not 100% convinced about this change. While I do appreciate the very hard work the analysts do (often  on par with authors, and sometimes even more!), the aim of the Author bios was to emphasis the authors expertise in this area to the readers. While it also helps if an analyst is an expert, they don't necessarily _have_ to be an expert—if guided by the author well as to exactly what they want, and there's also an argument that the Reviewers are experts and so in some cases add credibility here.

Maybe we should have an "Authors" section and a "Other team members" section beneath that (to still emphasis the author's expertise first)? Or just combine them into one "Chapter team" section (where we lose the Author's expertise, but overall the expertise of the team is sufficient).

SaptakS:
> I've a concern when the Analysts is the exact same as the author as the bio will be repeated twice.

I do agree and felt the same that there is too much duplicate information. I think I would lean towards "Chapter team" with some section in each member saying what was their role in the team. But then we should include reviewers and all.

I can see your concern about taking the focus away from author's expertise and agree with it, but would be curious to know @nrllh thoughts.

tunetheweb:
Discussed with @nrllh . We'll leave this for now.

PR #3867, opened 2024-11-11
Correct paragraph on WebP support in WordPress

adamsilverstein:
Unfortunately WebP "be default"  - which was merged for 6.1 - was later reverted, see https://make.wordpress.org/core/2022/09/11/webp-in-core-for-6-1/ - so while WordPress has the capability to output WebP for uploaded images, this feature is not turned on by default.


adamsilverstein:
ps. feel free to adjust text as you see fit, I took a pass.

adamsilverstein:
ps. Interesting to see that WordPress has the highest AVIF adoption even. though we only added support this year, I wonder how AVIF adoption grew in WP since 2022.

tunetheweb:
@sirjonathan presume you're OK with this?

sirjonathan (commented):
Great catch. Thanks @adamsilverstein

PR #3896, opened 2024-11-15
2024 Accessibilty chapter corrections

mgifford:
Typo and incorrect link.  There is a duplicate image, not the bar chart that was described.

Also accessibility or Accessible was misspelled.


mgifford:
@tunetheweb I don't know how to deal with the image. The Bar Chart hadn't been published before, so I suspect this is a manual process.  Cut/paste error, sorry.

tunetheweb (approved):

PR #3897, opened 2024-11-16
fix: correctly link bsky profiles

manniL:
This PR ensures that bsky profiles are correctly link (`bsky.app/profile/<HANDLE>` instead of `bsky.app/<HANDLE>`)


tunetheweb (approved):
LGTM! Thanks!!

PR #3919, opened 2024-11-21
Update 0_extract_cookies.sql to use new httparchive.crawl dataset

yohhaan:
Switch Cookies 2024 query to new httparchive.crawl dataset


tunetheweb (approved):
Nice! Thanks.

PR #3926, opened 2024-11-25
Support Python 3.13

tunetheweb:
Noticed a couple of issues when using Python 3.13:

- pandas 2.2.2 is not compatible
- `datetime.datetime.utcnow` is deprecated and down to be removed.

Both fixes are backwards compatible to Python 3.11 (and we demand 3.12 as minimum anyway).

### DIFF
--- a/src/requirements.txt
+++ b/src/requirements.txt
@@ -7,6 +7,6 @@ pytest-cov==5.0.0
 sqlfluff==1.4.5
 
 # sql/util/* dependencies
-pandas==2.2.2
+pandas==2.2.3
 google-cloud-bigquery==3.25.0
 requests==2.32.3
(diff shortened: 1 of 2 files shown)


max-ostapenko:
In workflows we are using `runs-on: ubuntu-20.04` which has Python 3.8.10.
E.g. [predeploy.yml](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/.github/workflows/predeploy.yml)

Let's change it to '[ubuntu-latest](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md)?

max-ostapenko:
Oh, right.
So we can get the environment ready just in 1 step instead of 3
:)

max-ostapenko (approved):

tunetheweb:
> Oh, right. So we can get the environment ready just in 1 step instead of 3 :)

True. But lets leave them separate for now as gives us greater control.

PR #3932, opened 2024-11-29
Update contributors.json

hadiamjad:


tunetheweb:
I've changed this to point to the privacy branch, as need that for test.

@max-ostapenko it's saying package/package.lock are changing. I suspect this is just because your branch is out of date? Maybe merging latest `main` into that will fix it?

max-ostapenko:
Will add it with direct commit.

PR #3933, opened 2024-11-29
Update privacy.md

hadiamjad:


tunetheweb:
@max-ostapenko don't think we need this PR do we as will be merge conflicts if we merge with yours? Can you update your branch with this info?

max-ostapenko:
@hadiamjad you need to add you contacts to `src/config/contributors.json`.
See example here: https://github.com/HTTPArchive/almanac.httparchive.org/pull/3930

PR #3960, opened 2024-12-05
Performance 2024: image prioritization and discoverability

rviscomi:
Updates the prioritization and discoverability queries and includes their results in the chapter, also adds a figure for % of unsized images

I did a bit of editorializing so definitely want @ines-akrap and @imeugenia's LGTM

Once published, I'll update the corresponding links in https://web.dev/articles/top-cwv to point to the 2024 results

### DIFF
--- a/src/content/en/2024/performance.md
+++ b/src/content/en/2024/performance.md
@@ -241,7 +241,7 @@ One of the most effective ways to optimize the LCP resource load delay is to ens
 
 Unfortunately, 35% of mobile websites do not have an LCP element that is statically discoverable in the document. While this is a slight improvement over the 39% we saw in 2022, it's still a significant blocker of LCP performance.
 
-As we'll explore in the following sections, there are three primary ways that websites prevent their LCP resources from being statically discoverable: lazy-loading, CSS background images, and client-side rendering.
+As we'll explore in the following sections, there are three primary ways that websites prevent their LCP resources from being statically discoverable: lazy loading, CSS background images, and client-side rendering.
 
 #### LCP lazy-loading
 
--- a/src/content/en/2024/performance.md
+++ b/src/content/en/2024/performance.md
@@ -305,7 +305,7 @@ Another one of the most effective ways to optimize the loading delay of LCP imag
 )
 }}
 
-Adoption of LCP image prioritization skyrocketed to 15% of mobile websites in 2024, up from just 0.03% in 2022! This massive leap is thanks in large part to WordPress implementing [core support](https://make.wordpress.org/core/2023/07/13/image-performance-enhancements-in-wordpress-6-3/) for `fetchpriority` in 2023.
+Adoption of LCP image prioritization skyrocketed to 15% of mobile websites in 2024, up from just 0.03% in 2022! This massive leap is thanks in large part to WordPress implementing <a hreflang="en" href="https://make.wordpress.org/core/2023/07/13/image-performance-enhancements-in-wordpress-6-3/">core support</a> for `fetchpriority` in 2023.
 
 As amazing as it is to see such rapid growth, there is still significant room for more sites to take advantage of this impactful one-line optimization.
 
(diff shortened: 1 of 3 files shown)


src/content/en/2024/performance.md
```diff
@@ -275,6 +292,23 @@ The chart below illustrates the distribution of client-side generated content. I
...
+{{ figure_markup(
+  caption="The percent of mobile pages that use `fetchpriority=high` on their LCP image.",
+  content="15%",
+  classes="big-number",
+  sheets_gid="731441901",
+  sql_file="lcp_async_fetchpriority.sql"
+)
+}}
+
+Adoption of LCP image prioritization skyrocketed to 15% of mobile websites in 2024, up from just 0.03% in 2022! This massive leap is thanks in large part to WordPress implementing [core support](https://make.wordpress.org/core/2023/07/13/image-performance-enhancements-in-wordpress-6-3/) for `fetchpriority` in 2023.
```
tunetheweb, line 308:
```suggestion
Adoption of LCP image prioritization skyrocketed to 15% of mobile websites in 2024, up from just 0.03% in 2022! This massive leap is thanks in large part to WordPress implementing <a hreflang="en" href="https://make.wordpress.org/core/2023/07/13/image-performance-enhancements-in-wordpress-6-3/">core support</a> for `fetchpriority` in 2023.
```

src/content/en/2024/performance.md
```diff
@@ -226,11 +226,26 @@ Typically, LCP element rendering takes a long time if the LCP element hasn't bee
...
+)
+}}
 
-One of the ways to optimize the LCP resource load delay is to ensure the resource can be discovered as early as possible. If you make the resource discoverable in the initial HTML document, it enables the LCP resource to begin downloading sooner. A big obstacle to LCP resource discoverability is lazy loading of the LCP resource.
+Unfortunately, 35% of mobile websites do not have an LCP element that is statically discoverable in the document. While this is a slight improvement over the 39% we saw in 2022, it's still a significant blocker of LCP performance.
 
-Overall, lazy-loading images is a helpful performance technique that should be used to postpone loading of non-critical resources until they are near the viewport. However, using lazy-loading on the LCP image will delay the browser from loading it quickly. That is why lazy-loading should not be used on LCP elements. In this section, we explore how many sites use this performance anti-pattern.
+As we'll explore in the following sections, there are three primary ways that websites prevent their LCP resources from being statically discoverable: lazy-loading, CSS background images, and client-side rendering.
```
kevinfarrugia, line 244: If an LCP image is lazy loaded then it is still statically discoverable and I don't think lazy-loaded LCP images are included in the 35%.

kevinfarrugia, line 244: Nit: lazy loading not lazy-loading. Lazy-loading would be used as an adjective, such as "a lazy-loaded image".

(Barry has taught me well 😅)

tunetheweb: > If an LCP image is lazy loaded (natively) then it is still statically discoverable

Effectively not. It's not discoverable by the preload scanner, and needs CSS to run layout to know if it's in viewport. Which effectively means it's not statically discoverable (i.e. it's the same as if it was hidden in CSS background image, or within JS that needs to run).

tunetheweb (approved):
LGTM but would like at least one author to approve before merging since we're putting words in their mouths.

src/content/en/2024/performance.md
kevinfarrugia: Fair enough, but the [custom metric definition of `is_statically_discoverable`](https://github.com/HTTPArchive/custom-metrics/blob/420b19bebef9649b989386773d64bf35db206d1b/dist/performance.js) doesn't take this into consideration: 

```js
const rawLcpElement = Array.from(rawDoc.querySelectorAll('picture source, img, svg image')).find(i => {
        let src = i.src;
        if (i.hasAttribute('srcset')) {
            src = splitSrcSet(i.srcset).find(src => src == lcpUrl);
        } else if (i.hasAttribute('href')) {
            src = i.getAttribute('href');
        }

        return new URL(src, location.href).href == lcpUrl;
    });
```

rviscomi: Good point @kevinfarrugia, I think it's at least worth mentioning that native lazy loading isn't included in the static discoverability stat. We might want to consider looking at a different high-level metric in the future that captures all of the causes of resource load delay.

rviscomi: Updated PTAL

ines-akrap (approved):

tunetheweb:
OK let's merge this before the live stream. We can make further adjustments as needed.

PR #3961, opened 2024-12-09
Adds banner for announcement of live stream event

SaptakS:
Adds a banner on the top of the page for the live stream event.

image

cc @tunetheweb @nrllh


tunetheweb (approved):

max-ostapenko:
IMHO it's better to skip translations and put content into base template.
We'll need to revert this on 11th.

Probably it's gonna look the same for next announcements.

tunetheweb:
I've moved it into the language base templates as want this cross-year but I'd like to keep the option for translating open to reuse this to resolve #1764

PR #4018, opened 2025-03-01
Latest ubuntu image

max-ostapenko:
As mentioned in https://github.com/HTTPArchive/httparchive.org/issues/1004
Also changed actions/setup-python@v5.4.0 to major version to get less dependabot PRs.

### DIFF
--- a/.github/workflows/code-static-analysis.yml
+++ b/.github/workflows/code-static-analysis.yml
@@ -35,7 +35,7 @@ jobs:
         uses: actions/checkout@v4
       - name: Set up Python 3.12
         if: ${{ matrix.language == 'python' }}
-        uses: actions/setup-python@v5
+        uses: actions/setup-python
         with:
           python-version: '3.12'
       - name: Install dependencies
--- a/.github/workflows/code-static-analysis.yml
+++ b/.github/workflows/code-static-analysis.yml
@@ -25,8 +25,8 @@ on:
 
 jobs:
   CodeQL-Build:
-    # CodeQL runs on ubuntu-20.04, windows-latest, and macos-latest
-    runs-on: ubuntu-20.04
+    # CodeQL runs on ubuntu-latest, windows-latest, and macos-latest
+    runs-on: ubuntu-latest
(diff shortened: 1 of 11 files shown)


.github/workflows/code-static-analysis.yml
```diff
@@ -35,14 +35,14 @@ jobs:
         uses: actions/checkout@v4
       - name: Set up Python 3.12
         if: ${{ matrix.language == 'python' }}
-        uses: actions/setup-python@v5.4.0
+        uses: actions/setup-python@v5
         with:
           python-version: '3.12'
```
tunetheweb, line 40: Should we update this too?

.github/workflows/predeploy.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       with:
         node-version: '20'
     - name: Set up Python 3.12
-      uses: actions/setup-python@v5.4.0
+      uses: actions/setup-python@v5
       with:
         python-version: '3.12'
```
tunetheweb, line 42: Ditto

.github/workflows/test_website.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       with:
         node-version: '20'
     - name: Set up Python 3.12
-      uses: actions/setup-python@v5.4.0
+      uses: actions/setup-python@v5
       with:
         python-version: '3.12'
```
tunetheweb, line 35: Ditto again

tunetheweb (approved):
LGTM, though question whether we should go to latest Python.

.github/workflows/code-static-analysis.yml
max-ostapenko: Dependabot updates this action every couple of months.
I believe it's safe to expect that the `python-version` parameter will not be changed in minor versions.

So keeping a major will make it a yearly update instead.

max-ostapenko: Let's try the latest ;) Updated.

PR #4019, opened 2025-03-02
Fix setup python

tunetheweb:
All actions failing. Looks like we need a version number.

FYI @max-ostapenko


tunetheweb:
See https://github.com/HTTPArchive/almanac.httparchive.org/actions/runs/13615604782 as an example

PR #4046, opened 2025-03-22
Fix three links

hidde:
Hi, I noticed these three links were going to the wrong place and added the right ones.


tunetheweb (approved):
Thanks @hidde !

PR #4047, opened 2025-03-22
fix: use the right links

hidde:
these three links were also going to the wrong place. I added links for two of them.

I could not find out which one was supposed to be in 'Best Sustainability Statements' (there are a lot of pages on the web that could qualify, haha) so I removed that one.


tunetheweb:
@ldevernay could you take a look?

burakguneli:
LGTM. Looks like those links were not very accurate. Thanks for the update!

PR #4094, opened 2025-05-05
Call for contributors

krlslman:
Added a "Call for Contributors" announcement for all languages.
If the default year is the current year, this call-to-action will be displayed.


src/templates/en/base.html
```diff
@@ -24,6 +24,11 @@
   You’re viewing a previous edition of the Web Almanac.
   Check out the <a href="/{{ lang }}/{{ DEFAULT_YEAR }}/{% if chapter_lang_exists(lang, DEFAULT_YEAR, slug) or page_lang_exists(lang, DEFAULT_YEAR, slug) %}{{ slug }}{% endif %}">{{ DEFAULT_YEAR }} edition!</a>
 </p>
+{%- elif year == DEFAULT_YEAR -%}
```
tunetheweb, line 27: This will still show after wqe publish this year's edition.

Let's add this to `src/templates/base/base.html`:

{% block call_for_contributors %}https://github.com/HTTPArchive/almanac.httparchive.org/discussions/4062{% endblock %}

And then use that to decide whether to show or not:

```suggestion
{%- elif self.call_for_contributors() | trim | length > 0 -%}
```

src/templates/en/base.html
```diff
@@ -24,6 +24,11 @@
...
+<p lang="en">
```
tunetheweb, line 28: Let's remove these `lang` attributes (and from line 23) when they match the template language:

```suggestion
<p>
```

This was my fault as I never removed them when translating the text. But let's do it properly now.

src/templates/en/base.html
```diff
@@ -24,6 +24,11 @@
...
+  📣 <a href="https://github.com/HTTPArchive/almanac.httparchive.org/discussions/4062">
```
tunetheweb, line 29: Use the block suggested above.

```suggestion
  📣 <a href="{{ self.call_for_contributors() }}">
```

src/templates/en/base.html
```diff
@@ -24,6 +24,11 @@
...
+  <b>Call for Contributors:</b></a> Help create the Web Almanac 2025 and be part of the team shaping this year’s edition!
```
tunetheweb, line 30: For English it should be "2025 Web Almanac". But let's make this a bit more generic.

Let's add the following to `src/server/__init__.py`:

```py
from datetime import datetime
app.jinja_env.globals['now'] = datetime.now
```

And then change this to use that, and also replace the "and" with "–":

```suggestion
  <b>Call for Contributors:</b></a> Help create the {{ now().year }} Web Almanac—be part of the team shaping this year’s edition!
```

tunetheweb (changes requested):
Look pretty good, but suggested some clean ups and things to make it more generic.

tunetheweb (approved):
Did some minor tweaks in https://github.com/HTTPArchive/almanac.httparchive.org/pull/4094/commits/f11cb0c5ab7b22dc5f9c22bbbaa6efbf163fca63 but this LGTM to me now.

Thanks!

Issue #4082, opened 2025-04-27
Performance 2025


nrllh:
# Performance 2025

![Performance illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/performance/hero_lg.jpg)

If you're interested in contributing to the Performance chapter of the 2025 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).  You can find more details about this year’s Call for Contributions [here](https://github.com/HTTPArchive/almanac.httparchive.org/discussions/4062).

💬 Please make sure to join [#web-almanac-Performance](https://join.slack.com/t/httparchive/shared_invite/zt-45sgwmnb-eDEatOhqssqNAKxxOSLAaA) on Slack for team coordination.

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| ---- | ------- | --------- | -------- | ------- |    -------  |
|  -  | - | - | - | -  | - |

<details>
<summary><b>Expand for more information about each role 👀</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
[...]

PR #2974, opened 2022-06-15
Performance 2022 Queries

konfirmed:
Progress on #2890 

# Queries:

### Web Getting Right (Not Custom)
- [x] % of sites that are using priority as an attribute on LCP element
- [x] % of sites where the LCP is element is preloaded

#### Gaming the Metrics (Custom)
- [x] Filtering for synthetic tests (does not catch hex codes, see https://www.simpleskincare.in/ for example)
- [x] LCP Animation & Overlay Hack
- [x] LCP Svg Overlay Hack
- [x] CLS Animation Hack
- [x] FID iFrame Hack

#### Antipatterns (Custom)

- [x] LCP Lazy Loaded

#### Antipatterns (Not Custom)

- [x] Are LCP Image Elements Responsive
- [x] LCP Element Resource Delay
- [x] TTFB by Category
- [x] Domain Sharding
- [x] FID Double Tap to Zoom Disabled
- [x] FID / INP Long Task Data (let's pick either FID or INP)

#### Web Vitals

- [x] By Device
- [x] By Country
- [x] By Network connection
- [x] By Rank
- [x] By Rank and device

PR #4145, opened 2025-06-24
[WIP] Performance 2025 Queries

tannerhodges:
## 📝 Description

Working on queries for https://github.com/HTTPArchive/almanac.httparchive.org/issues/4082.

## ✅ Migrated

The following queries have been migrated from 2024 to 2025 using the new `crawl` dataset: https://har.fyi/guides/migrating-to-crawl-dataset/

- bfcache_cachecontrol_nostore.sql
- bfcache_unload.sql
- cls_animations.sql
- cls_unsized_image_height.sql
- cls_unsized_images.sql
- font_resource_hints_usage.sql
- font_resource_hints_usage_trends.sql
- font_usage_mobile.sql
- lcp_host.sql

## 🚧 TODOs

Still working on migrating the following queries from 2024 to 2025:

- [ ] sql/2025/performance/generated_content.sql
- [ ] sql/2025/performance/generated_content_web_vitals.sql
- [ ] sql/2025/performance/inp_long_tasks.sql
- [ ] sql/2025/performance/inp_tbt.sql
- [ ] sql/2025/performance/js_bytes_rank.sql
- [ ] sql/2025/performance/lcp_async_fetchpriority.sql
- [ ] sql/2025/performance/lcp_bytes_distribution.sql
- [ ] sql/2025/performance/lcp_bytes_histogram.sql
- [ ] sql/2025/performance/lcp_element_data.sql
- [ ] sql/2025/performance/lcp_format.sql
- [ ] sql/2025/performance/lcp_host_3p.sql
- [ ] sql/2025/performance/lcp_initiator_type.sql
- [ ] sql/2025/performance/lcp_lazy.sql
- [ ] sql/2025/performance/lcp_lazy_secondary_pages.sql
- [ ] sql/2025/performance/lcp_lazy_technologies.sql
- [ ] sql/2025/performance/lcp_preload_discoverable.sql
- [ ] sql/2025/performance/lcp_resource_load_delay.sql
- [ ] sql/2025/performance/lcp_resource_type.sql
- [ ] sql/2025/performance/lcp_responsive_data.sql
- [ ] sql/2025/performance/lcp_wasted_bytes.sql
- [ ] sql/2025/performance/monthly_cls_lcp.sql
- [ ] sql/2025/performance/render_blocking_resources.sql
- [ ] sql/2025/performance/render_blocking_savings_fcp.sql
- [ ] sql/2025/performance/render_blocking_savings_lcp.sql
- [ ] sql/2025/performance/resource_hints_usage.sql
- [ ] sql/2025/performance/resource_hints_usage_2021.sql
- [ ] sql/2025/performance/rtt_distribution.sql
- [ ] sql/2025/performance/viewport_meta_zoom_disable.sql
- [ ] sql/2025/performance/web_vitals_by_device.sql
- [ ] sql/2025/performance/web_vitals_by_device_secondary_pages.sql
- [ ] sql/2025/performance/web_vitals_by_rank_and_device.sql
- [ ] sql/2025/performance/web_vitals_by_technology.sql

### DIFF
--- a/sql/2025/performance/bfcache_cachecontrol_nostore.sql
+++ b/sql/2025/performance/bfcache_cachecontrol_nostore.sql
@@ -1,0 +1,29 @@
+CREATE TEMP FUNCTION HAS_NO_STORE_DIRECTIVE(cache_control STRING) RETURNS BOOL AS (
+  REGEXP_CONTAINS(cache_control, r'(?i)\bno-store\b')
+);
+
+WITH requests AS (
+  SELECT
+    client,
+    LOGICAL_OR(HAS_NO_STORE_DIRECTIVE(JSON_VALUE(payload._cacheControl))) AS includes_ccns
+  FROM
+    `httparchive.crawl.requests`
+  WHERE
+    date = '2025-06-01' AND
+    is_main_document
+  GROUP BY
+    client,
+    page
+)
+
+SELECT
+  client,
+  COUNTIF(includes_ccns) AS pages,
+  COUNT(0) AS total,
+  COUNTIF(includes_ccns) / COUNT(0) AS pct
+FROM
+  requests
+GROUP BY
+  client
+ORDER BY
+  client
--- a/sql/2025/performance/bfcache_unload.sql
+++ b/sql/2025/performance/bfcache_unload.sql
@@ -1,0 +1,39 @@
+CREATE TEMPORARY FUNCTION getUnloadHandler(audit STRING)
+RETURNS BOOL LANGUAGE js AS '''
+try {
+  var $ = JSON.parse(audit);
+  return $.details?.items?.some(n => n.value?.toLowerCase() === "unloadhandler");
+} catch (e) {
+  return false;
+}
+''';
+
+WITH lh AS (
+  SELECT
+    client,
+    page,
+    rank,
+    getUnloadHandler(TO_JSON_STRING(lighthouse.audits.deprecations)) AS has_unload
+  FROM
+    `httparchive.crawl.pages`
+  WHERE
+    date = '2025-06-01'
+)
+
+SELECT
+  client,
+  _rank AS rank,
+  COUNTIF(has_unload) AS pages,
+  COUNT(0) AS total,
+  COUNTIF(has_unload) / COUNT(0) AS pct
+FROM
+  lh,
+  UNNEST([1000, 10000, 100000, 1000000, 10000000, 100000000]) AS _rank
+WHERE
+  rank <= _rank
+GROUP BY
+  client,
+  rank
+ORDER BY
+  rank,
+  client
--- a/sql/2025/performance/cls_animations.sql
+++ b/sql/2025/performance/cls_animations.sql
@@ -1,0 +1,27 @@
+WITH lh AS (
+  SELECT
+    client,
+    ARRAY_LENGTH(JSON_QUERY_ARRAY(lighthouse.audits.`non-composited-animations`.details.items)) AS num_animations
+  FROM
+    `httparchive.crawl.pages`
+  WHERE
+    date = '2025-06-01' AND
+    is_root_page
+)
+
+SELECT
+  percentile,
+  client,
+  APPROX_QUANTILES(num_animations, 1000)[OFFSET(percentile * 10)] AS num_animations,
+  COUNTIF(num_animations > 0) AS pages,
+  COUNT(0) AS total,
+  COUNTIF(num_animations > 0) / COUNT(0) AS pct
+FROM
+  lh,
+  UNNEST([10, 25, 50, 75, 90, 100]) AS percentile
+GROUP BY
+  percentile,
+  client
+ORDER BY
+  percentile,
+  client
--- a/sql/2025/performance/font_resource_hints_usage.sql
+++ b/sql/2025/performance/font_resource_hints_usage.sql
@@ -1,0 +1,65 @@
+CREATE TEMPORARY FUNCTION getResourceHints(custom_metrics STRING)
+RETURNS ARRAY<STRUCT<name STRING, href STRING>>
+LANGUAGE js AS '''
+var hints = new Set(['preload', 'prefetch', 'preconnect', 'prerender', 'dns-prefetch']);
+try {
+    var $ = JSON.parse(custom_metrics);
+    var almanac = $.other.almanac;
+    return almanac['link-nodes'].nodes.reduce((results, link) => {
+        var hint = link.rel.toLowerCase();
+        if (!hints.has(hint)) {
+            return results;
+        }
+        results.push({
+            name: hint,
+            href: link.href
+        });
+        return results;
+    }, []);
+} catch (e) {
+    return [];
+}
+''';
+
+WITH resource_hints AS (
+  SELECT DISTINCT
+    client,
+    page,
+    hint.name
+  FROM
+    `httparchive.crawl.pages`
+  LEFT JOIN
+    UNNEST(getResourceHints(TO_JSON_STRING(custom_metrics))) AS hint
+  WHERE
+    date = '2025-06-01' AND
+    is_root_page
+),
+
+font_requests AS (
+  SELECT
+    client,
+    page,
+    type
+  FROM
+    `httparchive.crawl.requests`
+  WHERE
+    date = '2025-06-01' AND
+    type = 'font' AND
+    is_root_page
+)
+
+SELECT
+  client,
+  name,
+  COUNT(DISTINCT page) AS pages,
+  SUM(COUNT(DISTINCT page)) OVER (PARTITION BY client) AS total,
+  COUNT(DISTINCT page) / SUM(COUNT(DISTINCT page)) OVER (PARTITION BY client) AS pct_hints
+FROM
+  resource_hints
+LEFT JOIN
+  font_requests
+USING (client, page)
+GROUP BY
+  client, name, type
+ORDER BY
+  pct_hints DESC;
--- a/sql/2025/performance/font_resource_hints_usage_trends.sql
+++ b/sql/2025/performance/font_resource_hints_usage_trends.sql
@@ -1,0 +1,94 @@
+CREATE TEMPORARY FUNCTION getResourceHints(custom_metrics STRING)
+RETURNS ARRAY<STRUCT<name STRING, href STRING>>
+LANGUAGE js AS '''
+var hints = new Set(['preload', 'prefetch', 'preconnect', 'prerender', 'dns-prefetch']);
+try {
+    var $ = JSON.parse(custom_metrics);
+    var almanac = $.other.almanac;
+    return almanac['link-nodes'].nodes.reduce((results, link) => {
+        var hint = link.rel.toLowerCase();
+        if (!hints.has(hint)) {
+            return results;
+        }
+        results.push({
+            name: hint,
+            href: link.href
+        });
+        return results;
+    }, []);
+} catch (e) {
+    return [];
+}
+''';
+
+WITH resource_hints AS (
+  SELECT DISTINCT
+    client,
+    date,
+    page,
+    hint.name AS name
+  FROM
+    `httparchive.crawl.pages`
+  LEFT JOIN
+    UNNEST(getResourceHints(TO_JSON_STRING(custom_metrics))) AS hint
+  WHERE
+    (date = '2025-06-01' OR date = '2024-06-01' OR date = '2023-06-01') AND
+    is_root_page
+),
+
+font_requests AS (
+  SELECT
+    client,
+    date,
+    page,
+    type
+  FROM
+    `httparchive.crawl.requests`
+  WHERE
+    (date = '2025-06-01' OR date = '2024-06-01' OR date = '2023-06-01') AND
+    type = 'font' AND
+    is_root_page
+),
+
+totals AS (
+  SELECT
+    client,
+    date,
+    COUNT(0) AS total_pages
+  FROM
+    `httparchive.crawl.pages`
+  WHERE
+    (date = '2025-06-01' OR date = '2024-06-01' OR date = '2023-06-01') AND
+    is_root_page
+  GROUP BY
+    client,
+    date
+)
+
+SELECT
+  client,
+  date,
+  name,
+  type,
+  COUNT(DISTINCT page) AS pages,
+  ANY_VALUE(total_pages) AS total,
+  COUNT(DISTINCT page) / ANY_VALUE(total_pages) AS pct
+FROM
+  resource_hints
+LEFT JOIN
+  font_requests
+USING (client, date, page)
+JOIN
+  totals
+USING (client, date)
+GROUP BY
+  client,
+  date,
+  name,
+  type
+HAVING
+  type IS NOT NULL
+ORDER BY
+  client,
+  date,
+  name DESC
--- a/sql/2025/performance/cls_unsized_image_height.sql
+++ b/sql/2025/performance/cls_unsized_image_height.sql
@@ -1,0 +1,26 @@
+WITH lh AS (
+  SELECT
+    client,
+    CAST(JSON_VALUE(unsized_image, '$.node.boundingRect.height') AS INT64) AS height
+  FROM
+    `httparchive.crawl.pages`,
+    UNNEST(JSON_QUERY_ARRAY(lighthouse.audits.`unsized-images`.details.items)) AS unsized_image
+  WHERE
+    date = '2025-06-01' AND
+    is_root_page
+)
+
+SELECT
+  percentile,
+  client,
+  APPROX_QUANTILES(height, 1000)[OFFSET(percentile * 10)] AS height,
+  COUNT(0) AS unsized_images
+FROM
+  lh,
+  UNNEST([10, 25, 50, 75, 90, 100]) AS percentile
+GROUP BY
+  percentile,
+  client
+ORDER BY
+  percentile,
+  client
--- a/sql/2025/performance/cls_unsized_images.sql
+++ b/sql/2025/performance/cls_unsized_images.sql
@@ -1,0 +1,27 @@
+WITH lh AS (
+  SELECT
+    client,
+    ARRAY_LENGTH(JSON_QUERY_ARRAY(lighthouse.audits.`unsized-images`.details.items)) AS num_unsized_images
+  FROM
+    `httparchive.crawl.pages`
+  WHERE
+    date = '2025-06-01' AND
+    is_root_page
+)
+
+SELECT
+  percentile,
+  client,
+  APPROX_QUANTILES(num_unsized_images, 1000)[OFFSET(percentile * 10)] AS num_unsized_images,
+  COUNTIF(num_unsized_images > 0) AS pages,
+  COUNT(0) AS total,
+  COUNTIF(num_unsized_images > 0) / COUNT(0) AS pct
+FROM
+  lh,
+  UNNEST([10, 25, 50, 75, 90, 100]) AS percentile
+GROUP BY
+  percentile,
+  client
+ORDER BY
+  percentile,
+  client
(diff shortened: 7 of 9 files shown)


sql/2025/performance/web_vitals_by_technology.sql
```diff
@@ -0,0 +1,275 @@
...
+    CONCAT(origin, '/') AS page,
+    CASE
+      WHEN device = 'phone' THEN 'mobile'
+      ELSE device
+    END AS client,
+
+    # TODO: Should we remove FID now that it's deprecated?
+    fast_fid,
+    avg_fid,
+    slow_fid,
```
tannerhodges: ❓ Now that FID is deprecated, should I go ahead and remove FID references from these queries?

sql/2025/performance/lcp_resource_type.sql
```diff
@@ -0,0 +1,33 @@
+# We are unable to track LCP for video elements: https://issues.chromium.org/issues/364860066
+# TODO: It looks like this ^^ was resolved 2024-11-27, so does that mean we can track LCP for video now?
```
tannerhodges: ❓ Is this comment about being unable to track LCP for video elements still relevant?

If we are tracking LCP for video elements now, does that mean I should change the `CASE` below to also account for videos?

sql/2025/performance/bfcache_cachecontrol_nostore.sql
```diff
@@ -0,0 +1,29 @@
...
+
+WITH requests AS (
+  SELECT
+    client,
+    LOGICAL_OR(HAS_NO_STORE_DIRECTIVE(JSON_VALUE(payload._cacheControl))) AS includes_ccns
+  FROM
+    `httparchive.crawl.requests`
+  WHERE
+    date = '2025-07-01' AND
+    is_main_document
```
tunetheweb, line 13: This will look at both home pages and secondary pages.

The sum below does say _pages_, rather than _sites_ so maybe this is OK?

Or maybe it would be better to either:
1) Limit to home pages (`AND is_root_page`)
2) Check if any site is using CCNS for either home page or secondary page?

Or, we could leave as is, since it's what we did last year and technically it's correct sicne you claim pages.

sql/2025/performance/bfcache_unload.sql
```diff
@@ -0,0 +1,39 @@
...
+  return false;
+}
+''';
+
+WITH lh AS (
+  SELECT
+    client,
+    page,
+    rank,
+    getUnloadHandler(TO_JSON_STRING(lighthouse.audits.deprecations)) AS has_unload
```
tunetheweb: I think you can pass this as JSON to avoid the conversion to STRING and then back again to JSON.

Also it looks like "unloadhandler" is not the value anymore.

```suggestion
CREATE TEMPORARY FUNCTION getUnloadHandler(audit JSON)
RETURNS BOOL LANGUAGE js AS '''
try {
  return audit?.details?.items?.some(n => n.value?.toLowerCase()?.includes("unload event listeners"));
} catch (e) {
  return false;
}
''';

WITH lh AS (
  SELECT
    client,
    page,
    rank,
    getUnloadHandler(lighthouse.audits.deprecations) AS has_unload
```

sql/2025/performance/bfcache_unload.sql
```diff
@@ -0,0 +1,39 @@
...
+  lh,
+  UNNEST([1000, 10000, 100000, 1000000, 10000000, 100000000]) AS _rank
+WHERE
+  rank <= _rank
+GROUP BY
+  client,
+  rank
+ORDER BY
+  rank,
+  client
```
tunetheweb: The reuse of `rank` makes it difficult to know what you are grouping by. Suggest to rename to ranking:

```suggestion
SELECT
  client,
  ranking,
  COUNTIF(has_unload) AS pages,
  COUNT(0) AS total,
  COUNTIF(has_unload) / COUNT(0) AS pct
FROM
  lh,
  UNNEST([1000, 10000, 100000, 1000000, 10000000, 100000000]) AS ranking
WHERE
  rank <= ranking
GROUP BY
  client,
  ranking
ORDER BY
  client,
  ranking
```

sql/2025/performance/cls_animations.sql
```diff
@@ -0,0 +1,27 @@
...
+  WHERE
+    date = '2025-07-01' AND
+    is_root_page
+)
+
+SELECT
+  percentile,
+  client,
+  APPROX_QUANTILES(num_animations, 1000)[OFFSET(percentile * 10)] AS num_animations,
+  COUNTIF(num_animations > 0) AS pages,
```
tunetheweb: This is "home pages" rather than "pages". Confusing with first query above where "pages" means total pages.

Suggest either rename this column to "home pages", or look at all pages, or look at "sites" (whether either the home page or the secondary page" uses non-composited animations

sql/2025/performance/font_resource_hints_usage.sql
```diff
@@ -0,0 +1,65 @@
...
+
+WITH resource_hints AS (
+  SELECT DISTINCT
+    client,
+    page,
+    hint.name
+  FROM
+    `httparchive.crawl.pages`
+  LEFT JOIN
+    UNNEST(getResourceHints(TO_JSON_STRING(custom_metrics))) AS hint
```
tunetheweb: You only need part of this, so don't pass it all.

```suggestion
CREATE TEMPORARY FUNCTION getResourceHints(custom_metrics_link_nodes JSON)
RETURNS ARRAY<STRUCT<name STRING, href STRING>>
LANGUAGE js AS '''
var hints = new Set(['preload', 'prefetch', 'preconnect', 'prerender', 'dns-prefetch']);
try {
    return custom_metrics_link_nodes.nodes.reduce((results, link) => {
        var hint = link.rel.toLowerCase();
        if (!hints.has(hint)) {
            return results;
        }
        results.push({
            name: hint,
            href: link.href
        });
        return results;
    }, []);
} catch (e) {
    return [];
}
''';

WITH resource_hints AS (
  SELECT DISTINCT
    client,
    page,
    hint.name
  FROM
    `httparchive.crawl.pages`
  LEFT JOIN
    UNNEST(getResourceHints(custom_metrics.other.almanac['link-nodes'])) AS hint
```

sql/2025/performance/font_resource_hints_usage_trends.sql
```diff
@@ -0,0 +1,94 @@
+CREATE TEMPORARY FUNCTION getResourceHints(custom_metrics STRING)
```
tunetheweb: Same here

sql/2025/performance/generated_content.sql
```diff
@@ -0,0 +1,73 @@
+#standardSQL
+CREATE TEMPORARY FUNCTION getGeneratedContent(custom_metrics STRING)
```
tunetheweb: Same here

sql/2025/performance/generated_content_web_vitals.sql
```diff
@@ -0,0 +1,141 @@
+#standardSQL
+CREATE TEMP FUNCTION IS_NON_ZERO(good FLOAT64, needs_improvement FLOAT64, poor FLOAT64) RETURNS BOOL AS (
+  good + needs_improvement + poor > 0
+);
+
+CREATE TEMPORARY FUNCTION IS_GOOD(good FLOAT64, needs_improvement FLOAT64, poor FLOAT64) RETURNS BOOL AS (
+  SAFE_DIVIDE(good, (good + needs_improvement + poor)) >= 0.75
+);
+
+CREATE TEMPORARY FUNCTION getGeneratedContent(custom_metrics STRING)
```
tunetheweb: And here

sql/2025/performance/inp_long_tasks.sql
```diff
@@ -0,0 +1,39 @@
+WITH long_tasks AS (
+  SELECT
+    client,
+    page,
+    ANY_VALUE(httparchive.core_web_vitals.GET_CRUX_INP(summary)) AS inp,
+    SUM(CAST(JSON_VALUE(item, '$.duration') AS FLOAT64)) AS long_tasks
```
tunetheweb: Slightly cleaner:

```suggestion
    SUM(CAST(JSON_VALUE(item.duration) AS FLOAT64)) AS long_tasks
```

sql/2025/performance/inp_long_tasks.sql
```diff
@@ -0,0 +1,39 @@
...
+)
+
+SELECT
+  client,
+  long_tasks,
+  inp
+FROM
+  meta
+WHERE
+  MOD(row, CAST(FLOOR(n / 1000) AS INT64)) = 0
```
tunetheweb: I'm not sure what this does, but I don't like it.

It _seems_ to return every 1000th row? But since it's done BEFORE the `ORDER BY` it's non-deterministic.

Do we need this?

```suggestion
```

sql/2025/performance/js_bytes_rank.sql
```diff
@@ -0,0 +1,17 @@
+SELECT
+  IF(_rank < 100000000, CAST(_rank AS STRING), 'all') AS rank,
+  client,
+  APPROX_QUANTILES(CAST(JSON_VALUE(summary.bytesJS) AS INT64), 1000)[OFFSET(500)] / 1024 AS js_kbytes
+FROM
+  `httparchive.crawl.pages`,
+  UNNEST([1000, 10000, 100000, 1000000, 10000000, 100000000]) AS _rank
```
tunetheweb: I like this IF statement  and use it myself. Can we use it in other queries using rank?

Also I like to return both, so I can sort by the number, but display the text. Also avoid reusing rank which is confusing:

```suggestion
  rank_grouping
  IF(_rank < 100000000, CAST(_rank AS STRING), 'all') AS ranking,
  client,
  APPROX_QUANTILES(CAST(JSON_VALUE(summary.bytesJS) AS INT64), 1000)[OFFSET(500)] / 1024 AS js_kbytes
FROM
  `httparchive.crawl.pages`,
  UNNEST([1000, 10000, 100000, 1000000, 10000000, 100000000]) AS _rank_grouping
```

sql/2025/performance/lcp_async_fetchpriority.sql
```diff
@@ -0,0 +1,109 @@
+CREATE TEMP FUNCTION getLoadingAttr(attributes STRING) RETURNS STRING LANGUAGE js AS '''
```
tunetheweb: Change all these functions to take JSON rather than STRING to save converting to STRING and back to JSON.

sql/2025/performance/lcp_element_data.sql
```diff
@@ -0,0 +1,129 @@
+CREATE TEMP FUNCTION getLoadingAttr(attributes STRING) RETURNS STRING LANGUAGE js AS '''
```
tunetheweb: Same

tunetheweb (commented):
Have only reviewed about half of the queries but starting to repeat myself so will let you work on the feedback so far (and apply that feedback to other similar queries where appropriate).

sql/2025/performance/web_vitals_by_technology.sql
tunetheweb: Yes, since we no longer populate these in CrUX. So they just return `null`.

sql/2025/performance/lcp_resource_type.sql
tunetheweb: It definitely works in some cases, but have reports of it not working in all cases. We've it on out TODO list to investigate.

But even if it does work for videos, it measures the first frame. Which arguably is still an "image". So maybe current classification if OK? Cause measuring whether it's a video is a little more complicated since it _could_ still be the poster image even on a `<video>` element.

sql/2025/performance/bfcache_cachecontrol_nostore.sql
tannerhodges: Yeah, my two cents is to leave this query as is (since it goes back to 2022 and seems virtually unchanged since then).

❓ Unless the authors prefer to change it this year? @hfhashmi @25prathamesh @himanshujariyal

sql/2025/performance/bfcache_unload.sql
tannerhodges: Aha! Thank you, I agree. I'll go through and clean up this year's functions to avoid the extra conversion.

Will post later once that's updated.

💡 In the meantime, I found where that value changed in Lighthouse https://github.com/GoogleChrome/lighthouse/pull/16333, released in [v12.4.0](https://github.com/GoogleChrome/lighthouse/releases/tag/v12.4.0).

Looks like the value change applied to the HTTP Archive data starting in `2025-03-01` (everything before that run still uses the old value).

- `2025-02-01` and before: `UnloadHandler`
- `2025-03-01` and after: `Unload event listeners are deprecated and will be removed.`

sql/2025/performance/cls_animations.sql
tannerhodges: 👍 I'll rename this. (For what it's worth, these names and queries are all copied from last year. I've just updated the queries to use the new tables this year.)

sql/2025/performance/font_resource_hints_usage.sql
tannerhodges: Ah, sounds good. I'll do this with the other custom functions as well so they only take the minimum data necessary.

Will post once that's updated.

sql/2025/performance/web_vitals_by_technology.sql
tannerhodges: 👌 Perfect, I've removed all "fid" references from the 2025 queries in 6cecaff2ffc51e039d958bb0224290c8a91f08a5.

sql/2025/performance/lcp_resource_type.sql
tannerhodges: Makes sense. For now, I'll leave this as is then and just remove my `TODO` comment (250578e47dad7015736750c86b54222d5e2a2204).

sql/2025/performance/bfcache_unload.sql
tannerhodges: Updated in two commits just so we have a record of how to handle the different Lighthouse values:

- 4c983e4 Update `getUnloadHandler()` to handle different Lighthouse versions
- 29fff55 Update `getUnloadHandler()` to only handle Lighthouse >= v12.4.0 values

This should be resolved now.

sql/2025/performance/bfcache_unload.sql
tannerhodges: Agreed. I've updated this everywhere we're using `_rank` in the 2025 queries.

Updated in 36bf92882ca7d6113603b7ad50b9020e43346bf6.

sql/2025/performance/cls_animations.sql
tannerhodges: Updated in 0dff265faf05c255c198cb3b3c84f4883b093d93.

sql/2025/performance/inp_long_tasks.sql
tannerhodges: 😕 This is another query that goes back to 2022 (see https://github.com/HTTPArchive/almanac.httparchive.org/pull/2974), and seems virtually unchanged since then.

It seems like an optimization to only sample 1000 rows for each client.

I assume it's not necessary, but I also don't know what impact removing it would have on time to run + compute cost?

Personally, I'd treat this the same as `bfcache_cachecontrol_nostore.sql` ([see comment](https://github.com/HTTPArchive/almanac.httparchive.org/pull/4145/#discussion_r2318695043)) and leave as is for the sake of continuity. But I'm happy to remove if you think it's worth removing.

tunetheweb: Let's remove this. It doesn't do anything as you've already selected all the data in the queries above!. Feel free to run both to confirm there's no real difference.

sql/2025/performance/font_resource_hints_usage.sql
tannerhodges: Updated in e9d425bcf2d4afa91e9584f1e6f12a95347ea052.

sql/2025/performance/font_resource_hints_usage_trends.sql
tannerhodges: Updated in eb8bcfd8e268c84404109d9c48cf11b6805f34e0.

sql/2025/performance/generated_content.sql
tannerhodges: Updated in 359f4438b6b9e803b6fcc292c8be03f4527a6d9f.

sql/2025/performance/generated_content_web_vitals.sql
tannerhodges: Updated in ef0adf2c5a5e4ecb87ae0393a52d95e6a0c83830.

sql/2025/performance/inp_long_tasks.sql
tannerhodges: Updated in e20e8fa887c9f4bf098693cdfa0fa33fc6c9cb6a.

sql/2025/performance/js_bytes_rank.sql
tannerhodges: Per `bfcache_unload.sql` ([see comment](https://github.com/HTTPArchive/almanac.httparchive.org/pull/4145/#discussion_r2318852496)), I've updated to use "ranking" in 36bf92882ca7d6113603b7ad50b9020e43346bf6.

Currently, the only other queries we're using `rank` in are:

- `render_blocking_savings_fcp.sql`
- `render_blocking_savings_lcp.sql`
- `web_vitals_by_rank_and_device.sql`

Personally, I'd prefer to keep these ^^ other queries as is to keep them simple (and since the current Google Sheet's charts are already setup to handle their rankings).

However, let me know if you think it's worth updating the other queries and I can make those changes as well.

Either way, I'll definitely use this `IF` statement in any new queries I write.

sql/2025/performance/lcp_async_fetchpriority.sql
tannerhodges: Updated in 0387a6cc020848e682e055f98af0bf50a6eee2c3.

sql/2025/performance/lcp_element_data.sql
tannerhodges: Updated in c496d3fd7756a2ca3ee44d199903a8cce7b8e7b3.

sql/2025/performance/inp_long_tasks.sql
tannerhodges: 😬 I think I found out why they were sampling:

Only able to save 309,518 of 5,287,925 rows. To save all rows, use the export to BigQuery table action instead.

I've tried exporting as CSV and re-importing to Google Sheets, but it seems like 309,518 rows is a hard limit for how much I can import…

❓ Is there a workaround so I can include all this data? Or should I keep the `MOD` sampling so the data fits in Google Sheets?

tunetheweb: Ugh well this tells us this is just a bad query. And sampling is a bad way to get around this.

It looks like they [didn't end up using this](https://almanac.httparchive.org/en/2024/performance#long-tasks):

> Due to the lack of same-source data for long tasks and INP, we decided not to correlate them.

So my vote would be to drop this query altogether. It's not a good query.

And even if it was good SQL (which it's not), the comparison is suspect (which looks to be what they discovered). Comparing the long tasks during page load with the INP during interactions (which might have lead to different long tasks) is extremely suspect.

sql/2025/performance/js_bytes_rank.sql
tunetheweb: SGTM. Resolving.

sql/2025/performance/inp_long_tasks.sql
tannerhodges: 🫡 Sounds good. Deleted in 4d4c0497f93eb2d300ad11a3a9b5a4ed1721a020.

tannerhodges:
🙏 Awesome, thanks @tunetheweb!

PR #4161, opened 2025-07-08
Translation of base file to Japanese

ksakae1216:
Please review for me.


src/templates/ja/2024/base.html
```diff
@@ -8,27 +8,27 @@
 {% block foreword %}
 {# TODO - translate #}<div lang="en">
```
tunetheweb:
```suggestion
<div>
```

tunetheweb (commented):
Should we translate "2024 Web Almanac General Co-Chair"?

PR #4173, opened 2025-07-23
Removes call for contributor banner

SaptakS:
As requested by @nrllh , this PR removes the call for contributor link from the top of the page. I have kept the template to render it, and just removed the link which effectively removes the banner. I think this would help us to use the same template logic in later years to show the call for contributor banner again.

### DIFF
--- a/src/templates/base/base.html
+++ b/src/templates/base/base.html
@@ -106,7 +106,7 @@
 
 {% endblock %}
 
-{% block call_for_contributors_link %}https://github.com/HTTPArchive/almanac.httparchive.org/discussions/4062{% endblock %}
+{% block call_for_contributors_link %}{% endblock %}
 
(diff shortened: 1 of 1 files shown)


src/templates/base/base.html
```diff
@@ -106,7 +106,7 @@
 
 {% endblock %}
 
-{% block call_for_contributors_link %}https://github.com/HTTPArchive/almanac.httparchive.org/discussions/4062{% endblock %}
+{% block call_for_contributors_link %}{% endblock %}
```
tunetheweb, line 111: Let's add a comment to show how this is used:

```suggestion
{# Add github link to discussion to enable the Call for Contributors link #}
{# See: https://github.com/HTTPArchive/almanac.httparchive.org/pull/4173/files #}
{% block call_for_contributors_link %}{% endblock %}
```

Issue #4073, opened 2025-04-27
Fonts 2025


nrllh:
# Fonts 2025

![Fonts illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/fonts/hero_lg.jpg)

If you're interested in contributing to the Fonts chapter of the 2025 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).  You can find more details about this year’s Call for Contributions [here](https://github.com/HTTPArchive/almanac.httparchive.org/discussions/4062).

💬 Please make sure to join [#web-almanac-fonts](https://join.slack.com/t/httparchive/shared_invite/zt-45sgwmnb-eDEatOhqssqNAKxxOSLAaA) on Slack for team coordination.

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| ---- | ------- | --------- | -------- | ------- |  -------   |
|  @charlesberret   | @charlesberret  | @bramstein @svgeesus @jmsole @RoelN    | @IvanUkhov  | -  | @turban1988 |

<details>
<summary><b>Expand for more information about each role 👀</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
[...]

PR #4175, opened 2025-07-24
Fonts 2025 queries

IvanUkhov:
Makes progress on #4073

# Fonts

## Resources

* 📄 [Planning document]
* 📊 [Results sheet]
* 📝 [Chapter content]

## Structure

The queries are split by the section where they are used:

* `design/` is about foundries and families,
* `development/` is about tools and technologies, and
* `performance/` is about hosting and serving.

Each file name starts with one of the following prefixes indicating the primary subject of the corresponding analysis:

* `fonts_` is about font files,
* `pages_` is about HTML pages,
* `scripts_` is about JavaScript scripts, and
* `styles_` is about CSS style sheets.

The prefix is followed by the property studied given in singular, potentially extended one or several suffixes narrowing down the scope, as in `fonts_size_by_table.sql` and `pages_link_relation.sql`.

## Content

Each query starts with a preamble indicating the section, question, and normalization type, as illustrated below:

```sql
-- Section: Performance
-- Question: What is the distribution of the file size broken down by table?
-- Normalization: Pages
```

Many queries rely on temporary functions for convenience and clarity. The functions that appear in several queries are extracted into a common file called `common.sql`. Whenever any of the functions defined in `common.sql` is used by a query, the query has the following pseudo-directive at the top:

```sql
-- INCLUDE https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/{year}/fonts/common.sql
```

The pseudo-directive has to be replaced with the content of `common.sql` prior to executing the query in question.

In addition, queries generally have parameters, as in `@date`, so as to be able to run them for different configurations. The values for the parameters will have to be supplied upon execution.

All the above is taken take of automatically if the queries are executed using `execute.py`, which we discuss next.

## Execution

The queries can be executed using the `execute.py` script. The results are first saved in local CSV files sitting next to the SQL files and then uploaded to the spreadsheet. In the spreadsheet, for each query, a separate sheet is created and named after the question the query answers, which is given in its preamble. If the CSV file already exists, the corresponding query is not executed. If cell A1 is already populated, the corresponding sheet is not updated.

First, ensure that the Application Default Credentials authorization strategy is configured, and that the HTTP Archive project is used as the quota project:

```shell
gcloud auth application-default login \
  --scopes https://www.googleapis.com/auth/cloud-platform,https://www.googleapis.com/auth/spreadsheets
gcloud auth application-default set-quota-project httparchive
```

Second, install the Python prerequisites for the script:

```shell
pip install -r requirements.txt
```

The script can be run for all or a subset of the queries as illustrated below:

```shell
python execute.py
python execute.py design/*.sql
python execute.py development/fonts_*.sql
```

By default, it operates in a dry-run mode: it does not run the queries but prints an estimate of the amount of data that would be processed by each query. To actually run the queries, pass the `--no-dry-run` option as follows:

```shell
python execute.py --no-dry-run
python execute.py --no-dry-run design/*.sql
python execute.py --no-dry-run development/fonts_*.sql
```

[Planning document]: https://docs.google.com/document/d/1jVc0vgmAY_lBxryItRBguXxEq77mvbaQ3UpbTweUoSI
[Results sheet]: https://docs.google.com/spreadsheets/d/1otdu4p_CCI70B4FVzw6k02frStsPMrQoFu7jUim_0Bg
[Chapter content]: https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content/en/2025/fonts.md


IvanUkhov:
@tunetheweb, I think you were the one reviewing the queries last year. If you do not mind, I would like to invite you to review this year, too, but please feel free to assign someone else. This year, we did not change anything. We just migrated the queries to `crawl` and added a Python script for execution. The instructions are in the readme.

sql/2025/fonts/design/styles_family.sql
```diff
@@ -0,0 +1,73 @@
...
+WITH
+families AS (
+  SELECT
+    client,
+    FAMILY_INNER(family) AS family,
+    COUNT(DISTINCT page) AS count,
+    ROW_NUMBER() OVER (PARTITION BY client ORDER BY COUNT(DISTINCT page) DESC) AS rank
+  FROM
+    `httparchive.crawl.parsed_css`,
+    UNNEST(FAMILIES(TO_JSON_STRING(css))) AS family
```
max-ostapenko:
```suggestion
CREATE TEMPORARY FUNCTION FAMILIES($ JSON)
RETURNS ARRAY<STRING>
LANGUAGE js
OPTIONS (library = ["gs://httparchive/lib/css-font-parser.js", "gs://httparchive/lib/css-utils.js"])
AS '''
try {
  const result = [];
  walkDeclarations($, (declaration) => {
    result.push(parseFontFamilyProperty(declaration.value)[0]);
  }, {
    properties: 'font-family',
    rules: (rule) => rule.type.toLowerCase() === 'font-face'
  });
  return result;
} catch (e) {
  return [];
}
''';

WITH
families AS (
  SELECT
    client,
    FAMILY_INNER(family) AS family,
    COUNT(DISTINCT page) AS count,
    ROW_NUMBER() OVER (PARTITION BY client ORDER BY COUNT(DISTINCT page) DESC) AS rank
  FROM
    `httparchive.crawl.parsed_css`,
    UNNEST(FAMILIES(css)) AS family
```

max-ostapenko: Need to change the function in order to avoid double-JSON parsing.

max-ostapenko: Please adjust in other places where `css` column is used.

IvanUkhov: Uh, interesting. I did not realize custom functions support JSON, too. Thank you for the note! I need to updates at a few places.

IvanUkhov: Updated. It simplified things quite a bit. Just perhaps one correction is that `$` is not a valid argument name in SQL.

sql/2025/fonts/common.sql
```diff
@@ -0,0 +1,149 @@
...
+    )
+  )
+);
+
+-- Extract the family name from a payload.
+CREATE TEMPORARY FUNCTION FAMILY(payload JSON) AS (
+  FAMILY_INNER(
+    COALESCE(
+      STRING(JSON_QUERY(payload, '$._font_details.names[16]')),
+      STRING(JSON_QUERY(payload, '$._font_details.names[1]'))
```
tunetheweb: Nit but can all these `JSON_QUERY`'s be simplified to:

```suggestion
      payload._font_details.names[16],
      payload._font_details.names[1]
```

sql/2025/fonts/common.sql
```diff
@@ -0,0 +1,149 @@
...
+
+-- Check if the font was successfully parsed given its payload.
+CREATE TEMPORARY FUNCTION IS_PARSED(payload JSON) AS (
+  payload._font_details.table_sizes IS NOT NULL
+);
+
+-- Check if the font is a variable font given its payload.
+CREATE TEMPORARY FUNCTION IS_VARIABLE(payload JSON) AS (
+  REGEXP_CONTAINS(
+    TO_JSON_STRING(payload._font_details.table_sizes),
```
tunetheweb, line 138: Is `TO_JSON_STRING` needed here?

```suggestion
   payload._font_details.table_sizes,
```

sql/2025/fonts/common.sql
```diff
@@ -0,0 +1,149 @@
...
+    '(?i)gvar|CFF2'
+  )
+);
+
+-- Extract the variable formats from a payload.
+CREATE TEMPORARY FUNCTION VARIABLE_FORMATS(payload JSON) AS (
+  REGEXP_EXTRACT_ALL(
+    TO_JSON_STRING(payload._font_details.table_sizes),
```
tunetheweb, line 146: And here?

```suggestion
   payload._font_details.table_sizes,
```

sql/2025/fonts/design/fonts_designer.sql
```diff
@@ -0,0 +1,54 @@
...
+    client,
+    NULLIF(TRIM(STRING(JSON_QUERY(payload, '$._font_details.names[9]'))), '') AS designer,
+    COUNT(DISTINCT page) AS count,
+    ROW_NUMBER() OVER (PARTITION BY client ORDER BY COUNT(DISTINCT page) DESC) AS rank
+  FROM
+    `httparchive.crawl.requests`
+  WHERE
+    date = @date AND
+    type = 'font' AND
+    is_root_page AND
```
tunetheweb, line 19: If we wanted to get greater coverage you could drop `is_root_page` and change:
```sql
COUNT(DISTINCT page) AS count
```
to
```sql
COUNT(DISTINCT root_page) AS count
```

sql/2025/fonts/design/fonts_designer.sql
```diff
@@ -0,0 +1,54 @@
...
+
+-- INCLUDE https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/{year}/fonts/common.sql
+
+WITH
+designers AS (
+  SELECT
+    client,
+    NULLIF(TRIM(STRING(JSON_QUERY(payload, '$._font_details.names[9]'))), '') AS designer,
+    COUNT(DISTINCT page) AS count,
+    ROW_NUMBER() OVER (PARTITION BY client ORDER BY COUNT(DISTINCT page) DESC) AS rank
```
tunetheweb, line 13: `rank` is a confusing column name since we also use this for [CrUX Rank](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide#rank)

Maybe change to just `row`? Or `row_number`? Note if changing you'll need to change the `QUALIFY` clause too.

sql/2025/fonts/design/fonts_family_by_script.sql
```diff
@@ -0,0 +1,49 @@
...
+  FROM
+    `httparchive.crawl.requests`
+  WHERE
+    date = @date AND
+    type = 'font' AND
+    is_root_page AND
+    IS_PARSED(payload) AND
+    -- Without the subsampling, the query throws “UDF out of memory.” To work
+    -- around this, a random half is removed, and the counts are doubled.
+    RAND() < 0.5
```
tunetheweb: This seems really suspicious! How do you know the randomness is even and so you should double?

Can't you use `rank` instead to restrict this to the top million? Or run desktop and mobile separately?

sql/2025/fonts/design/styles_family.sql
```diff
@@ -0,0 +1,72 @@
...
+    `httparchive.crawl.parsed_css`,
+    UNNEST(FAMILIES(css)) AS family
+  WHERE
+    date = @date AND
+    is_root_page
+  GROUP BY
+    client,
+    family
+  QUALIFY
+    rank <= 100
```
tunetheweb, line 43: Again, consider better name than `rank`.

sql/2025/fonts/performance/fonts_format_file.sql
```diff
@@ -0,0 +1,44 @@
...
+-- Normalization: Requests and fonts
+
+-- INCLUDE https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/{year}/fonts/common.sql
+
+WITH
+requests AS (
+  SELECT
+    client,
+    url,
+    FILE_FORMAT(STRING(summary.ext), STRING(summary.mimeType)) AS format,
```
tunetheweb, line 12: Do we need the `STRING` conversion? Aren't these returned as strings anyway?

sql/2025/fonts/execute.py
```diff
@@ -0,0 +1,247 @@
+#!/usr/bin/env python
```
tunetheweb, line 1: Is this `font` specific? Or should we put it in the `sql/util` folder?

tunetheweb (commented):
LGTM with some non-blocking comments.

sql/2025/fonts/common.sql
IvanUkhov: I tried with `[OFFSET(…)]` and it complained. It did not occur to me that `OFFSET` is not needed. Fixed. But `STRING` has to stay, as the type would be `JSON` otherwise. Thank you for the suggestion!

sql/2025/fonts/common.sql
IvanUkhov: Well, yes, if we stick to regular expressions. I did not want to rewrite that part.

sql/2025/fonts/common.sql
IvanUkhov: It is the same everywhere. The hack is that we just use a regular expression instead of dealing with JSON.

sql/2025/fonts/design/fonts_designer.sql
IvanUkhov: The query is specifically about root pages. Last year, we decided it would be confusing otherwise. I remember you explained that it could be one additional page or several, but that would be a strange selection, and it would also be more difficult to explain and motivate in the chapter.

This pattern repeat is several queries if I remember right.

sql/2025/fonts/design/fonts_designer.sql
IvanUkhov: Here we assume the writer knows the context. I think the probability of confusion is small. As with `count`, `total`, and `proportion`, I have been trying to stay parsimonious with names. So `rank` is the only name used for ranking results in all queries.

sql/2025/fonts/design/fonts_designer.sql
tunetheweb: Yeah it depends if you want to measure "home page usage" or "site usage". As often home pages aren't totally representative of site usage.

I mean it's only 2 pages for a site (home page and one other page) so even then it's not a complete representation of the whole site anyway. But it could be _more_ representative than just the home page if the home page is a special page (as it often is).

But maybe one for next year. Also maybe less relevant for fonts since home page likely has the brand fonts anyway.

sql/2025/fonts/design/fonts_family_by_script.sql
IvanUkhov: My understanding is that `RAND` given a pseudo-random number between zero and one. So any logic based on that will ensure there is no bias. Everything else would yield a biased selection, depending on how the data is stored and accessed by the underlying workers. But please do let me know if you know something speaking to the contrary.

Yes, I think running separately for desktop and mobile would also make this query manageable. I just did not want to replicate the logic.

sql/2025/fonts/performance/fonts_format_file.sql
IvanUkhov: I think all JSON accessors return a value of type JSON. Hence, I have this conversion everywhere.

sql/2025/fonts/execute.py
IvanUkhov: I tried to keep it general, and it could be useful for others. However, I am relying on the standardized header format (see the readme) for exporting the context to the spreadsheet. To make it 100% general, one would have to treat the header's absence as a special case.

sql/2025/fonts/design/fonts_family_by_script.sql
tunetheweb: Yes but it could randomly work out that you randomly select ALL the pages using a particular font even if only 50% of pages do actually use it. Or vice versa. You'd depending on randomness being sufficiently random for that not to happen (which is not completely unreasonable, but also not completely guaranteed).

So on this one I'd prefer we either limit rank to only look at part of the dataset, or run desktop and mobile separately (maybe that's possible even within the one query without it running out of memory?).

My other feedback is non-blocking, but this one I'm really not comfortable about as something we should encourage in our statistical analysis.

sql/2025/fonts/performance/fonts_format_file.sql
tunetheweb: Ah yes you're right.

sql/2025/fonts/execute.py
tunetheweb: OK let's keep it here.

sql/2025/fonts/design/fonts_family_by_script.sql
IvanUkhov: Subsampling using a random-number generator only fails when the sample size is small, which is known as the law of small numbers. Overall, the subsample is guaranteed to be representative of the population (the entire table). However, we group by two columns, which zooms in on subpopulations that can be very small. If you just scroll through the results, you will find a lot of such cases, and all of them are not trustworthy:

https://docs.google.com/spreadsheets/d/1otdu4p_CCI70B4FVzw6k02frStsPMrQoFu7jUim_0Bg/edit?gid=74546850#gid=74546850

To avoid misinterpretation, I will try to split it by device so that the entire table is visible in the results.

sql/2025/fonts/common.sql
```diff
@@ -0,0 +1,149 @@
...
+  return detectWritingScript(codepoints.map((character) => parseInt(character, 10)), 0.05);
+} else {
+  return [];
+}
+""";
+
+-- Infer scripts from a payload.
+CREATE TEMPORARY FUNCTION SCRIPTS(payload JSON) AS (
+  SCRIPTS_INNER(JSON_EXTRACT_STRING_ARRAY(payload, '$._font_details.cmap.codepoints'))
+);
```
max-ostapenko: You can skip array extraction, if JSON's value is array and you're using the value in JS function.

sql/2025/fonts/design/fonts_family_by_script.sql
max-ostapenko: Unwrapping custom functions (if possible) may help with performance.

sql/2025/fonts/common.sql
IvanUkhov: Oh, good point! Fixed.

sql/2025/fonts/design/fonts_family_by_script.sql
IvanUkhov: Hm, I got rid of JSON_EXTRACT_STRING_ARRAY, as Max suggested, and it worked. But I guess timing also plays a role. Maybe this morning, there were more resources available.

IvanUkhov:
(The linter is failing due to the code elsewhere.)

tunetheweb:
> (The linter is failing due to the code elsewhere.)

Fixing in #4196

tunetheweb:
That's fixed in `main` now if you can resync this branch @IvanUkhov .

After that are you good to merge this?

IvanUkhov:
Thank you. Rebased.

Well, I have not received any feedback from the lead. I would merge, if you are OK with potential follow-up PRs.

tunetheweb (approved):

tunetheweb:
> Thank you. Rebased.
> 
> Well, I have not received any feedback from the lead. I would merge, if you are OK with potential follow-up PRs.

Yeah lets do that.

Issue #4065, opened 2025-04-27
Capabilities 2025


nrllh:
# Capabilities 2025

![Capabilities illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2020/capabilities/hero_lg.jpg)

If you're interested in contributing to the Capabilities chapter of the 2025 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).  You can find more details about this year’s Call for Contributions [here](https://github.com/HTTPArchive/almanac.httparchive.org/discussions/4062).

💬 Please make sure to join [#web-almanac-capabilities](https://join.slack.com/t/httparchive/shared_invite/zt-45sgwmnb-eDEatOhqssqNAKxxOSLAaA) on Slack for team coordination.

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| ---- | ------- | --------- | -------- | ------- |    -------  |
|  @Dawntraoz   | @Dawntraoz @MichaelSolati | @webmaxru  | @guaca  | -  | @mikaelaraujo |

<details>
<summary><b>Expand for more information about each role 👀</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
[...]

PR #4243, opened 2025-10-11
Capabilities 2025 queries

christianliebel:
Makes progress on #4065 and #4104

Updated Capabilities queries from 2022, which now include the Built-in AI APIs and WebGPU.


sql/2025/generative-ai/fugu.sql
```diff
@@ -0,0 +1,35 @@
...
+SELECT
+  client,
+  fuguAPI,
+  COUNT(DISTINCT root_page) AS sites,
+  total,
+  COUNT(DISTINCT root_page) / total AS pct,
+  ARRAY_TO_STRING(ARRAY_AGG(DISTINCT page LIMIT 50), ' ') AS sample_urls
+FROM
+  `httparchive.crawl.pages`
+    JOIN
```
tunetheweb:
```suggestion
JOIN
```

tunetheweb:
Actually should we put these in the capabilities SQL folder? Assuming there is a capabilities chapter this year.

We could put in both but that seems needless duplication when they’re the exact same.

christianliebel:
@tunetheweb You are right, there is a capabilities chapter, at least planned. I moved the queries to the capabilities folder, which makes more sense. Let’s keep them there, and we will use a subset of the results for Generative AI.

PR #4254, opened 2025-10-20
Lint only changed SQL files on PRs

max-ostapenko:
Added diff logic to SQL linting task.
It finished linting SQL faster than it takes to download super-linter image.

1. SQLfluff as PIP dependency will run on sql diffs (+ full scan on sqlfluff version update or a manual trigger)
2. SQLfluff in super-linter is off.
3. SQL linting related to sqlfluff v3.5.0

### DIFF
--- a/src/package-lock.json
+++ b/src/package-lock.json
@@ -15,7 +15,7 @@
         "node-fetch": "3.3.2",
         "node-watch": "0.7.4",
         "prettier": "3.6.1",
-        "puppeteer": "24.29.0",
+        "puppeteer": "24.29.1",
         "rainbow-code": "2.1.7",
         "recursive-readdir": "2.2.3",
         "run-script-os": "1.1.6",
@@ -364,9 +364,9 @@
       "license": "MIT"
     },
     "node_modules/bare-events": {
-      "version": "2.8.1",
-      "resolved": "https://registry.npmjs.org/bare-events/-/bare-events-2.8.1.tgz",
-      "integrity": "sha512-oxSAxTS1hRfnyit2CL5QpAOS5ixfBjj6ex3yTNvXyY/kE719jQ/IjuESJBK2w5v4wwQRAHGseVJXx9QBYOtFGQ==",
+      "version": "2.8.2",
+      "resolved": "https://registry.npmjs.org/bare-events/-/bare-events-2.8.2.tgz",
+      "integrity": "sha512-riJjyv1/mHLIPX4RwiK+oW9/4c3TEUeORHKefKAKnZ5kyslbN+HXowtbaVEqt4IMUB7OXlfixcs6gsFeo/jhiQ==",
       "dev": true,
       "license": "Apache-2.0",
       "peerDependencies": {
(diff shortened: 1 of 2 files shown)


tunetheweb:
I've got some questions:

> It finished linting SQL faster than it takes to download super-linter image.

It took 46 seconds, rather than 1 minute 38 seconds. So yes faster but only a minute faster. Is it worth running something different for SQLFluff for that? What's the issue you're looking to solve here?

Also at the minute "Lint Code Base" will run anyway so we'll need to prevent that running otherwise there's not saving right? This is also a required action so we'd need to remove that and make it optional (not sure that's a big deal - we've already turned off Test Website since it didn't always run). We normally check them anyway so note sure we need the enforcement.

I'm also tried running a total lint:

- With this change it takes [7 mins 45 seconds](https://github.com/HTTPArchive/almanac.httparchive.org/actions/runs/18666113871)
- Using super-linter it takes [7 mins 48 seconds](https://github.com/HTTPArchive/almanac.httparchive.org/actions/runs/18666330965/job/53218345106)

So again no real saving. Both also seemed to pick up new errors btw. We should look to fix thise.

max-ostapenko:
I suggest using the latest, lightweight SQL linter - via PIP.
PIP and super-linter may temporarily differ in versions, and current errors come from new rules in v3.5.0.
We’ll be forced to fix these before dependabot’s PR can auto-merge, if we deduplicate.

BTW Both your runs use PIP, not super-linter.
A full super-linter run takes ~25 min. But to be fair, PIP is roughly 2× faster start to finish.

tunetheweb:
> I suggest using the latest, lightweight SQL linter - via PIP.

Yeah not adverse to that. If we can update this the other points above to not run both.

Since your diff code is relatively easy maybe we should aim tom remove superlinter altogether? Depends if we have Python or JS versions of all the linters.

> BTW Both your runs use PIP, not super-linter.
> A full super-linter run takes ~25 min. But to be fair, PIP is roughly 2× faster start to finish.

🤦‍♂️ Ooops. Silly.

> PIP and super-linter may temporarily differ in versions, and current errors come from new rules in v3.5.0.
> We’ll be forced to fix these before dependabot’s PR can auto-merge, if we deduplicate.

So in this case these are extra checks. So fixing (or skipping with a `--noqa` comment since they are old SQL and not worth worrying too much about) won't break the superlinter version. But yes I agree in theory that could happen — I can't remember if it has in the past—certainly not been a big problem as far as I remember.

max-ostapenko:
I don't think we should drop super-linter - we use ~10 linters and it's easy to just run all of them for the `src/` folder.
SQL is special - it has a dedicated contributors and it's own folder (we know when to trigger linting).

I've updated all the points now.

sql/2019/fonts/06_32.sql
```diff
@@ -20,4 +20,3 @@ FROM (
   ORDER BY
     freq / total DESC
 )
-LIMIT 100
```
tunetheweb, line 23: This seems like the wrong fix. 

Should just remove the 
Outer query (which is no longer needed) and add the LIMIt directly on the inner query 

which had the ORDER BY

tunetheweb: I think the same fix needs to be appled to all the other SQL files you've edited where we're removed the limits.

max-ostapenko: Done.

sql/2021/css/image_dimension_popularity.sql
```diff
@@ -1,59 +1,54 @@
...
+USING (client, page, img_url)
+WHERE
+  height IS NOT NULL AND
+  width IS NOT NULL
+GROUP BY
+  client,
+  height,
+  width
+ORDER BY
+  pct DESC
```
tunetheweb, line 54:
```suggestion
  pct DESC
LIMIT 500  
```

sql/2021/css/keyframes_positions.sql
```diff
@@ -24,36 +24,31 @@ try {
...
 )
-LIMIT 500
+USING (client)
+WHERE
+  date = '2021-07-01'
+GROUP BY
+  client,
+  position
+ORDER BY
+  pct DESC
```
tunetheweb, line 54:
```suggestion
  pct DESC
LIMIT 500
```

sql/2021/css/image_dimension_popularity.sql
```diff
@@ -1,59 +1,55 @@
...
+LIMIT 500
```
tunetheweb:
```suggestion
LIMIT 500
```

PR #4266, opened 2025-11-12
Adds page weight use-efficient-cache-lifetimes.sql Query, plus typo fix for a file name

dwsmart:
Adds use-efficient-cache-lifetimes.sql, which looks at the potential savings available from longer cache headers, using the lighthouse data

Also renames `response_media_file_type.distribution.sql` > `response_media_file_type_distribution.sql` for consistancy


sql/2025/page-weight/use-efficient-cache-lifetimes.sql
```diff
@@ -0,0 +1,18 @@
+-- looks at the distribution of wasted bytes from resources identified by the Lighthouse "Use Efficient Cache Lifetimes" audit
+-- https://developer.chrome.com/docs/performance/insights/cache
+SELECT
+  percentile,
```
tunetheweb, line 4: Should we pivot by root pages? Or want this across both page types?

dwsmart: yeah, your right, it should be just root, or both grouped into root and internal.

To be consistent with the rest of the chapter, I've updated the query to group by `is_root_page`

PR #4277, opened 2025-11-26
Add 2025 Methodology and remove unplanned chapters

tunetheweb:


src/templates/en/2025/methodology.html
```diff
@@ -0,0 +1,559 @@
...
+    COUNT(0) = MAX(COUNT(0)) OVER (PARTITION BY client);
+#}
+
+{% block desktopUserAgent %}Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/138.0.0.0 Safari/537.36 PTST/250604.160032{% endblock %}
+{% block mobileUserAgent %}Mozilla/5.0 (Linux; Android 8.1.0; Moto G (4)) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/138.0.0.0 Mobile Safari/537.36 PTST/250604.160032{% endblock %}
+
+{% block desktopConnectionSpeed %}Cable (5/1 Mbps 28ms RTT){% endblock %}
+{% block mobileConnectionSpeed %}4G (9 Mbps 170ms RTT){% endblock %}
+{% block desktopViewport %}1376 x 768px{% endblock %}
+{% block mobileViewport %}512 x 360px{% endblock %}
```
tunetheweb: @pmeenan can you confirm these are still the details we use? Can you also confirm where I can see this so I don't need to bug you ever year?

src/templates/en/2025/methodology.html
```diff
@@ -0,0 +1,559 @@
...
+{% block lighthouseVersion %}12.8.0{% endblock %}
+
+{% block lighthouseDesktopCPUSlowDown %}<em>N/A</em>{% endblock %}
+{% block lighthouseMobileCPUSlowDown %}1x/4x{% endblock %}
+{% block lighthouseDesktopDownloadThroughout %}1.6 Mbps{% endblock %}
+{% block lighthouseMobileDownloadThroughout %}1.6 Mbps{% endblock %}
+{% block lighthouseDesktopUploadThroughout %}0.768 Mbps{% endblock %}
+{% block lighthouseMobileUploadThroughout %}0.768 Mbps{% endblock %}
+{% block lighthouseDesktopRTT %}150 ms{% endblock %}
+{% block lighthouseMobileRTT %}150 ms{% endblock %}
```
tunetheweb, line 89: @pmeenan can you confirm these are still the details we use? Can you also confirm where I can see this so I don't need to bug you ever year?

src/templates/en/2025/methodology.html
pmeenan: Looks correct. Desktop config is [here](https://github.com/HTTPArchive/crawl/blob/main/Desktop.json) and mobile is [here](https://github.com/HTTPArchive/crawl/blob/main/Mobile.json)

src/templates/en/2025/methodology.html
pmeenan: Correct. Lighthouse's scores were calibrated to a specific profile that hasn't changed. It is independent of the test config and hard-coded [here](https://github.com/HTTPArchive/wptagent/blob/6eae276cc37b136a6ea260def24be6927e748be4/internal/traffic_shaping.py#L102)

tunetheweb: Doesn't that say different settings for desktop?

```
                    # 10Mbps, 40ms RTT
                    rtt = 40
                    in_bps = 10240000
                    out_bps = 10240000
```

src/templates/en/2025/methodology.html
tunetheweb: Doesn't that say desktop is 1920 x 1080 rather than 1376 x 768?

pmeenan: Argh. Yes. I had only looked at the connectivity settings.

tunetheweb: Also we have 28ms RTT but if I'm readying that properly desktop is actually 25ms RTT?

src/templates/en/2025/methodology.html
pmeenan: So it does. Sorry, I forgot we were testing lighthouse on desktop.

src/templates/en/2025/methodology.html
pmeenan: Yes. Usually we subtract the first-hop latency from the physical connection which would explain the difference but I don't think mobile has the same difference.

tunetheweb: Alright updated it to this:

```
{# Get from https://github.com/HTTPArchive/crawl/blob/main/Desktop.json #}
{% block desktopConnectionSpeed %}Cable (5/1 Mbps 25ms RTT){% endblock %}
{% block desktopViewport %}1376 x 768px{% endblock %}
{# Get from https://github.com/HTTPArchive/crawl/blob/main/Mobile.json #}
{% block mobileConnectionSpeed %}4G (9 Mbps 170ms RTT){% endblock %}
{% block mobileViewport %}512 x 360px{% endblock %}
```

src/templates/en/2025/methodology.html
tunetheweb: Updated to this:

```
{# Use this https://github.com/HTTPArchive/wptagent/blob/main/internal/traffic_shaping.py#L102 #}
{% block lighthouseDesktopCPUSlowDown %}<em>N/A</em>{% endblock %}
{% block lighthouseDesktopDownloadThroughout %}10.0 Mbps{% endblock %}
{% block lighthouseDesktopUploadThroughout %}10.0 Mbps{% endblock %}
{% block lighthouseDesktopRTT %}40 ms{% endblock %}
{% block lighthouseMobileCPUSlowDown %}1x/4x{% endblock %}
{% block lighthouseMobileDownloadThroughout %}1.6 Mbps{% endblock %}
{% block lighthouseMobileUploadThroughout %}0.75 Mbps{% endblock %}
{% block lighthouseMobileRTT %}150 ms{% endblock %}
```

src/templates/en/2025/methodology.html
tunetheweb: And now also with the correct desktop viewport :-)

```
{# Get from https://github.com/HTTPArchive/crawl/blob/main/Desktop.json #}
{% block desktopConnectionSpeed %}Cable (5/1 Mbps 25ms RTT){% endblock %}
{% block desktopViewport %}1920 x 1080px{% endblock %}
{# Get from https://github.com/HTTPArchive/crawl/blob/main/Mobile.json #}
{% block mobileConnectionSpeed %}4G (9 Mbps 170ms RTT){% endblock %}
{% block mobileViewport %}512 x 360px{% endblock %}
```

tunetheweb: And again. Sigh...

```
{# Get from https://github.com/HTTPArchive/crawl/blob/main/Desktop.json #}
{% block desktopConnectionSpeed %}Cable (5/1 Mbps 25ms RTT){% endblock %}
{% block desktopViewport %}1920 x 1080px{% endblock %}
{# Get from https://github.com/HTTPArchive/crawl/blob/main/Mobile.json #}
{% block mobileConnectionSpeed %}4G (9 Mbps 170ms RTT){% endblock %}
{% block mobileViewport %}512 x 360px{% endblock %}
```

tunetheweb: Oh I'm gonna add DPR as well as think that's important. And looks like it's 3x on mobile and 1x on desktop.

One more question, is there CPU throttling on WebPageTest? What does this mean in [the mobile settings](https://github.com/HTTPArchive/crawl/blob/main/Mobile.json)?

```
    "throttle_cpu":8,
```

pmeenan: That means 8x slowdown.

tunetheweb: Cool will add that to the table.

I lied about that being the last question. When I run a desktop WPT on webpagetest.httparchive.org it says:

```
First View only
Test runs: 1
Connectivity: 5000/1000 Kbps, 28ms Latency
```

so back to 28ms despite the settings being 25ms. Why’s that? Does it do some adjustment if the settings? Or is that description hard coded somewhere?

pmeenan: The UI has a different place where the config is defined and it looks like they don't perfectly align: https://github.com/HTTPArchive/webpagetest/blob/master/www/settings/connectivity.ini.sample

tunetheweb: Ok I’ll go with 25 as that’s what’s actually used in the test.

PR #4293, opened 2025-12-17
Update PDF URLs to use the new path

max-ostapenko:
The path has changed after replacing a direct GCS bucket backend with a proxy.

It's strange that the false paths in the tests didn't highlight it in earlier PRs.


tunetheweb (approved):
Oh good catch! I’ll also need access to the bucket to be able to upload new PDFs to it.

max-ostapenko:
Write was restricted even before.
So your access should be fine.

PR #4322, opened 2026-01-10
Yash Bio update

tunetheweb:


Yash-Vekaria:
@tunetheweb I see you updated my bio. Thanks.

But, I cannot see the website field.

Just checking to ensure that the website was updated to: https://www.yashvekaria.com/ instead of outdated URL: https://yash-vekaria.github.io that is no longer functional.

tunetheweb:
Ah I misunderstood the request. Reverted the bio change and updated the website instead.

Issue #4084, opened 2025-04-27
PWA 2025


nrllh:
# PWA 2025

![PWA illustration](https://raw.githubusercontent.com/HTTPArchive/almanac.httparchive.org/main/src/static/images/2019/pwa/hero_lg.jpg)

If you're interested in contributing to the PWA chapter of the 2025 Web Almanac, please reply to this issue and indicate which role or roles best fit your interest and availability: [author](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide), [reviewer](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide), [analyst](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Analysts'-Guide), and/or [editor](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Editors'-Guide).  You can find more details about this year’s Call for Contributions [here](https://github.com/HTTPArchive/almanac.httparchive.org/discussions/4062).

💬 Please make sure to join [#web-almanac-pwa](https://join.slack.com/t/httparchive/shared_invite/zt-45sgwmnb-eDEatOhqssqNAKxxOSLAaA) on Slack for team coordination.

## Content team

| Lead | Authors | Reviewers | Analysts | Editors | Coordinator |
| ---- | ------- | --------- | -------- | ------- |    -------  |
|  @diekus  | @diekus @MichaelSolati | @webmaxru @Schweinepriester @aarongustafson |  -  | @aarongustafson | @Yash-Vekaria  |

<details>
<summary><b>Expand for more information about each role 👀</b></summary>

- The **[content team lead](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Content-Team-Leads'-Guide)** is the chapter owner and responsible for setting the scope of the chapter and managing contributors' day-to-day progress.
- **[Authors](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Authors'-Guide)** are subject matter experts and lead the content direction for each chapter. Chapters typically have one or two authors. Authors are responsible for planning the outline of the chapter, analyzing stats and trends, and writing the annual report.
- **[Reviewers](https://github.com/HTTPArchive/almanac.httparchive.org/wiki/Reviewers'-Guide)** are also subject matter experts and assist authors with technical reviews during the planning, analyzing, and writing phases.
[...]

PR #4336, opened 2026-01-12
PWA 2025 queries

onurglr:
Progress on #4084

Updated SQL queries for the 2025 Web Almanac PWA chapter.


tunetheweb:
This looks like it's using June 2025 whereas the Web Almanac is supposed to be based on July 2025. Can you replace all the `2025-06-01` with `2025-07-01` and rerun the queries?

onurglr:
> This looks like it's using June 2025 whereas the Web Almanac is supposed to be based on July 2025. Can you replace all the `2025-06-01` with `2025-07-01` and rerun the queries?

The queries currently use June 1 because I only have access to 2022 reference data.
I can update them to use July 1 and rerun the queries; however, this will take some time. I’ll follow up once it’s done.

onurglr:
> OK I've updated the queries and rerun all the data. Merging this after CI completes.

Thanks Mr. Pollard , just confirming that I’m not expected to start any scans or update the Excel from now on, right?

tunetheweb:
Correct. All done already! Thanks for creating the initial queries as that made it a whole lot easier.

onurglr:
> Correct. All done already! Thanks for creating the initial queries as that made it a whole lot easier.

Thanks, I’m glad I could help

PR #4348, opened 2026-01-12
CDN 2025 chapter

nick-mccord:
Closes #4066

Staged version: https://cdn-2025-dot-webalmanac.uk.r.appspot.com/en/2025/cdn
Staged featured quotes: https://cdn-2025-dot-webalmanac.uk.r.appspot.com/en/2025/?feat=cdn#featured-chapter

PR #4349, opened 2026-01-13
Adding authors and enabling content

nick-mccord:


tunetheweb:
Already parts of #4348 (and needs to be done at the same time anyway).

PR #4354, opened 2026-01-13
Ecommerce 2025 SQL queries

tunetheweb:
Progress on #4072

### DIFF
--- a/sql/2025/ecommerce/core_web_vitals_by_platform.sql
+++ b/sql/2025/ecommerce/core_web_vitals_by_platform.sql
@@ -64,7 +64,7 @@ ON
   CONCAT(origin, '/') = url AND
   IF(device = 'desktop', 'desktop', 'mobile') = client
 WHERE
-  date = '2025-06-01'
+  date = '2025-07-01'
 GROUP BY
   client,
   ecomm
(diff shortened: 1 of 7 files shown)


tunetheweb:
@amandeepsinghvirdi these are all using the `2025-06-01` dataset but we're using the July dataset for the 2025 Web Almanac. Is it too late to change this?

amandeepsinghvirdi:
I'm checking with @onurglr regarding this.

onurglr:
> I'm checking with @onurglr regarding this.

@tunetheweb Do I need to do anything on my side? Sorry for the late reply, I wasn’t at my computer.

tunetheweb:
Nope all good.

onurglr:
> Nope all good.

I’m really sorry that my analyses are not useful since they were run in June. However, when I checked the 2025 queries that were created, I can see that I ran 10 queries for Ecommerce. The list is as follows, but when I check GitHub, it appears that only 7 queries have been added. I just wanted to point that out.

core_web_vitals_passingmetrics_byvendor_bydevice,
core_web_vitals_yoy,
median_lighthouse_score_ecommsites,
top_ecommerce,
top_payment_by_geo,
top_payment_yoy,
top_shipping_by_geo,
top_shipping_yoy,
top_shopsystem_by_geo,
top_vendors_crux_rank

tunetheweb:
OK copied those across from 2024 and updated. Though looks like they weren't used this year or last year. Still better to have them than not.

PR #4357, opened 2026-01-14
Security 2025 chapter edits

clarkio:
Refine language and improve clarity in security.md

### DIFF
--- a/src/content/en/2025/security.md
+++ b/src/content/en/2025/security.md
@@ -887,7 +887,7 @@ A boolean is <a hreflang="en" href="https://httpwg.org/specs/rfc8941.html#boolea
 
 #### Use of `document.domain`
 
-By using [`document.domain`](https://developer.mozilla.org/docs/Web/API/Document/domain), a developer is able to read the domain portion of the current document, as well as set a new domain (only subdomains of the current domain are allowed), after which the browser will use the new domain as origin for the same-origin policy checks. However, the use of this property is now deprecated and browsers may stop supporting the property soon.
+By using [`document.domain`](https://developer.mozilla.org/docs/Web/API/Document/domain), a developer is able to read the domain portion of the current document, as well as set a new domain (only superdomains of the current domain are allowed), after which the browser will use the new domain as origin for the same-origin policy checks. However, the use of this property is now deprecated and browsers may stop supporting the property soon.
 
 <figure>
   <table>


src/content/en/2025/security.md
```diff
@@ -223,17 +223,17 @@ The distribution of valid `max-age` values remains largely the same with the exc
 
 ## Cookies
 
-Cookies are a vital part of the web. They allow websites to save information for use over multiple stateless requests. In order to secure sites' cookies, there are many features built into browsers that (among much more) are reported on in the [Cookies chapter](./cookies). We specifically refer to the [Cookie attributes](./cookies#cookie-attributes), [Cookie prefixes](./cookies#cookie-prefixes) and [Persistence (expiration)](./cookies#persistence-expiration) sections.
+Cookies are a vital part of the web. They allow websites to save information for use over multiple stateless requests. In order to secure sites' cookies, there are many features built into browsers that are further reported on in the [Cookies chapter](./cookies). We specifically refer to the [Cookie attributes](./cookies#cookie-attributes), [Cookie prefixes](./cookies#cookie-prefixes) and [Persistence (expiration)](./cookies#persistence-expiration) sections in this chapter.
```
tunetheweb:
```suggestion
Cookies are a vital part of the web. They allow websites to save information for use over multiple stateless requests. In order to secure sites' cookies, there are many features built into browsers that are further reported on in the [Cookies](./cookies) chapter. We specifically refer to the [Cookie attributes](./cookies#cookie-attributes), [Cookie prefixes](./cookies#cookie-prefixes) and [Persistence (expiration)](./cookies#persistence-expiration) sections in this chapter.
```

src/content/en/2025/security.md
```diff
@@ -223,17 +223,17 @@ The distribution of valid `max-age` values remains largely the same with the exc
...
 
-CSP can be used to block specific resources from being loaded, which can help reduce the impact of potential XSS attacks. In addition CSP can also serve other purposes, such as enforcing the use of encrypted communication channels by means of the `update-insecure-requests` directive or controlling on which pages the current page can be loaded as a subresource using the `frame-ancestors` directive. This allows websites to defend against clickjacking attacks.
+CSP can be used to block specific resources from being loaded, which can help reduce the impact of potential cross-site script (XSS) attacks. In addition CSP can also serve other purposes, such as enforcing the use of encrypted communication channels by means of the `update-insecure-requests` directive or controlling on which pages the current page can load as a subresource using the `frame-ancestors` directive. This allows websites to defend against clickjacking attacks.
```
tunetheweb:
```suggestion
CSP can be used to block specific resources from being loaded, which can help reduce the impact of potential cross-site scripting (XSS) attacks. In addition CSP can also serve other purposes, such as enforcing the use of encrypted communication channels by means of the `update-insecure-requests` directive or controlling on which pages the current page can load as a subresource using the `frame-ancestors` directive. This allows websites to defend against clickjacking attacks.
```

src/content/en/2025/security.md
```diff
@@ -887,7 +887,7 @@ A boolean is <a hreflang="en" href="https://httpwg.org/specs/rfc8941.html#boolea
 
 #### Use of `document.domain`
 
-By using [`document.domain`](https://developer.mozilla.org/docs/Web/API/Document/domain), a developer was able to read the domain portion of the current document, as well as set a new domain (only superdomains of the current domain are allowed), after which the browser will use the new domain as origin for the same-origin policy checks. However, the use of this property is now deprecated and browsers may stop supporting the property soon.
+By using [`document.domain`](https://developer.mozilla.org/docs/Web/API/Document/domain), a developer is able to read the domain portion of the current document, as well as set a new domain (only subdomains of the current domain are allowed), after which the browser will use the new domain as origin for the same-origin policy checks. However, the use of this property is now deprecated and browsers may stop supporting the property soon.
```
tunetheweb: I'm not sure about this change. I think superdomain was intentional.

https://developer.mozilla.org/en-US/docs/Web/API/Document/domain#setting_the_domain
> It can only be set to the same or a parent domain. For example, if https://a.example.com and https://b.example.com both use `document.domain = "example.com";` then they have both modified their origin to have the same domain, and they can now access each other's DOM directly—despite being cross-origin, which would normally prevent such access.

Maybe should be more explicit (like MDN is) in case others think this is a typo?

src/content/en/2025/security.md
```diff
@@ -929,13 +929,13 @@ By using [`document.domain`](https://developer.mozilla.org/docs/Web/API/Document
...
 
 As previously mentioned, a Content Security Policy (CSP) can be effective against <a hreflang="en" href="https://owasp.org/www-community/attacks/Clickjacking">Clickjacking</a> attacks through the use of the `frame-ancestors` directive. Some of the top CSP header values include a `frame-ancestors` directive with a `'none'` or `'self'` value, thereby blocking embedding of the page overall or restricting the embeddings to pages of the same origin.
 
-Another way of defending against clickjacking attacks is through the [`X-Frame-Options` (XFO)](https://developer.mozilla.org/docs/Web/HTTP/Headers/X-Frame-Options) header. By setting the XFO developers can communicate that a document cannot be embedded in other documents ('DENY') or can only be embedded in documents of the same origin (`SAMEORIGIN`).
+Another way of defending against clickjacking attacks is through the [`X-Frame-Options` (XFO)](https://developer.mozilla.org/docs/Web/HTTP/Headers/X-Frame-Options) header. By setting the XFO header developers can communicate that a document cannot be embedded in other documents ('DENY') or can only be embedded in documents of the same origin (`SAMEORIGIN`).
```
tunetheweb:
```suggestion
Another way of defending against clickjacking attacks is through the [`X-Frame-Options` (XFO)](https://developer.mozilla.org/docs/Web/HTTP/Headers/X-Frame-Options) header. By setting the XFO header, developers can communicate that a document cannot be embedded in other documents ('DENY') or can only be embedded in documents of the same origin (`SAMEORIGIN`).
```

src/content/en/2025/security.md
clarkio: tbh I wasn't sure either but I saw no mention of "superdomain" in the MDN page which led me to believe subdomain may have been intended here. Who can we double check with on this one?

tunetheweb: @vikvanderlinden @GJFR 

But I'm pretty confident they meant superdomain so let's revert to that for now:

```suggestion
By using [`document.domain`](https://developer.mozilla.org/docs/Web/API/Document/domain), a developer is able to read the domain portion of the current document, as well as set a new domain (only superdomains of the current domain are allowed), after which the browser will use the new domain as origin for the same-origin policy checks. However, the use of this property is now deprecated and browsers may stop supporting the property soon.
```

tunetheweb (approved):

src/content/en/2025/security.md
vikvanderlinden: Yes, this is superdomain. Although maybe we can rephrase to more explicitly include the same domain as well?

From the linked MDN document:

```
...
It can only be set to the same or a parent domain. For example, if https://a.example.com and https://b.example.com both use

document.domain = "example.com";

then they have both modified their origin to have the same domain, and they can now access each other's DOM directly
...
```

tunetheweb: Lets leave as is and merge this.

PR #4356, opened 2026-01-14
WebAssembly 2025 chapter

tunetheweb:
Closes #4090

Staged version: https://webassembly-2025-dot-webalmanac.uk.r.appspot.com/en/2025/webassembly
Staged featured quotes: https://webassembly-2025-dot-webalmanac.uk.r.appspot.com/en/2025/?feat=webassembly#featured-chapter

PR #4362, opened 2026-01-15
Banner for WebAssembly report

nimeshgit:
Regarding https://github.com/HTTPArchive/almanac.httparchive.org/pull/4356,

With reference of our report Year 2021 / [WebAssembly](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/static/images/2021/webassembly), I have updated banner files. 

1. 3 large size (866x433) banners in format of avif, jpg and webp.
2. 3 small size (326x163) banners in format of avif, jpg and webp.


tunetheweb:
@nimeshgit thanks for doing this but after careful consideration I'd prefer to keep with the image we've used in the past. There are a number of reasons for this:
1. Consistency across years
2. Consistency in style, we generally avoid text in the hero images as they can cause a number of issues, including:
  - Future proofing (you have the year in the image)
  - Future proofing (you have Spec 3.0 in the image, but if/when 4.0 comes along this will look out of date and/or require updating)
  - Internationalisation as we translate out chapters.

Sorry for wasting your time asking you to create the various versions and now not accepting it.

PR #4401, opened 2026-01-29
WebAssembly edits

nrllh:
I have edited the chapter. Could you folks please take a look at it and take further adjustments? 

@nimeshgit : we still need the tool "almanac-wasm". Can you please provide it to us?


src/content/en/2025/webassembly.md
```diff
@@ -22,40 +22,22 @@ doi: 10.5281/zenodo.18258991
...
 
-### The mission
+WebAssembly has evolved from a web-centric optimization tool into a high-performance, universal bytecode format. Officially <a hreflang="en" href="https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/">a W3C standard since 2019</a>, the ecosystem reached a technical turning point with the release of <a hreflang="en" href="https://webassembly.github.io/spec/core/">Wasm Version 3.0 in December 2025</a>. This version standardizes several <a hreflang="en" href="https://webassembly.github.io/spec/core/">advanced features</a> — such as Garbage Collection, 64-bit address space, and Multiple Memories — allowing high-level languages like Java, Kotlin, and Dart to run natively and efficiently in both browser and standalone environments.
```
tunetheweb: There's repetition in both these paragraphs. Can you address?

src/content/en/2025/webassembly.md
```diff
@@ -22,40 +22,22 @@ doi: 10.5281/zenodo.18258991
...
-
-In essence, Wasm is moving beyond the browser to provide a secure, high-performance, and portable universal runtime for modern computing.
 
 ## Methodology
 
-We follow [the same methodology from the 2021 Web Almanac](../2021/webassembly#methodology), which is the first edition that included WebAssembly. WebAssembly is compiled into bytecode and distributed as binary format. The findings here infer the source language used in the modules. The accuracy of that analysis is covered in more detail in the respective sections.
+We follow the same methodology from [the 2021 Web Almanac](../2021/webassembly#methodology), where WebAssembly was introduced for the first time.
 
-### Limitations
+**Data Collection:** This chapter relies on this dataset provided by HTTP Archive Juli 2025 crawl data which is hosted on Google BigQuery. to identify WebAssembly modules by matching the `Content-Type` (`application/wasm`) and the `.wasm` file extension. Using this method, we identified 233,857 Wasm modules on desktop and 255,060 on mobile.
```
tunetheweb: We avoid absolute numbers because they are basically meaningless.

233,857 sounds a lot, but is it? It depends how many sites we crawl. If this is from top million that;'s a lot! If it's from top billion, not so much. The reader has no context here. That's why we use percentages to give that context.

It's maybe ok to use absolute numbers if you also provide context. For example: "Using this method, we identified 233,857 Wasm modules on desktop and 255,060 on mobile across X% of websites in out crawl".

But ultimately I don't get what the point of this paragraph is. Is it to show the scale of our analysis? Again, I'm not sure that's that meaningful without understand what that scale means.

src/content/en/2025/webassembly.md
```diff
@@ -22,40 +22,22 @@ doi: 10.5281/zenodo.18258991
...
 
-- **Data Acquisition & Retrieval** Our analysis relies on the almanac-wasm tool to fetch WASM binaries using parameters from the initial HTTP Archive crawl. However, retrieval depends on third-party server availability; 404/403 errors or resource updates since the initial crawl may occur. We assume the retrieved file is identical to the one present during the original recording.
+**Analysis:** In addition to the HTTP Archive dataset, we use the <a hreflang="en" href="#/">almanac-wasm</a> tool to download and validate the WebAssembly modules identified from the HTTP Archive for local analysis. This tool extracts metadata from these downloaded files, allowing us to identify programming languages, libraries, and specific features used within the Wasm modules.
```
tunetheweb: Where's the href supposed to point to here?

```suggestion
**Analysis:** In addition to the HTTP Archive dataset, we use <a hreflang="en" href="#/">almanac-wasm</a> a tool to download and validate the WebAssembly modules identified from the HTTP Archive for local analysis. This tool extracts metadata from these downloaded files, allowing us to identify programming languages, libraries, and specific features used within the Wasm modules.
```

src/content/en/2025/webassembly.md
```diff
@@ -22,40 +22,22 @@ doi: 10.5281/zenodo.18258991
...
 
-  - **Limitations**: This is not a definitive approach. "Stripped" binaries (using tools like *wasm-strip*) or obfuscated code remove the debug info and metadata required for identification.
-  - **Example**: Compilers like TinyGo generate unique structures and WASI imports that often diverge from standard *go_exec.js* patterns, resulting in these binaries being classified as "Unknown".
+**Limitations:** Our tool `àlmanac-wasm` focuses on static analysis of Wasm modules and does not execute them. Therefore, we cannot capture dynamic behaviors or runtime features that may be present during actual execution in a browser or standalone environment. Additionally, some Wasm modules may be obfuscated or minified, which can limit our ability to accurately identify their characteristics.
```
tunetheweb: Accented character

```suggestion
**Limitations:** Our tool `almanac-wasm` focuses on static analysis of Wasm modules and does not execute them. Therefore, we cannot capture dynamic behaviors or runtime features that may be present during actual execution in a browser or standalone environment. Additionally, some Wasm modules may be obfuscated or minified, which can limit our ability to accurately identify their characteristics.
```

Also what's the difference between `almanc-wasm` and https://github.com/HTTPArchive/wasm-stats ?

src/content/en/2025/webassembly.md
```diff
@@ -68,7 +50,7 @@ We follow [the same methodology from the 2021 Web Almanac](../2021/webassembly#m
   )
 }}
 
-We see that 0.35% of desktop sites and 0.28% of mobile sites are using WebAssembly. This is approximately 43,000 sites in our dataset for both, but with a larger dataset for mobile the relative percentage is lower.
+Our analysis shows while WebAssembly's adoption in 2021 0.04% was, we find that in 2025, it has grown to 0.35% on desktop and 0.28% on mobile, representing approximately 43,000 sites for each platform within our dataset.
```
tunetheweb:
```suggestion
Our analysis shows while WebAssembly's [adoption in 2021 was 0.04%](../2021/webassembly#how-many-modules), in 2025 it has grown to 0.35% on desktop and 0.28% on mobile, representing approximately 43,000 sites for each platform within our dataset.
```

src/content/en/2025/webassembly.md
```diff
@@ -82,52 +64,50 @@ We see that 0.35% of desktop sites and 0.28% of mobile sites are using WebAssemb
   )
 }}
 
-Interestingly, looking over previous Web Almanac years, WebAssembly requests have increased drastically from ~0.05% in 2021 to 0.28%/0.35% this year. There was also a slight dip from 2024.
+We find that WebAssembly adoption has grown from 0.04% in 2021, although rates have remained consistent over the last two years. In 2025, we found WebAssembly modules on 0.35% of desktop sites and 0.28% of mobile sites, representing approximately 43,000 sites. Furthermore, mobile adoption was lower than desktop adoption in all observed years, averaging a 30% difference.
```
tunetheweb, line 62: Repeat of above. Remove one.

src/content/en/2025/webassembly.md
```diff
@@ -82,52 +64,50 @@ We see that 0.35% of desktop sites and 0.28% of mobile sites are using WebAssemb
...
 
 ### WebAssembly by rank
 
 {{ figure_markup(
   image="webassembly-by-rank.png",
-  caption="Page Ranking (Group) for Web Assembly 2025",
+  caption="Adoption of WebAssembly by site rank",
```
tunetheweb: We generally use the chart title for the caption:

```suggestion
  caption="WebAssembly usage site rank",
```

src/content/en/2025/webassembly.md
```diff
@@ -82,52 +64,50 @@ We see that 0.35% of desktop sites and 0.28% of mobile sites are using WebAssemb
...
-
-- **The "Long Tail" Effect**: While the percentage of adoption drops significantly as you move down there ranks, there is still usage across all ranks.
+We see a strong correlation between site popularity and WebAssembly adoption. Usage is most concentrated among the top 1,000 websites, reaching 2% on desktop and 1.27% on mobile. These top-tier sites frequently host complex applications—such as design tools or heavy media editors—that require the high performance Wasm provides.
 
-- **Platform Parity**: There is a very close correlation between desktop and mobile usage across all ranking groups. This confirms that Wasm is being used as a cross-platform solution rather than being restricted to desktop-only environments.
+Adoption rates decrease as site rank declines, following a consistent distribution pattern. For sites outside the top 10 million, adoption is approximately 0.33% for desktop and 0.28% for mobile. While desktop usage remains higher in the top ranking groups, the gap narrows significantly in the long tail, suggesting that for the majority of the web, WebAssembly is deployed as a cross-platform resource rather than being restricted to specific environments.
```
tunetheweb: Two different points here (both interesting!) so let's separate so they don't get lost in amongst each other:

```suggestion
Adoption rates decrease as site rank declines, following a consistent distribution pattern. For sites outside the top 10 million, adoption is approximately 0.33% for desktop and 0.28% for mobile.

While desktop usage remains higher in the top ranking groups, the gap narrows significantly in the long tail, suggesting that for the majority of the web, WebAssembly is deployed as a cross-platform resource rather than being restricted to specific environments.
```

src/content/en/2025/webassembly.md
```diff
@@ -82,52 +64,50 @@ We see that 0.35% of desktop sites and 0.28% of mobile sites are using WebAssemb
...
   )
 }}
 
-There are over 300,000 wasm requests in our dataset, which come down to 32,197 unique wasm file requests on desktop and 29,997 on mobile. The large number of requests shows many sites are requesting multiple (hundreds in come cases!) WASM files.
+
+Overall, we recorded 303,496 WebAssembly requests on desktop and 308,971 on mobile. Although more desktop sites utilize WebAssembly, the total volume of requests is slightly higher on mobile.
+
+Furthermore, we identified 157,967 unique URLs on desktop and 165,870 on mobile. To estimate the number of unique binaries, we grouped modules by identical filename and response size. Using this method, we found 87,596 unique Wasm modules on desktop and 84,851 on mobile. These findings indicate that approximately 72% of WebAssembly requests serve duplicate modules, highlighting substantial reuse of libraries across the web.
+
+
```
tunetheweb: Nit, remove unnecessary extra new line:

```suggestion

```

src/content/en/2025/webassembly.md
```diff
@@ -82,52 +64,50 @@ We see that 0.35% of desktop sites and 0.28% of mobile sites are using WebAssemb
...
   sql_file="counts.sql",
   sheets_gid="540023407"
   )
 }}
 
-There are over 300,000 wasm requests in our dataset, which come down to 32,197 unique wasm file requests on desktop and 29,997 on mobile. The large number of requests shows many sites are requesting multiple (hundreds in come cases!) WASM files.
+
+Overall, we recorded 303,496 WebAssembly requests on desktop and 308,971 on mobile. Although more desktop sites utilize WebAssembly, the total volume of requests is slightly higher on mobile.
+
+Furthermore, we identified 157,967 unique URLs on desktop and 165,870 on mobile. To estimate the number of unique binaries, we grouped modules by identical filename and response size. Using this method, we found 87,596 unique Wasm modules on desktop and 84,851 on mobile. These findings indicate that approximately 72% of WebAssembly requests serve duplicate modules, highlighting substantial reuse of libraries across the web.
```
tunetheweb: Here we use absolute numbers again, but I think that's OK cause the point here is the amount of duplication. So OK to leave this as is.

src/content/en/2025/webassembly.md
```diff
@@ -136,17 +116,11 @@ There are over 300,000 wasm requests in our dataset, which come down to 32,197 u
   )
 }}
 
-We observed that the MIME type `application/wasm` is used in 293,470 requests on desktop clients and 301,127 requests on mobile clients.
+The standard `application/wasm` MIME type was identified in 293,470 desktop and 301,127 mobile requests. Instances of missing or incorrect MIME types (such as `text/html` or `text/plain`) were low, affecting 3.2% of desktop and 2.4% of mobile requests. These represent a significant decline compared to 2021, indicating improved awareness and adherence to proper server configuration.
```
tunetheweb:
```suggestion
The `application/wasm` MIME type was identified in 293,470 desktop and 301,127 mobile requests. Instances of missing or incorrect MIME types (such as `text/html` or `text/plain`) were low, affecting 3.2% of desktop and 2.4% of mobile requests. These represent a significant decline compared to 2021, indicating improved awareness and adherence to proper server configuration.
```

src/content/en/2025/webassembly.md
```diff
@@ -158,6 +132,13 @@ Small modules of WebAssembly often found using AssemblyScript or Rust and Large
   )
 }}
 
+
+
```
tunetheweb:
```suggestion

```

src/content/en/2025/webassembly.md
```diff
@@ -158,6 +132,13 @@ Small modules of WebAssembly often found using AssemblyScript or Rust and Large
...
+WebAssembly module sizes vary drastically based on their specific use cases. We observed that the bottom 50% of modules are quite small, ranging between 2 KB and 14 KB. These are typically "micro-utilities" like Base64 encoders or checksum calculators, often written in AssemblyScript or Rust to handle performance-critical tasks where JavaScript lacks precision.
+
+Conversely, at the 90th percentile, sizes increase significantly to 381 KB on desktop and 316 KB on mobile. These larger binaries usually represent full desktop-grade applications ported to the web—such as Adobe Photoshop or Google Earth—compiled from heavier languages like C++ or C# to handle complex 3D rendering and logic.
```
tunetheweb: 381 KB still sounds small to me for those! JS delivers MB of JS and mobile apps are often 100s of mega bytes. Any thoughts on why _only_ 381 KB at 90th percentile?

src/content/en/2025/webassembly.md
```diff
@@ -158,6 +132,13 @@ Small modules of WebAssembly often found using AssemblyScript or Rust and Large
...
+
+
```
tunetheweb:
```suggestion

```

src/content/en/2025/webassembly.md
```diff
@@ -136,17 +116,11 @@ There are over 300,000 wasm requests in our dataset, which come down to 32,197 u
...
 ### Module size
 
-The smallest WebAssembly modules are likely used for specific functions, such as "Micro-Utility" like _Base64 Encoder/Decoder_ or a _CRC32 Checksum_ utility—These are typically used for performance-critical calculations or polyfills where JavaScript might be too slow or lack the specific precision needed.
-
-Larger modules are probably full applications compiled to WebAssembly. These are massive modules where an entire desktop-grade codebase (often millions of lines of C++ or Rust) is compiled to run in the browser. For example, Adobe Photoshop Web, AutoCAD, and Google Earth. Large modules  handle complex image rendering, layer management, and 3D engine calculations directly on the client side browser.
-
-Small modules of WebAssembly often found using AssemblyScript or Rust and Large modules of WebAssembly often found using languages like C++ or C# (.net). You can get more details on the Wasm Language Usage section.
 
 {{ figure_markup(
   image="raw-response-sizes.png",
```
tunetheweb, line 121: Is "raw" the right word here? I read that as the uncompressed size, but later on I see we look at uncompressed side so by "raw" I presume you mean "network bytes"? We should change this to be clear what "raw" means.

src/content/en/2025/webassembly.md
```diff
@@ -168,7 +149,9 @@ Small modules of WebAssembly often found using AssemblyScript or Rust and Large
   )
 }}
 
-These WebAssembly modules differ considerably in size, with the smallest being just a few kilobytes, and the largest one is 228.102 MB in desktop's client and 166.415 MB for mobile client.
+When examining uncompressed sizes, we observe that while the median module remains lightweight at approximately 30 KB on both platforms, the largest binaries at the 90th percentile are significantly heavier on desktop (897 KB) than on mobile (756 KB).
```
tunetheweb: Is Wasm not typically binary? What compression does it use? Gzip? Brotli? I thought those only really worked on text resources (maybe explaining the relatively low different between these two charts?) and they shouldn't be further compressed for sending on the wire?

src/content/en/2025/webassembly.md
```diff
@@ -168,7 +149,9 @@ Small modules of WebAssembly often found using AssemblyScript or Rust and Large
...
+
+
```
tunetheweb, line 153:
```suggestion

```

src/content/en/2025/webassembly.md
```diff
@@ -179,9 +162,11 @@ These WebAssembly modules differ considerably in size, with the smallest being j
   )
 }}
 
+Beyond these standard distributions, the dataset contains significant outliers. The largest single WebAssembly module identified measured 228 MB on desktop and 166 MB on mobile, indicating the deployment of large-scale client-side applications.
```
tunetheweb: Or just badly optimised applications?

Also I'm not sure what "large-scale" here means? Just "large"?

src/content/en/2025/webassembly.md
```diff
@@ -195,27 +180,14 @@ Our crawl identified a modest number of modules, it is possible to analyze and l
...
-
-- **Library : Microsoft (23.2%)** represents the massive footprint of the Microsoft ecosystem on the web, primarily driven by Blazor WebAssembly. Blazor allows developers to build interactive web UIs using C# and .NET instead of JavaScript. The high percentage reflects many enterprise and business applications that have been ported to the web using Microsoft's specialized Wasm runtime for the .NET framework.
 
-- **Library : RXEngine (6.2%)** is a more specialized entry, often associated with high-performance execution engines used for specific industries like gaming or advanced data processing. While more niche than the top two, its 6.2% share indicates it is a popular choice for developers who need a pre-built, optimized engine to handle computationally intensive tasks (such as real-time analytics or complex UI interactions) without building the entire infrastructure from scratch.
+We find that System (43%), Microsoft (23%), RXEngine (6%), and Dotnet (6%) are the most popular libraries or frameworks used in WebAssembly modules, indicating Microsoft's dominance within this ecosystem, driven specifically by the Dotnet and Blazor frameworks.
```
tunetheweb: What is System btw? And who owns that? Seems weird not to mention the biggest one - or is that Microsoft too?

Links for all these would also be helpful as Googling "System Wasm" for example unsurprisingly isn't helpful.

Finally why do you mention "Dotnet" and "Blazor" but not Microsoft which is 23%?

src/content/en/2025/webassembly.md
```diff
@@ -249,14 +223,10 @@ The initial release of WebAssembly was considered an Minimal Viable Product (MVP
...
 
 ## Conclusions
 
-There is a significant increase in the number of webpages using this technology for serverless, containerization, machine learning components and plug-n-play types of applications. The future of WebAssembly could be as a niche web technology, but as an entirely mainstream runtime on a wide range of other platforms. WebAssembly runtime (multi-language, lightweight, secure) are making it a popular choice for a wider range of non-browser applications for agnostic platforms.
+We observe a significant increase in WebAssembly adoption over the last five years, rising from 0.04% in 2021 to 0.35% in 2025, though growth has stabilized in the last two years. Usage is most prevalent on high-ranking websites and decreases significantly among less popular pages. We find that WebAssembly is currently deployed for two distinct purposes: handling specific utility functions (such as encryption or checksums) and powering full standalone applications. Furthermore, our findings highlight the widespread adoption of Microsoft’s frameworks, indicating their significant role in driving the current WebAssembly ecosystem.
```
tunetheweb: We (over)use "we observe" a lot in this chapter. Can we think of alternative wording here to avoid that repitition? To me just going with this reads better:

```suggestion
WebAssembly adoption has significantly increased over the last four years, rising from 0.04% in 2021 to 0.35% in 2025, though growth has stabilized in the last two years. Usage is most prevalent on high-ranking websites and decreases significantly among less popular pages. We find that WebAssembly is currently deployed for two distinct purposes: handling specific utility functions (such as encryption or checksums) and powering full standalone applications. Furthermore, our findings highlight the widespread adoption of Microsoft’s frameworks, indicating their significant role in driving the current WebAssembly ecosystem.
```

I also changed "five years" to "four years" as talking about 2021 to 2025.

tunetheweb (changes requested):

src/content/en/2025/webassembly.md
nrllh: I think, it comes from https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2025/webassembly/total_sizes_by_percentile.sql#L6C41-L6C71 

So it should be network bytes?

src/content/en/2025/webassembly.md
nrllh: My understanding from the data we have is that one WebAssembly application often relies on multiple other small modules, which increases the total number of modules and causes also these statistics here at 90th percentile .

src/content/en/2025/webassembly.md
nrllh: @nimeshgit could you please help us to understand this.

src/content/en/2025/webassembly.md
nrllh: I am not super confident, but after a bit Googling, System is the library used in most Microsoft's Wasm frameworks. 

@nimeshgit can you please help us to clarify it?

Do you have further data we can understand these libraries (and their origin) better?

nimeshgit: Yes, I would also agree with you on your googling :) because The term "System" is extremely common for project names, particularly within IT, software development, engineering, and enterprise contexts. It is frequently used to denote a comprehensive, integrated solution rather than just a single tool. 

For Microsoft technologies, It has the default namespace named System as well.  It is likely that most of them are owned by Microsoft but We can not confirm that all the "System" wasm modules are owned by Microsoft with current analysis.

To step further and confirm whether System wasm belongs to Microsoft Technology or not, We need to cross check with other analysis techniques like what is the its source of the language ? is that co-relates with Microsoft Technologies or not ? or may be binary comparision with various versions of existing Microsoft's System wasm libraries.

We will have link as per your suggession.

Dotnet may also include mono stacks, stripped down obfuscation blazor stacks, native c++ stacks and or other 3rd party technologies as well. To understand the very specific usage to the technology i.e. dotnet or blazor, We had seperated them here.  If We give under single Microsoft technology, We would lose the detailed analysis. However It is also good to have analysis stats for Microsoft technology usage as the bigger picture.

src/content/en/2025/webassembly.md
nimeshgit: ** number stats are updated, can we please update string as below :

**Data Collection:** This chapter relies on this dataset provided by HTTP Archive Juli 2025 crawl data which is hosted on Google BigQuery. to identify WebAssembly modules by matching the `Content-Type` (`application/wasm`) and the `.wasm` file extension. For language usage stats, we identified 64.3% of desktop and 72.8% of mobile Wasm modules to analyse lanaguage stats, the rest Wasm modules fall under failures when downloading wasm file and wasm validation.

src/content/en/2025/webassembly.md
tunetheweb: OK so System, Microsoft, and DotNet are ALL Microsoft (with caveats about System maybe not being all Microsoft)? If so that makes sense why you say Microsoft is such a big placer but I definitely did not get that from the reading. Maybe that's obvious to Wasm experts but I think it could be made much clearer.

nimeshgit:
> I have edited the chapter. Could you folks please take a look at it and take further adjustments?
> 
> @nimeshgit : we still need the tool "almanac-wasm". Can you please provide it to us?

please, have the link [almanac-wasm-stats](https://github.com/nimeshgit/almanac-wasm-stats)
We are updating stack still that includes documentation and tests.

Later on We would be raising pull request and will merge the stack under HTTPArchive and enable in the pipeline with the help of Patrick and Barry.

src/content/en/2025/webassembly.md
nimeshgit: yes Barry, raw means uncompressed, the term was introduced in 2021.

src/content/en/2025/webassembly.md
nimeshgit: Yes, Barry you have an interesting point of view !
If We think about JS code with bytecode, JS is the human readable, high level source code, while bytecode is a low level, intermediate representation of that code that is machine agnostic. 

we know the modern JS engines, such as Google's V8, convert JS source code into bytecode internally as part of the execution process; this shows the bytecode's efficiency compare to JS' in small size.

src/content/en/2025/webassembly.md
nimeshgit: Barry, The Wasm is fundamentally a binary instruction format, not typically a text format. the format sent over the wire is the compact binary format (wasm). 

It has built in compression.

It is observed that applying Gzip or Brotli to an already highly optimized binary file yields minimal additional size savings (often less than 10%, sometimes even negligible), Because of the minimal benefit of compression, Barry the Wasm files are typically served with an identity encoding, meaning no further general purpose compression is applied

tunetheweb: Is it worth commenting on this?

At the moment we show raw (which I think should be renamed "compress" as per other conversation to make it more obvious) and uncompressed, but never really explain that it's not really expected for them to be compressed. And that that explains why the charts aren't that different. By presenting both, it's almost like the opposite and we think it is something to be looked at, almost like we expect there to be interesting insights here?

So I went back to the 2021 chapter that we seem to be basing this on and it has this to say:

"...one of the benefits of Wasm bytecode is that it’s highly compressible, and size over the wire is what matters for download speed and billing reasons. Let’s check sizes of raw response bodies as sent by servers instead:"

That's a good narrative for why they included both.

But the 2025 chapter doesn't seem to have a similar narrative and just says "here's the raw bytes, here's the compressed bytes" and I'm not sure what the readers's supposed to take away from that.

Additionally it contradicts what you say about compression (and what I believed too).

Again, as a reader I'm confused as to what to take away from this section?

src/content/en/2025/webassembly.md
nimeshgit: please have url : https://github.com/nimeshgit/almanac-wasm-stats

[Further reference](https://github.com/HTTPArchive/almanac.httparchive.org/pull/4401#issuecomment-3823882643)

src/content/en/2025/webassembly.md
nimeshgit: Please have updated break-down stats with details in [excel sheet under language usage](https://docs.google.com/spreadsheets/d/16z2MNwq8FFbuNYcJJZceML6rB5VAmBXNNHZy5FZuf8g/edit?gid=101432539#gid=101432539).

src/content/en/2025/webassembly.md
nimeshgit: [With reference to our draft doc](https://docs.google.com/document/d/1FwWmk0MFjPnXa1UHNzpWNrNAbuqub5CS3G02OLVpIE4/edit?tab=t.0),

Tool : almanac-wasm-stats has enhancements as below

1. It can take inputs in **url and respective user-agent strings** that improves download task.
2. Accept huge number of input in the format of **BigQuery's JSONL result**.
3. **Validate Wasm** and provide insights with Binary Toolkit that helps to improve stats (ref. wasm2wat)
4. **Run and Trace tasks activities concurrently** i.e. wasm file downloading, validating and populating stats.
5. **Enhances language identifiers** for old rust implimentation ([wasm-stats](https://github.com/HTTPArchive/wasm-stats)) and added new languages : Scala, Dotnet/Mono, Go & TinyGo, TeaVM based languages, Kotlin -- This reduces the language usage : "Unknown" numbers and improves language stats.
6. Produce **full Wasm language usage stats** along with validation and download failures.
7. Tool's **plug-n-play architecture** that helps to introduce new stats with WebAssembly Toolkit / SDK  in JSON existing stats format for future enhancements.

src/content/en/2025/webassembly.md
nimeshgit: I think It is okay to have year 2021 reference.

src/content/en/2025/webassembly.md
nimeshgit: Hi Nurullah and Barry, 
It seems to resolve this concern and to make it easy to understand,
It would be good to have the compression method chart along with its details.

What do you say, guys ?

src/content/en/2025/webassembly.md
nimeshgit: Link
Microsoft : https://learn.microsoft.com/en-us/aspnet/core/blazor/webassembly-build-tools-and-aot?view=aspnetcore-10.0

nimeshgit: We have run tool offline, found the libraries + framework stats which are specifically used under Microsoft eco system,  These numbers are interesting, please have a look from [here](https://docs.google.com/spreadsheets/d/16z2MNwq8FFbuNYcJJZceML6rB5VAmBXNNHZy5FZuf8g/edit?gid=1181339291#gid=1181339291)

src/content/en/2025/webassembly.md
```diff
@@ -22,40 +22,20 @@ doi: 10.5281/zenodo.18258991
...
+WebAssembly (Wasm) has evolved from a web-centric optimization tool into a high-performance, universal bytecode format. Officially <a hreflang="en" href="https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/">a W3C standard since 2019</a>, the ecosystem reached a technical turning point with the release of <a hreflang="en" href="https://webassembly.github.io/spec/core/">Wasm Version 3.0 in December 2025</a>. This version standardizes several <a hreflang="en" href="https://webassembly.github.io/spec/core/">advanced features</a> — such as Garbage Collection, 64-bit address space, and Multiple Memories — allowing high-level languages like Java, Kotlin, and Dart to run natively and efficiently in both browser and standalone environments.
```
tunetheweb:
```suggestion
WebAssembly (Wasm) has evolved from a web-centric optimization tool into a high-performance, universal bytecode format. Officially <a hreflang="en" href="https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/">a W3C standard since 2019</a>, the ecosystem reached a technical turning point with the release of <a hreflang="en" href="https://webassembly.github.io/spec/core/">Wasm Version 3.0 in December 2025</a>. This version standardizes several <a hreflang="en" href="https://webassembly.github.io/spec/core/">advanced features</a>—such as Garbage Collection, 64-bit address space, and Multiple Memories—allowing high-level languages like Java, Kotlin, and Dart to run natively and efficiently in both browser and standalone environments.
```

src/content/en/2025/webassembly.md
```diff
@@ -22,40 +22,20 @@ doi: 10.5281/zenodo.18258991
...
-
-### Limitations
+We follow the same methodology from [the 2021 Web Almanac](../2021/webassembly#methodology), where WebAssembly was introduced for the first time.
 
-- **Data Acquisition & Retrieval** Our analysis relies on the almanac-wasm tool to fetch WASM binaries using parameters from the initial HTTP Archive crawl. However, retrieval depends on third-party server availability; 404/403 errors or resource updates since the initial crawl may occur. We assume the retrieved file is identical to the one present during the original recording.
+**Data Collection:** This chapter relies on this dataset provided by HTTP Archive Juli 2025 crawl data which is hosted on Google BigQuery. to identify WebAssembly modules by matching the `Content-Type` (`application/wasm`) and the `.wasm` file extension. Using this method, we found at least one WebAssembly module on 43,000 sites, representing 0.35% of all sites analyzed.
```
tunetheweb:
```suggestion
**Data Collection:** This chapter relies on this dataset provided by HTTP Archive July 2025 crawl data which is hosted on Google BigQuery. to identify WebAssembly modules by matching the `Content-Type` (`application/wasm`) and the `.wasm` file extension. Using this method, we found at least one WebAssembly module on 43,000 sites, representing 0.35% of all sites analyzed.
```

src/content/en/2025/webassembly.md
```diff
@@ -22,40 +22,20 @@ doi: 10.5281/zenodo.18258991
...
 
-- **Source Language Identification** To identify source languages, we utilize the <a hreflang="en" href="https://github.com/WebAssembly/wabt">WebAssembly Binary Toolkit (`wasm2wat`)</a> (see [the also MDN guide](https://developer.mozilla.org/docs/WebAssembly/Guides/Text_format_to_Wasm)) and Rust-based parsers to analyze imports, exports, and custom sections for compiler metadata and language fingerprints.
+**Analysis:** In addition to the HTTP Archive dataset, we use <a hreflang="en" href="#/">almanac-wasm</a> a tool to download and validate the WebAssembly modules identified from the HTTP Archive for local analysis. This tool extracts metadata from these downloaded files, allowing us to identify programming languages, libraries, and specific features used within the Wasm modules.
```
tunetheweb: What is the href pointing to?

src/content/en/2025/webassembly.md
tunetheweb: This was resolved without resolving the issue.

src/content/en/2025/webassembly.md
tunetheweb: This was resolved, without resolving the issue.

src/content/en/2025/webassembly.md
tunetheweb: I presume you mean compressed not "uncompressed"? Can we make it clear in the text what this is?

src/content/en/2025/webassembly.md
tunetheweb: Can we add something like this to the text?

src/content/en/2025/webassembly.md
tunetheweb: I'm not sure what you mean by that? I'm still very confused here..

src/content/en/2025/webassembly.md
tunetheweb: This comment was resolved without resolving the issue.

src/content/en/2025/webassembly.md
tunetheweb: Can we make this clearer in the text then? I'm I'm confused then readers will like be too.

tunetheweb (changes requested):
I'm still very confused and not sure this PR make this text any less confusing. Can we address the comments I've raised @nrllh and @nimeshgit ?

src/content/en/2025/webassembly.md
nimeshgit: lets talk on slack tomorrow afternoon about this.

src/content/en/2025/webassembly.md
nimeshgit: lets resolve on slack tomorrow afternoon.

src/content/en/2025/webassembly.md
nimeshgit: Yes, It is confusing bit !

You would know, We have various techniques for compression i.e. br, gzip, zstd. We can see in year 2021 along with the chart.

We know and according to our past research with benchmarks; for the best performance to achive, 

Image

apart from content-encoding in header, We are also using optimization tools and techniques offine in test/release pipelines.

Wasm are also practiced to compressed further with recommended tools like Brotli -- I believe without such compression, as a developer like myself and many colleagues / friends outside use to term such "raw" wasm as  "uncompressed" wasm -- they are compressed but not compressed with LZ77 algorithm, Huffman coding, 2nd-order context modeling and or other latest techniques.

@nrllh  and @tunetheweb , This was the reason, 
We have detail explaination in our Year 2021 article about how to optimize compression with Brotli after raw compression session.

We would also have compression detail session for both clients I am sure that would help to understand the compression technique evolution for WebAssembly so far.

nimeshgit: reference :  https://facebook.github.io/zstd/#benchmarks
https://docs.google.com/spreadsheets/d/16z2MNwq8FFbuNYcJJZceML6rB5VAmBXNNHZy5FZuf8g/edit?gid=1745126816#gid=1745126816

src/content/en/2025/webassembly.md
```diff
@@ -136,17 +116,11 @@ There are over 300,000 wasm requests in our dataset, which come down to 32,197 u
   )
 }}
 
-We observed that the MIME type `application/wasm` is used in 293,470 requests on desktop clients and 301,127 requests on mobile clients.
+The standard `application/wasm` MIME type was identified in 293,470 desktop and 301,127 mobile requests. Instances of missing or incorrect MIME types (such as `text/html` or `text/plain`) were low, affecting 3.2% of desktop and 2.4% of mobile requests. These represent a significant decline compared to 2021, indicating improved awareness and adherence to proper server configuration.
 
-Some requests lacked a `Content-Type` header, and some had incorrect MIME types, such as `text/html` or `text/plain`. These account for 3.2% and 2.4% of requests, respectively. Compared to 2022, these percentages have dropped significantly, indicating increased awareness in setting the correct MIME type for WASM applications.
```
nimeshgit: This observervation shows about developer awareness when building wasm, can we have this as

Some requests lacked a `Content-Type` header, and some had incorrect MIME types, such as `text/html` or `text/plain`, they are 3.2% and 2.4% of requests  respectively on desktop and mobile. Compared to 2022, these percentages have dropped significantly, indicating increased awareness in setting the correct MIME type for Wasm applications.

src/content/en/2025/webassembly.md
nrllh: I removed the part ".. indicating.."

nimeshgit:
Barry, we have merged @nrllh updates plus your suggetions and compression methods for desktop and mobile clients along with couple of edits.

src/content/en/2025/webassembly.md
nimeshgit: can we mark this as resolved fixed ?

nimeshgit (approved):

tunetheweb (approved):
OK let's merge this.

Thanks for improving the text!

PR #4409, opened 2026-02-09
suggestions of pull request #4401

nimeshgit:
This includes suggestions of pull requests #4401  along with below updates.

- Added compression methods stats for desktop and mobile client.
- Migrated sql from year 2021 to get raw size, stripped size and optimized size of the web assembly.
- Updated the language usage stats, It has duplicate records counts and details in percentage for number of Wasm module downloaded, validated and analysed.


nimeshgit:
@tunetheweb , as discussed with Nurullah, let me update couple of parts here.

nrllh:
As discussed on Slack, let’s please work on #4401. We should not create a new PR for this, so I’m closing this one.

PR #4432, opened 2026-03-11
Add cloudbuild.yaml for Google Cloud Build configuration

max-ostapenko:
Added a deployment workflow draft - based on main push trigger.
Removed duplication with the existing GH actions.

### DIFF
--- a/.github/workflows/test_website.yml
+++ b/.github/workflows/test_website.yml
@@ -98,4 +98,4 @@ jobs:
         run: gh pr merge --squash --auto "$PR_URL"
         env:
           PR_URL: ${{github.event.pull_request.html_url}}
-          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+          GITHUB_TOKEN: ${{ secrets.AUTO_MERGE_PAT }}
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,15 +3,15 @@ updates:
   - package-ecosystem: "npm"
     directory: "/src/"
     schedule:
-      interval: "daily"
+      interval: "monthly"
     ignore:
       - dependency-name: "prettier"
         update-types: ["version-update:semver-patch"]
   - package-ecosystem: "pip"
     directory: "/src/"
     schedule:
-      interval: "daily"
+      interval: "monthly"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "monthly"
--- a/.github/workflows/add-to-release-notes.yml
+++ b/.github/workflows/add-to-release-notes.yml
@@ -16,6 +16,6 @@ jobs:
     if: github.repository == 'HTTPArchive/almanac.httparchive.org'
     steps:
     - name: Update release notes
-      uses: release-drafter/release-drafter@v6.4.0
+      uses: release-drafter/release-drafter@v7.1.1
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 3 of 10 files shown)


cloudbuild.yaml
```diff
@@ -0,0 +1,142 @@
...
+
+substitutions:
+  # Update when a new Prince version is released: https://www.princexml.com/latest/
+  # python:3.12 is based on Debian Bookworm (12), so use the debian12 package.
+  _PRINCE_PACKAGE: 'prince_16.1-1_debian12_amd64.deb'
+
+steps:
+  # ─────────────────────────────────────────────────────────────────────────
+  # Step 1: Install dependencies, generate all chapters, run full test suite.
+  # Uses python:3.12 (Debian Bookworm) and installs Node 20 at runtime.
```
tunetheweb: Node 20's quite old now. Let's got to 24.

Can probably also update Python to 3.14 but not sure if that's tied to the OS based on comment above, and changing that would require a PrinceXML update so let's hold off on that.

```suggestion
  # Uses python:3.12 (Debian Bookworm) and installs Node 24 at runtime.
```

cloudbuild.yaml
```diff
@@ -0,0 +1,142 @@
...
+  - name: 'python:3.12'
+    id: 'build-and-test'
+    entrypoint: 'bash'
+    dir: 'src'
+    args:
+      - '-c'
+      - |
+        set -e
+        # Install Node.js 20 (nodesource setup script also runs apt-get update)
+        curl -fsSL https://deb.nodesource.com/setup_20.x | bash -
```
tunetheweb:
```suggestion
        # Install Node.js 24 (nodesource setup script also runs apt-get update)
        curl -fsSL https://deb.nodesource.com/setup_24.x | bash -
```

.github/workflows/predeploy.yml
```diff
@@ -1,94 +0,0 @@
-######################################
```
tunetheweb, line 1: Let's keep this for now, but disable it. Just until we get confidence this all works.

.github/workflows/linter.yml
```diff
@@ -10,9 +10,6 @@ name: Lint Code Base
 on:
   workflow_dispatch:
   pull_request:
-  push:
-    branches:
-      - main
```
tunetheweb, line 15: Any reason not keep this? I kinda like it as a backup in case something is missed in a PR review.

.github/workflows/test_website.yml
```diff
@@ -11,9 +11,6 @@ on:
   pull_request:
     paths-ignore:
       - 'sql/**'
-  push:
-    branches:
-      - main
```
tunetheweb, line 16: Any reason not keep this? I kinda like it as a backup in case something is missed in a PR review.

.github/workflows/predeploy.yml
max-ostapenko: Valid!

.github/workflows/linter.yml
max-ostapenko: I had in mind that we need a complete sequential flow - if something fails to stop the deployment.
If we run one thing on GitHub and another on Cloud Build this doesn't happen.
So the idea is to migrate everything required to deployment workflow.

tunetheweb: Agree it's not needed. I just like it as an extra backup check so I can go into the Actions tab and see exactly when it started failing on the `main` branch. Similar to how we're seeing in har.fyi now.

Of course, ideally, it should never make it to `main` if it's going to cause an error. But sometimes things aren't caught in PRs...

So, while I agree it's not strictly necessary, and not part of the deploy process, I see no harm to keep it?

PR #4487, opened 2026-06-23
Update cloudbuild to Node 24

tunetheweb:
We tried this once before but [had to roll it back](https://github.com/HTTPArchive/almanac.httparchive.org/commit/fb38949c25535944d68936a392d57ce22341de3d).

Can't remember why?

However, since then [we upgraded the OS](https://github.com/HTTPArchive/almanac.httparchive.org/commit/16f9e2538718c48d67e8690eef952fcb19d0468e) so maybe it works now?

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -29,8 +29,8 @@ steps:
         apt-get update -qq
         apt-get install -y python3 python3-venv curl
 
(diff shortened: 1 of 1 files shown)


max-ostapenko (approved):
Let's try 🤞

Issue #1300, opened 2020-09-17
Translate content into German


tunetheweb:
These are the core templates - without which we cannot release any translated chapters. They are in the [language specific templates directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/templates):
- [ ] base.html - 
- [ ] base_chapter.html - 
- [ ] error.html - 

**2024**
- [ ] base.html - 
- [ ] contributors.html - 
- [ ] index.html - 
- [ ] table_of_contents.html - 

**2022**
- [ ] base.html - 
- [ ] contributors.html - 
- [ ] index.html - 
- [ ] table_of_contents.html - 

**2021**
- [ ] base.html - 
- [ ] base.html - foreword
- [ ] contributors.html - 
- [ ] index.html - 
- [ ] table_of_contents.html - 

**2020**
- [ ] base.html - 
- [ ] contributors.html - 
- [ ] index.html  - 
- [ ] table_of_contents.html - 

**2019**
- [ ] base.html - 
- [ ] contributors.html - 
- [ ] index.html  - 
- [ ] table_of_contents.html - 

These are the chapters to be translated, in rough order of popularity. They exist in the [content directory](https://github.com/HTTPArchive/almanac.httparchive.org/tree/main/src/content):

**2024**
- [ ] css - 
- [ ] javascript - 
- [ ] markup - 
- [ ] structured-data - 
- [ ] fonts - 
- [ ] media - 
- [ ] webassembly - 
- [ ] third-parties - 
- [ ] seo -  
- [ ] accessibility - 
- [ ] performance - 
- [ ] privacy - 
- [ ] security - 
- [ ] cms - 
- [ ] jamstack - 
- [ ] sustainability - 
- [ ] page-weight - 
- [ ] cdn - 
- [ ] ecommerce -
- [ ] http - 
- [ ] cookies - 

**2022**
- [ ] css - 
- [ ] javascript - 
- [ ] markup - 
- [ ] structured-data - 
- [ ] fonts - 
- [ ] media - 
- [ ] webassembly - 
- [ ] third-parties - 
- [ ] interoperability - 
- [ ] seo -  
- [ ] accessibility - 
- [ ] performance - 
- [ ] privacy - 
- [ ] security - 
- [ ] mobile-web - 
- [ ] capabilities - 
- [ ] pwa - 
- [ ] cms - 
- [ ] jamstack - 
- [ ] sustainability - 
- [ ] page-weight - 
[...]

PR #4496, opened 2026-06-26
Translate third-parties (2025) to German

mika943:
Part of https://github.com/HTTPArchive/almanac.httparchive.org/issues/1300


tunetheweb (approved):
I can't speak to the Germin but this LGTM from a technical perspective.

@max-ostapenko not sure if you want to review the German?

PR #4498, opened 2026-06-26
Translate base.html to German

mika943:
Part of #1300

PR #4509, opened 2026-07-07
Enable German for 2019-2024

tunetheweb:
Part of #1300

Since @mika943 did the hard work in #4498 I've used those to do the base templates for the other years (minus the foreword).

### DIFF
--- a/src/templates/de/2019/base.html
+++ b/src/templates/de/2019/base.html
@@ -2,7 +2,7 @@
 
 {% block methodology_stat_1 %}5,8M{% endblock %}
 {% block methodology_stat_2 %}20,9 TB{% endblock %}
-{% block total_websites %}fast 9 Millionen{% endblock %}
+{% block total_websites %}fast 6 Millionen{% endblock %}
 {% block dataset %}Juli 2019{% endblock %}
 
 {% block foreword %}
--- a/src/templates/de/2019/base.html
+++ b/src/templates/de/2019/base.html
@@ -1,0 +1,28 @@
+{% extends "%s/base.html" % lang %}
+
+{% block methodology_stat_1 %}5,8M{% endblock %}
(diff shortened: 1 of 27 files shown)


src/templates/de/2019/base.html
```diff
@@ -0,0 +1,28 @@
+{% extends "%s/base.html" % lang %}
+
+{% block methodology_stat_1 %}5,8M{% endblock %}
+{% block methodology_stat_2 %}20,9 TB{% endblock %}
+{% block total_websites %}fast 9 Millionen{% endblock %}
```
max-ostapenko:
```suggestion
{% block total_websites %}fast 6 Millionen{% endblock %}
```

src/templates/de/2024/base.html
```diff
@@ -0,0 +1,41 @@
+{% extends "%s/base.html" % lang %}
+
+{% block methodology_stat_1 %}16.9M{% endblock %}
```
max-ostapenko:
```suggestion
{% block methodology_stat_1 %}16,9M{% endblock %}
```

max-ostapenko (dismissed):

src/templates/de/2019/base.html
tunetheweb: Good spot!

tunetheweb:
Hmmm... this "Dismiss stale pull request approvals when new commits are pushed" setting seems a bit much:

image

So if you approve a PR with some suggestions and I accept those suggestions, I lose your approval?

Going to merge this anyway, but do you think we should untick this box @max-ostapenko ? Or do we need it for automation for some reason?

max-ostapenko:
@tunetheweb I don't think it impacts automation, more of a security feature. 
Feel free to turn off.

tunetheweb:
Done!

PR #4511, opened 2026-07-11
disable Google Signals and ad personalization in Analytics

max-ostapenko:
@tunetheweb Looking at #4508 - maybe just turn-off GA ad features?

### DIFF
--- a/src/templates/base.html
+++ b/src/templates/base.html
@@ -18,6 +18,8 @@
         function gtag() {
             dataLayer.push(arguments);
         }
+        gtag('set', 'allow_google_signals', false);
+        gtag('set', 'allow_ad_personalization_signals', false);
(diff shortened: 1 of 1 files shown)


src/templates/base.html
```diff
@@ -18,6 +18,8 @@
         function gtag() {
             dataLayer.push(arguments);
         }
+        gtag('set', 'allow_google_signals', false);
+        gtag('set', 'allow_ad_personalization_signals', false);
```
tunetheweb, line 22: The second definitely, since we don't use ads.

The first could be useful to understand our demographics more. But we don't use it enough, and also don't have a cookie banner to ask this permission so probably should just turn it off.

tunetheweb (approved):

PR #4512, opened 2026-07-15
Translate accessibility statement to German #HSFDPMUW

JulianGStudium:
This file provides a comprehensive accessibility statement for the Web Almanac, detailing commitments, measures, conformance status, and feedback mechanisms regarding digital accessibility

Makes progress on #1300


tunetheweb (approved):
LGTM from a technical point of view.

@max-ostapenko you OK with the German?

max-ostapenko:
@tunetheweb all good!

PR #4528, opened 2026-08-06
Translation of 2025 base.html foreword to Japanese

ksakae1216:
Please review for me.

I think this is the last translation for fiscal year 2025.
Please let me know if there are any other files that need to be translated.
Thank you very much for your support throughout this fiscal year as well.


tunetheweb (approved):
Thank you for all your efforts on this, yet again! Fantastic work.

The only other file is the Methodology, which is mostly a repeat of previous years with some tweaks.

I don't think you translated it for every year so up to you whether you want to. The chapters are the main things IMHO.

PR #4541, opened 2026-09-21
docs: use neutral Media Session artwork URLs

RichardJansma:
## Summary

- replace `via.placeholder.com` artwork URLs in the 2022 Media Session example with reserved `www.example.com` paths
- keep the English, Japanese, and Portuguese versions of the identical code sample aligned

## Why

The existing third-party placeholder endpoint is unavailable. We encountered disappearing placeholder services ourselves, which is why Jansma VIP offers Placeholder.photo and intends to maintain it for the long term.

For this documentation sample, the maintainer suggested reserved `www.example.com` paths to avoid any live-service dependency, so this PR follows that preference.

## Validation

- only the artwork URL strings changed
- the declared sizes and `image/png` types are unchanged
- `git diff --check` passes


tunetheweb (approved):
Thank you!

And very generous in taking the time to submit this neutral PR without even pushing your own service in the text!

RichardJansma:
Thank you! I’m very happy to help.

The goal of this campaign is to find old references to services such as `via.placeholder.com`, `placeholder.com`, and `placehold.it`, and help projects replace or remove them in whatever way fits best.

I built my own service as a dependable long-term replacement that I can maintain myself, so I feel comfortable recommending it where a live placeholder service is genuinely useful. But the main goal is fixing broken references, whether that means using my service, another service, a local asset, or a neutral example URL like in this PR.

For example, in [ng-select PR #2888](https://github.com/ng-select/ng-select/pull/2888), the best solution was simply to remove the obsolete placeholder-related properties altogether.

I enjoy finding these forgotten issues and helping maintainers resolve them. Thanks again for reviewing and merging this!

